PR #25, opened 2018-07-05
Merge hca-module-v5 into develop

hannes-ucsc:
Like we talked about, @mikebaumann and @caaespin.


hannes-ucsc:
I have verbal agreement from @caaespin and @mikebaumann to merge this.

PR #40, opened 2018-07-11
Add donor id to indexes

jyotheeshwar:


indexer/project/hca/extractors.py
```diff
@@ -101,6 +101,12 @@ def v5_2(metadata_file: dict) -> dict:
             "hca_id": jmespath.search(hca_id, metadata_file),
             "_type": "specimens"
         }
+        # append donor information if exists
+        described_by = "content.describedBy || `null`"
+        if "donor_organism" in jmespath.search(described_by, metadata_file):
+            biomaterial['isDonorOrganism'] = "true"
```
caaespin: @jyotheeshwar can you instead create a field called `donor_id` with the biomaterial_id? This would allow for a more straightforward query. Doing a boolean is useful, but it would complicate the query that UX is requesting. Set it to None if not present. Since this is only a simple field, would you mind doing pure python vs jmespath?

jyotheeshwar:
@caaespin updated the PR to address your comments, as well as address #54 

I didn't change the query from `jmespath` though for consistency's sake. Is there reason to not use it?

jyotheeshwar:
@caaespin also added code to address #89

PR #103, opened 2018-07-17
Feature/fix filetypes cols

mjkrause:


webservice/bagitutils.py
```diff
@@ -212,9 +213,13 @@ def participants_and_max_files_in_sample_and_protocols(self):
                 if protocol is not None:
                     native_protocols.add(protocol)
 
-        return participants, max(samples.values()), native_protocols
+            # Track all file types in manifest.
+            for file_type_ in row[BoardwalkColumns.FILE_TYPE].split(','):
```
coverbeck: Why the `split()`? I downloaded the manifest from dev, and I don't see any commas in that column.

I may be missing something, or maybe the input is different for the export case, but this seems unnecessary.

webservice/bagitutils.py
```diff
@@ -322,3 +330,22 @@ def native_column_name(native_protocol, suffix):
...
+    @staticmethod
+    def get_suffix(s):
+        """Lookup-table hack to address the problem of more than one file type 
+        occurring in a given file type column (2018-07-16).
+        :parameter s: (str) from a row from the manifest TSV file
+        :returns suffix: (str) holds integer corresponding to a file type"""
+
+        s = s.lower()
+
+        if s.endswith('fastq.gz') or s.endswith('fastq'): # unaligned read
```
coverbeck: Did you verify these extensions with Francisco? I talked about this with Lon and he suggested there could also be fasta and sam file types in bioinformatics, but if U of C isn't handling them, then it's OK.

webservice/bagitutils.py
mjkrause: I did some testing. Both lists, `row[BoardwalkColumns.FILE_URLS]` and `row[BoardwalkColumns.FILE_TYPE]`, I tested only had one item, and this may always be the case. We could take it out based on that. That said, I would suggest erring on the safe side and leave on the off-chance that for some reason that there's more than 1 item in that list. After all the split is silent. What do you think?

webservice/test/manifest_bag/sample.tsv
```diff
@@ -1,7 +1,7 @@
-entity:sample_id	donor_uuid	file_dos_uri1	file_dos_uri2	file_dos_uri3	file_path1	file_path2	file_path3	file_type1	file_type2	file_type3	gs_url1	gs_url2	gs_url3	participant	s3_url1	s3_url2	s3_url3	specimen_uuid	upload_file_id1	upload_file_id2	upload_file_id3
-s1	d1							crai	cram		gs://junk.crai	gs://junk.cram		d1				sp1			
-s2	d1							crai	cram					d1	s3://junk.crai	s3://junk.cram		sp1			
-s3	d2							crai	cram		gs://junk.crai	gs://junk.cram		d2				sp1			
-s4	d3							crai	cram		gs://junk.crai	gs://junk.cram		d3				sp1			
-s5	d4							crai	cram	bam	gs://junk.crai	gs://junk.cram	gs://junk.bam	d4				sp2			
-s6	d5							crai			gs://junk.crai			d5				sp3			
+entity:sample_id	bai_url1	bam_url2	crai_url1	cram_url2	donor_uuid	file_dos_uri1	file_dos_uri2	file_path1	file_path2	file_type1	file_type2	gs_url1	gs_url2	participant	s3_url1	s3_url2	specimen_uuid	upload_file_id1	upload_file_id2
```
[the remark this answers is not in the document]
mjkrause: This refers to point 1, b. and c. in Francisco's email: "Aligned read index files (CRAI, BAI, etc): file_dos_uri1, file_path1, file_type1, etc
Aligned read files (CRAM, BAM, etc): file_dos_uri2, file_path2, file_type2, etc"

That means that if CRAMs and BAMs are in the same manifest (which is the case in the mock TSV) they will end up in the same column. Unless we create separate column number that denote BAMs and separate those from CRAM, or unless the situation of CRAMs and BAMs do never occur in the wild...

In addition, here's my convo with Francisco from this afternoon:
"Hi Francisco, thanks for your email reply and explanation (the one Charles sent out). Could you please elaborate on what "others" refers to in the file types? Alternatively, if time permits, could you sent me a list of all the file types you have considered? Thanks.

fortuno [2:17 PM]
Sure! Right now we only collect those three data types in the bdBag: aligned reads, unaligned reads and aligned reads index
for aligned reads we have the following types considered: BAM and CRAM.
for unaligned reads we consider: FASTQ or BAM (careful, BAM can be aligned reads or unaligned reads).
for index we consider: BAI or CRAI.
those are the different types we consider in our dictionary right now.

Michael Krause (UCSC) [2:20 PM]
awesome, thanks for the info, sounds good.

fortuno [2:21 PM]
if you want to know more data types in addition to the one we collect right now in the bdbag but we still have in our dictionary, you can go here:
https://dcp.bionimbus.org/DD
under `Data File` table you have all the file types we can collect. Inside each of them, you have the property `data_format` with the list of possible data file formats

Michael Krause (UCSC) [2:27 PM]
great, I'm just trying to be thorough and cover all cases. but the most important things is that the two full-stacks have the same types, so for now I'll leave it with those you currently use."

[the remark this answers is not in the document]
coverbeck: I feel like we are talking past each other. :) I don't think you get my point, and I'm probably misunderstanding you.

I am referring to the column NAMES, not the values.

You have a column with the NAME `bai_url1` in the first row. Per the email, the .bai file VALUES should be in columns with the NAME `file_dos_uri1`, `file_path1`, etc. I hope that's clear; if not, let's talk in person tomorrow. Or where he says you should create column NAMEs such as `bai_url1`.

Or maybe this is just arbitrary metadata you are using for testing and it's OK?

webservice/test/test_bagitutils.py
```diff
@@ -105,6 +108,25 @@ def test_get_suffix(self):
...
+         "file_type2"."""
+        bag = BagHandler(data=self.data3, bag_info={}, bag_name='manifest')
+        participants, max_files_in_sample, protocols, file_types = \
+            bag.participants_and_max_files_in_sample_and_protocols()
+        self.assertEqual(len(participants), 1)
+        self.assertEqual(len(protocols), 1)
+        self.assertEqual(max_files_in_sample, 1)
+
+        samples = bag.samples(protocols, file_types)
+        self.assertEqual(len(samples[0].keys()), 9)
```
coverbeck: Might be nice to instead/also assert that `file_type1` and `file_type0` don't exist, which is clearer than the magic number 9.

webservice/test/test_bagitutils.py
```diff
@@ -105,6 +108,25 @@ def test_get_suffix(self):
         self.assertTrue(suffix, '0')
         self.assertRaises(ValueError, bag.get_suffix, 'not_a_filetype')
 
+    def test_only_one_cram_in_data(self):
+        """The data we're loading here only contains one row, and that row
+        contains a CRAM file. We tests whether "samples" has a column name
```
coverbeck: `test` not `tests`.

webservice/bagitutils.py
coverbeck: You still have the `split`.

mjkrause: To clarify this point: `BoardwalkColumns.FILE_TYPE` contains a comma-separated list of all file types in the manifest. Before iterating over it in the for-loop we need the split. So AFAIK it's not up for vote...

coverbeck: Can you please show an example where there is a comma-separated list. This is a column, for each row, the value is a single file type.

webservice/test/test_bagitutils.py
mjkrause: fixed

PR #151, opened 2018-07-31
Reformat test data files as JSON

alex-hancock:


test/indexer/test_hca_indexer.py
```diff
@@ -26,19 +26,20 @@ def setUpModule():
...
+    def _get_data_files(filename, updated=False):
         data_prefix = "data/"
+        metadata_suffix = ".metadata.json"
+        manifest_suffix = ".manifest.json"
+        if updated:
+            filename += ".updated"
 
-        with open(data_prefix + metadata_file, 'r') as infile:
+        with open(data_prefix + filename + metadata_suffix, 'r') as infile:
             metadata = json.loads(infile.read())
```
hannes-ucsc: Use `json.load(f)` instead. It's less code but also much more efficient for big files.

PR #157, opened 2018-08-02
Added end-point to delete bundles (fixes #45)

### LINKED ISSUE #45: Implement Handlers for Delete Events on the Indexer
As a user I would like to see data removed from the DSS also be removed from our indexers (and therefore, disappear from the portal. 

DOD: We have a way to retrieve a list of ES ids that match the bundle id from the event. Update the documents in ES. Make the logic of retrieving ES IDs as configurable at deploy time.  Unit tests included.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-75)
┆Project Name: azul
┆Issue Number: AZUL-75

jyotheeshwar:
Sample payload format:

```
{
            "match": {
                              "bundle_uuid":"9dec1bd6-ced8-448a-8e45-1fc7846d8995",
                              "bundle_version":"2018-03-29T154319.834528Z",
                              "index": "browser_files_dev"
                           }
}
```

### DIFF
--- a/src/azul/indexer.py
+++ b/src/azul/indexer.py
@@ -33,9 +33,6 @@ class BaseIndexer(ABC):
             es_client.indices.create(index=index_name, body=self.properties.settings, ignore=[400])
             es_client.indices.put_mapping(index=index_name, doc_type="doc", body=self.properties.mapping)
 
-        errored_documents = defaultdict(int)
-        conflict_documents = defaultdict(int)
-
         # Collect the documents to be indexed
         indexable_documents = {}
         for transformer in self.properties.transformers:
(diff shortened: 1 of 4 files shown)


jyotheeshwar:
@hannes-ucsc updated to handle conflicts.

jyotheeshwar:
Note that I'm only retrying once, unlike the indexer, as hitting Elasticsearch with multiple retries for a deletion operation seems like an overkill, and I would expect upstream code to handle the error appropriately.

src/azul/indexer.py
```diff
@@ -183,3 +183,72 @@ def merge(cur_bundles, new_bundles):
...
+
+        for hit in hits:
+            current_doc_id = hit['_id']
+            hit_doc = es_client.get(index=index,
+                                    doc_type="doc",
+                                    id=current_doc_id)
+
+            current_hit_doc = hit_doc['_source']
+            for bundle in current_hit_doc['bundles']:
+                if bundle['version'] == bundle_version:
```
hannes-ucsc: Don't you also need to compare the bundle UUID? Two distinct bundles (different UUID) could have the same version.

src/azul/indexer.py
```diff
@@ -183,3 +183,72 @@ def merge(cur_bundles, new_bundles):
...
+                    bundle['contents'] = None
```
hannes-ucsc: Why aren't you just deleting the bundle entry from the `bundles` dict?

src/azul/indexer.py
```diff
@@ -183,3 +183,72 @@ def merge(cur_bundles, new_bundles):
...
+
+            update_doc = {
+                "doc": current_hit_doc
+            }
+            try:
+                doc_version = hit_doc['_version']
+                update_call(current_doc_id, update_doc, doc_version)
+            except ConflictError as e:
+                on_conflict(current_doc_id, update_doc, e)
+            except ElasticsearchException as e:
```
hannes-ucsc: Don't swallow exceptions like that unless you have a good reason to do so. Embellishing logs is not a good reason. You can log and reraise if you want to embellish the logs. But you should default to letting the natural exception handling runs its course. If you log an exception, use `log(…,exc_info=True)` or `log(…,exc_info=e)`. Prefer the former, catch without `as e` and use a bare `raise`. 

https://docs.python.org/3/library/exceptions.html

src/azul/indexer.py
```diff
@@ -183,3 +183,72 @@ def merge(cur_bundles, new_bundles):
...
+
+        search_query = {
+            "query": {
+                "match": {
+                    "bundles.uuid.keyword": bundle_uuid
+                }
+            }
+        }
+
+        current_doc = es_client.search(index=index,
```
hannes-ucsc: This variable name is misleading. The search() method returns a response with potentially multiple documents.

hannes-ucsc:
@jyotheeshwar, following up on our discussion from Friday: We can't rely on loading the deleted bundle's files from DSS in order to determine the specimens and files that bundle contributed to. DSS will give us a 404. We will have to rely on the ES index to give us that information. Let's discuss this in person today or tomorrow.

mpanik:
@hannes-ucsc this should be ready for review (writing on behalf of @jyotheeshwar, since he is away for the rest of the week). 
@alex-hancock's solution #202 solved the issue I had running Jyo's tests

src/azul/indexer.py
```diff
@@ -176,8 +196,60 @@ def merge(cur_bundles, new_bundles):
...
+        bundle_version = dss_notification['match']['bundle_version']
+
+        es_client = self.properties.elastic_search_client
+
+        search_query = {
+            "query": {
+                "bool": {
+                    "must": [
+                        {
+                            "term": {
```
hannes-ucsc: I think this query could be simplified using one `terms` branch instead of two `term` branches.

src/azul/indexer.py
```diff
@@ -176,8 +196,60 @@ def merge(cur_bundles, new_bundles):
...
+        current_doc = es_client.search(doc_type="doc",
+                                       body=search_query)
+
+        hits = current_doc['hits']['hits']
+
+        updated_hits = []
+
+        for hit in hits:
+            current_doc_id = hit['_id']
+            hit_doc = es_client.get(index=hit['_index'],
```
hannes-ucsc: Why don't you fetch the source with query? This 1+N approach is really inefficient.

hannes-ucsc:
See https://github.com/DataBiosphere/azul/issues/45#issuecomment-414800044

mpanik:
Closing this in favor of https://github.com/DataBiosphere/azul/pull/268

PR #193, opened 2018-08-14
Refactor and consolidate elasticsearch client usage

mpanik:
I've decided to go with [AWSRequestsAuth](https://github.com/DavidMuller/aws-requests-auth)
few reasons:
- seems more up to date (in contrary with the other library which last commit was on Feb 8, 2016)
- has neat built-in feature to retrieve aws credentials with help of boto library

@hannes-ucsc  is `es_check()` safe to delete, from the name of the function it suppose to be health check for `es`, which we already have

### DIFF
--- a/src/azul/es.py
+++ b/src/azul/es.py
@@ -1,25 +1,30 @@
+import logging
+from functools import lru_cache
+
+from aws_requests_auth.boto_utils import BotoAWSRequestsAuth
 from elasticsearch import Elasticsearch, RequestsHttpConnection
-from aws_requests_auth import boto_utils
-from aws_requests_auth.aws_auth import AWSRequestsAuth
 
 from azul import config
 from azul.deployment import aws
 
+logger = logging.getLogger(__name__)
 
-def es_client():
-    host = config.es_endpoint[0]
-    port = config.es_endpoint[1]
-    ssl_required = True if port == 443 else False
-    if ssl_required:
-        aws_auth = AWSRequestsAuth(aws_host=host,
-                                   aws_region=aws.region_name,
-                                   aws_service='es',
-                                   **boto_utils.get_credentials())
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             http_auth=aws_auth,
-                             use_ssl=True,
-                             verify_certs=True,
-                             connection_class=RequestsHttpConnection)
-    else:
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             connection_class=RequestsHttpConnection)
+
+class ESClientFactory:
+
+    @classmethod
+    def get(cls):
+        host, port = config.es_endpoint
+        return cls._create_client(host, port, config.es_timeout)
+
+    @classmethod
+    @lru_cache(maxsize=32)
+    def _create_client(cls, host, port, timeout):
+        logger.debug(f'Creating ES client [{host}:{port}]')
+        common_params = dict(hosts=[dict(host=host, port=port)], timeout=timeout)
+        if host.endswith(".amazonaws.com"):
+            aws_auth = BotoAWSRequestsAuth(aws_host=host, aws_region=aws.region_name, aws_service='es')
+            return Elasticsearch(http_auth=aws_auth, use_ssl=True, verify_certs=True,
+                                 connection_class=RequestsHttpConnection, **common_params)
+        else:
+            return Elasticsearch(**common_params)
--- a/test/service/test_request_validation.py
+++ b/test/service/test_request_validation.py
@@ -41,7 +41,7 @@ class FacetNameValidationTest(WebServiceTestCase):
         self.assertEqual(response.json(), expected_json)
 
     def test_health_es_unreachable(self):
-        with mock.patch.dict(os.environ, AZUL_ES_ENDPOINT='nonexisting-index.us-east-1.es.amazonaws.com:80'):
+        with mock.patch.dict(os.environ, AZUL_ES_ENDPOINT='nonexisting-index.com:80'):
             url = self.base_url + "health"
             response = requests.get(url)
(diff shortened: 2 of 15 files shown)


src/azul/es.py
```diff
@@ -1,24 +1,25 @@
...
-                                   **boto_utils.get_credentials())
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             http_auth=aws_auth,
-                             use_ssl=True,
-                             verify_certs=True,
-                             connection_class=RequestsHttpConnection)
-    else:
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             connection_class=RequestsHttpConnection)
+class ElasticsearchClient:
```
hannes-ucsc: Can we rename this to `ESClientFactory` or `ElasticsearchClientFactory`?

src/azul/es.py
```diff
@@ -1,24 +1,25 @@
...
+        host, port = config.es_endpoint
+        ssl_required = True if port == 443 and host.endswith(".amazonaws.com") else False
+        return ElasticsearchClient._create_client(host, port, ssl_required, 60)
+
+    @staticmethod
+    @lru_cache(maxsize=32)
+    def _create_client(host, port, ssl_required, timeout):
+        common_params = dict(hosts=[dict(host=host, port=port)], timeout=timeout)
+        if ssl_required:
+            aws_auth = BotoAWSRequestsAuth(aws_host=host, aws_region='us-east-1', aws_service='es')
```
hannes-ucsc: The hard-coded region is actually the cause for https://github.com/DataBiosphere/azul/issues/194. Can you be in touch with @achave11 because he has a fix for this but no PR yet.

src/azul/es.py
```diff
@@ -1,24 +1,25 @@
...
-                             connection_class=RequestsHttpConnection)
-    else:
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             connection_class=RequestsHttpConnection)
+class ElasticsearchClient:
+
+    @staticmethod
+    def get():
+        host, port = config.es_endpoint
+        ssl_required = True if port == 443 and host.endswith(".amazonaws.com") else False
```
hannes-ucsc: Shouldn't that be `or` instead of `and`?

src/azul/es.py
```diff
@@ -1,24 +1,25 @@
...
+        return ElasticsearchClient._create_client(host, port, ssl_required, 60)
+
+    @staticmethod
+    @lru_cache(maxsize=32)
+    def _create_client(host, port, ssl_required, timeout):
+        common_params = dict(hosts=[dict(host=host, port=port)], timeout=timeout)
+        if ssl_required:
```
hannes-ucsc: I think this is conflating SSL with AWS Elasticsearch. The latter implies the former but not vice versa. I think the host.endswith(".amazonaws.com") logic should be in this method and the ssl_required argument should be eliminated and instead be derived the host and port arguments. The use of BotoAWSRequestsAuth should be triggered solely by the host name.

src/azul/es.py
```diff
@@ -1,24 +1,25 @@
...
-    else:
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             connection_class=RequestsHttpConnection)
+class ElasticsearchClient:
+
+    @staticmethod
+    def get():
+        host, port = config.es_endpoint
+        ssl_required = True if port == 443 and host.endswith(".amazonaws.com") else False
+        return ElasticsearchClient._create_client(host, port, ssl_required, 60)
```
hannes-ucsc: Might make sense to factor the timeout into an environment variable like AZUL_ES_TIMEOUT and a Config.es_timeout wrapper around it. Up to you.

src/azul/es.py
```diff
@@ -1,25 +1,29 @@
...
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             http_auth=aws_auth,
-                             use_ssl=True,
-                             verify_certs=True,
-                             connection_class=RequestsHttpConnection)
-    else:
-        return Elasticsearch(hosts=[{'host': host, 'port': port}],
-                             connection_class=RequestsHttpConnection)
+class ESClientFactory:
+    logger = logging.getLogger(__name__)
```
hannes-ucsc: We can't add a logger to every class that wants to log something. Please stick with standard pattern of using a per-module logger that is initialized at the beginning of the module.

src/azul/es.py
```diff
@@ -1,25 +1,29 @@
...
+
+    @staticmethod
```
hannes-ucsc: I wouldn't ask this normally but since you have to change the logger, I might as well: Please make this a `@classmethod` and …

src/azul/es.py
```diff
@@ -1,25 +1,29 @@
...
+    def get():
+        host, port = config.es_endpoint
+        return ESClientFactory._create_client(host, port, config.es_timeout)
```
hannes-ucsc: … use `cls` to refer to other static/class methods. I can explain why on a call if you would like.

src/azul/es.py
```diff
@@ -1,25 +1,29 @@
...
+
+    @staticmethod
```
hannes-ucsc: Once you fix the logger reference, this doesn't *need* to be `@classmethod` but you might as well change for consistency's sake.

test/service/test_request_validation.py
```diff
@@ -41,7 +41,7 @@ def test_health(self):
         self.assertEqual(response.json(), expected_json)
 
     def test_health_es_unreachable(self):
-        with mock.patch.dict(os.environ, AZUL_ES_ENDPOINT='nonexisting-index.us-east-1.es.amazonaws.com:80'):
```
hannes-ucsc: :-)

PR #204, opened 2018-08-16
Solved bug of sort criteria being assigned internal definition. (fixes #169)

### LINKED ISSUE #169: Incorrect Default Sort Criteria is being Sent on /specimens response
When we hit /specimens with no sort parameter specified, the server returns a paging object that sets the default sort.

This object has the default sort as `entity_id`. entity_id is the old name and it should be the specimen id field.

```
"pagination": {
    "count": 1,
    "total": 414,
    "size": 1,
    "search_after": "ff938e1c-f587-416d-8c3d-951bc7847ecb",
    "search_after_uid": "doc#ff938e1c-f587-416d-8c3d-951bc7847ecb",
    "search_before": null,
    "search_before_uid": null,
    "pages": 414,
    "sort": "entity_id",
    "order": "desc"
  },
```

When we send sort=entity_id back to you when a facet is selected we get an error "cant sort by this field entity_id".

I have temporarily disable the front end from reading in the bad default sort until this is fixed.

Thanks!
D

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-125)
┆Project Name: azul
┆Issue Number: AZUL-125

achave11:
* Added reverse lookup object for transformer.
* Added test to confirm validity in pagin object sort.

### DIFF
--- a/src/azul/service/responseobjects/elastic_request_builder.py
+++ b/src/azul/service/responseobjects/elastic_request_builder.py
@@ -459,13 +459,15 @@ class ElasticTransformDump(object):
             filters = {"file": {}}
         filters = filters['file']
 
-        translation_dict = request_config['translation']
+        translation = request_config['translation']
+        inverse_translation = {v: k for k, v in translation.items()}
+
         for facet in filters.keys():
-            if facet not in translation_dict:
+            if facet not in translation:
                 raise BadArgumentException(f"Unable to filter by undefined facet {facet}.")
 
         facet = pagination["sort"]
-        if facet not in translation_dict:
+        if facet not in translation:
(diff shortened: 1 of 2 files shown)


test/service/test_request_validation.py
```diff
@@ -186,3 +186,26 @@ def test_mixed_multiple_filter_facet_of_piechart(self):
         response = requests.get(url)
         self.assertEqual(400, response.status_code, response.json())
         self.assertEqual(self.filter_facet_message, response.json())
+
+    def test_summary_endpoint(self):
```
achave11: @hannes-ucsc this needs to go?

PR #215, opened 2018-08-20
Fix git dirty check for version endpoint

mpanik:


src/azul/__init__.py
```diff
@@ -97,7 +97,7 @@ def domain_name(self) -> str:
     def git_status(self):
         return {
             'commit': os.environ['azul_git_commit'],
-            'dirty': bool(os.environ['azul_git_dirty']),
+            'dirty': os.environ['azul_git_dirty'] == 'True',
```
hannes-ucsc: I'd prefer this to be strict. Anything but `'True'` or `'False'` should raise an exception.

Consider adding a utility function to `azul/__init__.py` that does that so we can reuse it in other places.

mpanik:
I've added additional test to verify dirtiness

PR #225, opened 2018-08-21
Add FIFO queue between transformers and index writer

hannes-ucsc:


hschmidt:
test

hannes-ucsc:
See #364 for adding coverage in unit tests. The other big gap in test coverage is `lambdas/indexer/app.py`. #262 will improve that but it is only an integration test so it won't be run on PRs. I don't know how to cover app.py better in a unit test without elaborately mocking of SQS that is.

PR #231, opened 2018-08-22
 Set ES index based on deployment name (fixes #220)

### LINKED ISSUE #220: Derive ES index name from deployment name
https://github.com/DataBiosphere/azul/blob/2269bbec413e63ef3eba650f95e72971201f6c0c/environment#L166

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-156)
┆Project Name: azul
┆Issue Number: AZUL-156

Ben-Wu:
Set index name based on deployment name to allow for a single ES domain with separated indexes.

@hannes-ucsc Can you explain what this snippet in `request_config.json` is for and if changes need to be made here for this issue?

https://github.com/DataBiosphere/azul/blob/dbde42ea5f4913d530bb9ee2cc421eca0be9c3df/src/azul/service/config/request_config.json#L27-L35

### DIFF
--- a/environment
+++ b/environment
@@ -52,14 +52,8 @@ azul_vars=(
     # Elasticsearch operation timeout in seconds
     AZUL_ES_TIMEOUT
 
-    # The name of the index containing Specimen entities
-    AZUL_SPECIMEN_INDEX
-
-    # The name of the index containing File entities
-    AZUL_FILE_INDEX
-
-    # The name of the index containing Donor entities
-    AZUL_DONOR_INDEX
+    # The template for the name of the ES index to use
+    AZUL_ES_INDEX_NAME_TEMPLATE
 
     # The name of the bucket where TerraForm maintains its state, allowing
     # multiple developers to collaboratively use TerraForm in a single AWS
@@ -159,6 +153,7 @@ _set AZUL_SUBDOMAIN_TEMPLATE "{lambda_name}"
 _set AZUL_RESOURCE_PREFIX "azul-"
 _set AZUL_ES_INDEX "azul-$AZUL_DEPLOYMENT_STAGE"
 _set AZUL_ES_DOMAIN "azul-index-$AZUL_DEPLOYMENT_STAGE"
+_set AZUL_ES_INDEX_NAME_TEMPLATE "azul_{entity_type}_$AZUL_DEPLOYMENT_STAGE"
 _set AZUL_ES_INSTANCE_COUNT 2
 _set AZUL_ES_INSTANCE_TYPE "m4.xlarge.elasticsearch"
 _set AZUL_ES_VOLUME_SIZE 35
@@ -166,14 +161,6 @@ _set AZUL_ES_TIMEOUT 60
 _set AZUL_INDEX_WORKERS 16
 _set AZUL_DSS_WORKERS 8
 _set AZUL_SUBSCRIBE_TO_DSS 1
-
-# FIXME: the index name should contain the stage name so we can share an ES
-# instance between deployments. However, the indexer seems to be hard-wired
-# to this particular index name.
-
-_set AZUL_FILE_INDEX "browser_files_dev"
-_set AZUL_SPECIMEN_INDEX "browser_specimens_dev"
-_set AZUL_DONOR_INDEX "browser_donor_dev"
 
 # This is the same TF bucket as the one DSS uses
 _set AZUL_TERRAFORM_BACKEND_BUCKET_TEMPLATE "org-humancellatlas-dss-config-{account_id}"
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -93,6 +93,9 @@ class Config:
     def es_index(self) -> str:
         return os.environ['AZUL_ES_INDEX']
 
+    def es_index_name(self, entity_type) -> str:
+        return os.environ['AZUL_ES_INDEX_NAME_TEMPLATE'].format(entity_type=entity_type)
+
     @property
     def domain_name(self) -> str:
         return os.environ['AZUL_DOMAIN_NAME']
--- a/src/azul/base_config.py
+++ b/src/azul/base_config.py
@@ -1,7 +1,7 @@
 from abc import ABC
-import os
 from typing import Any, Iterable, Mapping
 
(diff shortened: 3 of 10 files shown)


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -151,7 +152,7 @@ def create_request(
         # Create the Search Object
         es_search = Search(
             using=es_client,
-            index=os.getenv(index, 'browser_files_dev'))
+            index=azul_config.es_index_name(index) or 'azul_files_dev')
```
hannes-ucsc: It's ok to eliminate the fallback. There is no point to it. If we wanted a fallback we would implement it in es_index_name.

Here and everywhere else.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -6,6 +6,7 @@
 
 from elasticsearch_dsl import A, Q, Search
 
+from azul import config as azul_config
```
hannes-ucsc: The convention is that this is imported under the original name `config`.

To avoid the conflict with service.config could you import that as service_config?

src/azul/base_config.py
```diff
@@ -31,4 +31,4 @@ def entities(self) -> Iterable[str]:
     def index_names(self) -> Iterable[str]:
         entities = self.entities
         environment = os.getenv("STAGE_ENVIRONMENT", "dev")
-        return ["browser_{}_{}".format(entity, environment) for entity in entities]
+        return ["azul_{}_{}".format(entity, environment) for entity in entities]
```
hannes-ucsc: We don't use STAGE_ENVIRONMENT anymore. All references to that variable need to go.

Additionally, we want config.es_index_name() to be the only place where index names are generated. Duplicating that logic anywhere else is beside the point of the PR.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -611,8 +614,8 @@ def transform_autocomplete_request(
         if not filters:
             filters = {"file": {}}
 
-        index = 'AZUL_FILE_INDEX' if entry_format == 'file' \
-            else 'AZUL_DONOR_INDEX'
+        index = 'files' if entry_format == 'file' \
```
hannes-ucsc: That backslash should go.

src/azul/transformer.py
```diff
@@ -39,7 +39,7 @@ class ElasticSearchDocument:
     def __init__(self, elastic_search_id: str, content: Document, entity_name: str, _type: str="doc") -> None:
         self.elastic_search_id = elastic_search_id
         self._content = content
-        self.index = "browser_{}_{}".format(entity_name, os.getenv("STAGE_ENVIRONMENT", "dev"))
+        self.index = "azul_{}_{}".format(entity_name, os.getenv("STAGE_ENVIRONMENT", "dev"))
```
hannes-ucsc: See above. No duplication of the name generation logic please.

environment
```diff
@@ -159,6 +153,7 @@ _set AZUL_SUBDOMAIN_TEMPLATE "{lambda_name}"
 _set AZUL_RESOURCE_PREFIX "azul-"
 _set AZUL_ES_INDEX "azul-$AZUL_DEPLOYMENT_STAGE"
 _set AZUL_ES_DOMAIN "azul-index-$AZUL_DEPLOYMENT_STAGE"
+_set AZUL_ES_INDEX_NAME "azul_{index_name}_$AZUL_DEPLOYMENT_STAGE"
```
hannes-ucsc: s/index_name/entity_type

src/azul/__init__.py
```diff
@@ -93,6 +93,9 @@ def es_volume_size(self) -> int:
     def es_index(self) -> str:
         return os.environ['AZUL_ES_INDEX']
 
+    def es_index_name(self, index_name) -> str:
```
hannes-ucsc: s/index_name/entity_type

environment
hannes-ucsc: But the values stay as is (plural, for backwards compatibility). E.g. `files`, `specimen`.

hannes-ucsc:
Re: request_config.json. I don't know. I would scan the code base for references. I know that we don't have a donor index, and probably will never add one.

Ben-Wu:
Ready for re-review.  `request_config.json` is only used in `create_autocomplete_request()` in `elastic_request_builder.py`.  I changed it to use the entity type instead of the old index name env variable.

test/indexer/test_projects.py
```diff
@@ -65,9 +65,9 @@ def test_hca_extraction(self):
 
         @eventually(5.0, 0.5)
         def _assert_number_of_files():
-            total_files = self.es_client.count(index="browser_files_dev", doc_type="doc")
+            total_files = self.es_client.count(index="azul_files_dev", doc_type="doc")
```
hannes-ucsc: `config.es_index_name('files')`

test/indexer/test_projects.py
```diff
@@ -65,9 +65,9 @@ def test_hca_extraction(self):
...
             self.assertEqual(776, total_files["count"])
-            total_specimens = self.es_client.count(index="browser_specimens_dev", doc_type="doc")
+            total_specimens = self.es_client.count(index="azul_specimens_dev", doc_type="doc")
```
hannes-ucsc: Ditto.

test/indexer/test_projects.py
```diff
@@ -86,7 +86,7 @@ def test_no_duplicate_files_in_specimen(self):
         module_logger.info("End computation %s",
                            datetime.now().isoformat(timespec='microseconds'))
         # Check values in ElasticSearch
-        results = self.es_client.get(index="browser_specimens_dev",
+        results = self.es_client.get(index="azul_specimens_dev",
```
hannes-ucsc: Ditto.

Ben-Wu:
I deployed it and it seems to be working.  `/repository/files` and `/repository/specimens` work.

@hannes-ucsc Any ideas of specific things to look out for?  I'm still learning how everything fits together.

PR #237, opened 2018-08-23
Refactor MetadataDownloader to use download_bundle_metadata 

### LINKED ISSUE #143: Replace MetadataDownloader with download_bundle_metadata from metadata-api
Use this https://github.com/HumanCellAtlas/metadata-api/blob/1b7192cecbef43b5befecc4153bf2e2f4db5bb16/src/humancellatlas/data/metadata/helpers/dss.py#L14 instead of MetadataDownloader. Azul already depends on metadata-api. The biggest obstacle here is probably going to be mocking of this functionality for the tests.

* MetadataDownloader is a class but doesn't need to be
* The retries it implements aren't necessary anymore (most are handled by DSSClient as a first line of defense, the worker retries by SQS are a second line of defense)

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-112)
┆Project Name: azul
┆Issue Number: AZUL-112

mpanik:
(resolves #143)

We should probably still keep `MetadataDownloader` as a wrapper for `metadata` library.  In case of `download_bundle_metadata` change, we won't need to update all call of this method, just this one in `MetadataDownloader`.


mpanik:
```
self.dss_client = config.dss_client(dss_endpoint=dss_host)
```
@hannes-ucsc Not sure if there is reason, but it ended up as a `config.dss_endpoint` anyway. I made it to use `config.dss_endpoint` directly

src/azul/downloader.py
```diff
@@ -10,139 +10,40 @@
...
 
-from hca.dss import SwaggerAPIException
+from humancellatlas.data.metadata.helpers.dss import download_bundle_metadata
 
 from azul import config
 
 log = logging.getLogger(__name__)
 
 
 class MetadataDownloader(object):
```
hannes-ucsc: I think this was sort of pointless even before this PR because it was instantiated and a single method was called on it before the instance went out of scope. That's smells of an anti-pattern.

Anyways, now it really is anemic and I think it should go away.

src/azul/downloader.py
```diff
@@ -10,139 +10,40 @@
...
         the metadata_files are actual contents of the metadata files and the
         data_files are the metadata describing the files.
-
-        :param request: The contents of the DSS event notification
-        :param replica: The replica to which pull the bundle from
         """
-        bundle_uuid = request['match']['bundle_uuid']
-        bundle_version = request['match']['bundle_version']
-        metadata_files, manifest = self.__get_bundle(bundle_uuid, bundle_version, replica)
-        num_workers = int(os.environ['AZUL_DSS_WORKERS'])
```
hannes-ucsc: This configuration should be retained. `download_bundle_metadata` has a `num_workers` argument that you can use. You may have to upgrade the dependency to get that argument.

src/azul/downloader.py
```diff
@@ -10,139 +10,40 @@
...
-        with ThreadPoolExecutor(num_workers) as tpe:
-            def get_metadata(item):
-                file_name, manifest_entry = item
-                return file_name, self.__get_file(manifest_entry['uuid'], manifest_entry['version'], replica)
-            metadata_files = dict(tpe.map(get_metadata, metadata_files.items()))
+        _, manifest, metadata_files = download_bundle_metadata(client=self.dss_client, replica=self.replica,
```
hannes-ucsc: Typically, we wrap all arguments or none.

src/azul/indexer.py
```diff
@@ -19,20 +19,21 @@ class BaseIndexer(ABC):
     """
     def __init__(self, properties: BaseIndexProperties) -> None:
         self.properties = properties
+        self.es_client = ESClientFactory.get()
```
hannes-ucsc: I'd prefer to avoid this attribute because `.get()` caches client instances already. To newcomers reading the code I don't want to create the impression that stashing the client as an instance attribute is somehow a efficiency gain. It is but only minuscule one and that's not worth people starting to follow this pattern all over the code base. Local variable is ok, just not an instance attribute.

src/azul/indexer.py
```diff
@@ -19,20 +19,21 @@ class BaseIndexer(ABC):
...
+        self.metadata_downloader = MetadataDownloader()
```
hannes-ucsc: This attribute would go away if you eliminated the class.

hannes-ucsc: In general we should aim to minimize state unless additional state considerably increases computational efficiency or readability.

PR #240, opened 2018-08-24
Solidify concurrent specimen test (fixes #230)

### LINKED ISSUE #230: HCA indexer tests does not assert bundle count
I introduced a bug in the bundle merging code that caused the indexer to overwrite one bundle's contribution to a specimen with another bundle's contribution instead of merging them. The resulting specimen document only had one bundle in it instead of two. The test didn't catch that. The fact that this line only ever looks at the first bundle in the document is a smoking gun:

https://github.com/DataBiosphere/azul/blob/668b1cbd95410cca9ba3ab2f2982acd03e52e953/test/indexer/test_hca_indexer.py#L178

Also, assertion is skipped if result is empty.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-160)
┆Project Name: azul
┆Issue Number: AZUL-160

alex-hancock:


hannes-ucsc:
The last commit should have started in caps.

PR #269, opened 2018-08-30
Feature/mkrause/38 bdbag to s3

mjkrause:
S3FileHandler has methods that we plan to use when we create a persistent BDBag in AWS S3.


webservice/s3_file_handler.py
```diff
@@ -0,0 +1,136 @@
...
+                ClientMethod='get_object',
+                Params={
+                    'Bucket': bucket_name,
+                    'Key': key_name
+                }
+            )
+
+            response = requests.get(url)
+            _presigned_url = response.url
+            presigned_url = _presigned_url.encode('utf-8')
```
coverbeck: Why are you encoding this? I'm not saying you shouldn't, but could you explain why?

I'm sure there are lots of exceptions, but in general I would avoid encoding until the very last minute, when sending to some other system.

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,136 @@
...
+        :rtype: JSON / Python dict
+        """
+        if self.bucket_exists(bucket_name) == 200:
+            self.bucket_name = bucket_name
+            return self.bucket.create_bucket(
+                Bucket=bucket_name,
+                CreateBucketConfiguration={'LocationConstraint': self.region}
+            )
+
+    def list_objects_in_bucket(self, bucket):
```
coverbeck: I'm not sure we need this one, but it doesn't hurt.

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        response = self.session.list_objects_v2(Bucket=bucket)
+        if response.has_key('Contents'):
+            L = [response['Contents'][x]['Key']
+                 for x in range(len(response['Contents']))]
+            return [str(r) for r in L]  # convert to utf-8 (is unicode)
+        else:
+            return []
+
+    def get_bucket_list(self):
```
coverbeck: Not sure we need this one, but it doesn't hurt.

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,136 @@
...
+        if status_code == 200:
+            url = self.bucket.generate_presigned_url(
+                ClientMethod='get_object',
+                Params={
+                    'Bucket': bucket_name,
+                    'Key': key_name
+                }
+            )
+
+            response = requests.get(url)
```
coverbeck: What happens if this fails?

mjkrause: I added a bunch of tests addressing that. One that's missing is when the generation of the presigned URL fails. I do some more research to find out what error is actually raised, but could for a simple case just assume some `ClientError`.

mjkrause: DOne with that.

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,136 @@
...
+        except ClientError as e:
+            # If a client error is thrown, then check that it was a 404 error.
+            # If it was a 404 error, then the bucket does not exist.
+            error_code = int(e.response['Error']['Code'])
+            if error_code == 403:
+                return 403  # access to bucket not authorized
+            elif error_code == 404:
+                return 404  # bucket does not exist
+
+    def create_bucket(self, bucket_name):
```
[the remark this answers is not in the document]
mjkrause: Yes, the bucket was created in the console. Maybe we need this method again once we implement automated deployment, but I take it out for now.

webservice/s3_file_handler.py
mjkrause: I left it in for now.

webservice/s3_file_handler.py
mjkrause: Same as above.

PR #277, opened 2018-08-30
Add reverse lookup of entity index

alex-hancock:
Also deprecate environment variable `AZUL_ES_INDEX` and remove unused `local-import.py`


environment
```diff
@@ -147,10 +139,8 @@ fi
 
 _set AZUL_DOMAIN_NAME "$AZUL_DEPLOYMENT_STAGE.explore.data.humancellatlas.org"  # FIXME: project-specific, is this the right place for this?
 _set AZUL_SUBDOMAIN_TEMPLATE "{lambda_name}"
-_set AZUL_RESOURCE_PREFIX "azul-"
-_set AZUL_ES_INDEX "azul-$AZUL_DEPLOYMENT_STAGE"
+_set AZUL_PREFIX "azul"
```
hannes-ucsc: I think we should maintain two independent prefixes. This part should look like

```
AZUL_RESOURCE_PREFIX="azul"
AZUL_INDEX_PREFIX="azul"
```

src/azul/__init__.py
```diff
@@ -120,6 +122,10 @@ def lambda_env(self) -> Mapping[str, str]:
             'XDG_CONFIG_HOME': '/tmp'  # The DSS CLI caches downloaded Swagger definitions there
         }
 
+    def validate_terms(self, terms, char):
```
hannes-ucsc: `_validate_terms` (protected)

src/azul/__init__.py
```diff
@@ -48,8 +48,9 @@ def num_dss_workers(self) -> int:
         return int(os.environ['AZUL_DSS_WORKERS'])
 
     def qualified_resource_name(self, resource_name):
-        prefix = os.environ['AZUL_RESOURCE_PREFIX']
-        return f"{prefix}{resource_name}-{self.deployment_stage}"
+        prefix = os.environ['AZUL_PREFIX']
+        self.validate_terms((prefix, self.deployment_stage), "-")
```
hannes-ucsc: Let's validate the deployment stage name in `.deployment_stage()`. That would simplify the signature of `validate_terms`. The error message should include the name of the offending environment variable. And let's be very strict: the term should match this regex:

`[a-z][a-z0-9]{2,14}`

test/indexer/test_hca_indexer.py
```diff
@@ -68,7 +68,7 @@ def check_bundle_correctness(es_results):
                 index_id = result_dict["_id"]
                 expected_ids = set()
                 # FIXME: https://github.com/DataBiosphere/azul/issues/273
```
hannes-ucsc: The FIXME comment should be removed

src/azul/__init__.py
```diff
@@ -120,6 +123,10 @@ def lambda_env(self) -> Mapping[str, str]:
             'XDG_CONFIG_HOME': '/tmp'  # The DSS CLI caches downloaded Swagger definitions there
         }
 
+    def _validate_term(self, env_var):
+        assert re.fullmatch("[a-z][a-z0-9]{2,14}", os.environ[env_var]) is not None, \
+               "Separator character found in {env_var}: {os.environ[env_var]}."
```
hannes-ucsc: This should be an f-string.

src/azul/__init__.py
```diff
@@ -90,7 +92,8 @@ def es_volume_size(self) -> int:
         return int(os.environ['AZUL_ES_VOLUME_SIZE'])
 
     def es_index_name(self, entity_type) -> str:
-        return os.environ['AZUL_ES_INDEX_NAME_TEMPLATE'].format(entity_type=entity_type)
+        self._validate_term('AZUL_INDEX_PREFIX')
+        return f"{os.environ['AZUL_INDEX_PREFIX']}_{entity_type}_{self.deployment_stage}"
```
hannes-ucsc: Too much logic in the interpolation.

PR #304, opened 2018-09-06
Feature/mkrause/38 test endpoint (resolves #292)

### LINKED ISSUE #292: Implement BDBag upload to S3 storage in endoint
This story is based on #152. Here we implement the functionality in the endpoint `/repository/files/export`.
This story is done if we use curl with the appropriate query parameters to (1) upload a metadata BDBag to S3, and (2) to create presigned URL to that location and return it by that route.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-195)
┆Project Name: azul
┆Issue Number: AZUL-195
┆Epic: Epic 8: “Handoff” From Portal to Workspace

mjkrause:
This PR modifies the export function of Boardwalk's webservice. In webservice.py the route `/repository/files/export` calls method `get_manifest`. That endpoint originally created a tab-separated values (TSV) file from a metadata manifest, and downloaded it to a user's local system.

In this PR we add a query parameter to it so the user can specify whether the original functionality should be executed, or whether the metadata in form of a zipped BDBag should be uploaded to Amazon AWS S3 for temporary storage. The two options can be invoked with strings `tsv` and `bdbag`, respectively, as query parameters.

A test instance of the service is running at `https://mkrause2.ucsc-cgp-dev.org`. To test functionality do

- upload metadata as BDBag to S3 and return presigned URL to it: `curl 'https://mkrause2.ucsc-cgp-dev.org/api/v1/repository/files/export?filters=%7B%7D&output_formats=bdbag' -H 'accept: application/json'`
- test to return summary of data in database: `curl 'https://mkrause2.ucsc-cgp-dev.org/api/v1/repository/files/summary?filters=%7B%7D' -H 'accept: application/json'`
- test to return a metadata file: `curl 'https://mkrause2.ucsc-cgp-dev.org/api/v1/repository/files/export?filters=%7B%7D&output_formats=tsv' -H 'accept: application/json'`


webservice/webservice.py
```diff
@@ -305,8 +306,11 @@ def get_manifest():
     """
     # Setup logging
     logger = logging.getLogger("dashboardService.webservice.get_manifest")
+    # Create query parameters.
     filters = request.args.get('filters', '{"file": {}}')
     logger.debug("Filters string is: {}".format(filters))
+    output_formats = request.args.get('output_formats', '')
```
coverbeck: Did you see my comment about the name of the parameter in the [ticket](https://ucsc-cgl.atlassian.net/browse/AZUL-152?focusedCommentId=14263&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14263)? I think you should just call this `format`.

It should not be plural, unless you process multiple formats on one call, which you do not.

Also, can you document at the top of method the query parameter and values.

webservice/webservice.py
```diff
@@ -322,7 +326,64 @@ def get_manifest():
     # Get the response back
     logger.info("Creating the API response")
     response = es_td.transform_manifest(filters=filters)
-    # Return the excel file
+    if output_formats == 'tsv':  # returns metadata as TSV file
```
coverbeck: You should default to tsv, i.e., if the parameter is not specified, you should assume it's tsv. Otherwise you will break existing code that it calling this method without the query parameter.

webservice/webservice.py
```diff
@@ -322,7 +326,64 @@ def get_manifest():
...
+        return response
+    elif output_formats == 'bdbag':  # returns presigned URL to a BDBag in S3
```
coverbeck: You need to handle the case where an invalid value is passed in.

webservice/webservice.py
```diff
@@ -322,7 +326,64 @@ def get_manifest():
...
+        response_bag = bdbag_response(response)
+        return jsonify(response_bag)
+
+
+def bdbag_response(response_obj):
+    logger = logging.getLogger('dashboardService.webservice.bdbag_response')
+    # Create and return the BDbag folder.
+    bag_name = 'manifest'
+    bag_info = {'organization': 'UCSC Genomics Institute',
+                'data_type': 'TOPMed',
```
coverbeck: Why is this hard-coded to TOPMed?

webservice/webservice.py
```diff
@@ -322,7 +326,64 @@ def get_manifest():
...
+                     bag_info=bag_info,
+                     bag_name=bag_name)
+    zipped_bag = bag.create_bag()  # path to compressed bag
+    logger.info("Creating a compressed BDbag containing manifest.")
+
+    # Import bucket environment variable, launch instance of S3-file handler,
+    # and upload the BDBag file to S3.
+
+    # Transfer parameters.
+    aws_region = 'us-west-2'
```
coverbeck: Jesse added a parameter for the AWS region. You should use that instead of hard-coding it.

webservice/webservice.py
```diff
@@ -322,7 +326,64 @@ def get_manifest():
...
+        logger.info("Uploaded BDbag {} to S3 bucket {}.".format(bucket_key,
+                                                                azul_s3_bucket))
+        # Generate presigned URL to access that s3 location.
+        response = s3.create_presigned_url(azul_s3_bucket, bucket_key)
+        response['s3_upload_status_code'] = r['status_code']
+        if response['status_code'] == 200:
+            logger.info("Successfully created presigned URL for bucket {}.".
+                        format(''.join([azul_s3_bucket, '/', bucket_key])))
+        else:
+            logger.debug("Failed to create presigned URL for bucket {}.".
```
coverbeck: This should be `logger.error(...`

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,134 @@
...
+                        'Key': key_name
+                    }
+                )
+            except ClientError as e:
+                result['status_code'] = \
+                    {'status_code': e.response['Error']['Code']}
+                return result
+
+            response = requests.get(url)
+            if response.status_code == 200:
```
coverbeck: Please see my comment in [the ticket](https://ucsc-cgl.atlassian.net/browse/AZUL-152?focusedCommentId=14139&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14139) (formatting doesn't look right in the GitHub issue) about the exact format of response.

This is why I asked you in [this comment](https://ucsc-cgl.atlassian.net/browse/AZUL-152?focusedCommentId=14263&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14263) to to summarize the query parameter and the response body format, so that we could agree on it and you could implement exactly that.

webservice/webservice.py
```diff
@@ -322,7 +326,64 @@ def get_manifest():
...
+                        format(''.join([azul_s3_bucket, '/', bucket_key])))
+    else:
+        logger.error("Upload to S3 bucket {} failed: {}".format(
+            azul_s3_bucket, r['status_code']))
+        response = {
+            'presigned_url': None,
```
coverbeck: If there is an error, your HTTP response status code needs to indicate the error. If I'm reading this correctly, you will return a 200 status code with the error status code in the body. If that is the case, please return a 4xx or 5xx as appropriate.

I don't think you should have presigned_url in the response body. The response of an error does not have to match the response body of success.

See these links for some ideas: 

https://medium.com/@suhas_chatekar/return-well-formed-error-responses-from-your-rest-apis-956b5275948

https://apigee.com/about/blog/technology/restful-api-design-what-about-errors (although don't do it the Facebook way from this link).

webservice/webservice.py
mjkrause: This is a left-over from the past. I left them empty now. Perhaps we should give the user the option to fill those before the bag is created in the UI?

webservice/webservice.py
mjkrause: Ah, that's great. I meant to add it because it's needed here. Glad it's done already.

webservice/webservice.py
mjkrause: done

webservice/webservice.py
mjkrause: Agreed, I renamed it to `format`.

webservice/s3_file_handler.py
mjkrause: If `status_code` isn't 200, then it returns the HTTP error status code, and both, `presigned_url` and `s3_upload_status_code` have value `null`.

coverbeck: A successful response should match the Fence response that I linked to in JIRA.

mjkrause: My design rested on the notion to return the same JSON, no matter what status code. I changed the happy path to Fence's successful response:
`{url: <url>}
`
so no status code (200 is implied I suppose). 
But upon encountering an error don't we want to return a more detailed JSON, providing the error? If so, how much do we need to go into detail about the error? According to the Apigee article a lot, and I will try that, but hope I'm not overdoing it. I have so far included small messages in the body, but no formal error codes or status code explanations. Let's discuss the desired level of detail. Couldn't figure out much how Fence handles this so far.

webservice/webservice.py
mjkrause: Pl. see my response to the above comment.

mjkrause: Also, some unit testing with mocking for the webservice would really help...

webservice/s3_file_handler.py
mjkrause: BTW, I needed to change the URL of the Boardwalk instance. Use 
`curl 'https://mkrause3.ucsc-cgp-dev.org/api/v1/repository/files/export?filters=%7B%7D&format=bdbag' -H 'accept: application/json'`
to see the response.

coverbeck: The HTTP response code itself has has to indicate success or failure. A user should not have to look at the response body to determine whether there was success of failure. Maybe you're saying that and doing that, but to me it looked like, if there was an error, you would return a status code of 200, with a 4xx or a 5xx in the response body. If I'm misreading that, and the status code of the response is 4xx or 5xx, then you are good.

Regarding the error response:

1. I would not spend a lot of time dealing with it, something close to what you have now is fine, although I don't think it helps to have an empty presigned url property.
2. I think the error response format should be generic, so that it could be used for errors for any API. That way clients can write generic error handlers, and not have to customize the error handling for each different API.

coverbeck: BTW, here's how Fence handles a 4xx: https://github.com/uc-cdis/fence/blob/master/fence/errors.py#L8 and 
https://github.com/uc-cdis/fence/blob/master/fence/error_handler.py#L32

They are returning an HTML response; I think it's plenty to just return a message as they have in errors.py

mjkrause: I've modified all responses to client. The response does not contain any empty strings anymore. The happy path just contains the URL. In case of an error it returns the status code, the error code (if available) and a short message.
I also added a variable that sets the expiration time of the presigned URL, with a default value of 120 s to contain misuse potential.

webservice/requirements.txt
```diff
@@ -1,48 +1,66 @@
...
-elasticsearch-dsl>=5.0.0,<6.0.0
+elasticsearch-dsl==5.4.0
 enum34==1.1.6
+flake8==3.5.0
 Flask==0.12.2
 Flask-APScheduler==1.5.0
 Flask-Cors==3.0.2
 Flask-Elasticsearch==0.2.5
 Flask-Excel==0.0.5
+Flask-Login==0.4.0
```
coverbeck: Why did the case of this change? It was `flask-login` before. Is the requirements.txt not case-sensitive (I don't know)?

webservice/requirements.txt
```diff
@@ -51,10 +69,13 @@ python-dateutil==2.6.0
 python-editor==1.0.3
 pytz==2016.10
 requests==2.19.1
+s3transfer==0.1.13
 singledispatch==3.4.0.3
 six==1.10.0
 SQLAlchemy==1.1.5
+SQLAlchemy-Utils==0.33.3
```
coverbeck: I think Jesse ended up removing SQLAlchemy, and you have now reintroduced a bunch of them. I suspect it's because you don't have his changes... Also, it seems like this file changed a lot in general. Is that because you did a `pip freeze`? Fine if you did, just trying to understand.

I think you should rebase so that you get Jesse's latest changes, then add the libraries you explicitly need, then run `pip freeze`.

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,149 @@
...
+                    ClientMethod='get_object',
+                    Params={
+                        'Bucket': bucket_name,
+                        'Key': key_name
+                    },
+                    ExpiresIn=expiration_time
+                )
+            except ClientError as e:
+                result['status_code'] = \
+                    e.response['ResponseMetadata']['HTTPStatusCode']
```
coverbeck: Will ResponseMetaData always exist? I'm asking, I don't know.

webservice/s3_file_handler.py
```diff
@@ -0,0 +1,149 @@
...
+                result['error_code'] = e.response['Error']['Code']
+                result['msg'] = e.response['Error']['Message']
```
coverbeck: Will `Error` always exist? I'm asking; I don't know.

webservice/webservice.py
```diff
@@ -305,8 +306,14 @@ def get_manifest():
     """
     # Setup logging
     logger = logging.getLogger("dashboardService.webservice.get_manifest")
+    # Create query parameters.
     filters = request.args.get('filters', '{"file": {}}')
     logger.debug("Filters string is: {}".format(filters))
+
+    # NOTE: "format" here is a HTTP query parameter, not a Python command.
+    if not format:
```
coverbeck: Because `format` has a default value on line 296, this line will never be true, will it?

webservice/webservice.py
```diff
@@ -292,7 +293,7 @@ def get_order():
 
 
 @webservicebp.route('/repository/files/export', methods=['GET'])
-def get_manifest():
+def get_manifest(format='tsv'):
     """
     Creates and returns a manifest based on the filters pased on
```
coverbeck: In a previous comment, I had asked that you document the format parameter up here. Please list it, the possible values, and the default, so somebody doesn't have to go through the code.

webservice/webservice.py
```diff
@@ -322,7 +329,79 @@ def get_manifest():
...
+    elif format == 'bdbag':  # returns presigned URL to a BDBag in S3
+        response_bag = bdbag_response(response)
+        return jsonify(response_bag)
+
+
+def bdbag_response(response_obj):
+    """
+    Create a response for BDBag upload to S3. Returns just the presigned URL of
+    the S3 location if all goes well. Otherwise the HTTP status code. This that
+    can go wrong include
```
coverbeck: It looks like you didn't finish your thought here.

webservice/webservice.py
```diff
@@ -322,7 +329,79 @@ def get_manifest():
...
+                               'but could not generate presigned URL.'}
+    else:
+        logger.error("Upload to S3 bucket {} failed: {}".format(
+            azul_s3_bucket, r['status_code']))
+        response = {
+            'msg': 'BDBag upload to S3 bucket {} failed.'
+                .format(azul_s3_bucket),
+            'status': r['status_code']
+        }
     return response
```
coverbeck: You are not setting the status code of the response, which means it will always return 200 by default.

You need something like `return response, response.status` (not verified, but just to give you an idea of what I'm talking about).

webservice/webservice.py
```diff
@@ -322,7 +329,79 @@ def get_manifest():
...
+                     bag_info=bag_info,
+                     bag_name=bag_name)
+    zipped_bag = bag.create_bag()  # path to compressed bag
+    logger.info("Creating a compressed BDbag containing manifest.")
+
+    # Import bucket environment variable, launch instance of S3-file handler,
+    # and upload the BDBag file to S3.
+
+    # Transfer parameters.
+    aws_region = 'us-west-2'
```
coverbeck: You are not using the parameter. This should not be hard-coded.

webservice/requirements.txt
mjkrause: This was likely re-written before this commit: https://github.com/DataBiosphere/azul/pull/304/commits/2f8bd791aeed41298f365e67cbf29e2534be4602, where I installed needed packages. 

That likely also changed the order of the list. Perhaps `requirements.txt` was hand-edited before?

It seems that installation using pip is case-insensitive (https://stackoverflow.com/questions/26503509/is-pypi-case-sensitive)

webservice/requirements.txt
mjkrause: Yes, my branch was before Jesse's merge to feature/commons. I suggest we rebase once this PR is approved.

webservice/s3_file_handler.py
mjkrause: Good catch. I don't think we can rely on it. Check out https://github.com/boto/botocore/blob/develop/botocore/exceptions.py#L390. I can check for existence before.

webservice/s3_file_handler.py
mjkrause: Looks like `Error` is always a key in the `ClientError` dictionary. See https://github.com/boto/botocore/blob/develop/botocore/exceptions.py#L377

webservice/requirements.txt
coverbeck: I think you should do it to get the PR approved, as the dependencies are part of the PR approval.

webservice/webservice.py
mjkrause: Fixed in https://github.com/DataBiosphere/azul/pull/304/commits/c26408249d016cd4a89853a439eb64059020d072.

webservice/webservice.py
mjkrause: Not sure how to fix this. In my instance, AWS_REGION isn't an environment variable. I suggest we complete this PR, merge with feature/commons, then read out the region here, and redeploy the instance.

coverbeck: This is still not fixed.

webservice/webservice.py
```diff
@@ -322,7 +339,84 @@ def get_manifest():
...
+    bucket_key = str(uuid.uuid4())
+    azul_s3_bucket = os.getenv("AZUL_S3_BUCKET")
+    access_key_id = os.getenv("AWS_ACCESS_KEY_ID")
+    secret_key = os.getenv("AWS_SECRET_ACCESS_KEY")
+    url_expiration_time = 120  # in seconds
```
coverbeck: You should try to get this from a variable, and default to whatever Fence defaults to (which may be 120). Do it the same way I suggested for the AWS region. You don't have to add the variable to cgp-deployment, but it should be there just in case we ever need to change the value. That way we wouldn't need to do a commit, a PR, a code review, do a Docker build on Quay, redeploy just to change the value from x to y.

webservice/s3_file_handler.py
mjkrause: Addressed in https://github.com/DataBiosphere/azul/pull/304/commits/2e3f7f8f5936a9537f4d96b8a7882d35560119ca.

webservice/requirements.txt
coverbeck: Resolving this; this was my bad, as SQLAlchemy had been removed from cgp-dashboard, not this repo.

webservice/webservice.py
```diff
@@ -291,22 +292,38 @@ def get_order():
...
     :return: A manifest that the user can use to download the files in there
     """
     # Setup logging
     logger = logging.getLogger("dashboardService.webservice.get_manifest")
+    # Create query parameters.
     filters = request.args.get('filters', '{"file": {}}')
     logger.debug("Filters string is: {}".format(filters))
+
+    # NOTE: "format" here is a HTTP query parameter, not a Python command.
+    format = request.args.get('format', '')
```
coverbeck: You could just do `format = request.args.get('format', 'tsv')`, then get rid of the next two lines

webservice/webservice.py
```diff
@@ -322,7 +339,85 @@ def get_manifest():
...
-    # Return the excel file
+    if format == 'tsv':  # returns metadata as TSV file
+        return response
+    elif format == 'bdbag':  # returns presigned URL to a BDBag in S3
+        response_bag = bdbag_response(response)
+        if 'status' in response_bag:
+            http_status = response_bag['status']
+        else:
+            http_status = 200
+        return jsonify(response_bag), http_status
```
coverbeck: You do not handle the case where an invalid format is specified. You need another else, and you should return a status of 400 (bad request). Probably `return '', 400`.

webservice/webservice.py
```diff
@@ -322,7 +339,85 @@ def get_manifest():
...
+                     bag_info=bag_info,
+                     bag_name=bag_name)
+    zipped_bag = bag.create_bag()  # path to compressed bag
+    logger.info("Creating a compressed BDbag containing manifest.")
+
+    # Import bucket environment variable, launch instance of S3-file handler,
+    # and upload the BDBag file to S3.
+
+    # Transfer parameters.
+    azul_s3_aws_region = os.getenv("AZUL_S3_AWS_REGION", 'us-west-2')
```
coverbeck: This is minor, and I don't feel too strongly if you leave it, but in Python, aren't you supposed to use either double quotes or single quotes for strings? Here and throughout the file there is a mixture, some of which may proceed your code...

webservice/webservice.py
```diff
@@ -322,7 +339,85 @@ def get_manifest():
...
+    azul_presigned_url_expiration = \
+        os.getenv("AZUL_PRESIGNED_URL_EXPIRATION", 3600)  # in seconds
+    bucket_key = str(uuid.uuid4())
+    azul_s3_bucket = os.getenv("AZUL_S3_BUCKET")
```
coverbeck: It might be best to have default values for this and the next two. Not that the default values for these will work, like the previous two, but even so, the failures might be more obvious than if you have None as the values

webservice/webservice.py
coverbeck: I was looking trying to write the UI code, and for some reason it is already setting the format to `tarball`. I don't know where that comes from. In any case, to handle that and the case my comment describes, you need to do one of two things:

1. Treat everything that is not `bdbag` as tsv, e.g., make line 342 be `if format != 'bdbag'`
2. Add the test for `tarball`, e.g., change line 342 to be `if format == 'tsv' or format == 'tarball'` (then you still need a final else).

Whichever you pick, you should update the doc at the top accordingly.

webservice/webservice.py
mjkrause: Good call. Done.

webservice/webservice.py
mjkrause: I noticed that. I like the approach described here:https://stackoverflow.com/questions/56011/single-quotes-vs-double-quotes-in-python, but I, too, don't feel too strongly about it. At least there's some consistency about it.
I abided by that style in the two methods touched in this PR.

webservice/webservice.py
mjkrause: Done.

webservice/webservice.py
```diff
@@ -291,22 +292,37 @@ def get_order():
...
         - name: filters
           in: query
           type: string
           description: Filters to be applied when generating the manifest
+          
+        - name: format
+          in: query
+          type: string
+          values: 'tsv' (default), 'bdbag'
+          description: Output format. Either TSV file or a compressed BDBag
```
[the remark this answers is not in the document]
mjkrause: Sure, I noticed that I forgot that and was about to put it in when I was pulled away...

PR #308, opened 2018-09-06
Set commonmark dependency version to 0.7.5 (fixes #306)

### LINKED ISSUE #306: CommonMark dependency is breaking builds and deployments
We are using `hca==4.1.0` (https://github.com/HumanCellAtlas/dcp-cli) which is using dependency `commonmark >= 0.7.4, < 1`.

The problem is the commonmark package was broken in 0.8.0 and isn't fixed yet so when install it via pip, the module will not be found.  This is causing deployments and travis builds to fail.

example: https://travis-ci.org/DataBiosphere/azul/builds/425407139

Our options are:
1. use `hca==4.3.3` which has `commonmark >= 0.7.4, < 0.8`.  I haven't checked if there are any issues with upgrading.

2. explicitly set `commonmark==0.7.5` which is easy but azul itself doesn't use commonmark so it's strange to have it in it's `requirements.txt`

3. wait until it's fixed let builds fail

I think we should go with 2 until commonmark is fixed and then we can take it out of `requirements.txt`

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-203)
┆Project Name: azul
┆Issue Number: AZUL-203

Ben-Wu:
Follow-up issue: https://github.com/DataBiosphere/azul/issues/309

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ aws-requests-auth==0.4.2
 boto==2.48.0
 boto3==1.7.48
 botocore==1.10.48
+commonmark==0.7.5
 dataclasses<=0.6
 elasticsearch>=5.0.0,<6.0.0
 elasticsearch-dsl>=5.0.0,<6.0.0


requirements.txt
```diff
@@ -2,6 +2,7 @@ aws-requests-auth==0.4.2
 boto==2.48.0
 boto3==1.7.48
 botocore==1.10.48
+commonmark==0.7.5
```
hannes-ucsc: Thank you for figuring this out, @Ben-Wu. The only thing I would have done differently is to add a comment above this line pointing at the GH issue that will make this workaround obsolete.

PR #314, opened 2018-09-11
Project Tab Indexer r4

shiroyuki:
This is to update the code to work with the latest HumanCellAtlas/metadata-api#20 which is based on [v9.0.0 Project Schema](http://schema.staging.data.humancellatlas.org/type/project/9.0.0/project).

The changes will map, extract and index the data with project contact and publication.

### DIFF
--- a/src/azul/project/hca/transformers.py
+++ b/src/azul/project/hca/transformers.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 import logging
-from typing import Any, List, Mapping, MutableMapping, Sequence, Dict, Iterable
+from typing import Any, List, Mapping, MutableMapping, Sequence, Set, Iterable
 
 from humancellatlas.data import metadata as api
 from humancellatlas.data.metadata import AgeRange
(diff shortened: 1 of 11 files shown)


src/azul/project/hca/transformers.py
```diff
@@ -14,18 +14,54 @@
...
 def _project_dict(bundle: api.Bundle) -> dict:
     project: api.Project
+    laboratories: List[str]
+    contributors: List[str]
+
     project, *additional_projects = bundle.projects.values()
     reject(additional_projects, "Azul can currently only handle a single project per bundle")
+
+    laboratories = set()
+    contributors = set()
```
Ben-Wu: I think we should call this institutions because contributors is ambiguous especially when `project.contributors` is something else.

src/azul/project/hca/transformers.py
```diff
@@ -14,18 +14,54 @@
...
+
+    for contributor in project.contributors:
+        if contributor.laboratory:
+            laboratories.add(contributor.laboratory)
+
+        if contributor.contact_name:
+            contributors.add(contributor.institution)
+
     return {
```
Ben-Wu: According to the spec it looks like we should also have project description and a list of contact names

src/azul/project/hca/transformers.py
```diff
@@ -14,18 +14,54 @@
...
         'project_title': project.project_title,
-        'project_shortname': project.project_shortname,
-        'laboratory': sorted(list(project.laboratory_names)),
-        'contributors': sorted(list(project.contributors)),
+        'project_shortname': project.project_short_name,
+        'laboratory': sorted(laboratories),
+        'contributors': sorted(contributors),
         'document_id': project.document_id,
+        'contact': _project_contact_dict(project.contributors),
+        'publications': sorted(project.publications),
```
Ben-Wu: Does publication get stored properly? Do you have to turn it into a dict?

shiroyuki: For the record, yes, it is.

src/azul/project/hca/transformers.py
shiroyuki: According to the conversation with Michael yesterday, this will become a list of `ProjectContact` (HumanCellAtlas/metadata-api).

PR #320, opened 2018-09-14
Feature project tab indexer r5

shiroyuki:
Add additional dynamic mappings for `project.contributors` and `project.publications`.


src/azul/indexer.py
```diff
@@ -33,12 +33,9 @@ def index(self, dss_notification: Mapping[str, Any]) -> None:
                                                                version=bundle_version,
                                                                num_workers=config.num_dss_workers)
 
+        self.define_indexes()
```
Ben-Wu: Could you pass `es_client` into this function?  It seems unnecessary to do `ESClientFactory.get()` in `define_indexes`

src/azul/indexer.py
```diff
@@ -147,3 +144,13 @@ def on_error(doc_id, error):
             log.warning('%s.%s: conflicts: %r', bundle_uuid, bundle_version, conflict_documents)
             raise RuntimeError('%s.%s: Failed to index bundle. Failures: %i, conflicts: %i.' %
                                (bundle_uuid, bundle_version, len(errored_documents), len(conflict_documents)))
+
+    def define_indexes(self):
```
Ben-Wu: Why do we need to make this into a function? Do we need a separated `es_client`?

PR #321, opened 2018-09-14
Add new projects facets to web service responses (fixes #300)

### LINKED ISSUE #300: Add new projects facets to web service responses
Contributors and project title were added to projects.  They should be in the service response.

Add facets to `request_config.json` and `make_projects` in `hca_response_v5.py`

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-200)
┆Project Name: azul
┆Issue Number: AZUL-200

Ben-Wu:
Added `project_title`, `project_description`, `contributors`, and `publications`

### DIFF
--- a/src/azul/service/responseobjects/hca_response_v5.py
+++ b/src/azul/service/responseobjects/hca_response_v5.py
@@ -18,6 +18,7 @@ from jsonobject import (DictProperty,
                         StringProperty)
 
 from azul.service.responseobjects.utilities import json_pp
+from azul.json_freeze import freeze, thaw
 
 module_logger = logging.getLogger("dashboardService.elastic_request_builder")
 
@@ -331,7 +332,11 @@ class KeywordSearchResponse(AbstractResponse, EntryFetcher):
                 merged_dict[key] = list(merged_dict[key] + [value])
             elif isinstance(value, list):
                 cleaned_list = list(filter(None, chain(value, merged_dict[key])))
-                merged_dict[key] = list(set(cleaned_list))
+                if len(cleaned_list) > 0 and isinstance(cleaned_list[0], dict):  # make dicts hashable
+                    hashable_dicts = [freeze(d) for d in cleaned_list]
+                    merged_dict[key] = [thaw(d) for d in list(set(hashable_dicts))]
+                else:
+                    merged_dict[key] = list(set(cleaned_list))
             elif value is None:
                 merged_dict[key] = []
         merged_dict[identifier] = dict_id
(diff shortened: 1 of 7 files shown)


src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -331,7 +332,11 @@ def _merge(self, dict_1, dict_2, identifier):
                 merged_dict[key] = list(merged_dict[key] + [value])
             elif isinstance(value, list):
                 cleaned_list = list(filter(None, chain(value, merged_dict[key])))
-                merged_dict[key] = list(set(cleaned_list))
+                if len(cleaned_list) > 0 and isinstance(cleaned_list[0], dict):  # make dicts hashable
+                    cleaned_list = [freeze(d) for d in cleaned_list]
```
shiroyuki: While it is harmless in this context at this point, reassigning a variable with a certain data structure to a different data structure could be prone to bugs. I would not recommend implementing this pattern.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -331,7 +332,11 @@ def _merge(self, dict_1, dict_2, identifier):
...
+                    merged_dict[key] = [thaw(d) for d in list(set(cleaned_list))]
```
shiroyuki: I hope there must be the better way to produce a unique list of dictionary objects. This would cause Python to create new objects and therefore additional memory consumption.

This is a low priority but I wonder if you can write a helper method to produce a list of unique `dict` objects by iterating object instead of using `freeze` and `thaw`.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -340,6 +345,13 @@ def _merge(self, dict_1, dict_2, identifier):
     def return_response(self):
         return self.apiResponse
 
+    @staticmethod
+    def to_camel_case(s):
```
shiroyuki: We could declare this as `def to_camel_case(s : str)`. Also, `s` is not descriptive. Something generic like `text` also works in this case.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -340,6 +345,13 @@ def _merge(self, dict_1, dict_2, identifier):
...
+        if not s:
```
shiroyuki: Is there a case that `s` can be evaluated as `False`? If not, I don't think this defensive code is necessary here. Or, if you are worry about the possible empty `s`, I would go with `assert s` instead.

PR #333, opened 2018-09-20
Upgrade to nginx 1.15

coverbeck:
Also removed Consonance, which is not being used.

Needed to explicitly create man1 and man7 directories, or
Postgres installation would fail. See
https://github.com/debuerreotype/debuerreotype/issues/10 for details.


coverbeck:
https://charles.ucsc-cgp-dev.org

PR #336, opened 2018-09-21
 Add auth decorator (resolves #291)

### LINKED ISSUE #291: Implementation - Azul webservice authorization implementation
This is a place holder for the work that will be necessary but that is not yet described.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-194)
┆Project Name: azul
┆Issue Number: AZUL-194
┆Epic: Google authn for Boardwalk

jessebrennan:
This auth decorator passes the session and any authorization header on to the authorization request check to dashboard. (also some headers that ensure the request will make it through session protection).

If dashboard give the OK, then the decorated calls are allowed to run.

### DIFF
--- a/webservice/webservice.py
+++ b/webservice/webservice.py
@@ -12,8 +12,10 @@ from responseobjects.utilities import json_pp
 from bagitutils import BagHandler
 from s3_file_handler import S3FileHandler
 
+import functools
 
 # Setting up logging
+
 base_path = os.path.dirname(os.path.abspath(__file__))
 logging.config.fileConfig('{}/config/logging.conf'.format(base_path))
 bp_logger = logging.getLogger("dashboardService.webservice")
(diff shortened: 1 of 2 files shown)


webservice/webservice.py
```diff
@@ -25,9 +27,33 @@
...
+        # set the X-Forwarded-For and User-agent to prevent session protection on dashboard from causing problems
+        headers['X-Forwarded-For'] = request.headers.get('X-Forwarded-For', request.remote_addr)
+        if 'User-Agent' in request.headers:
+            headers['User-Agent'] = request.headers['User-Agent']
+        # make call to dashboard
+        res = requests.get(request.url_root + 'authorization', headers=headers)
+        if res.status_code == 204:
+            return f(*args, **kwargs)
+        else:
+            return 'Error with authorization: {}'.format(res.text), res.status_code
```
coverbeck: I don't think you should provide more details and just return a plain 403 or 401, assuming that's all that ever comes back.

webservice/webservice.py
```diff
@@ -25,9 +27,33 @@
...
+        # add cookie to request for auth
+        if 'Cookie' in request.headers:
+            headers['Cookie'] = request.headers['Cookie']
+        # set the X-Forwarded-For and User-agent to prevent session protection on dashboard from causing problems
+        headers['X-Forwarded-For'] = request.headers.get('X-Forwarded-For', request.remote_addr)
+        if 'User-Agent' in request.headers:
+            headers['User-Agent'] = request.headers['User-Agent']
+        # make call to dashboard
+        res = requests.get(request.url_root + 'authorization', headers=headers)
+        if res.status_code == 204:
```
coverbeck: Optional: Although endpoint returns 204, it could be safer to check for 2xx. Sort of torn on this one... Your call.

webservice/webservice.py
jessebrennan: Ok will do!

webservice/webservice.py
jessebrennan: I'm going to leave it unless you convince me otherwise. I have no problem changing though.

coverbeck: Sounds good, if I can't can come up with a convincing argument, then leave it as is. :)

jessebrennan:
@coverbeck this is ready for a re-review. Also if you're wondering about the __init__.py file I added, it was so that I could get testing to run properly. Unit tests are added.

PR #360, opened 2018-09-28
Config signature for v4 for encryption

jessebrennan:
The default signature version would not allow presigned urls for
SSE-KMS encrypted objects in s3 to be generated.
Updating fo s3v4 resolves the problem.


jessebrennan:
Just tested and this works on my local deployment.

PR #374, opened 2018-10-03
Redirect clients to download content from S3

### LINKED ISSUE #356: Download manifest crashes in `dev` for certain (large?) searches
On https://dev.explore.data.humancellatlas.org/, (which points to https://service.integration.explore.data.humancellatlas.org for the back end)

1. Select Organ: Brain and File Format: Bam
1. Select "Download Manifest"
1. Select "Download Manifest"
1. the download fails with `{"message": "Internal server error"}`

The link in the browser window after the fail is listed below.

https://service.integration.explore.data.humancellatlas.org/repository/files/export?filters=%7B%22file%22:%7B%22organ%22:%7B%22is%22:%5B%22brain%22%5D%7D,%22fileFormat%22:%7B%22is%22:%5B%22bam%22%5D%7D%7D%7D&format=tarball

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-230)
┆Project Name: azul
┆Issue Number: AZUL-230

shiroyuki:
(Resolves #356)

In order to resolve the issue around the limit on the size of the response payload for AWS Lambda (6 MB at the time of implementation) and ensure that the response is achievable within 30 seconds, this updates makes the manifest API utilize a private S3 bucket as an intermediate storage for downloadable content and redirect the client to that downloadable content on S3 via an associated pre-signed URL with HTTP 302.

#### Potential Known Issues

Iterating through large result from ElasticSearch may affect how much data we can write to S3 before the response time exceeds 30 seconds.

#### New Environment Variables

| Environment Variable | Description | Required | Default Value |
| --- | --- | --- | --- |
| `AZUL_S3_BUCKET` | The name of S3 bucket used by Azul | No | `azul-storage-$AZUL_DEPLOYMENT_STAGE` |

#### Permission Updates

The web service lambda function `azul-service-*` will have `s3::GetObject` and `s3::PutObject` on the corresponding S3 bucket (as specified in `AZUL_S3_BUCKET`).

#### Infrastructure Updates

The Terraform script has been updated with provisioning a new S3 bucket as specified in `AZUL_S3_BUCKET`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ install:
   - pip install coveralls
 
 script:
+  # Travis VM images contain a /etc/boto.cfg which ultimately causes moto to look for the Python bindings for GCE.
+  # Setting BOTO_CONFIG seems to override that.
   - export BOTO_CONFIG=/dev/null
   - make test
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - pip install coveralls
 
 script:
+  - export BOTO_CONFIG=/dev/null
   - make test
 
 after_success:
--- a/lambdas/service/lambda-policy.json.template.py
+++ b/lambdas/service/lambda-policy.json.template.py
@@ -32,6 +32,14 @@ emit({
                 "es:DescribeElasticsearchDomain"
             ],
             "Resource": f"arn:aws:es:{aws.region_name}:{aws.account}:domain/{config.es_domain}"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject",
+                "s3:GetObject"
+            ],
+            "Resource": f"arn:aws:s3:::{config.s3_bucket}/*"
         }
     ]
 })
--- a/src/azul/service/responseobjects/storage_service.py
+++ b/src/azul/service/responseobjects/storage_service.py
@@ -1,0 +1,50 @@
+from logging import getLogger
+from typing import Optional
+import boto3
+from azul import config
+
+logger = getLogger(__name__)
+
+
+class StorageService:
+
+    def __init__(self):
+        self.__bucket_name = config.s3_bucket
+        self.__client = None  # the default client will be assigned later to allow patching.
+
+    @property
+    def client(self):
+        if not self.__client:
+            self.__client = boto3.client('s3')
+        return self.__client
+
+    def set_client(self, client):
+        self.__client = client
+
+    def get(self, object_key: str):
+        try:
+            return self.client.get_object(Bucket=self.__bucket_name, Key=object_key)['Body'].read().decode()
+        except self.client.exceptions.NoSuchKey:
+            raise GetObjectError(object_key)
+
+    def put(self, object_key: str, data: bytes, content_type: Optional[str] = None) -> str:
+        params = {'Bucket': self.__bucket_name, 'Key': object_key, 'Body': data}
+
+        if content_type:
+            params['ContentType'] = content_type
+
+        self.client.put_object(**params)
+
+        return object_key
+
+    def get_presigned_url(self, key: str) -> str:
+        return self.client.generate_presigned_url(ClientMethod='get_object',
+                                                  Params=dict(Bucket=self.__bucket_name, Key=key))
+
+    def create_bucket(self, bucket_name: str = None):
+        self.client.create_bucket(Bucket=(bucket_name or self.__bucket_name))
+        logger.warning(f'{type(self).__name__}: Created a bucket called "{bucket_name or self.__bucket_name}"')
+
+
+class GetObjectError(RuntimeError):
+    pass
--- a/environment
+++ b/environment
@@ -102,6 +102,10 @@ azul_vars=(
     # like the number of shards in Elasticsearch.
     #
     AZUL_INDEXER_CONCURRENCY
+
+    # The name of the S3 bucket where the manifest API stores the downloadable
+    # content requested by client.
+    AZUL_S3_BUCKET
 )
 
 unset ${azul_vars[*]}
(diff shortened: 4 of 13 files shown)


hannes-ucsc:
@shiroyuki have you considered [moto](https://github.com/spulec/moto) for the tests?

shiroyuki:
@hannes-ucsc No, I have not. Would you suggest to try to use `moto` instead?

src/azul/service/responseobjects/storage_service.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self.__s3.put_object(**params)
+
+        return object_key
+
+    def delete(self, object_key: str):
+        self.__s3.delete_object(Bucket=self.__bucket_name, Key=object_key)
+
+    def get_presigned_url(self, key: str) -> str:
+        return self.__s3.generate_presigned_url(ClientMethod='get_object',
```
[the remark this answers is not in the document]
shiroyuki: @hannes-ucsc, `terraform/s3.tf.json.template.py` already set the expiration of any objects with that prefix to 1 day. Do you want to enforce the expiration on any pre-signed URL too?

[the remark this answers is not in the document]
hannes-ucsc: The URL should have an explicit, configurable expiration. The S3 object should expire after twice that amount to ensure that any ongoing HTTP connections aren't negatively affected by the object being deleted. It's a heuristic, so feel free to come up with a better one. I just don't want to use the default URL expiration (1h) here and not tie the object expiration to it.

src/azul/deployment.py
```diff
@@ -6,38 +6,38 @@
 import botocore.session
 
 
-def memoized_property(f):
+def memorized_property(f):
```
Ben-Wu: I don't think this is a typo.  I think it means memoization as in caching

hannes-ucsc: Yup.

lambdas/service/lambda-policy.json.template.py
```diff
@@ -32,6 +32,14 @@
                 "es:DescribeElasticsearchDomain"
             ],
             "Resource": f"arn:aws:es:{aws.region_name}:{aws.account}:domain/{config.es_domain}"
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject",
+                "s3:GetObject"
```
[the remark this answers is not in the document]
shiroyuki: I added this initially for testing. With Moto (or even Minio for local development), this is not needed when the service works with AWS S3. I'll remove this.

requirements.txt
```diff
@@ -5,6 +5,7 @@ botocore==1.10.48
 dataclasses<=0.6
 elasticsearch>=5.0.0,<6.0.0
 elasticsearch-dsl>=5.0.0,<6.0.0
+google-compute-engine
```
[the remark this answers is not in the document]
shiroyuki: Somehow the test breaks when we start using **moto**. I added this to dissolve the failing tests but I can remove this now and investigate further on the root cause.

shiroyuki:
@hannes-ucsc Can you review this again?

shiroyuki:
I forgot to mention this. The backend service will need `s3:GetObject` in order to get the pre-signed URL as the service has to set `ClientMethod` to `get_object`.

```python
url = s3.generate_presigned_url(
    ClientMethod='get_object',
    Params={
        'Bucket': 'bucket-name',
        'Key': 'key-name'
    }
)
```

> "Anyone with valid security credentials can create a presigned URL. However, in order to successfully access an object, the presigned URL must be created by someone who has permission to perform the operation that the presigned URL is based upon."
> -- https://docs.aws.amazon.com/AmazonS3/latest/dev/ShareObjectPreSignedURL.html

shiroyuki:
The code now has been rebased.

terraform/authentication.tf.json.template.py
```diff
@@ -34,4 +34,4 @@
             }
         }
     ]
-} if config.subscribe_to_dss else None)
+} if (config.subscribe_to_dss and False) else None)
```
hannes-ucsc: Remove.

.travis.yml
```diff
@@ -16,6 +16,7 @@ before_install:
 
 install:
   - pip install -U setuptools>=40.1.0
+  - pip install -U google-compute-engine
```
hannes-ucsc: Add comment explaining that this is a dependency for moto.

hannes-ucsc:
@shiroyuki, to ask for a review, please request a review using the GitHub API and assign the PR to the reviewer.

shiroyuki:
Reverted the accidental changes back and added comment on the requirement of `google-compute-engine`.

.travis.yml
```diff
@@ -18,6 +18,8 @@ install:
   - pip install -U setuptools>=40.1.0
   - pip install -r requirements.dev.txt
   - pip install coveralls
+  # moto requires this package for EC2 mock, which is triggered even thought we aren't using it.
```
hannes-ucsc: Now that I have your explanation for this, I started feeling puzzled by this and did some research. I think this is a better fix for this problem: https://github.com/travis-ci/travis-ci/issues/7940#issuecomment-311411559

hannes-ucsc: Also note that you should document every non-obvious fix. Had you done that in the first place, we could have avoided yet another review round-trip. Just sayin' …

shiroyuki:
Based on recommendation from @hannes-ucsc, to fix the broken test on Travis CI, `BOTO_CONFIG` will be set to `/dev/null` so that we don't have to install `google-compute-engine`. This will be ready for review again once the tests pass on Travis CI.

shiroyuki:
@hannes-ucsc I tried to unset the environment variable but it doesn't seem to work. So, I revert back the change to the `/dev/null` approach.

hannes-ucsc:
@shiroyuki That makes sense, come to think of it. The variable isn't set when Travis runs the build, we are setting it to stop boto from looking at the default location in /etc.

Issue #276, opened 2018-08-30
Bug with `specimens/summary` and `files/summary` not matching aggregates.


achave11:
Query to `files_index`
```JSON
{
  "took": 6,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "hits": {
    "total": 14056,
    "max_score": 1
},
  "aggregations": {
    "total_specimens": {
      "value": 183
    },
    "total_files": {
      "value": 14278
    }
  }
}
```

Query to `specimens_index`
```JSON
{
  "took": 6,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "hits": {
    "total": 184,
    "max_score": 1
},
  "aggregations": {
    "total_specimens": {
      "value": 183
    },
    "total_files": {
      "value": 14278
    }
  }
}
```

Note that the `aggregations` do not match to the total number of `hits`. Ideally this numbers should be matching across the board.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-182)
┆Project Name: azul
┆Issue Number: AZUL-182

hannes-ucsc:
@achave11  and I spent some time looking into the root cause for why the specimens and files summaries are `permanently` inconsistent. The fix is essentially to only use primary keys to aggregate over when producing entity counts. For example, the specimen count is currently driven by a field called `biomaterial_id` which is supposed to the a unique identifier for specimens, but in fact, we indexed bundles that contain distinct (based on `document_id`) specimens with the same `biomaterial_id`. The fix for the specimen count is to count distinct `document_id` values, not distinct `biomaterial_id` values. There is a similar problem for projects which currently aggregates over `project_short_name`.

Once we fix those issues, the files and specimens summaries will be eventually consistent. They can still be temporarily inconsistent due concurrent indexing processes. We would like to maintain two separate summary endpoints so we have a way of telling whether the entity-specific indices are inconsistent.

@NoopDog proposed a different solution, which was to have one summary endpoint that hits all indices in separate queries and cherry picks the authoritative index for a particular the summary element. I prefer this solution because it is more efficient and exposes permanently inconsistent indices.

@NoopDog, this means that you should use `/summary/specimens` together with `/specimens` on the Specimens tab and `/summary/files` together with `/files` on the Files tab.

achave11:
[...]

PR #394, opened 2018-10-09
Bug fixes over summary responses (fixes #368)

### LINKED ISSUE #368: Bundle contributions to specimens are not currently updatable
When aggregating over `total_estimated_cells` in specimens in order to obtain the total number of cells in the azul `specimens` index and a given specimen contains multiple bundle contributions, the specimen will aggregate an incorrect number of cell counts. This is because the aggregates do not distinguish between the multiple bundles contributing to the specimen. Therefore adding the exact same field ('total_estimated_cells') however many times there is bundles contributing to that specimen.

The expected behavior is to have only one specimen contribution per bundle, and update the specimen based on how recent the contributing bundle is. Therefore, the bundles with a more recent version would have precedence over the others, and thereby allowing the specimen to update to the most recent properties.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-238)
┆Project Name: azul
┆Issue Number: AZUL-238

### LINKED ISSUE #400: Add `donor_document_id` for superior accuracy on donor count
We discover that `donor_biomatirial_id` which is the parameter that we use to display the number of different donors in our system is not unique across all the bundles in azul. Making queries on this value to obtain the different donors yields inaccurate responses.
The field `document_id` is distinct among each donor. Making it a field in the azul index as `donor_document_id` will allow us to distinguish distinct donors in our aggregations. Therefore, giving us an accurate result when requesting total donors count.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-261)
┆Project Name: azul
┆Issue Number: AZUL-261

achave11:
* Added `donor_document_id` field to indexer, for specimens. (fixes #400)
* Added `document_id` as a singleton to indexer, for specimens.
* Aggregate donorCount using new field `donor_document_id`, yields accurate donor count for summary.
* `document_id` aggregate for accurate specimen count (connected to #276)
* Added inline script to total_cells aggregate in order to obtaining accurate cell counts.

**NOTES:**
In order for this changes to be reflected in the summary responses, one must reindex.
This changes include additions and modifications to `specimen` fields which in return yield more accurate results when creating a summary response.

### DIFF
--- a/src/azul/project/hca/transformers.py
+++ b/src/azul/project/hca/transformers.py
@@ -69,6 +69,7 @@ def _specimen_dict(specimen: api.SpecimenFromOrganism) -> MutableMapping[str, An
                 merged_specimen[key].add(value)
     merged_specimen = {k: list(sorted(v, key=lambda x: (x is not None, x))) for k, v in merged_specimen.items()}
     merged_specimen['biomaterial_id'] = specimen.biomaterial_id
+    merged_specimen['document_id'] = specimen.document_id
     return merged_specimen
 
 
@@ -149,7 +150,6 @@ class BiomaterialVisitor(api.EntityVisitor):
             # As more facets are required by the browser, handle each biomaterial as appropriate
             self.biomaterial_lineage.append(
                 {
-                    'document_id': entity.document_id,
                     'has_input_biomaterial': entity.has_input_biomaterial,
                     '_source': api.schema_names[type(entity)],
                     **(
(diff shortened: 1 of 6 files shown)


hannes-ucsc:
Also, isn't there a ticket for the `donor_document_id` change?

theathorn:
Obsoleted by PR #433.

PR #404, opened 2018-10-11
Only indexes bundles `2018-10-10` inclusive (fixes #398)

### LINKED ISSUE #398: Don't index old bundles in DSS integration
Add a `range` clause on `version` to the subscription query that limits notifications and reindexing to only those bundles with a version lexicographically greater than `2018-10-10`

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-259)
┆Project Name: azul
┆Issue Number: AZUL-259

achave11:
This fix only indexes bundles with a `manifest.version` field greater than `2018-10-10` inclusive, until the present day in which the query is executed `now`

### DIFF
--- a/src/azul/project/hca/__init__.py
+++ b/src/azul/project/hca/__init__.py
@@ -20,7 +20,15 @@ dss_subscription_query = {
                                   if config.dss_endpoint != "https://dss.data.humancellatlas.org/v1"
                                   else "files.biomaterial_json")
                     }
-                }
+                },
+                {
+                    "range": {
+                        "manifest.version": {
+                            "gte": "2018-10-10"
+                        }
+                    }
+                } if config.dss_endpoint == "https://dss.integration.data.humancellatlas.org/v1"
+                else {}
             ]
         }
     }


src/azul/project/hca/__init__.py
```diff
@@ -20,6 +20,14 @@
                                   if config.dss_endpoint != "https://dss.data.humancellatlas.org/v1"
                                   else "files.biomaterial_json")
                     }
+                },
+                {
+                    "range": {
+                        "manifest.version": {
+                            "gte": "2018-10-10T000000.000000Z",
```
hannes-ucsc: Can you check if just `2018-10-10` works, too?

src/azul/project/hca/__init__.py
```diff
@@ -20,6 +20,14 @@
...
+                            "lte": "now"
```
hannes-ucsc: Remove. `lte` is optional, I think.

src/azul/project/hca/__init__.py
```diff
@@ -20,6 +20,13 @@
                                   if config.dss_endpoint != "https://dss.data.humancellatlas.org/v1"
                                   else "files.biomaterial_json")
                     }
+                },
```
hannes-ucsc: This needs to be conditional. It should only be active if we're hitting the integration instance of DSS. See above for example.

src/azul/project/hca/__init__.py
```diff
@@ -20,6 +20,15 @@
...
+                {
+                    "range": {
+                        "manifest.version": {
+                            "gte": ("2018-10-10"
+                                    if config.dss_endpoint == "https://dss.integration.humancellatlas.org/v1"
+                                    else "0000-00-00")
+                        }
+                    }
```
hannes-ucsc: Use `*([{"foo": bar}] if condition else [])`

src/azul/project/hca/__init__.py
```diff
@@ -20,7 +20,15 @@
...
-                }
+                },
+                {
+                    "range": {
+                        "manifest.version": {
+                            "gte": "2018-10-10"
+                        }
+                    }
+                } if config.dss_endpoint == "https://dss.integration.data.humancellatlas.org/v1"
+                else {}
```
hannes-ucsc: Nope. See my previous comment. I don't want an empty dict in the else case. I want nothing. It was all there on a platter for you and you still ignored it: 

Use *([{"foo": bar}] if condition else [])

hannes-ucsc:
Never mind, I'l take it from here.

PR #405, opened 2018-10-11
Exclude project 100k pilot project in staging from being indexed

hannes-ucsc:
Only the project entity will be suppressed. All other entities from
bundles in that project will be indexed.


hannes-ucsc:
Please ignore.

PR #406, opened 2018-10-11
Fix: project contributors without email fail indexing

hannes-ucsc:
@Ben-Wu Please check if I captured your original intentions correctly. Python3 can't sort heterogeneous lists so we ~can~ cannot mix strings and None values in the same list.


hannes-ucsc:
If determinism for test comparisons is the only goal, I have a better solution in one of my branches.

If there are two or more entries where `email` is `None`, you would still not get a deterministic ordering, correct?

hannes-ucsc:
Obsoleted by #433.

Issue #88, opened 2018-07-13
Shopping cart: support for interactive building of collections


cricketsloan:
From https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/dcp/34:

As a user of the DCP, I will approach it and it's vast amounts of data and, after searching for the data that meets the requirements of my study or analysis, I will want to take those search results and treat them as a unit.

A user can create a persistent list of data, view existing lists, export or download the list. This feature is analogous to a shopping cart capability on an e-commerce site.

Original User Story for #88:

As a user I would like to iteratively add entities to a continually growing list, and then save that list.

This could be done using a Dynamo DB table, and providing some REST api to allow users to do simple CRUD operations on their current collection, and upon clicking save, we would save that collection via the collections API.

┆Issue is synchronized with this [Jira Epic](https://ucsc-cgl.atlassian.net/browse/AZUL-64)
┆containerName: azul
┆Issue Number: AZUL-64

cricketsloan:
➤ Trevor Heathorn commented:

Will eventually be blocked by #31.

cricketsloan:
➤ Trevor Heathorn commented:

From CGP Tech Leads Meeting on 7/12:
Work on making collections support in Orange for use by download shopping cart, be able to mint into blue box collections, using a collections JSON schema.

theathorn:
- shopping cart to save search results
- shopping cart per entity type (Files, Projects, ...)
- click button in UI to make a named collection (or use Collection ID)
- share, rename a collection
- once saved, user can't alter content of collection

theathorn:
See https://docs.google.com/document/d/1dvz7ij5JDAHSFZmpAHsw3_xbDge-EEhuLcvA5SIkhc8/

theathorn:
Blue Box Collections API: https://dss.integration.data.humancellatlas.org/

theathorn:
*User Story #1*
As a user I wish to select files according to a faceted search and save the results in a "shopping cart".
I can conduct additional searches and add or delete files to the shopping cart.
When I'm ready I can download all the files in the shopping cart.
See the GDC Data Portal (https://portal.gdc.cancer.gov/projects) for an example.
This can be achieved by creating an (immutable) Collection for the selected files which will return a UUID or url for the Collection.
The Collection UUID/url can be passed to the HCA CLI to perform bulk file download.

*User Story #2*
[...]

PR #417, opened 2018-10-12
RFC: Cart API

shiroyuki:
(related to #88)


rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+| `cart_name` | str | Cart Name
+| `entity_type` | str | Entity Type |
+
+### `CartItem`
+| Property | Type | Description |
+| --- | --- | --- |
+| `id` | UUID-like `str` | Cart Item ID, auto-generated by the web API |
+| `cart_id` | UUID-like `str` | Associated `Cart.id` |
+| `entity_id` | UUID-like `str` | Entity ID |
+| `bundle_id` | UUID-like `str` | Bundle ID |
```
hannes-ucsc: `bundle_uuid` to be consistent.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+| Issue | Status |
+| ----- | ------ |
+| #88   | Draft  |
+
+## Data Models
+
+### `Cart`
+| Property | Type | Description |
+| --- | --- | --- |
+| `id` | UUID-like `str` | Cart ID, auto-generated by the web API |
```
hannes-ucsc: Type should just be UUID. `-like` is fuzzy and `str` is an implementation detail.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+| `user_id` | UUID-like `str` | The ID of the user that creates the collection |
+| `cart_name` | str | Cart Name
```
hannes-ucsc: Lets introduce the notion of a default cart whose name is null.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+  * Generally, the response on `GET` request will return a single JSON object of the corresponding model if the ID is specified or the list of JSON object otherwise. *There will be some exceptions.*
+
+#### CRUD APIs for `Cart`
+
+| Method | Path | Description |
+| --- | --- | --- |
+| `POST` | `/resources/carts/` | Create a new cart |
+| `GET` | `/resources/carts/` | Retrieve a list of carts |
+| `GET` | `/resources/carts/{id}` | Retrieve a single cart (see more detail below) |
+| `PUT`/`PATCH` | `/resources/carts/{id}` | Either overwrite the whole cart or partially update the cart |
```
hannes-ucsc: Omit PATCH or specify how it works.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+| `DELETE` | `/resources/carts/{id}` | Delete a single cart (see more detail below) |
+
+#### `GET /resources/carts/{id}`
+
+Instead of returning `Cart` information, the response will be also returning the list of associated `CartItem`, as shown below:
```
hannes-ucsc: /carts/{id} should only return information about the cart itself.

/carts/{id}/items should be used for cart items.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+
+* The endpoint only deletes the target cart if the user creates the target cart and responds with **HTTP 200**.
+  * Otherwise, the endpoint must respond with **HTTP 403**.
+* The endpoint responds **HTTP 410** (gone) if the cart is unknown.
+
+### CRUD APIs for `CartItem`
+
+| Method | Path | Description |
+| --- | --- | --- |
+| `POST` | `/resources/cart-items/` | Create a new cart item |
```
hannes-ucsc: See above.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Data Models
+
+### `Cart`
+| Property | Type | Description |
+| --- | --- | --- |
+| `id` | UUID-like `str` | Cart ID, auto-generated by the web API |
+| `user_id` | UUID-like `str` | The ID of the user that creates the collection |
+| `cart_name` | str | Cart Name
+| `entity_type` | str | Entity Type |
```
hannes-ucsc: Carts should be able to mix items of a different types.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,84 @@
...
+
+### `CartItem`
+| Property | Type | Description |
+| --- | --- | --- |
+| `id` | UUID-like `str` | Cart Item ID, auto-generated by the web API |
```
hannes-ucsc: Cart items should not have their own id. The tuple `(cart_id, entity_id, bundle_uuid, bundle_version)` uniquely identifies a cart item.

rfc/cart-api-v1.md
shiroyuki: What do you think about adding a new field `default` (`bool`)? In this approach,

1. any cart can be the default cart,
2. the backend will set `cart_name` to `"default"` if the user has no default cart or `"untitled"` otherwise,
3. for data inspection, the new field would make it obvious which cart is default without the knowledge on how the backend service chooses the default cart.

One downside for either approaches is that if there exists more than two "default" carts (by accident or direct data manipulation), the service will have to choose which one is the default cart?

rfc/cart-api-v1.md
shiroyuki: According to the comment from @theathorn (https://github.com/DataBiosphere/azul/issues/88#issuecomment-421520215), the cart should be per entity type. Is this requirement still standing?

rfc/cart-api-v1.md
shiroyuki: So, in this case, `id` is a string representation of `sha1(cart_id, entity_id, bundle_uuid, bundle_version)`?

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+
+* The endpoint will only delete the target cart if the user creates the target cart and responds with **HTTP 200**.
+  * Otherwise, the endpoint must respond with **HTTP 403**.
+* The endpoint will respond **HTTP 410** (gone) if the cart is unknown.
+
+### CRUD APIs for `CartItem`
+
+| Method | Path | Description |
+| --- | --- | --- |
+| `POST` | `/resources/carts/{cart_id}/items/` | Create a new cart item |
```
Ben-Wu: It would be useful to allow creation of multiple items in one request for adding results of a search.
We could use take a request body that has an items property that is an array of item objects.  To add a single item it would be an array with one element to stay consistent.

shiroyuki: We will have a batch API for that if the UI can add items in batch.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+| `DELETE` | `/resources/carts/{cart_id}/items/{id}` | Delete a single cart item |
+
+#### `POST /resources/carts/{cart_id}/items/`
+
+| Param Type | Param Name | Association |
+| --- | --- | --- |
+| URL Path | `cart_id` | `Cart.id` |
+
+* The request body will be similar to `CartItem` without `cart_id`.
+* The endpoint will also respond **HTTP 409** if the corresponding cart is unknown.
```
Ben-Wu: Why do we use 409 here?  Doesn't 404 make more sense if we can't find the cart?

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+    cart_name: str // Required
+}
+```
+
+and the response is basically the JSON representation of `Cart`.
+
+#### `DELETE /resources/carts/{id}`
+
+* The endpoint will only delete the target cart if the user creates the target cart and responds with **HTTP 200**.
+  * Otherwise, the endpoint must respond with **HTTP 403**.
```
Ben-Wu: Do you mean respond with 403 if the cart doesn't belong to the user?

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+* The endpoint will respond **HTTP 410** (gone) if the cart is unknown.
```
Ben-Wu: How do we distinguish between a gone resource and one that didn't exist?  Unless instead of deleting we add a deleted field.  To keep it simple I think we can just use 404

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+    id: str, // Cart ID
+    cart_name: str,
+    default: bool,
+    items: List[CartItem] // Each CartItem object will not have `cart_id`.
+}
+```
+
+##### Default Cart
+
+The default cart is a special case and will be given back to the client ONLY IF `id` is `default` (all lowercase).
```
Ben-Wu: Just to clarify `default` is only used for this endpoint right?  So we can't do `PUT /resources/carts/default` to edit the default cart

Ben-Wu: nvm missed the line a few lines down

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+There are two data models required for Cart API: `Cart` and `CartItem`. They are
+designed in the way that allows multiple types of entity in a single `Cart`.
+
+### `Cart`
+| Property | Type | Description | Note |
+| --- | --- | --- | --- |
+| `id` | UUID | Cart ID, auto-generated by the web API | - |
+| `user_id` | UUID | The ID of the user that creates the collection | - |
+| `cart_name` | `str` | Cart Name: If it is `NULL`, the cart is the default one for a corresponding user. | **NEED MORE DISCUSSION** |
+| `default` | `bool` | Flag to indicate whether this cart is the default one for a corresponding user (`False` by default) | **NEED MORE DISCUSSION** |
```
Ben-Wu: FYI default is a reserved word in dynamo so something like `default_cart` or `is_default` should be used.  Any preferences?

shiroyuki: Let's go with `default_cart`.

rfc/cart-api-v1.md
shiroyuki: Yes. 👍

rfc/cart-api-v1.md
shiroyuki: I agree with your suggestion on 404. Let's use that.

rfc/cart-api-v1.md
shiroyuki: Sorry... you are right on that. Please change this to 404.

rfc/cart-api-v1.md
Ben-Wu: Writing down our earlier discussion: In addition to a batch api we may have to take a query in the response, make the query in the backend, and then save the results to the cart.  This is to deal with pagination because if the user wants to "save all results to cart" the client may not have all the results loaded.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,179 @@
...
+For the details on the default cart, see `GET /resources/carts/{id}`.
+
+### `CartItem`
+| Property | Type | Description |
+| --- | --- | --- |
+| `id` | `str` | Cart Item ID, hashed by the combination of `` |
+| `cart_id` | UUID | Associated `Cart.id` |
+| `entity_id` | UUID | Entity ID |
+| `entity_type` | `str` | Entity Type |
+| `bundle_uuid` | UUID | Bundle ID |
```
Ben-Wu: How should we handle entities with more than one bundle (i.e. non-files)

shiroyuki: If entity can have one or many bundles,

``` Entity --(1)----(1..n)--> Bundle ```

then, each cart item is per entity per bundle per bundle version.

Comment (Ben-Wu) [reply rfc/cart-api-v1.md]: That's a lot of cart items.  
I forgot that we can also store lists/sets which may work depending on the use case of the bundle information.  I don't actually remember why we decided to store the bundle here.  I don't think DSS collections use it.

File: rfc/cart-api-v1.md [hunk 16]
```diff
@@ -0,0 +1,179 @@
...
+| Property | Type | Description | Note |
+| --- | --- | --- | --- |
+| `id` | UUID | Cart ID, auto-generated by the web API | - |
+| `user_id` | UUID | The ID of the user that creates the collection | - |
+| `cart_name` | `str` | Cart Name: If it is `NULL`, the cart is the default one for a corresponding user. | **NEED MORE DISCUSSION** |
+| `default_cart` | `bool` | Flag to indicate whether this cart is the default one for a corresponding user (`False` by default) | **NEED MORE DISCUSSION** |
+
+For the details on the default cart, see `GET /resources/carts/{id}`.
+
+### `CartItem`
```
Ben-Wu: I think we should also store some user-readable data about the item so the user can distinguish items in the cart.
Storing all the metadata for each entity type probably isn't feasible so something generic across the entity types may be good enough.  e.g. `ItemName`

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+
+Then, for each batch, `λ3` will put each ready-to-write `CartItem` to DynamoDB. At the end of each batch, `λ3` will notify `λ4` that it finished writing the given batch via **the job part complete notification queue**.
+
+###### job Updates
+
+For each notice in **the job part complete notification queue**, `λ4` will either create a new `CartItemBatchJob` or try to update the corresponding one with the updated number of remaining parts.
+
+> **The job part complete notification queue** is a FIFO queue to minimize race condition.
+
+> The approach to avoid race conditions is based on [the doc on optimistic locking for DynamoDB with Java](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html) and [the use of `ConditionExpression` in `DynamoDBResource.put_item`](https://github.com/boto/boto3/issues/1550) where `CartItemBatchJob.remaining_parts` is acting as version.
```
Ben-Wu: I think this is a more relevant link https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.ConditionalUpdate

The java optimistic locking isn't really relevant to boto3

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+
+> With [the size limit on SQS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html), the message publication will be more frequent than ES scanning.
+
+Then, for each batch, `λ3` will put each ready-to-write `CartItem` to DynamoDB. At the end of each batch, `λ3` will notify `λ4` that it finished writing the given batch via **the job part complete notification queue**.
+
+###### job Updates
+
+For each notice in **the job part complete notification queue**, `λ4` will either create a new `CartItemBatchJob` or try to update the corresponding one with the updated number of remaining parts.
+
+> **The job part complete notification queue** is a FIFO queue to minimize race condition.
```
Ben-Wu: A FIFO queue should be able to prevent the race conditions.  It would require a second queue though: one non-fifo to trigger the lambda and within the lambda we receive messages from the fifo.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+
+When `λ2` receives a message with a requested filter, it will run the query, iterate the result and send a list of ready-to-write `CartItem` to `λ3` (in multiple batches) via **the job part queue**.
+
+> With [the size limit on SQS](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html), the message publication will be more frequent than ES scanning.
+
+Then, for each batch, `λ3` will put each ready-to-write `CartItem` to DynamoDB. At the end of each batch, `λ3` will notify `λ4` that it finished writing the given batch via **the job part complete notification queue**.
+
+###### job Updates
+
+For each notice in **the job part complete notification queue**, `λ4` will either create a new `CartItemBatchJob` or try to update the corresponding one with the updated number of remaining parts.
```
Ben-Wu: The item should be created when the request is created so that it initializes the counter with the total number of batches and the start time and any other metadata we might need.  This job update should just decrement the counter

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+
+TL;DR:
+* Respond **HTTP 202** on successfully accepting a request.
+* Use `POST /resources/cart-item-batch-jobs/{job_id}` to check the async job status.
+
+The response body for HTTP 202 will be `CartItemBatchJob`.
+
+##### Technical Considerations
+
+As this operation may require more than 30 seconds to run, the batch operation is divided into three major parts: processing requests, job updates, and job reports.
```
Ben-Wu: Processing requests can be broken into multiple parts: handling client request, creating cart item batches, writing batches to dynamo.  Basically λ 1 to 3 in the diagram.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+
+> See [the overall diagram describing the process](https://github.com/DataBiosphere/azul/raw/rfc/cart-api/rfc/DNAstack%20-%20General%20-%20Batch%20Cart%20Item%20Creation%20API%20Diagram.png).
+
+###### Processing requests
+
+> See "Creation Request" in the diagram.
+
+When the endpoint `λ1` receives a request with filters, the endpoint will delegate the query to `λ2` via **the request queue**. At this point, the client will not be able to figure out whether the new cart items based on the query are in the cart. The client will have to track the progress by sending request to `λ5` for an update.
```
Ben-Wu: Using λ as identifiers is kind of hard to understand.  Should note that λ1 is the `/resources/carts/{cart_id}/items/query` endpoint which returns a request id that is sent to λ5 which is the `/resources/cart-item-batch-jobs/{job_id}` endpoint

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+| Param Type | Param Name | Association |
+| --- | --- | --- |
+| URL Path | `cart_id` | `Cart.id` |
+| URL Path | `item_id` | `CartItem.id` |
+
+* The endpoint will also respond **HTTP 404** if the corresponding cart is not found.
+
+### Other APIs
+
+#### `POST /resources/carts/{cart_id}/items/query`
```
Ben-Wu: Suggestion: change to `POST /batch/carts/{cart_id}/items/query` as it isn't a CRUD operation.

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+
+> **The job part complete notification queue** is a FIFO queue to minimize race condition.
+
+> The approach to avoid race conditions is based on [the doc on optimistic locking for DynamoDB with Java](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html) and [the use of `ConditionExpression` in `DynamoDBResource.put_item`](https://github.com/boto/boto3/issues/1550) where `CartItemBatchJob.remaining_parts` is acting as version.
+
+###### Job Reports
+
+The client at any time may send a request to `λ5` (`POST /resources/cart-item-batch-jobs/{job_id}`) to see the progress.
+
+#### `POST /resources/cart-item-batch-jobs/{job_id}`
```
Ben-Wu:
```suggestion
#### `GET /resources/carts/{cart_id}/items/request/{job_id}`
```
Or something like that using the carts resource

rfc/cart-api-v1.md
```diff
@@ -0,0 +1,251 @@
...
+| `id` | `str` | Cart Item ID, hashed by the combination of `` |
+| `cart_id` | UUID | Associated `Cart.id` |
+| `entity_id` | UUID | Entity ID |
+| `entity_type` | `str` | Entity Type |
+| `bundle_uuid` | UUID | Bundle ID |
+| `bundle_version` | `str` | Bundle Version |
+
+> Cart Item ID is a hash string of the combination of `cart_id`, `entity_id`, `bundle_uuid`, `bundle_version`. The hash algorithm is probably `sha256` but it is not finalized.
+
+### `CartItemBatchJob`
```
Ben-Wu: May be useful to add a cart id here so we can query for in progress jobs for a given cart.  But this is dependent on how the frontend is designed.

hannes-ucsc:
If so, just confirm in a comment and reassign to me, please.

Ben-Wu:
The RFC and implementation have diverged a bit.  We'll update this and reassign

Ben-Wu:
We also plan on moving the markdown files to https://github.com/DataBiosphere/azul/pull/598 to make it easier to maintain.  Is that OK with you @hannes-ucsc?

Ben-Wu:
Updated RFC moved to https://github.com/DataBiosphere/azul/blob/feature/benwu/cart-save-all-sfn/rfc/cart-api-v1.md

hannes-ucsc:
Superseded by #598.

PR #472, opened 2018-10-19
Use manual aggregations instead of ES for project summaries (fixes #432)

### LINKED ISSUE #432: Stop mixing manual and ES aggregates for project summaries
Some project summary items as included at the end each hit in the /project response are computed via native ES aggregates. Others are computed manually in the service by iterating over the ES response. Apparently this was done because of duplicates among the entities whose fields are being aggregated. For example, a project document can contain multiple mentions of the same file and the same specimen, if that file/specimen is contributed by multiple bundles in the same project.

Once the fixes for #391 land, those duplicates will go away. Nevertheless, we should still aggregate manually across the board. The reason for this is that using native ES aggregations is inefficient and non-deterministic. The way it works is aggregating over all project documents, via a term aggregation on the project ID field and then having sub-aggregations on the entity fields. The problem is that the number of terms must be hard-coded in the request (currently 99999) and that the ES response includes one bucket per project in the index, not just per project in the current page. We could reduce the number of term buckets returned to an N < 99999, but that would introduce a non-determinism: there is no guarantee that the N buckets include one bucket for each hit in the page. There is another problem with using a low value for N: the buckets are [approximate](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-approximate-counts).

Maybe there is a way to achieve this natively with ES aggregation but I am not aware of one. The aggregation code should be generic. Take a look at the aggregators and accumulators in azul/indexer/transformer.py

Ben-Wu:
Created a method in `ProjectSummaryResponse` that will get the all the distinct values of a given key in a list of dictionaries.  This replaces the bucket aggregations previously done in ElasticSearch.

@hannes-ucsc I found that the way aggregations are done in `transformers.py` to be too complex for this case.  Let me know if you want aggregation here to be more generic and moved out of the class but I think it's unnecessary because there are no other use cases for this code at the moment.


src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -329,70 +329,34 @@ class ProjectSummaryResponse(BaseSummaryResponse):
...
         """
-        for project_bucket in project_buckets['buckets']:
-            if project_bucket['key'] == project_id:
-                return project_bucket[agg_key]['value']
-        return -1
+        terms = set()
+        for item in items:
+            try:
+                terms |= set(item[property_name])
+            except KeyError:
```
Ben-Wu: I made the assumption that it's acceptable that some items not have the property since this is the case for the real data.  This also means that there is no error if we try to look for a property that does not exist in any of the items.  I don't see an issue doing this but I wanted to confirm that this is acceptable.

hannes-ucsc: I think the real data should not have missing properties. If it does, that would constitute a bug. I want to avoid the ambiguity between a property missing from an entity (aggregate entity), and it's value being `None` (`[None]`). We should invariably guarantee that every entity always has all properties so they behave more like instance of a class (with a fixed set of attributes that can be none but are always present) instead of a dictionary where an entry can be absent or None.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -412,22 +376,25 @@ def __init__(self, project_id, raw_response):
 
         for hit in raw_response['hits']['hits']:
             if hit['_id'] == project_id:
+                hit_contents = hit['_source']['contents'].to_dict()
```
hannes-ucsc: This loop bugs me. I wanted to eliminate it but failed to do so because of the aggregates being computed by ES. Instead, I added the assert below. With your fixes in place I think it can now be eliminated.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -412,22 +376,25 @@ def __init__(self, project_id, raw_response):
...
                 total_cell_count, organ_cell_count = self.get_cell_count(hit)
+                donor_count = len(self.get_distinct_terms(hit_contents['specimens'], 'donor_biomaterial_id'))
+                species_list = self.get_distinct_terms(hit_contents['specimens'], 'genus_species')
```
hannes-ucsc: Eliminating the loop above probably makes the variable redundant. Generally speaking though, unless there is `species_set` and a `species_list` in the same scope, I prefer to not represent the type in the variable name, except to distinguish between single values ("entity") and collections ("entities"). So here that would be just `species`. It's unfortunate that `species` is its own plural but we can gloss over that.

src/azul/service/responseobjects/hca_response_v5.py
shiroyuki: In this case, suppose that one of the value from `item[property_name]` is `None`. Should `terms` eventually include or exclude `None`?

For example,
* `[None]` → `[]`?
* `[123, None ,456]` → `[123, 456]`?

shiroyuki: After reconsidering the case, I don't know if it is possible to see `None` in `item[property_name]`. Please disregard my comment.

Ben-Wu:
Set accumulator approach looks alright to me

Ben-Wu:
Ready for rereview.  Just checked the endpoints after deploying and reindexing.  Responses look good to me.

PR #484, opened 2018-10-23
Added url field to hits.files at /files (resolves #453)

### LINKED ISSUE #453: Include DSS URL in `contents.files`
┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-304)
┆Project Name: azul
┆Issue Number: AZUL-304

GPelayo:
Added the `hits.files.url` field to /files endpoint response.

Preview:
https://service.geryl.dev.explore.data.humancellatlas.org/repository/files

### DIFF
--- a/src/azul/service/responseobjects/elastic_request_builder.py
+++ b/src/azul/service/responseobjects/elastic_request_builder.py
@@ -412,6 +412,7 @@ class ElasticTransformDump(object):
                           filters=None,
                           pagination=None,
                           post_filter=False,
+                          include_file_urls=False,
                           entity_type='files'):
         """
         This function does the whole transformation process. It takes
@@ -428,6 +429,7 @@ class ElasticTransformDump(object):
         :param pagination: Pagination to be used for the API
         :param post_filter: Flag to indicate whether to do a post_filter
         call instead of the regular query.
+        :param include_file_urls: Show file url field in request
         :param entity_type: the string referring to the entity type used to get
         the ElasticSearch index to search
         :return: Returns the transformed request
@@ -545,6 +547,15 @@ class ElasticTransformDump(object):
         if entity_type == 'projects':  # Add project summaries to each project hit
             self.add_project_summaries(final_response['hits'], es_response)
 
+        if include_file_urls:
+            try:
+                for h in final_response['hits']:
+                    for data_file in h['files']:
+                        data_file['url'] = f"{config.dss_endpoint}/files/{data_file['uuid']}" \
+                                           f"?replica=aws&version={data_file['version']}"
+            except KeyError as e:
+                logging.error('Unable to add file url. Due to the following exception:')
+                logging.error(str(e))
         return final_response
 
     def transform_manifest(
--- a/src/azul/service/responseobjects/elastic_request_builder.py
+++ b/src/azul/service/responseobjects/elastic_request_builder.py
@@ -548,9 +548,10 @@ class ElasticTransformDump(object):
             self.add_project_summaries(final_response['hits'], es_response)
 
         if include_file_urls:
-            for h in final_response['hits']:
-                for data_file in h['files']:
-                    data_file['url'] = f"{config.dss_endpoint}/files/{data_file['uuid']}?replica=aws&version={data_file['version']}"
+            for hit in final_response['hits']:
+                for data_file in hit['files']:
+                    query_params = f"?replica=aws&version={data_file['version']}"
+                    data_file['url'] = f"{config.dss_endpoint}/files/{data_file['uuid']}{query_params}"
         return final_response
 
     def transform_manifest(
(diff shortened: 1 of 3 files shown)


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -552,6 +554,15 @@ def transform_request(self,
...
             self.add_project_summaries(final_response['hits'], es_response)
 
+        if include_file_urls:
+            try:
+                for h in final_response['hits']:
+                    for data_file in h['files']:
+                        data_file['url'] = f"{config.dss_endpoint}/files/{data_file['uuid']}" \
+                                           f"?replica=aws&version={data_file['version']}"
+            except KeyError as e:
+                logging.error('Unable to add file url. Due to the following exception:')
```
hannes-ucsc: Please don't write these blanket `except` clauses. I can count five dictionary lookups in the body of the `try:`, each one could cause a `KeyError`. Correct me if I am wrong but of neither of them should be swallowed like this because they would all constitute an error that needs to be surfaced very explicitly by returning a non-200 response. The only thing that a blanket clause like this achieves, is communicate that we are unsure about our invariants.

hannes-ucsc: Also, please take a look at how to log an exception: See `exc_info` in https://docs.python.org/3/library/logging.html#logging.debug

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -552,6 +554,15 @@ def transform_request(self,
         if entity_type == 'projects':  # Add project summaries to each project hit
             self.add_project_summaries(final_response['hits'], es_response)
 
+        if include_file_urls:
+            try:
+                for h in final_response['hits']:
+                    for data_file in h['files']:
+                        data_file['url'] = f"{config.dss_endpoint}/files/{data_file['uuid']}" \
```
hannes-ucsc: No backslashes for line continuations please.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -435,6 +436,7 @@ def transform_request(self,
         :param pagination: Pagination to be used for the API
         :param post_filter: Flag to indicate whether to do a post_filter
         call instead of the regular query.
+        :param include_file_urls: Show file url field in request
```
hannes-ucsc: In English, *URL* is an acronym and needs to be all-caps. For variables `url` is fine.

src/azul/service/responseobjects/elastic_request_builder.py
GPelayo: So should I remove the try-except and allow the `KeyError` to resurface as a 500 error?

hannes-ucsc: Yes, unless you can show that a KeyError is expected to occur, where and under which circumstances. In that case the enclosing `try:` block should only contain a single statement with a single dictionary lookup.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -545,6 +547,10 @@ def transform_request(self,
         if entity_type == 'projects':  # Add project summaries to each project hit
             self.add_project_summaries(final_response['hits'], es_response)
 
+        if include_file_urls:
+            for h in final_response['hits']:
+                for data_file in h['files']:
+                    data_file['url'] = f"{config.dss_endpoint}/files/{data_file['uuid']}?replica=aws&version={data_file['version']}"
```
hannes-ucsc: This line looks like it's too long. If so, please wrap but without using backslash.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -545,6 +547,10 @@ def transform_request(self,
         if entity_type == 'projects':  # Add project summaries to each project hit
             self.add_project_summaries(final_response['hits'], es_response)
 
+        if include_file_urls:
+            for h in final_response['hits']:
```
```suggestion
            for hit in final_response['hits']:
```

test/service/test_request_builder.py
```diff
@@ -640,6 +640,16 @@ def test_project_summaries(self):
...
 
+    def test_transform_request_with_file_url(self):
+        response_json = EsTd().transform_request(filters={"file": {}},
+                                                 pagination={'order': 'desc',
+                                                             'size': 10,
+                                                             'sort': 'specimenId'},
+                                                 post_filter=True,
+                                                 include_file_urls=True,
+                                                 entity_type='files')
+        self.assertTrue(all(['url' in file_data.keys() for hit in response_json['hits'] for file_data in hit['files']]))
```
hannes-ucsc: Please assert that the URL is valid, i.e. can be parsed.

test/service/test_request_builder.py
```diff
@@ -640,6 +642,26 @@ def test_project_summaries(self):
...
+                                                 pagination={'order': 'desc',
+                                                             'size': 10,
+                                                             'sort': 'specimenId'},
+                                                 post_filter=True,
+                                                 include_file_urls=True,
+                                                 entity_type='files')
+        bundle_files = [file_data for hit in response_json['hits'] for file_data in hit['files']]
+        for file_data in bundle_files:
+            self.assertIn('url', file_data.keys())
+            actual_url = urlparse(file_data['url'])
```
hannes-ucsc: Up to here would have been sufficient.

PR #496, opened 2018-10-25
Switched back to sha256 (resolves #494)

### LINKED ISSUE #494: Switch back to sha256 for files
Originally the manifest emitted a column with the sha256 for each file. When I implemented the download-manifest functionality in the dcp-cli, I used the sha256 for validating the file integrity.

During the indexer aggregation effort, I then removed the `sha256`, `etag`, and `crc32c` fields from the `files` entity to save index space, and only left the `sha1` field. I changed the manifest endpoint to emit the `sha1` instead of the `sha256`. That broke the dcp-cli and I opened https://github.com/HumanCellAtlas/dcp-cli/issues/199 to fix that. However, fixing it would then also require a new release and everyone to upgrade.

We should switch back to `sha256` for file entities in the index and in the manifest.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-325)
┆Project Name: azul
┆Issue Number: AZUL-325

GPelayo:
Reverted changes from https://github.com/DataBiosphere/azul/commit/74bd8bd512e15efe5cc42700d3193bf66a7368dc


src/azul/project/hca/transformers.py
```diff
@@ -76,6 +76,7 @@ def _file_dict(f: api.File) -> MutableMapping[str, Any]:
         'indexed': f.manifest_entry.indexed,
         'name': f.manifest_entry.name,
         'sha1': f.manifest_entry.sha1,
+        'sha256': f.manifest_entry.sha256,
```
hannes-ucsc: We want to replace `sha1` with `sha256`.

hannes-ucsc: Emphasis on "replace"

GPelayo: But if I remove the sha1 field the test_index_correctness test fails

test/service/data_generator/td_mapping.json
```diff
@@ -89,6 +89,17 @@
                   "analyzer": "autocomplete",
                   "search_analyzer": "standard"
                 },
+                "sha256": {
```
hannes-ucsc: REPLACE sha1 with sha256.

test/service/test_response.py
```diff
@@ -36,6 +36,7 @@ def project_value(v):
                         "name": "billion.key",
                         "read_index": "blue",
                         "sha1": "fc5923256fb9dd349698d29228246a5c94653e80",
+                        "sha256": "0166ec04250562942e3613be6c4b96ee2cea4967",
```
hannes-ucsc: REPLACE sha1 with sha256

PR #501, opened 2018-10-26
Manifest API Optimization (1st round)

### LINKED ISSUE #483: Manifest download occasionally times out on staging
Requesting a manifest on staging without any filters, takes close to 30s, sometimes over 30s and the request times out.

I think there are a few low-hanging fruit that have the potential to get us down to 15s or so. I would like to try the following optimizations.

1) Add logging statements so we see how long each phase of the request processing takes in a real lambda environment.

2) Eliminate the aggregates from the ES request.

3) Use [source filtering](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-source-filtering.html) to request only the fields from ES that are needed for preparing the manifest. I would propose to use store fields but I don't think our mapping specifies any stored fields. But keep that in mind as an possible alternative to source filtering.

5) "Stream" the TSV to S3. The point of this would be to overlap the ES I/O with the S3 I/O such that while we are waiting for a chunk to be written to S3, we're already fetching the next result from ES. This obviously involves threads. I can see two possible approaches: 1) a single PUT where boto3's `upload_fileobj()` method is given a custom implementation of `RawIOBase`. 2) a multipart PUT where each chunk is a part of the multipart upload. Each approach has caveats. 1) is tricky because the RawIOBase instance needs to be seekable which contradicts the streaming nature of this thing. 2) requires a finalizing CompleteMultipartUpload request and I don't know if that adds considerable latency. I doubt it and that's why I'd try 2) first.

4) Use `search_after` instead of the scroll API (or the `scan()` helper in elasticsearch_dsl)

I would like to try them in the order listed, as soon as we're down to 15s, let's file a PR and reconvene. Always measure the impact of each improvement.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-321)
┆Project Name: azul
┆Issue Number: AZUL-321

shiroyuki:
* Implement the support for source_filter (fixes #483)

### DIFF
--- a/src/azul/service/responseobjects/elastic_request_builder.py
+++ b/src/azul/service/responseobjects/elastic_request_builder.py
@@ -142,7 +142,6 @@ class ElasticTransformDump(object):
                        es_client,
                        req_config,
                        post_filter: bool=False,
-                       source_filter: List=None,
                        entity_type='files'):
         """
         This function will create an ElasticSearch request based on
@@ -156,7 +155,6 @@ class ElasticTransformDump(object):
         config
         :param post_filter: Flag for doing either post_filter or regular
         querying (i.e. faceting or not)
-        :param List source_filter: Source Filter (``_source``)
         :param entity_type: the string referring to the entity type used to get
         the ElasticSearch index to search
         :return: Returns the Search object that can be used for executing
@@ -167,18 +165,20 @@ class ElasticTransformDump(object):
         facet_config = {key: field_mapping[key]
                         for key in req_config['facets']}
         # Create the Search Object
-        es_search = Search(
-            using=es_client,
-            index=config.es_index_name(entity_type, aggregate=True))
+        es_search = Search(using=es_client, index=config.es_index_name(entity_type, aggregate=True))
         # Translate the filters keys
         filters = ElasticTransformDump.translate_filters(
             filters, field_mapping)
         # Get the query from 'create_query'
         es_query = ElasticTransformDump.create_query(filters)
         # Do a post_filter using the returned query
-        es_search = es_search.query(es_query) if not post_filter else es_search.post_filter(es_query)
+        if post_filter:
+            es_search = es_search.post_filter(es_query)
+        else:
+            es_search = es_search.query(es_query)
 
-        if source_filter:
+        source_filter = ElasticTransformDump.create_source_filter(req_config)
+        if source_filter is not None:
             es_search = es_search.source(include=source_filter)
 
         # Iterate over the aggregates in the facet_config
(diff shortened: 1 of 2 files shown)


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -137,11 +138,12 @@ def create_aggregate(filters, facet_config, agg):
...
-    def create_request(
-        filters, es_client,
-        req_config,
-        post_filter=False,
-        entity_type='files'):
+    def create_request(filters,
+                       es_client,
+                       req_config,
+                       post_filter: bool=False,
+                       source_filter: List=None,
```
hannes-ucsc: List of what? `str`, right?

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -154,6 +156,7 @@ def create_request(
         config
         :param post_filter: Flag for doing either post_filter or regular
         querying (i.e. faceting or not)
+        :param List source_filter: Source Filter (``_source``)
```
hannes-ucsc: No need to duplicate type hints in doc strings. Type hints should always be in the signature.

Also, explaining a parameter called `source_filter` with the words "Source Filter" is ironic, don't you think?

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -584,7 +590,11 @@ def transform_manifest(
             filters = {"file": {}}
         # Create an ElasticSearch request
         filters = filters['file']
-        es_search = self.create_request(filters, self.es_client, request_config, post_filter=False)
+        es_search = self.create_request(filters,
+                                        self.es_client,
+                                        request_config,
+                                        post_filter=False,
+                                        source_filter=['bundles.*', 'contents.files.*'])
```
hannes-ucsc: The manifest fields are defined in `request_config.json`. Clearly there is coupling between this source filter and that configuration. Can you see if you can derive the source filter from the configuration? If that's too involved file a ticket and comment on the coupling here, referencing the ticket.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -173,8 +176,11 @@ def create_request(
         # Get the query from 'create_query'
         es_query = ElasticTransformDump.create_query(filters)
         # Do a post_filter using the returned query
-        es_search = es_search.query(
-            es_query) if not post_filter else es_search.post_filter(es_query)
+        es_search = es_search.query(es_query) if not post_filter else es_search.post_filter(es_query)
```
hannes-ucsc: Can you convert this to a if-else statement? I feel that would make this important conditional more prominent.

hannes-ucsc:
Should the ticket reference be `(connected to …)` instead of `(fixes …)`?

shiroyuki:
https://github.com/databiosphere/azul/issues/483#issuecomment-434014474

hannes-ucsc:
I think c1439f7 misses the point. We definitely don't want to duplicate the config file (since its an abomination in the first place) and we don't want to create two separate sections in the config file that are tightly coupled.

Superseded by #511

shiroyuki:
So, if I understand correctly, you would like me to revert the code to the one like #511 but instead of having the hardcoded list, you want the code to generate `source_filter` with `manifest.(bundles|files)`.? (I'll work toward that assumption for the time being.)

Here is for the next step (after this PR).

Later in this process, there are two approach to disable aggregations:
1. add a new argument something like `aggregation_enabled` to make `create_request` ignore aggregations, or
2. remove `facet` from `request_config` where the empty facet list will effectively disable aggregation.

The first approach is in line with #511. What do you think?

PR #509, opened 2018-10-29
Use sha256 instead of sha1 (fixes #494)

### LINKED ISSUE #494: Switch back to sha256 for files

achave11:


hannes-ucsc:
Please don't ask for a review on PRs that don't build.

hannes-ucsc:
Well done.

PR #511, opened 2018-10-30
Use source filter to reduce ES reponse size for manifests

hannes-ucsc:
(connected to #483)


hannes-ucsc:
I tested this on my own deployment which indexes `staging` and the manifest response comes back a lot quicker. More like 10 seconds rather than 30.

PR #520, opened 2018-10-30
Implemented s3 multipart upload (WIP)

shiroyuki:
(connected to #483)

### DIFF
--- a/src/azul/service/responseobjects/storage_service.py
+++ b/src/azul/service/responseobjects/storage_service.py
@@ -1,9 +1,13 @@
+from contextlib import contextmanager
+from dataclasses import dataclass
 from logging import getLogger
-from typing import Optional
+from threading import Lock, Thread
+from typing import Optional, List
 import boto3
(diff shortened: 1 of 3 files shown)


src/azul/service/responseobjects/storage_service.py
```diff
@@ -37,14 +41,156 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                part = self._create_new_part(data)
+            else:
+                part = last_part
+                part.content.append(data)
+
+        if not part.is_uploadable:
+            return
+
+        # logger.info(f'multipart_upload/push: Part {part.part_number}: uploading in a background thread')
+        Thread(target=self._upload_part, args=(part,)).start()
```
hannes-ucsc: Starting threads like this, and not joining them is a deadly sin.

Nowadays you shouldn't have to use threads directly anymore. It's almost always more advisable to use a ThreadPoolExecutor.

PR #525, opened 2018-10-31
Implemented S3 Multipart Upload with Executor (WIP)

shiroyuki:
(connected to #483)


src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Buffer(StringIO):
+
+    def __init__(self, limit_size: int, inbetween_callback: Callable):
+        super(Buffer, self).__init__()
+        self.limit_size = limit_size
+        self.inbetween_callback = inbetween_callback
+        self.__remaining_size = 0
+        self.__total_size = 0
+        self.__pointer_lock = Lock()
```
hannes-ucsc: What's the lock for? There should only be one writer.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def write(self, s: str):
+        self.__pointer_lock.acquire()
+        byte_count = len(s.encode())
```
hannes-ucsc: Decide if Buffer deals with bytes or with strings. You shouldn't have to convert back and forth. For bytes there is also BytesIO.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -37,14 +43,144 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
 
         return object_key
 
+    @contextmanager
+    def multipart_upload(self, object_key: str, content_type: str):
```
hannes-ucsc: I would eliminate this function and turn MultipartUploadHandler into a context manager by adding `__enter__` and `__exit__` methods. It should be cleaner. What smells here is the fact that you have to pass the thread pool and the upload ID to the the MultipartUploadHandler constructor. That indicates that MultipartUploadHandler should manage that state and not this function.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -37,14 +43,144 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
 
     def create_bucket(self, bucket_name: str = None):
         self.client.create_bucket(Bucket=(bucket_name or self.__bucket_name))
-        logger.warning(f'{type(self).__name__}: Created a bucket called "{bucket_name or self.__bucket_name}"')
+
+
+class MultipartUploadHandler:
+
+    def __init__(self, bucket_name, object_key, upload_id, thread_pool):
+        self.__resource = boto3.resource('s3')
```
hannes-ucsc: Please don't use the private (`__`) mechanism. It's overkill unless your are writing a library with classes designed for inheritance that is going to be used across teams. It makes the code hard to read.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -37,14 +43,144 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        self.__bucket_name = bucket_name
+        self.__object_key = object_key
+        self.__upload_id = upload_id
+        self.__handler = self.__resource.MultipartUpload(self.__bucket_name, self.__object_key, self.__upload_id)
```
hannes-ucsc:
```suggestion
        self.mp_upload = self.__resource.MultipartUpload(self.__bucket_name, self.__object_key, self.__upload_id)
```

It's really confusing that a class whose name ends in `Handler` has a field called `handler`. In general you should avoid calling things just `handler` because that's just too generic.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -37,14 +43,144 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        self.__next_part_number = 1
+        self.__closed = False
+        self.__parts = []
+        self.__futures = []
+        self.__thread_pool = thread_pool
+        self.__new_part_lock = Lock()
+        self.__sync_lock = Lock()
```
hannes-ucsc: It's usually a bad smell to use locks conjunction with ThreadPoolExecutor. The whole point of the `concurrent` package is provide an abstraction that lets you write concurrent code WITHOUT having to manage the locks yourself.

These locks can be avoided if you make _upload_part return the part and then use future.get() to get the return value.

src/azul/service/responseobjects/storage_service.py
shiroyuki: I agree. In your opinion, what do you think of turning `MultipartUploadHandler` into a subclass of this class?

src/azul/service/responseobjects/storage_service.py
shiroyuki: So, if I understand you correctly, in the `push` method, instead of using locks, it will be something like:

```python
def push(self, data: bytes):
    part = self.thread_pool.submit(self.resume_last_part).result()
    if not part.is_uploadable:
        return
    part.uploaded = True
    self.__futures.append(self.__thread_pool.submit(self._upload_part, part))
```

where that part `self.thread_pool.submit(self.resume_last_part).result()` is safe enough to guarantee that the we will either resume the last part, which is previously not uploadable, or create a new part without race condition, right?

src/azul/service/responseobjects/storage_service.py
hannes-ucsc: Dubious. A MultipartUploadHandler is not a StorageService. A StorageService has an MultipartUploadHandler.

src/azul/service/responseobjects/storage_service.py
hannes-ucsc: I don't know what you mean by "resuming" a part. Once you have enough data for a part you should offload uploading that part as a task in the thread pool. Once all parts have been uploaded (as indicated by all futures indicating success), the MP upload can be completed by the main thread.

src/azul/service/responseobjects/buffer.py
hannes-ucsc: Good point. It probably won't. Consider wrapping the instance of this class in an `io.TextIOWrapper(write_through=True, encoding="utf8")` so this class doesn't have to worry about encoding.

shiroyuki:
@hannes-ucsc: Change Summary:

* The single-part upload code is restored and used (and the multi-part one is disabled).
* All Lock objects have been removed.
* `MultipartUploadHandler` is now a context manager and `StorageService.multipart_upload` has been subsequently removed.
* Most of private properties are now public.
* `Buffer` is renamed to `BytesBuffer` and based on `BytesIO`. As a result, `TextIOWrapper` is introduced to allow `csv.writer` and `BytesBuffer` to work together.

shiroyuki:
@hannes-ucsc Sorry for the delay. I forgot to give you an update on the benchmark. According to the new benchmarks, removing locks indeed improves the performance. However, the result shows no significant improvement gained from the multipart upload code over the single-part upload code.

The screenshots below are the benchmark in the different setups.

> * SPU: the single-part upload (`s3_client.put_item`)
> * MPU: the multipart upload
> * `[\d]x`: the number of worker threads (e.g., `4x` means four threads.

screen shot 2018-11-19 at 3 46 56 pm
screen shot 2018-11-19 at 3 47 08 pm
screen shot 2018-11-19 at 3 47 21 pm
screen shot 2018-11-19 at 3 47 36 pm

shiroyuki:
@hannes-ucsc This is the benchmark result based on the elapsed times extracted from the logs.

| Max Worker | Sample Size | Min      | Max      | Mean     | Median   | STD      |
| ---------- | ----------- | -------- | -------- | -------- | -------- | -------- |
|          0 |          50 |   9.7470 |  12.6719 |  10.8423 |  10.5247 |   0.8235 | 
|          1 |          51 |  10.8765 |  13.6470 |  12.1834 |  12.1125 |   0.5224 | 
|          2 |          50 |  11.2203 |  13.6915 |  12.0126 |  11.9559 |   0.4550 | 
|          4 |          50 |   9.6316 |  11.2974 |  10.4415 |  10.4074 |   0.2949 | 
|         12 |          50 |  10.0290 |  12.1640 |  10.6353 |  10.5174 |   0.4391 |

shiroyuki:
@hannes-ucsc After going through some reconsideration on not merging, I think it is better to be merged as the multipart upload code will perform better as the data grows.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,38 @@
+from logging import getLogger
+from io import BytesIO
+from typing import Callable
+
+logger = getLogger(__name__)
+
+
+class BytesBuffer(BytesIO):
```
hannes-ucsc: The name is very generic so as a reader I don't get the essential purpose. Document the purpose in one or two sentences.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+           # ...
+
+    where the context pattern will automatically shutdown upon exiting the context.
+    """
+
+    def __init__(self, object_key, content_type: str):
+        self.bucket_name = config.s3_bucket
+        self.object_key = object_key
+        self.upload_id = None
+        self.handler = None
```
hannes-ucsc: Unused.

hannes-ucsc: Please check all attributes in all classes that you introduce for unused fields.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,38 @@
...
+from dataclasses import dataclass
 from logging import getLogger
-from typing import Optional
+from typing import Optional, List
 import boto3
 from azul import config
 
 logger = getLogger(__name__)
+AWS_S3_DEFAULT_MINIMUM_PART_SIZE = 5242880  # 5 MB; see https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
+MULTIPART_UPLOAD_MAX_WORKERS = 12
```
hannes-ucsc: Didn't you conclude that 4 gave the fastest overall time?

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,38 @@
...
 
 
 class StorageService:
 
     def __init__(self):
-        self.__bucket_name = config.s3_bucket
-        self.__client = None  # the default client will be assigned later to allow patching.
+        self.bucket_name = config.s3_bucket
```
hannes-ucsc: See I don't understand why `bucket_name` is public while `_client` is protected. It could have disastrous consequences if someone modified `bucket_name` on an existing instance. Assuming I am correct, if we go to the extend of distinguishing between public and protected, we need to do so correctly, otherwise all that syntactic noise that the distinction introduces just makes things worse.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,38 @@
...
+        self._client = None  # the default client will be assigned later to allow patching.
 
     @property
```
hannes-ucsc: Check out `@lru_cache` from `functools`. It makes lazy init a one-liner (at a significant increase in spatial expense, but the concision and elimination of explicit state is worth it.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,38 @@
...
     def client(self):
-        if not self.__client:
-            self.__client = boto3.client('s3')
-        return self.__client
+        if not self._client:
+            self._client = boto3.client('s3')
+        return self._client
 
     def set_client(self, client):
```
hannes-ucsc: Unused. Remove.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,38 @@
...
+MULTIPART_UPLOAD_MAX_WORKERS = 12
 
 
 class StorageService:
 
     def __init__(self):
-        self.__bucket_name = config.s3_bucket
-        self.__client = None  # the default client will be assigned later to allow patching.
+        self.bucket_name = config.s3_bucket
+        self._client = None  # the default client will be assigned later to allow patching.
```
hannes-ucsc: I don't think the comment is accurate anymore.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,38 @@
...
+        if not self._client:
+            self._client = boto3.client('s3')
+        return self._client
 
     def set_client(self, client):
-        self.__client = client
+        self._client = client
 
-    def get(self, object_key: str):
+    def get(self, object_key: str) -> str:
```
hannes-ucsc: Not happy about the asymmetry between get() and put(). The former returns a string (using the default encoding, which I think should instead be explicitly fixed to utf-8) and the latter takes bytes. Unless there is a compelling reason for the asymmetry, I'd rather avoid it. Sorry for not catching that earlier.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        self.handler = None
+        self.next_part_number = 1
+        self.closed = False
+        self.content_type = content_type
+        self.parts = []
+        self.futures = []
+        self.thread_pool = None
+
+    def __enter__(self):
+        return self.start()
```
hannes-ucsc: I don't like I can use this class either as a context manager or by calling the `start` and `shutdown` methods explicitly. AFAICT, we only use one approach in this project. Please inline start() and shutdown() to enforce use as a CM. 

And to preempt the "But we may want to use it in this other way in the future" response: We're not writing a general library for other developers. We're trying to achieve specific functionality at minimal cost (time, amount of code, cost of maintenance, amount of test code, amount of documentation). We don't get paid for code that offers additional flexibility or features that aren't needed for delivering the actual features we were asked to implement.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+    def __exit__(self, type, value, traceback):
```
hannes-ucsc: You need to distinguish between normal exit and exceptional exit. You do not want to call `complete()` when this is being called with an exception. The context manager PEP makes that clear.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        Aborts a multipart upload.
+
+        :param raise_exception: Flag to raise exception on successful abort.
+        :raises UnexpectedMultipartUploadAbort: An abort due to unexpected
+                                                error occurred
+
+        If ``raise_exception`` is ``True``, :class:`UnexpectedMultipartUploadAbort`
+        will be raised.
+        """
+        if not self.is_active:
```
hannes-ucsc: I think `abort` should raise when called on an inactive handler. I don't see how idempotence is useful here.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+            return
+
+        self.mp_upload.abort()
+        self.mp_upload = None
+        self.closed = True
```
hannes-ucsc: Can you see if you can eliminate `closed` and instead piggy-back on the presence/absence of another instance attribute?

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+            self.futures.append(self.thread_pool.submit(self._upload_part, last_part))
+
+        for _ in as_completed(self.futures):
+            pass  # Blocked until all uploads are done.
+
+        self.mp_upload.complete(MultipartUpload={"Parts": [part.to_dict() for part in self.parts]})
+        self.mp_upload = None
+        self.closed = True
+
+    def abort(self, raise_exception:bool=True):
```
hannes-ucsc: `raise_exception` is only ever passed `False` so why do we even have it?

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        self.closed = True
+
+        logger.warning('Upload %s: Aborted', self.upload_id)
+        if raise_exception:
+            raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}')
+
+    def push(self, data: bytes):
+        part = self._get_next_part(data)
+
+        if not part.is_uploadable:
```
hannes-ucsc:
```suggestion
        if not part.is_complete:
```

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+    content: List[bytes]
+    uploaded: bool  # If true, the content is either being uploaded or already pushed to S3 (with etag defined).
+
+    @property
+    def already_uploaded(self):
+        return self.uploaded or self.etag is not None
+
+    @property
+    def is_uploadable(self):
+        return len(b''.join(self.content)) >= AWS_S3_DEFAULT_MINIMUM_PART_SIZE
```
hannes-ucsc: This is inefficient. If you only want the length, what you want to do is `sum(map(len,self.content)))`

hannes-ucsc: Then again I don't see the need for tracking incomplete parts in the first place.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +43,192 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
 class GetObjectError(RuntimeError):
     pass
+
+
+class EmptyMultipartUploadError(RuntimeError):
+    pass
+
+
+class UnexpectedMultipartUploadAbort(RuntimeError):
+    pass
```
hannes-ucsc: EOL missing at EOF

src/azul/service/responseobjects/storage_service.py
shiroyuki: i forgot to commit that change.

src/azul/service/responseobjects/storage_service.py
shiroyuki: When I implement this part. The only compelling reason that I can remember is probably because the code would eventually needs to work with the string version of the data, not bytes. However, I can see your concern here and I now think this definitely will confuse other people. I will change this to bytes.

src/azul/service/responseobjects/storage_service.py
shiroyuki: Originally, I only made `_client` protected for lazy loading. However, if `self._client` was removed as the result of using `lru_cache` on `self.client` (property), should we still keep `bucket_name` public?

src/azul/service/responseobjects/storage_service.py
shiroyuki: Originally, this code is to prevent `MultipartUploadHandler` from uploading less-than-5MB parts. Since we have ByteBuffer, there are two approaches to decide the fate of this property.
* Heavily rely on ByteBuffer to prevent that from happening
* Keep this as the final safeguard against unintentional upload.

src/azul/service/responseobjects/storage_service.py
shiroyuki: Originally, the attribute is more about implying whether the part has enough data to be uploaded. Could you elaborate why `is_complete` is better suitable here?

shiroyuki:
@hannes-ucsc Here is the summary of changes:
* `BytesBuffer` is now `FlushableBuffer` (with docs).
* Make `StorageService.get` to return `bytes` instead of `str`.
* `MultipartUploadHandler.__exit__` now handles exceptions.
* `Part.is_uploadable` and the ability for `MultipartUploadHandler` to resume non-uploadable parts have been removed in order to simplify the implementation.
> The implementation relies on AWS to reject the completion of the upload (`boto3.exceptions.ClientError` with message `EntityTooSmall`) to trigger auto-abort.

src/azul/service/responseobjects/storage_service.py
hannes-ucsc: It's good to have a second line of defense against coding errors. But the defense shouldn't consist of duplicating logic or trying to gracefully handle the scenario you're trying to prevent. That would only exacerbate the problem by hiding it. The defense should be to raise an assertion error so we are made aware of the problem and can fix it. Besides, the assertion of an invariant is simpler and easier to implement than gracefully handling a violation of an invariant.

src/azul/service/responseobjects/storage_service.py
shiroyuki: The updated code will raise `InactiveMultipartUploadAbort` if the handler is inactive.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,162 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+        self.mp_upload.abort()
+        self.mp_upload = None
+        self.thread_pool.shutdown()
+
+        logger.warning('Upload %s: Aborted', self.upload_id)
+
+    def push(self, data: bytes):
+        part = self._create_new_part(data)
+
```
hannes-ucsc: Remove

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,162 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        self.thread_pool.shutdown()
+
+    def abort(self):
+        if not self.is_active:
+            raise InactiveMultipartUploadAbort()
+
+        self.mp_upload.abort()
+        self.mp_upload = None
+        self.thread_pool.shutdown()
+
```
hannes-ucsc: Remove

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,162 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                raise UploadPartSizeOutOfBoundError(f'{self.bucket_name}/{self.object_key}')
+            raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}')
+
+        self.mp_upload = None
+        self.thread_pool.shutdown()
+
+    def abort(self):
+        if not self.is_active:
+            raise InactiveMultipartUploadAbort()
+
```
hannes-ucsc: Remove

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,162 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        according to the AWS documentation) in the same module.
+        """
+        if not self.is_active:
+            return
+
+        if not self.parts:
+            self.abort()
+            raise EmptyMultipartUploadError(f'{self.bucket_name}/{self.object_key}')
+
+        for _ in as_completed(self.futures):
```
hannes-ucsc: You need to check on the status of each future. What if any of them raised an exception?

src/azul/service/responseobjects/storage_service.py
shiroyuki: You meant removing `abort`?

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+from io import BytesIO
+from typing import Callable
+
+logger = getLogger(__name__)
+
+
+class FlushableBuffer(BytesIO):
+    """
+    A buffer that flushes the output to a callback function (``inbetween_callback``),
+    either if there exists the size of the fragment is large enough
```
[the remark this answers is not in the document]
hannes-ucsc: Be sure to accurately describe what this class does. A good way to do that is specify the invariant. As in "the callback is invoked zero or more times with an argument that is N bytes long, followed by exactly one invocation with an argument that is between 1 and N bytes long."

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+    (over ``chunk_size``) on flush or when the buffer is closed.
+
+    :param chunk_size: The minimum size of buffer (byte length)
```
hannes-ucsc: I thought we agreed for chunk_size to be the exact size, not the minimum size of each chunk (except the last chunk).

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+    :param inbetween_callback: The callback function to receive flushed output
+    """
+
+    def __init__(self, chunk_size: int, inbetween_callback: Callable):
+        super(FlushableBuffer, self).__init__()
+        self.__chunk_size = chunk_size
+        self.__inbetween_callback = inbetween_callback
```
hannes-ucsc: This name is annoyingly long. `callback` is fine as this is the only callback in the entire class.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+        self.__remaining_size = 0
+
+    def write(self, b: bytes):
+        super().write(b)
+        byte_count = len(b)
```
hannes-ucsc: Don't extract a variable. `len(b)` is just as expressive.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def close(self):
+        if self.__remaining_size > 0:
+            logger.warning(f'Clearing the remaining buffer (approx. {self.__remaining_size} B)')
+            self.__inbetween_callback(self.getvalue())
+            self.__remaining_size = 0
+            # As the buffer is closed, the pointer doesn't need to be reset.
+        super().close()
+
+    def __clean_up(self):
```
hannes-ucsc: Only one callsite. Inline.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def write(self, b: bytes):
+        super().write(b)
+        byte_count = len(b)
+        self.__remaining_size += byte_count
+        self.__clean_up()
+
+    def close(self):
+        if self.__remaining_size > 0:
+            logger.warning(f'Clearing the remaining buffer (approx. {self.__remaining_size} B)')
```
hannes-ucsc: How does this call for a warning?

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+            self.__inbetween_callback(self.getvalue())
+            self.__remaining_size = 0
+            # As the buffer is closed, the pointer doesn't need to be reset.
+        super().close()
+
+    def __clean_up(self):
+        if self.__remaining_size < self.__chunk_size:
```
hannes-ucsc: No bail-outs, please.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+        byte_count = len(b)
+        self.__remaining_size += byte_count
+        self.__clean_up()
+
+    def close(self):
+        if self.__remaining_size > 0:
+            logger.warning(f'Clearing the remaining buffer (approx. {self.__remaining_size} B)')
+            self.__inbetween_callback(self.getvalue())
+            self.__remaining_size = 0
+            # As the buffer is closed, the pointer doesn't need to be reset.
```
hannes-ucsc: What pointer? Use consistent language.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,55 @@
...
+        super().close()
+
+    def __clean_up(self):
+        if self.__remaining_size < self.__chunk_size:
+            return
+        value = self.getvalue()
```
hannes-ucsc: Consider getbuffer() to avoid the copy.  But only if the change is easy.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -191,6 +194,39 @@ def _translate(self, untranslated, keyname):
...
+        with MultipartUploadHandler(object_key, content_type) as multipart_upload:
+            with FlushableBuffer(AWS_S3_DEFAULT_MINIMUM_PART_SIZE, multipart_upload.push) as buffer:
+                buffer_wrapper = TextIOWrapper(buffer, encoding="utf-8", write_through=True)
+                writer = csv.writer(buffer_wrapper, dialect='excel-tab')
+
+                writer.writerow(list(self.manifest_entries['bundles'].keys()) +
+                                list(self.manifest_entries['contents.files'].keys()))
+                for hit in self.es_search.scan():
+                    self._iterate_hit(hit, writer)
+                    buffer.flush()
```
hannes-ucsc: This shouldn't be necessary. Please remove or motivate in a comment.

src/azul/service/responseobjects/buffer.py
shiroyuki: Somehow, invoking `getbuffer()` causes the object to be unresizable.

```BufferError: Existing exports of data: object cannot be re-sized```

This will prevent the buffer from truncating/resetting itself.

File: src/azul/service/responseobjects/buffer.py (hunk 28 above)
Comment (shiroyuki) [reply src/azul/service/responseobjects/buffer.py]: Sorry, I missed that.

File: src/azul/service/responseobjects/buffer.py (hunk 29 above)
Comment (shiroyuki) [reply src/azul/service/responseobjects/buffer.py]: Done

File: src/azul/service/responseobjects/buffer.py (hunk 35 above)
Comment (hannes-ucsc) [reply src/azul/service/responseobjects/buffer.py]: See docs on `getbuffer()`. You might have to `del` the return value before calling `truncate()`

Comment (shiroyuki) [reply src/azul/service/responseobjects/buffer.py]: I tried that. It works. But I also looked into a different way. The just updated code relies on the stream position `BytesIO.seek` and the memory efficiency should be the same. Can you check it out and let me know what you think?

File: src/azul/service/responseobjects/storage_service.py [hunk 37]
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}')
+
+        try:
+            self.mp_upload.complete(MultipartUpload={"Parts": [part.to_dict() for part in self.parts]})
+        except self.mp_upload.meta.client.exceptions.ClientError as e:
+            logger.error('Upload %s: Error detected while completing the upload. %s', self.upload_id, e)
+            self.__abort()
+            if 'EntityTooSmall' in e.args[0]:
+                raise UploadPartSizeOutOfBoundError(f'{self.bucket_name}/{self.object_key}')
+            raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}')
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -191,32 +194,39 @@ def _translate(self, untranslated, keyname):
...
-            hit_dict = hit.to_dict()
-            assert len(hit_dict['contents']['files']) == 1
-            file = hit_dict['contents']['files'][0]
-            file_fields = self._translate(file, 'contents.files')
-            for bundle in hit_dict['bundles']:
-                # FIXME: If a file is in multiple bundles, the manifest will list it twice. `hca dss download_manifest`
-                # would download the file twice (https://github.com/DataBiosphere/azul/issues/423).
-                bundle_fields = self._translate(bundle, 'bundles')
-                writer.writerow(bundle_fields + file_fields)
+    def _push_content(self) -> str:
```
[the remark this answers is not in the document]
shiroyuki: Done. In `environment`, I set the default to `0` to use MPU.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -226,10 +236,9 @@ def __init__(self, es_search, manifest_entries, mapping):
         """
         The constructor takes the raw response from ElasticSearch and creates
         a csv file based on the columns from the manifest_entries
-        :param raw_response: The raw response from ElasticSearch
+        :param es_search: The ElasticSearch DSL Search object
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
 
     def create_bucket(self, bucket_name: str = None):
-        self.client.create_bucket(Bucket=(bucket_name or self.__bucket_name))
-        logger.warning(f'{type(self).__name__}: Created a bucket called "{bucket_name or self.__bucket_name}"')
+        self.client.create_bucket(Bucket=(bucket_name or self.bucket_name))
+
+
+class MultipartUploadHandler:
+    """
+    S3 Multipart Upload Handler
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+    This class is to facilitate multipart upload to S3 storage. The class itself
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+    is also a context manager.
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+    Here is the sample usage.
+
+    .. code-block:: python
+
+       with MultipartUploadHandler('samples.txt', 'text/plain'):
+           handler.push(b'abc')
+           handler.push(b'defg')
+           # ...
+
+    where the context pattern will automatically shutdown upon exiting the context.
```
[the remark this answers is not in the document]
shiroyuki: Done. I also added the description of what would happen if an exception was raised within the context.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                                                                  Key=self.object_key,
+                                                                  ContentType=self.content_type)
+        self.upload_id = api_response['UploadId']
+        self.mp_upload = boto3.resource('s3').MultipartUpload(self.bucket_name, self.object_key, self.upload_id)
+        self.thread_pool = ThreadPoolExecutor(max_workers=MULTIPART_UPLOAD_MAX_WORKERS)
+        return self
+
+    def __exit__(self, etype, value, traceback):
+        if etype:
+            logger.error('Upload %s: Error detected within the MPU context.\n\n%s',
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                         self.upload_id,
+                         '\n'.join(format_exception(etype, value, traceback)))
+            self.__abort()
+            return
+        self.__complete()
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+    def __complete(self):
+        if not self.parts:
+            self.__abort()
+            raise EmptyMultipartUploadError(f'{self.bucket_name}/{self.object_key}')
```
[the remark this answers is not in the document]
shiroyuki: Originally, I want to prevent empty object from being uploaded. This is removed per your recommendation until there is such a use case that we have to handle.

src/azul/service/responseobjects/storage_service.py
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +40,135 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+        for future in as_completed(self.futures):
+            exception = future.exception()
+            if exception is not None:
+                logger.error('Upload %s: Error detected while uploading a part (%s: %s).', self.upload_id,
+                             type(exception).__name__, exception)
+                self.__abort()
+                raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}')
```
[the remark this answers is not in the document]
shiroyuki: Done.

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,60 @@
...
+        self.__remaining_size += len(b)
+
+        if self.__remaining_size >= self.__chunk_size:
+            offset = 0
+            while offset + self.__chunk_size <= self.remaining_size:
+                self.seek(offset)
+                self.__callback(self.read(self.__chunk_size))
+                offset += self.__chunk_size
+
+            # Get the remainer before resetting the pointer.
```
hannes-ucsc:
```suggestion
            # Get the remainder before resetting the pointer.
```

src/azul/service/responseobjects/buffer.py
```diff
@@ -0,0 +1,60 @@
...
+        self.__callback = callback
+        self.__remaining_size = 0
+
+    def write(self, b: bytes):
+        super().write(b)
+        self.__remaining_size += len(b)
+
+        if self.__remaining_size >= self.__chunk_size:
+            offset = 0
+            while offset + self.__chunk_size <= self.remaining_size:
```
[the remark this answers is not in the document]
shiroyuki: I will go with your suggestion. One little change I made is to have the condition as `self.__remaining_size >= self.__chunk_size` for readability.

src/azul/service/responseobjects/buffer.py
shiroyuki: The typos are fixed.

environment
```diff
@@ -179,6 +183,7 @@ _set AZUL_ES_TIMEOUT 60  # matches AWS' own timeout on the ELB sitting in front
 _set AZUL_DSS_WORKERS 8
 _set AZUL_INDEXER_CONCURRENCY 64
 _set AZUL_SUBSCRIBE_TO_DSS 1
+_set AZUL_DISABLE_MULTIPART_MANIFESTS 0 # Enable the multipart upload by default.
```
hannes-ucsc:
```suggestion
_set AZUL_DISABLE_MULTIPART_MANIFESTS 0
```

I think that's obvious.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+        else:
+            self.__complete()
+
+    def __complete(self):
+        for future in as_completed(self.futures):
+            exception = future.exception()
+            if exception is not None:
+                logger.error('Upload %s: Error detected while uploading a part.',
+                             self.upload_id,
+                             exc_info=(type(exception), exception, None))
```
hannes-ucsc:
```suggestion
                             exc_info=exception
```

Is equivalent.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}') from exception
+
+        try:
+            self.mp_upload.complete(MultipartUpload={"Parts": [part.to_dict() for part in self.parts]})
+        except self.mp_upload.meta.client.exceptions.ClientError as exception:
+            logger.error('Upload %s: Error detected while completing the upload.',
+                         self.upload_id,
+                         exc_info=(type(exception), exception, None))
+            self.__abort()
+            if 'EntityTooSmall' in exception.args[0]:
```
hannes-ucsc: Please eliminate this conditional and the UploadPartSizeOutOfBoundError class. If you use `from …` when raising the exception all the necessary information will be available and there is no need to further differentiate the type of the outer exception.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                raise UploadPartSizeOutOfBoundError(f'{self.bucket_name}/{self.object_key}')
+            else:
+                raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}')
```
hannes-ucsc:
```suggestion
                raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}') from exception
```

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
 
 class GetObjectError(RuntimeError):
     pass
+
+
+class EmptyMultipartUploadError(RuntimeError):
+    pass
+
+
+class UnexpectedMultipartUploadAbort(RuntimeError):
```
hannes-ucsc: Give this class a constructor that requires bucket name and key as arguments and compose the message to the super constructor. That way you can avoid having to compose the message `f'{self.bucket_name}/{self.object_key}` at each call site.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+    pass
```
hannes-ucsc: Give this class a constructor that requires bucket name and key as arguments and compose the message to the super constructor. That way you can avoid having to compose the message `f'{self.bucket_name}/{self.object_key}` at each call site.

src/azul/service/responseobjects/storage_service.py
hannes-ucsc:
```suggestion
class MultipartUploadError(RuntimeError):
```

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+    def to_dict(self):
+        return dict(PartNumber=self.part_number, ETag=self.etag)
 
 
 class GetObjectError(RuntimeError):
     pass
+
+
+class EmptyMultipartUploadError(RuntimeError):
```
hannes-ucsc: Remove. Unused.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+
+    def push(self, data: bytes):
+        part = self._create_new_part(data)
+        self.futures.append(self.thread_pool.submit(self._upload_part, part))
+
+    def _create_new_part(self, data: bytes):
+        part = Part(part_number=self.next_part_number, etag=None, content=data)
+        self.parts.append(part)
+        self.next_part_number += 1
+
```
hannes-ucsc: remove

src/azul/service/responseobjects/storage_service.py
```diff
@@ -39,12 +39,137 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None)
...
+                             exc_info=(type(exception), exception, None))
+                self.__abort()
+                raise UnexpectedMultipartUploadAbort(f'{self.bucket_name}/{self.object_key}') from exception
+
+        try:
+            self.mp_upload.complete(MultipartUpload={"Parts": [part.to_dict() for part in self.parts]})
+        except self.mp_upload.meta.client.exceptions.ClientError as exception:
+            logger.error('Upload %s: Error detected while completing the upload.',
+                         self.upload_id,
+                         exc_info=(type(exception), exception, None))
```
hannes-ucsc: See above

src/azul/service/responseobjects/storage_service.py
```diff
@@ -1,34 +1,34 @@
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from dataclasses import dataclass
+from functools import lru_cache
 from logging import getLogger
 from typing import Optional
 import boto3
 from azul import config
 
 logger = getLogger(__name__)
+AWS_S3_DEFAULT_MINIMUM_PART_SIZE = 5242880  # 5 MB; see https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
```
hannes-ucsc: I don't think we need to make this configurable, btw.

environment
```diff
@@ -110,6 +110,10 @@ azul_vars=(
     # Determine if CloudWatch alarms should be deployed (1 yes, 0 no)
     # CloudWatch alarms should not be enabled for personal deployments
     AZUL_ENABLE_CLOUDWATCH_ALARMS
+
+    # Boolean value, 1 to use the single-part upload (SPU) in Manifest API, 0 to
```
hannes-ucsc:
```suggestion
    # Boolean value, 1 to upload a manifest in a single request to S3, 0 to upload the manifest
    # in multiple concurrent requests for equal parts of a smaller size. This allows the 
    # manifest generation code to start uploading the manifest to S3 while the manifest
    # data is still being fetched from Elasticsearch, shortening the overall time needed 
    # to generate and upload the manifest.
```

Wrapped at 80 columns.

src/azul/service/responseobjects/storage_service.py
shiroyuki: Ok then.

PR #554, opened 2018-11-14
Fix / add index for program, project, and consent codes

jessebrennan:


indexer/test/dcc/test-files/gen3/example_metadata.json
```diff
@@ -198,6 +198,11 @@
       "age_at_index": "",
       "year_of_death": ""
     },
+    "program": {
+        "dbgap_accession_number": "topmed",
+        "id": "45b91a48-1961-5b9f-9b0f-766bd30665d8",
+        "name": "topmed"
+    },
     "project": {
```
mikebaumann: It seems like there should be a `case.concent_codes` entry in your test input data so your test ensures that if the consent codes are there, they are indexed. Also, check/take-into-account whether consent_codes is a string or array.

jessebrennan: There is an entry in the data. I'm not sure about the second part of your comment

PR #558, opened 2018-11-16
Generate manifest using step functions (fixes #546)

### LINKED ISSUE #546: Use step functions to produce manifest
The manifest endpoint has been bumping against API Gateway's 29s limit on request duration. So far we've managed to get by with various tweaks but those are not a long term solution. 

The manifest endpoint should kick off a step function execution and return a 301 redirect to itself with the execution ID encoded in an opaque token. The 301 response should also have a Retry-After header set to, say, 10s. The step function begins populating the manifest. For now the step function should have a single state in which it scans the ES results exhaustively and writes them to S3. In the future the step function could have a loop and perform a multipart upload (one per loop iteration). For that to work we would need to use `search_after` with the ES request.

```
GET /files/export 
301 Moved Permanently
Retry-After: 10
Location: /files/export?token=<base64_encoded_execution_id>
```

The client waits 10s and then requests the manifest endpoint again. The service checks if the execution is still running. Let's say it is and another self redirect is returned:

```
GET /files/export?token=<base64_encoded_execution_id>
301 Moved Permanently
Retry-After: 10
Location: /files/export?token=<base64_encoded_execution_id>
```

The client waits another 10s and then requests the manifest endpoint again. The service checks if the execution is still running. Let's say it is now done. We can assume that the manifest in S3 is now fully populated. The service returns a 302 redirect to a signed URL of the manifest in the S3 bucket.

```
GET /files/export?token=<base64_encoded_execution_id>
302 Moved Temporarily
Location: some-bucket.s3.amazon-aws.com/manifests/manifest-<UUID>.tsv
```

The token might have to include the manifest UUID as well..

We might also need to add a `&wait=1` parameter to the manifest endpoint URL. If it is set, the service sleeps 10s before returning and doesn't return a `Retry-After` header. This is for clients (like web browsers) that generally don't support `Retry-After`.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-351)
┆Project Name: azul
┆Issue Number: AZUL-351

Ben-Wu:


Ben-Wu:
Just a note - I'll look into test coverage in `app.py`

terraform/step_function.tf.json.template.py
```diff
@@ -0,0 +1,70 @@
+import json
+
+from azul import config
+from azul.deployment import aws
+from azul.template import emit
+
+
+def create_lambda_name(basename):
+    return f"arn:aws:lambda:{aws.region_name}:{aws.account}:function:{config.service_name}-{basename}"
```
hannes-ucsc: I wonder if this should be moved to `Config.get_lambda_arn(function_name, suffix)`

src/azul/__init__.py
```diff
@@ -208,6 +208,14 @@ def token_queue_name(self):
     def document_queue_name(self):
         return config.qualified_resource_name('documents', suffix='.fifo')
 
+    @property
+    def manifest_lambda_basename(self):
```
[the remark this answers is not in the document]
Ben-Wu: I would be more comfortable leaving this in one place as this is used in three different places: the step function terraform, the chalice config, and the decorator of the pure lambda.  Keeping the constant makes it easier to find everywhere the resource is used also it's possible we may need another lambda function for the manifest in the future so we would change this name to `manifest_something`.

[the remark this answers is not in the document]
hannes-ucsc: OK, sounds good.

Ben-Wu:
Cleaned things up by renaming and moving some things around.  Set `StepFunctionHelper` as a class attribute of `ManifestService` to allow mocking without passing it to the constructor.

src/azul/service/responseobjects/step_function_helper.py
```diff
@@ -0,0 +1,59 @@
+import json
+
+from azul.deployment import aws
+
+
+class StepFunctionHelper:
```
Ben-Wu: This class isn't really necessary because it has no state or class attributes.  These methods can just be functions in the file.  Is there any preference on which style to use?  I don't have any strong opinions in either direction.

src/azul/service/responseobjects/manifest_service.py
```diff
@@ -0,0 +1,60 @@
...
+
+from azul import config
+from azul.service.responseobjects.step_function_helper import StateMachineError, StepFunctionHelper
+
+
+class ManifestService:
+    """
+    Class containing logic for starting and checking the status of manifest generation jobs
+    """
+    manifest_endpoint = '/repository/manifest/files'
```
hannes-ucsc:
```suggestion
    manifest_endpoint = '/manifest/files'
```

src/azul/service/responseobjects/manifest_service.py
```diff
@@ -0,0 +1,60 @@
...
+
+        :param filters: filters to use for the manifest
+        :return: The id of the execution of the state machine
+        """
+        return self.step_function_helper.start_execution(
+            config.manifest_state_machine_name,
+            execution_input={'filters': filters}
+        )['executionArn'].split(':')[-1]
+
+    def get_manifest_status(self, params, wait, local=False):
```
hannes-ucsc: I'm uncomfortable with how the logic of parsing request input and generating the response output is spread over three layers: two functions in `app.py` and this method. They all seem to be tightly coupled. The fact that you have to extract the `manifest_endpoint` and use it in the route in app.py is the bad smell here.

I also understand that we don't want `app.py` to grow without bound. I do want all routes to be defined in a single `app.py` module so I think the way forward is to have the functions in `app.py` only extract the URL path and request parameters but do everything else in the service layer.

src/azul/service/responseobjects/step_function_helper.py
hannes-ucsc: This wrapper is annoyingly anemic. It doesn't do much but requires a lot of documentation. I realize that you need this for the mocks but we should be true to ourselves by not inflating this class' importance with fluffy documentation.

src/azul/service/responseobjects/step_function_helper.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Wrapper around boto3 SFN client to handle resource name generation and state machine executions
+    """
+    def state_machine_arn(self, state_machine_name):
+        return f'arn:aws:states:{aws.region_name}:{aws.account}:stateMachine:{state_machine_name}'
+
+    def execution_arn(self, state_machine_name, execution_name):
+        return f'arn:aws:states:{aws.region_name}:{aws.account}:execution:{state_machine_name}:{execution_name}'
+
+    def start_execution(self, state_machine_name, execution_name=None, execution_input=None):
```
hannes-ucsc: Correct me if I am wrong but we don't use the cases were execution name and input are optional. Let's make those parameters mandatory. That would be adding to the value of this class.

src/azul/service/responseobjects/step_function_helper.py
```diff
@@ -0,0 +1,59 @@
...
+        """
```
hannes-ucsc: Remove this docstring except maybe for the reference to the boto3 method. This stuff is all obvious and can be looked up in the Boto3 documentation.

src/azul/service/responseobjects/step_function_helper.py
```diff
@@ -0,0 +1,59 @@
...
+                'startDate': datetime(2015, 1, 1)
+            }
+        """
+        execution_params = {
+            'stateMachineArn': self.state_machine_arn(state_machine_name),
+            'input': json.dumps(execution_input)
+        }
+        if execution_name is not None:
+            execution_params['name'] = execution_name
+        return aws.stepfunctions.start_execution(**execution_params)
```
hannes-ucsc: The return value isn't actually used anywhere so this function should return nothing. Less to test, less to document, less to worry about.

src/azul/service/responseobjects/manifest_service.py
```diff
@@ -0,0 +1,60 @@
...
+                'Status': 302,
+                'Location': execution_output['Location']
+            }
+        elif execution['status'] == 'RUNNING':
+            wait_times = [1, 1, 2, 6, 10]
+            base_url = 'http://localhost:8000' if local else config.service_endpoint()
+            return {
+                'Status': 301,
+                'Retry-After': wait_times[max(0, min(wait, len(wait_times) - 1))],
+                'Location': f'{base_url}{self.manifest_endpoint}?token={self.encode_params(params)}&wait={wait}'
```
hannes-ucsc: See if you can use `app.current_request.context['path']` instead of `self.manifest_endpoint`.

src/azul/service/responseobjects/manifest_service.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if execution['status'] == 'SUCCEEDED':
+            execution_output = json.loads(execution['output'])
+            return {
+                'Status': 302,
+                'Location': execution_output['Location']
+            }
+        elif execution['status'] == 'RUNNING':
+            wait_times = [1, 1, 2, 6, 10]
+            base_url = 'http://localhost:8000' if local else config.service_endpoint()
```
hannes-ucsc: See if you can avoid hard-coding the endpoint using `app.current_request.headers['host']`.

This means you'd probably have to pass `app.current_request` from `app.py` down to the service but I think that's OK for this particular purpose. I still think `app.py` should extract the URL parameters and pass them explicitly as separate method arguments.

src/azul/service/responseobjects/step_function_helper.py
hannes-ucsc: Class is fine. With a module, you'd have to import each function individually or use `*` import which is considered harmful.

lambdas/service/app.py
```diff
@@ -533,6 +536,106 @@ def get_manifest():
...
+        - name: wait
+          in: query
+          type: any
+          description: Integer indicating amount of time to wait before checking status
+    :return: Response with location of the generated manifest
+    """
+    logger = logging.getLogger("dashboardService.webservice.get_manifest")
+
+    if app.current_request.query_params is None:
+        app.current_request.query_params = {}
```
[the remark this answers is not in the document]
Ben-Wu: Looks like I forgot to remove this when I copied it from the old manifest function

src/azul/service/responseobjects/step_function_helper.py
Ben-Wu: We don't use the case where execution_input is None but right now execution_name is always not given because aws will generate a uuid for us

src/azul/service/responseobjects/step_function_helper.py
Ben-Wu: The return value is used to get the execution id which is used to create the retry token

src/azul/service/responseobjects/manifest_service.py
Ben-Wu: That works.  Just need to check the `X-Forwarded-Proto` header for http vs. https and assume if the header is missing then it's http.

src/azul/service/responseobjects/manifest_service.py
Ben-Wu: Yes that cleans things up a bit

src/azul/service/responseobjects/manifest_service.py
Ben-Wu: The reason I made two functions in `app.py` is so I can directly call the second function from a unit test without having to do `requests.get`.  We should be able to move most of what's in `app.py` into `manifest_service.py` and just have one function in each file

src/azul/service/responseobjects/manifest_service.py
```diff
@@ -0,0 +1,60 @@
...
+            return {
+                'Status': 301,
+                'Retry-After': wait_times[max(0, min(wait, len(wait_times) - 1))],
```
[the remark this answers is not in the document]
Ben-Wu: Does that mean also remove `Retry-After` from the response? Or just use a constant value for it?

src/azul/service/responseobjects/manifest_service.py
hannes-ucsc: I meant only when `local` is `True`. When `local` is `False`, the existing method to synthesize the URL is sufficient.

src/azul/service/responseobjects/step_function_helper.py
Ben-Wu: We have the ARN of the state machine but not of the execution of the state machine that we just created.  Name is an ID but the official documentation calls it execution name

src/azul/service/responseobjects/manifest_service.py
Ben-Wu: Wouldn't it be simpler to remove local and always use `app.current_request.headers['host']` to create the URL.  That way the the URL can be created in `app.py` and passed here as an argument and there's just one way to create the URL.

Ben-Wu: Is there some case where `app.current_request.headers['host']` doesn't give us what we want?

Ben-Wu: One thing that just occured to me is that if we don't use a `local` param we can't really test the case where the path is the actual deployed path.

hannes-ucsc: No, you're right. I was worried about someone forging the `Host` header but I checked that API Gateway validates the header so it should be safe to use.

src/azul/service/responseobjects/step_function_helper.py
hannes-ucsc: I prefer this over returning a value:

```suggestion
        assert self.execution_arn(state_machine_name, execution_name) == aws.stepfunctions.start_execution(**execution_params)['executionArn']
```

Just not all on one line.

terraform/step_function.tf.json.template.py
```diff
@@ -0,0 +1,65 @@
...
+                    ]
+                })
+            }
+        },
+        "aws_sfn_state_machine": {
+            "manifest_state_machine": {
+                "name": config.manifest_state_machine_name,
+                "role_arn": "${aws_iam_role.state_machine_iam_role.arn}",
+                "definition": json.dumps({
+                    "Comment": "Write all entities matching a filter to DynamoDB",
```
hannes-ucsc: ?

hannes-ucsc: Do we even need the comments in the definition? I think they are superfluous so if not, please remove.

lambdas/service/app.py
```diff
@@ -533,6 +537,96 @@ def get_manifest():
...
+    parameters:
+        - name: filters
+          in: query
+          type: string
+          description: Filters to be applied when generating the manifest
+        - name: token
+          in: query
+          type: string
+          description: An opaque string describing the manifest generation job
+    :return: Response with either a 302 status and the location of the generated manifest,
```
```suggestion
    :return: A 200 response with a JSON body describing the status of the manifest.
```

lambdas/service/app.py
```diff
@@ -533,6 +537,96 @@ def get_manifest():
...
+        or a 301 status and a URL to re-check the status of the job
```
```suggestion
```

lambdas/service/app.py
```diff
@@ -533,6 +537,96 @@ def get_manifest():
...
+
+        If manifest is still generating the response will look like:
```
```suggestion
        If the manifest is still being generated, the response will look like:
```

lambdas/service/app.py
```diff
@@ -533,6 +537,96 @@ def get_manifest():
...
+        {
+            "Status": 301,
+            "Retry-After": 2,
+            "Location": "https://retry.url"
+        }
+        Status represents an HTTP status code (301 moved permanently)
```
```suggestion
        The `Status` field emulates an HTTP status code (301 moved permanently). Note that the actual HTTP response is still 200. The body of the 200 response merely emulates HTTP in order to bypass the default user agent behavior which, in many cases, disobeys Retry-After. It is intended to be processed by client-side Javascript such that the recommended delay can be handled in Javascript rather relying on a native implementation by the browser.
```

lambdas/service/app.py
```diff
@@ -533,6 +537,96 @@ def get_manifest():
...
+        Retry-After is the recommended number of seconds to wait before rechecking the status of the manifest
+        Location is the URL to make a GET request to in order to recheck the status
```
hannes-ucsc: Empty lines between paragraphs and periods at the end of sentences. Code literals like `Location` should be in back ticks. Preformatted sections like JSON fragments should be between triple back-ticks.

Ben-Wu:
@hannes-ucsc I copied the docstring into a swagger editor and this is what it looks like.  Backslashes are used in the docstring to create empty lines in swagger.  

I don't love the way the documentation looks in code with the indentation and backslashes.  I couldn't find a way to automatically generate swagger through chalice.  The `TODO` in `app.py` is for a flask only library so that todo should probably be changed.  If we want to use sphinx I'm pretty sure we would need to completely change everything so I think it's better to not follow the swagger format until we figure out how we want to do documentation.

screen shot 2018-12-03 at 7 01 38 pm

Ben-Wu:
Removed the swagger specific formatting in the docstring and it's now in the state of @hannes-ucsc's patch

Issue #58, opened 2018-07-13
Authentication on the webservice


cricketsloan:
We need to integrate authentication in the web service.

I think this should be using whatever framework blue provides for authentication

#### Checkpoints for research and planning

- [x] Figure out the authentication process
  - [x] @shiroyuki to discuss with Trent Smith on the authentication process and how to work with Fusillade
  - [x] @shiroyuki to implement a prototype
  - [x] @shiroyuki to discuss with @NoopDog on the exact workflow (required a RFC)
- [x] Figure out the token verification procedure
  - [x] @shiroyuki to implement a prototype

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-45)
┆Project Name: azul
┆Issue Number: AZUL-45
┆Epic: Interactive building of collections

cricketsloan:
➤ Trevor Heathorn commented:

Include Michael Baumann in design discussions.

theathorn:
This ticket is to support Authentication in the HCA Data Portal and Browser.

theathorn:
From @briandoconnor : This is google ID login via Oauth/OIDC.

shiroyuki:
@theathorn, @hannes-ucsc: Azul will rely on Fusillade + Auth0 which currently only give two options to the users: login with Google ID or HCA's Auth0.

**Other Updates:**
* The prototype code is on the `experimentals/auth` branch.
* I think the prototype code is good enough to turn into a production code. So, I'll clean it up.
* The next stop would be a RFC on the authentication procedure and discuss with @NoopDog on how we want the frontend and backend working together.

shiroyuki:
@NoopDog The RFC is available at https://github.com/DataBiosphere/azul/blob/experiments/auth/rfc/azul-0058-authentication.md.

Per our discussion:
* We will go with option 2 (Begin with Azul redirecting to Fusillade; end with Fusillade redirecting back to UI)
* For the time being, the subject of the given JWT (`sub`) is the user identifier.

theathorn:
@hannes-ucsc wrote:
Hi Trevor, there are a few open questions wrt to authn/authz for Azul/data-browser. Would you be OK with arranging a meeting between me and Brian O, Charles, Michael B, Juti, Dave and me? Agenda items would be

* Should the HCA Orange team attempt to implement authz/authn generically or is it ok to bake in a dependency on Fussilade?
* Specifically, should Azul be involved in authentication exchanges?
[...]

PR #563, opened 2018-11-22
RFC and Prototype for Authentication Integration

shiroyuki:
(Related to #58)

RFC: https://allspark.dev.data.humancellatlas.org/dcp-ops/docs/wikis/RFC:-Authentication-Flow


src/azul/security/verifier.py
```diff
@@ -0,0 +1,98 @@
...
+        verification_options = dict(key=public_key,
+                                    issuer=issuer,
+                                    audience=config.access_token_audience_list,
+                                    algorithms=allowed_algorithms)
+        return jwt.decode(token, **verification_options)
+    except KeyError:
+        raise InvalidTokenError('missing information')
+    except PyJWTError:
+        logger.warning(f"Detected a JWT with INVALID SIGNATURE (given: '{token}').", exc_info=True)
+        raise UnverifiableTokenError(token)
```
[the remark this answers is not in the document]
shiroyuki: I agree that we can just raise `InvalidTokenError` here. But I have some idea to utilize this error.

Initially when I implement these exceptions, I thought of how the caller code or maintainers should respond according to a type of error.

| Exception | Interpretation |
| --- | --- |
| `InvalidTokenError` | The token is broken. No security risk. |
| `NonDecodableTokenError` | The token is not decodable. Probably random token. Low security risk |
| `UnverifiableTokenError` | The token is decodable but fails the verification. Expired token? Tampered token? |

So, in the last case, I wonder if we should just log the messages in the warning level and treat as normal invalid token error. Or later on, we should also notify to the DevOps team.

What do you think, @hannes-ucsc?

src/azul/security/verifier.py
```diff
@@ -0,0 +1,98 @@
...
+
+def get_jwks_uri(openid_provider):
+    if openid_provider.endswith(gserviceaccount_domain):
+        return f"https://www.googleapis.com/service_accounts/v1/jwk/{openid_provider}"
+    else:
+        return get_openid_config(openid_provider)["jwks_uri"]
+
+
+def convert_base64_string_to_int(value:str) -> int:
+    return int.from_bytes(urlsafe_b64decode(value + "==="), byteorder="big")
```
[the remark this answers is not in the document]
shiroyuki: But I'll fix that.

shiroyuki:
I would agree with you that option 2 makes the code more manageable and portable. I'll wait for more feedback from @NoopDog.

hannes-ucsc:
Can you resolve the conflicts. Also where is the wheel you mentioned earlier?

shiroyuki:
@hannes-ucsc The unpacked wheels have been pushed to `lambdas/service/vendor`. I'm working on the merge conflict.

shiroyuki:
@hannes-ucsc Merge conflict resolved.

shiroyuki:
Switched back to "in progress" temporarily to add missing tests."

shiroyuki:
For clarification, the review request is for the RFC on the GitLab.

shiroyuki:
**Summary of notable changes up to d49b8ba**

* As patching (with `unittest.mock`) doesn't work when for the endpoint tests,
  * the base class for app test cases will add three endpoints to provide JWKS URI, public keys and generate for-testing-only JWTs to make the app a temporary authentication broker,
  * when the tests run, the private and public keys will be automatically generated once where the key files (actually PEM files) are listed in the `.gitignore` file,
  * the verifier always accept issuers from `http://127.0.0.1:\d{1,5}`.

What have not been addressed (still in progress)?

* Weird padding in `azul.security.verifier.convert_base64_string_to_int`

Please note that these are files added only for deployment.

* `lambdas/service/vendor/.libs_cffi_backend/libffi-d78936b1.so.6.0.4`
* `lambdas/service/vendor/_cffi_backend.cpython-36m-x86_64-linux-gnu.so`
* `lambdas/service/vendor/cffi*`
* `lambdas/service/vendor/cryptography*`

lambdas/service/app.py
```diff
@@ -769,13 +806,67 @@ def files_proxy(uuid):
         dss_response.raise_for_status()
 
 
+def self_url(endpoint_path=None):
```
hannes-ucsc: This method already exists on `develop`.

.gitignore
```diff
@@ -7,6 +7,10 @@
 #
 # - No leading slash means the pattern matches at any depth.
 
+# OpenSSL Private-Public Keys for Testing
```
hannes-ucsc:
```suggestion
# OpenSSL private/public keys for testing authentication
```

But can we we delete these during test clean-up?

hannes-ucsc:
So now there are two reviews. GitHub must have troubles today.

lambdas/service/app.py
shiroyuki: The duplicate has been removed.

shiroyuki:
Summary of Changes:
* Replaced uncompressed wheel/source files with the compressed version.
* Updated `lambdas/service/Makefile` to uncompress the files before `make deploy`.
* The test keys are created and destroy on test app's start and end respectively. (i.e., the keys will be clean up after the test.)
* Resolved conflicts.

test/app_test_case.py
```diff
@@ -1,17 +1,26 @@
...
+from jwt import encode as jwt_encode
 import requests
 from threading import Thread
+from chalice import Response
 # noinspection PyPackageRequirements
 from chalice.local import LocalDevServer
 # noinspection PyPackageRequirements
 from chalice.config import Config as ChaliceConfig
-from azul import config
+from azul import config as azul_config
```
hannes-ucsc: I want this to be`config` consistently across the code base. If some other name conflicts with that, I want that other thing to be renamed.

shiroyuki: ok

shiroyuki: The alias has been removed.

shiroyuki:
@hannes-ucsc Sorry, I forgot to check those wheel files in.

Summary of changes (as of 29f87045461dacf246841ef66ac3127474b5ab76):
* Simplified the deployment process (that part that uncompresses the wheel files).
* Added the missing wheel files.
* Moved key generation/removal to setUpClass and tearDownClass to speed up the tests.

.gitignore
```diff
@@ -7,6 +7,10 @@
 #
 # - No leading slash means the pattern matches at any depth.
 
+# OpenSSL Private-Public Keys for testing authentication
```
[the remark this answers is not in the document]
hannes-ucsc: Why is this STILL here?

src/azul/__init__.py
```diff
@@ -309,6 +309,17 @@ def cart_item_state_machine_name(self):
     def cart_api_ip_whitelist(self):  # TODO: Remove when authentication is added
         return os.environ['AZUL_CART_API_IP_WHITELIST'].split(' ')
 
+    @property
+    def access_token_issuer(self):
+        return "https://humancellatlas.auth0.com/"
```
```suggestion
        return "https://humancellatlas.auth0.com"
```

If at all possible. We use base URLs that don't end in slash because …

src/azul/__init__.py
```diff
@@ -309,6 +309,17 @@ def cart_item_state_machine_name(self):
...
+
+    @property
+    def access_token_audience_list(self):
+        return [
+            f"https://{self.deployment_stage}.data.humancellatlas.org/",
+            f"{self.access_token_issuer}userinfo"
```
hannes-ucsc: … it makes this look nicer:

```suggestion
            f"{self.access_token_issuer}/userinfo"
```

src/azul/security/auth.py
```diff
@@ -0,0 +1,102 @@
...
+from urllib.parse import urlencode
+
+from azul import config
+from azul.security.verifier import (verify,
+                                    InvalidTokenError,
+                                    NonDecodableTokenError)
+
+logger = getLogger(__name__)
+
+fusillade_callback_url = f'{config.service_endpoint()}/auth/callback'
```
hannes-ucsc:
```suggestion
fusillade_callback_url = {config.service_endpoint() + '/auth/callback'
```

src/azul/security/auth.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def get_fusillade_hostname():
+    # This can be moved to config.
```
hannes-ucsc: Please do that.

src/azul/security/auth.py
```diff
@@ -0,0 +1,102 @@
...
+    deployment_stage = get_fusillade_stage()
+    fusillade_subdomain =f'auth.{deployment_stage}' if deployment_stage else 'auth'
+    return f'{fusillade_subdomain}.data.humancellatlas.org'
+
+
+def get_fusillade_stage() -> str:
```
hannes-ucsc: I don't like expressing this mapping from deployment stage to Fusillade stage as code. We use `deployments/*/environment` for expressing differences between deployments.

src/azul/security/auth.py
```diff
@@ -0,0 +1,102 @@
...
+    return authenticate(authorization_token)
+
+
+def authenticate(authorization_token:str) -> Dict[str, Any]:
+    try:
+        bearer_token_prefix = "Bearer "
+        assert authorization_token.startswith(bearer_token_prefix), "not_bearer_token"
+        access_token = authorization_token[len(bearer_token_prefix):]
+        assert access_token, "missing_bearer_token"
+    except AssertionError as e:
```
hannes-ucsc: This is misuse of assertions.

src/azul/security/verifier.py
```diff
@@ -0,0 +1,97 @@
...
+from base64 import urlsafe_b64decode
+from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.backends import default_backend
+import jwt
+from jwt.exceptions import DecodeError, PyJWTError
+import requests
+
+from azul import config
+
+logger = getLogger(__name__)
```
hannes-ucsc:
```suggestion
logger = logging.getLogger(__name__)
```

shiroyuki:
@hannes-ucsc I think the problem might be due to Python 3.7, which I install via `brew`. But I'll investigate further.

hannes-ucsc:
Indeed, it is caused by Python 3.7. @danielsotirhos ran into a similar issue.

shiroyuki:
@hannes-ucsc In that case, I think we don't need these wheels.

shiroyuki:
I'll remove those wheels.

common.mk
```diff
@@ -34,3 +34,7 @@ endif
 # The template output file depends on the template file, of course, as well as the environment. To be safe we force the
 # template creation. This is what the fake .FORCE target does. It still is necessary to declare a target's dependency on
 # a template to ensure correct ordering.
+
+vendor/%.dist-info: $(AZUL_HOME)/wheels/%-*-*-*.whl
```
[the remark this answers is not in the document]
hannes-ucsc: Yes, please remove.

src/azul/security/auth.py
hannes-ucsc: Reads more clearly due to the whitespace around +

shiroyuki: Oh... I see. This will be changed in the next commit.

shiroyuki:
This pattern rule has been removed.
``` Makefile
vendor/%.dist-info: $(AZUL_HOME)/wheels/%-*-*-*.whl
	unzip $< -d vendor
	test -x $@
```

shiroyuki:
The wheels are now removed.

src/azul/security/verifier.py
```diff
@@ -0,0 +1,111 @@
...
+
+from azul import config
+
+logger = logging.getLogger(__name__)
+
+allowed_algorithms = ('RS256',)
+gserviceaccount_domain = "iam.gserviceaccount.com"
+
+# Shared HTTP session
+session = requests.Session()
```
hannes-ucsc: No side-effects at module loading/import time, please. Please fix in a follow-up PR.

environment
```diff
@@ -226,6 +229,8 @@ _set AZUL_CART_API_IP_WHITELIST  # TODO: Remove when authentication is added
 
 _set azul_dss_query_prefix
 
+_set AZUL_FUSILLADE_HOSTNAME "auth.dev.data.humancellatlas.org"
```
hannes-ucsc: Other services endpoints are configured via a "endpoint" variable e.g., a base URL to which we append. This would allow for more flexible configuration.

hannes-ucsc:
Sorry, I found merge errors after merging and backed this out again. Please file a new PR with 

- `AZUL_FUSSILADE_ENDPOINT` instead of AZUL_FUSILLADE_HOSTNAME
- no side effect at module import time
- and the merge error fixed (I'll point it out) in a review comment

test/app_test_case.py
```diff
@@ -65,6 +163,10 @@ def base_url(self):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        generate_test_keys()
+        cls._path_to_app = os.path.join(config.project_root, 'lambdas', cls.lambda_name())
```
hannes-ucsc: Merge error, please remove.

test/app_test_case.py
```diff
@@ -65,6 +163,10 @@ def base_url(self):
...
+        sys.path.append(cls._path_to_app)
```
hannes-ucsc: Merge error, please remove.

test/app_test_case.py
```diff
@@ -74,6 +176,8 @@ def setUpClass(cls):
     @classmethod
     def tearDownClass(cls):
         cls.app_module = None
+        sys.path.remove(cls._path_to_app)
```
hannes-ucsc: Merge error, please remove.

shiroyuki:
Got it. I'll look into it now.

PR #573, opened 2018-11-29
[WIP] Gitlab Dev Environment CD

NoopDog:
Need to discuss:
1.  Can we skip virtual env as the environment is ephemeral?
1. How to control the env that is deployed to (source environment / git branch)?
1. Tagging
1. Reindex vs delete and reindex
1. Running tests

### DIFF
--- a/.gitlab-ci.lower-env.yml
+++ b/.gitlab-ci.lower-env.yml
@@ -16,6 +16,8 @@ build_noopdog:
   environment:
     name: noopdog
   script:
+    - virtualenv -p python3 .venv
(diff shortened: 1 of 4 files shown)


NoopDog:
Open questions are:
@hannes-ucsc 

1. How should we do the tagging? Need to discuss how the DSS way could get into our workflow.
1. Need to confirm if we want to do builds that require an IAM update manually rather than giving the AllSpark user the capability to do that.

hannes-ucsc:
Superseded by #636.

PR #583, opened 2018-12-05
Save query/URL shortener

Ben-Wu:
An example URL is http://dnastack2.dev.url.data.humancellatlas.org/aJL

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -31,9 +31,9 @@ changes:
       - clients
       - environment
     notes: >
-           The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL.
+           The `/url` endpoint was added to return a URL that redirects to a given long URL.
            Documentation: /lambdas/service/app.py#L701
-           AZUL_URL_REDIRECT_DOMAIN_NAME and AZUL_URL_REDIRECT_S3_BUCKET should be copied from the example
+           AZUL_URL_REDIRECT_BASE_DOMAIN_NAME and AZUL_URL_REDIRECT_FULL_DOMAIN_NAME should be copied from the example
            deployment environment to local deployments
 
 
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -29,11 +29,9 @@ changes:
       - clients
     notes: >
            The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL.
-           Documentation:
-           https://github.com/DataBiosphere/azul/blob/2ab5c91d13e576a0e831298681e221ab0048bbcd/lambdas/service/app.py#L701
-           Running `make terraform` will delete the previously existing azul-storage- bucket.
-           This will fail if the bucket is not empty so manual deletion is required.
-           The storage bucket should have only been used for manifests and has a 1 day expiration policy.
+           Documentation: /lambdas/service/app.py#L701
+           Running `make terraform` after upgrading will delete the previously existing `azul-storage-...` bucket.
+           This will fail if the bucket is not empty and manual deletion will be required.
 
   - title: Make the manifest endpoint dual mode (standard HTTP and lifted 200)
     issues:
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -27,11 +27,13 @@ changes:
       - terraform
       - deploy
       - clients
-    notes: The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL
-           Documentation here -
+    notes: >
+           The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL.
+           Documentation:
+           https://github.com/DataBiosphere/azul/blob/2ab5c91d13e576a0e831298681e221ab0048bbcd/lambdas/service/app.py#L701
            Running `make terraform` will delete the previously existing azul-storage- bucket.
            This will fail if the bucket is not empty so manual deletion is required.
-           The storage bucket should have only been used for manifests and had a 1 day expiration policy.
+           The storage bucket should have only been used for manifests and has a 1 day expiration policy.
 
   - title: Make the manifest endpoint dual mode (standard HTTP and lifted 200)
     issues:
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -9,6 +9,8 @@ meta:
     - terraform: (requires `make terraform`)
     - requirements: (requires `pip -Ur requirements.dev.txt`)
     - clients: the service API changed in a way that requires clients to be updated
+    - environment: an environment variable was added to the `/deployments/.example.local` deployment
+                   and needs to be copied to local deployments
 
 changes:
   - title: Emulate HTTP redirects and retry-after in DSS file proxy
@@ -27,11 +29,13 @@ changes:
       - terraform
       - deploy
       - clients
+      - environment
     notes: >
            The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL.
            Documentation: /lambdas/service/app.py#L701
-           Running `make terraform` after upgrading will delete the previously existing `azul-storage-...` bucket.
-           This will fail if the bucket is not empty and manual deletion will be required.
+           AZUL_URL_REDIRECT_DOMAIN_NAME and AZUL_URL_REDIRECT_S3_BUCKET should be copied from the example
+           deployment environment to local deployments
+
 
   - title: Make the manifest endpoint dual mode (standard HTTP and lifted 200)
     issues:
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -20,6 +20,19 @@ changes:
     notes: Refer to docstring for service API handler `/fetch/dss/files/{uuid}`.
            Note that the DSS file proxy endpoint changed from `/dss/files/{uuid}` to `/fetch/dss/files/{uuid}`
 
+  - title: Add endpoint to shorten URLs
+    issues:
+      - https://github.com/databiosphere/azul/issues/66
+    upgrade:
+      - terraform
+      - deploy
+      - clients
+    notes: The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL
+           Documentation here -
+           Running `make terraform` will delete the previously existing azul-storage- bucket.
+           This will fail if the bucket is not empty so manual deletion is required.
+           The storage bucket should have only been used for manifests and had a 1 day expiration policy.
+
   - title: Make the manifest endpoint dual mode (standard HTTP and lifted 200)
     issues:
       - https://github.com/DataBiosphere/azul/issues/567
--- a/deployments/.example.local/environment
+++ b/deployments/.example.local/environment
@@ -19,8 +19,8 @@ _set AZUL_SUBSCRIBE_TO_DSS 0
 _set AZUL_DOMAIN_NAME "dev.explore.data.humancellatlas.org"
 _set AZUL_SUBDOMAIN_TEMPLATE "{lambda_name}.$AZUL_DEPLOYMENT_STAGE"
 
-_set AZUL_URL_REDIRECT_DOMAIN_NAME "dev.url.data.humancellatlas.org"
-_set AZUL_URL_REDIRECT_S3_BUCKET "$AZUL_DEPLOYMENT_STAGE.$AZUL_URL_REDIRECT_DOMAIN_NAME"
+_set AZUL_URL_REDIRECT_BASE_DOMAIN_NAME "dev.url.data.humancellatlas.org"
+_set AZUL_URL_REDIRECT_FULL_DOMAIN_NAME "$AZUL_DEPLOYMENT_STAGE.$AZUL_URL_REDIRECT_BASE_DOMAIN_NAME"
 
 # All personal deployments should share an ES domain with `dev`
 #
--- a/deployments/.example.local/environment
+++ b/deployments/.example.local/environment
@@ -19,6 +19,9 @@ _set AZUL_SUBSCRIBE_TO_DSS 0
 _set AZUL_DOMAIN_NAME "dev.explore.data.humancellatlas.org"
 _set AZUL_SUBDOMAIN_TEMPLATE "{lambda_name}.$AZUL_DEPLOYMENT_STAGE"
 
+_set AZUL_URL_REDIRECT_DOMAIN_NAME "dev.url.data.humancellatlas.org"
+_set AZUL_URL_REDIRECT_S3_BUCKET "$AZUL_DEPLOYMENT_STAGE.$AZUL_URL_REDIRECT_DOMAIN_NAME"
+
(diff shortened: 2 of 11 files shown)


terraform/s3.tf.json.template.py
```diff
@@ -19,6 +16,25 @@
...
+            },
+            "aws_s3_bucket_object": {
+                "default_query_redirect": {
+                    "bucket": config.s3_redirect_bucket,
+                    "acl": "public-read",
+                    "key": "error.html",
+                    "content": " ",
+                    # TODO: Should not redirect to dev but there is no prod subdomain
+                    #  and preview doesn't have a 404 page
+                    "website_redirect": "https://dev.data.humancellatlas.org/404/"
```
Ben-Wu: `data.humancellatlas.org/404/` doesn't exist right now so I put `dev`.  We could also use the 404 of the www subdomain or not redirect and show the s3 403 key not found page.

lambdas/service/app.py
```diff
@@ -688,3 +693,68 @@ def _file_url(uuid, **kwargs):
     return (config.service_endpoint() + '/dss/files' +
             '/' + urllib.parse.quote(uuid, safe='') +
             '?' + urllib.parse.urlencode(kwargs))
+
+
+@app.route('/query', methods=['POST'], cors=True)
```
Ben-Wu: I don't feel great about this path so any suggestions would be appreciated.  `query` kind of makes sense to me because I see it as creating a query resource.  `/resources/query`?

hannes-ucsc: Let's just do`/urls`.

terraform/s3.tf.json.template.py
hannes-ucsc: I think the decision to redirect to the front-end was a mistake. We shouldn't assume which frontend is hitting us. Can we just return a generic, body-less 404 response?

src/azul/__init__.py
```diff
@@ -221,6 +225,10 @@ def manifest_lambda_basename(self):
     def manifest_state_machine_name(self):
         return config.qualified_resource_name('manifest')
 
+    @property
+    def url_shortener_whitelist(self):
+        return ['.*humancellatlas.org']
```
```suggestion
        return [r'.*humancellatlas\.org']
```

hannes-ucsc:
Sorry, I think I messed up your branch by adding a merge from develop at the tip. I clicked the "Update branch button" thinking it would rebase but instead it merged.

If you rebase, the merge should disappear.

I've since disabled that branch requirement—it doesn't make sense since we build both the PR branch and a simulated merged against the base branch.

terraform/s3.tf.json.template.py
Ben-Wu: We can let s3 return a 404 by making the entire bucket public and not setting an error document.  The response body isn't empty though.

It will look like this:
```
404 Not Found

    Code: NoSuchKey
    Message: The specified key does not exist.
    Key: 404
    RequestId: ...
    HostId: ...
```

I can't find a way to return an empty 404 but I think this is fine.

Ben-Wu: We can also keep the bucket private and return 403 if the object isn't found

Ben-Wu:
We can use the same bucket for manisfest and urls.  I assume that we wouldn't want public read access for the entire bucket so that would mean that we would return 403s for invalid shortened urls instead of 404.

Ben-Wu:
Switched to `azul-public-` and `azul-private-` buckets

lambdas/service/app.py
```diff
@@ -688,3 +693,68 @@ def _file_url(uuid, **kwargs):
...
+                                    WebsiteRedirectLocation=url)
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'InvalidRedirectLocation':
+                    raise BadRequestError('Invalid URL given')
+                raise
+            return get_url_response(key)
+        if existing_url == url:
+            return get_url_response(key)
+        key_length += 1
+    raise ChaliceViewError('Could not create URL')
```
```suggestion
    raise ChaliceViewError('Could not create shortened URL')
```

lambdas/service/app.py
```diff
@@ -688,3 +693,68 @@ def _file_url(uuid, **kwargs):
...
+
+
+def encode_url(url):
+    url_hash = hashlib.sha1(bytes(url, encoding='utf-8')).digest()
```
hannes-ucsc: We should consider normalizing the URL. Probably not now, though.

https://pypi.org/project/url-normalize/

CHANGELOG.yaml
```diff
@@ -12,6 +12,21 @@ meta:
...
+    issues:
+      - https://github.com/databiosphere/azul/issues/66
+    upgrade:
+      - terraform
+      - deploy
+      - clients
+    notes: >
+           The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL.
+           Documentation:
+           https://github.com/DataBiosphere/azul/blob/2ab5c91d13e576a0e831298681e221ab0048bbcd/lambdas/service/app.py#L701
```
hannes-ucsc: This link won't work, unfortunately, because it points to the commit on the feature branch, not `develop` where it will ultimately end up. This is why I am looking forward to generating the changelog from the commit history where we could dynamically generate these links on the fly.

Just do

```suggestion
           /lambdas/service/app.py#L701
```

or leave this out and file it as a separate PR after this PR has landed. I think the former is more pragmatic.

CHANGELOG.yaml
```diff
@@ -12,6 +12,21 @@ meta:
...
+           Running `make terraform` will delete the previously existing azul-storage- bucket.
```
hannes-ucsc:
```suggestion
           Running `make terraform` after upgrading will delete the previously existing `azul-storage-…` bucket.
```

CHANGELOG.yaml
```diff
@@ -12,6 +12,21 @@ meta:
...
+           This will fail if the bucket is not empty so manual deletion is required.
```
hannes-ucsc:
```suggestion
           This will fail if the bucket is not empty and manual deletion will be required.
```

CHANGELOG.yaml
```diff
@@ -12,6 +12,21 @@ meta:
...
+           The storage bucket should have only been used for manifests and has a 1 day expiration policy.
```
hannes-ucsc: I think I know why you mention this (to say that the bucket is likely to be empty) but I would leave it out.

lambdas/service/app.py
```diff
@@ -688,3 +693,68 @@ def _file_url(uuid, **kwargs):
...
+        except GetObjectError:
+            try:
+                storage_service.put(key,
+                                    data=bytes(url, encoding='utf-8'),
+                                    ACL='public-read',
+                                    WebsiteRedirectLocation=url)
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'InvalidRedirectLocation':
+                    raise BadRequestError('Invalid URL given')
+                raise
```
hannes-ucsc: For clarity and symmetry I tend to put this in an `else` branch, even it is redundant.

lambdas/service/app.py
```diff
@@ -688,3 +693,68 @@ def _file_url(uuid, **kwargs):
...
+
+    def get_url_response(path):
+        return {'url': f'http://{config.s3_public_bucket}.s3-website-{aws.region_name}.amazonaws.com/{path}'}
+
+    key_length = 3
+    while key_length <= len(encoded_url):
+        key = f'url/{encoded_url[:key_length]}'
+        try:
+            existing_url = storage_service.get(key)
+        except GetObjectError:
```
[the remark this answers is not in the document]
Ben-Wu: I agree that `GetObjectError` isn't needed but I do think it's a little cleaner and easier to follow to except `GetObjectError` than `storage_service.client.exceptions.NoSuchKey`.  I don't feel strongly either way though so I can remove it.

lambdas/service/app.py
Ben-Wu: I realize now that we can just except `storage_service.client.exceptions.InvalidRedirectLocation` which makes more sense since we can except `NoSuchKey` above anyway.

Ben-Wu:
I can do that. Shouldn't be a problem

lambdas/service/app.py
```diff
@@ -735,3 +740,66 @@ def _file_url(uuid, **kwargs):
...
+        - name: url
+          in: body
+          type: string
+          description: URL to shorten
+
+    :return: A 200 response with JSON body containing the shortened URL:
+
+    ```
+    {
+        "url": "http://shortened.url/ABC"
```
hannes-ucsc: We can probably have a more realistic example now that we have a custom domain name.

terraform/s3.tf.json.template.py
```diff
@@ -19,6 +16,27 @@
                             "days": 1
                         }
                     }
+                },
+                "public_bucket": {
+                    "bucket": config.s3_public_bucket,
+                    "acl": "public-read",
+                    "website": {
+                        "index_document": "error.html"
```
hannes-ucsc: Does this file need to exist in the bucket? If so can we create it with Terraform?

https://www.terraform.io/docs/providers/aws/r/s3_bucket_object.html

hannes-ucsc: If it's intentional that the file doesn't exist, a comment should explain that.

Ben-Wu: It's intentionally non-existent so that the URL will give a 404.  It's there just because it's a required field.  We can also create an object through terraform if we wanted to.

Ben-Wu:
@hannes-ucsc The custom domain isn't working right now.  I'll fix it tomorrow and let you know when you can test it.  I also think the way the custom domain is built right now isn't going to work for non-personal deployments.  I think dev would end up as `dev.dev.explore...` and prod would have `prod` in it.  The solution I have in mind requires adding another environment variable.  Would avoiding this be preferred?

Ben-Wu:
As far as I can tell, to use an alias to the bucket, the subdomain needs to be the same as the bucket name, e.g. `azul-public-deployment.dev.explore.data.humancellatlas.com`.  If we want `deployment.dev.explore.data.humancellatlas.com` we would need to use a CNAME.  I'm not sure if this will cause any problems for the non-personal deployments.

Ben-Wu:
@hannes-ucsc Deployed and can be tested at https://service.dnastack2.dev.explore.data.humancellatlas.org/url

lambdas/service/app.py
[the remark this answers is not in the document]
hannes-ucsc: > follow to except GetObjectError than storage_service.client.exceptions.NoSuchKey

It's shorter but misleading since it creates the impression that we're catching any error that occurred during GetObject. Accuracy over concision, I'd say. If you'd like to keep the custom exception, you could also rename GetObjectError to NoSuchObjectError or similar.

CHANGELOG.yaml
```diff
@@ -20,6 +22,21 @@ changes:
...
+  - title: Add endpoint to shorten URLs
+    issues:
+      - https://github.com/databiosphere/azul/issues/66
+    upgrade:
+      - terraform
+      - deploy
+      - clients
+      - environment
+    notes: >
+           The `/url` endpoint was added to accept URLs and return a URL redirecting to the given URL.
```
hannes-ucsc:
```suggestion
           The `/url` endpoint was added to return a URL that redirects to a given long URL.
```

environment
```diff
@@ -107,6 +107,12 @@ azul_vars=(
     # content requested by client.
     AZUL_S3_BUCKET
 
+    # URL pointing to the static website hosted at the S3 bucket for URL redirection
```
hannes-ucsc: Domain names aren't URLs. Those are distinct concepts.

deployments/.example.local/environment
```diff
@@ -19,6 +19,9 @@ _set AZUL_SUBSCRIBE_TO_DSS 0
 _set AZUL_DOMAIN_NAME "dev.explore.data.humancellatlas.org"
 _set AZUL_SUBDOMAIN_TEMPLATE "{lambda_name}.$AZUL_DEPLOYMENT_STAGE"
 
+_set AZUL_URL_REDIRECT_DOMAIN_NAME "dev.url.data.humancellatlas.org"
```
hannes-ucsc: This contradicts your statement from Slack that bucket name and the domain name have to be the same for CNAMEs.

environment
hannes-ucsc: Also, isn't there a dependency between the two? They must be the same, correct? Why do we need two variables in that case.

lambdas/service/app.py
```diff
@@ -779,3 +783,69 @@ def self_url(endpoint_path=None):
...
+        - name: url
+          in: body
+          type: string
+          description: URL to shorten
+
+    :return: A 200 response with JSON body containing the shortened URL:
+
+    ```
+    {
+        "url": "http://url.data.humancellatlas.org/ABC"
```
```suggestion
        "url": "http://url.data.humancellatlas.org/b3N"
```

So it doesn't look like hex.

deployments/.example.local/environment
Ben-Wu: I.e. `AZUL_URL_REDIRECT_S3_BUCKET` is used as both the bucket name and the actual redirection url. I know that isn't clearly stated.

Ben-Wu:
Changed environment variables to `AZUL_URL_REDIRECT_BASE_DOMAIN_NAME` and `AZUL_URL_REDIRECT_FULL_DOMAIN_NAME`.  I'm definitely having trouble figuring out good names for these.

Ben-Wu:
@hannes-ucsc Did you somehow look at an outdated version of the terraform file? 
https://github.com/DataBiosphere/azul/pull/583/files#diff-cb3704d9564e517b7af9212f939a547e
The CNAME is created there and put in the correct zone.

hannes-ucsc:
> Did you somehow look at an outdated version of the terraform file?

I must have.

hannes-ucsc:
I had to back this out. It gave the following error during `make terraform`:

```
Error: Error applying plan:

1 error(s) occurred:

* aws_route53_record.url_redirect_record: 1 error(s) occurred:

* aws_route53_record.url_redirect_record: [ERR]: Error building changeset: InvalidChangeBatch: [RRSet of type CNAME with DNS name dev.url.data.humancellatlas.org. is not permitted at apex in zone dev.url.data.humancellatlas.org.]
	status code: 400, request id: 20646bbe-ff62-11e8-aca2-8987dd949512
```

Doesn't make sense to my why it's possible for a SOA and an A record of the same name to coexist, but not a SOA and a CNAME.

https://stackoverflow.com/questions/20215729/rrset-of-type-cname-with-dns-name-foo-com-is-not-permitted-at-apex-in-zone-bar

We could try alias records again or do `dev.dev.url.data.humancellatlas.org`.

Ben-Wu:
CNAME would be better than alias because we can make it shorter.  I think `dev.dev.url.data.humancellatlas.org` is acceptable since it's dev anyway.  Although prod would need to be `prod.url.data.humancellatlas.org` or `short...` or `s...` or something like that.

PR #589, opened 2018-12-06
Added Bundle Counter Script

GPelayo:


scripts/count-bundles.py
```diff
@@ -0,0 +1,184 @@
...
+                    'project_uuid': '0ec2b05f-ddbe-4e5a-b30f-e81f4b1e330c',
+                    'expected_count': 2244
+                }
+            ]
+        },
+    }
+
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--stage', '-s')
```
hannes-ucsc: Probably overkill given all the hard-coded stuff above, no?

PR #593, opened 2018-12-07
Make the manifest endpoint dual mode (fixes #567)

### LINKED ISSUE #567: Make the manifest endpoint dual mode (standard HTTP and lifted 200)
There are two clients that consume the manifest endpoint: users downloading a manifest and the matrix service. We don't want to encumber the matrix service with the lifted 200 responses that we create so the data browser can handle the redirects in Javascript.

Add a URL param that enables the lifted responses. By default the response should be standard HTTP 301 and 302s with `Location` and `Retry-After` headers.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-363)
┆Project Name: azul
┆Issue Number: AZUL-363

Ben-Wu:


lambdas/service/app.py
```diff
@@ -619,7 +641,15 @@ def start_manifest_generation():
     retry_url = f'{protocol}://{base_url}{endpoint_path}'
 
     try:
-        return manifest_service.get_manifest_status(token, retry_url)
+        status_code, retry_after, location = manifest_service.get_manifest_status(token, retry_url, browser_request)
```
hannes-ucsc: I think it's fine if the service method returns the JSON rather than a tuple with the JSON being generated here.

lambdas/service/app.py
```diff
@@ -583,6 +603,7 @@ def start_manifest_generation():
     ```
     {
         "Status": 302,
+        "Retry-After": 0,
```
hannes-ucsc: I'd prefer not to have Retry-After present so as to avoid ambiguity. 302 implies that the client doesn't have to wait.

hannes-ucsc: Also I just told @achave11 to remove it from the equivalent of this for the DSS file proxy.

hannes-ucsc: But I am open to discussion about this.

Ben-Wu:
Yes the browser parameter needs to be appended.  How could we use the Accept header?

lambdas/service/app.py
Ben-Wu: I have the Retry-After here for consistency because the service method is returning a tuple so the view function just builds a response from that without checking the values.

Ben-Wu: But that isn't relevant if the service returns a whole response instead of a tuple which I'll probably change to

hannes-ucsc:
I think we should use a different URL path to enable the lifted responses. I considered Accept but that would prevent us from letting clients specify different manifest formats. I considered a special request header which is easier to do than append a parameter to the URL (no ambiguity wrt to ? vs & as the separator). The path seems the most robust mechanism. It also fits nicely with your auto-detection of the path from the request headers.

I think we should use `/fetch/manifest/files` or `/xhr/manifest/files` or `/browser/manifest/files`. Fetch is the substitute API for XHR. "Browser" seems ambiguous and generic.

hannes-ucsc:
I'll probably ask @achave11 to change `/dss/files` to `/fetch/files` as well.

PR #616, opened 2018-12-15
URL shortener V2

Ben-Wu:


environment
```diff
@@ -107,6 +107,17 @@ azul_vars=(
     # content requested by client.
     AZUL_S3_BUCKET
 
+    # Name of the Route 53 zone used for shortened URLs.
+    # This hosted zone will have to be created manually prior to running
+    # `make terraform`. Personal deployments typically share a zone with the
+    # `dev` deployment.
+    AZUL_URL_REDIRECT_BASE_DOMAIN_NAME
```
hannes-ucsc: Document the empty case and wrap on 80 chars.

terraform/s3.tf.json.template.py
```diff
@@ -1,9 +1,32 @@
-import boto3
-import json
 from azul import config
-from azul.deployment import aws
 from azul.template import emit
 
+# Route 53 records are only created for dev environments (not staging, intergration, or prod)
+data = {
```
hannes-ucsc: I don't think this extracted variable adds any clarity. Quite the opposite, actually.

The alternative is to extract the entire argument do emit() into a variable and then use `if` statements to conditionally modify it before its passed to emit(). But can we stick with precedent here and just inline the variable?

Ben-Wu:
Rebased, deployed, and tested both url shortening and manifest

hannes-ucsc:
One more conflict, but only a trivial one. I rebased, resolved and force pushed. I don't think this needs to be deployed and tested again.

Issue #629, opened 2018-12-21
Implement default cart behaviour


Ben-Wu:
This was brought up in the original RFC by @hannes-ucsc

https://github.com/DataBiosphere/azul/blob/develop/rfc/cart-api-v1.md

This may require more speccing out.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-399)
┆Project Name: azul
┆Issue Number: AZUL-399
┆Epic: Interactive building of collections

PR #631, opened 2018-12-24
Implemented the default cart logic

shiroyuki:
(Related to #629)

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -954,9 +954,9 @@ def get_cart(cart_id):
     """
     Get the cart of the given ID belonging to the user
 
-    The default cart can be accessible by using the actual cart UUID or
-    setting the requesting cart ID to "default". If the default cart does
-    not exist, the endpoint will create one automatically.
+    The default cart is accessible under its the actual cart UUID or by passing
+    "default" as the cart ID. If the default cart does not exist, the endpoint
+    WILL NOT create the default one automatically.
 
     This endpoint returns a 404 error if the cart does not exist or does not belong to the user.
 
@@ -965,10 +965,13 @@ def get_cart(cart_id):
         "CartId": str
     }
     """
-    cart_id = None if cart_id == 'default' else cart_id
     user_id = get_user_id()
+    cart_item_manager = CartItemManager()
     try:
-        cart = CartItemManager().get_cart(user_id, cart_id)
+        if cart_id == 'default':
+            cart = cart_item_manager.get_default_cart(user_id)
+        else:
+            cart = cart_item_manager.get_cart(user_id, cart_id)
     except ResourceAccessError:
         raise NotFoundError('Cart does not exist')
     else:
@@ -1051,9 +1054,9 @@ def get_items_in_cart(cart_id):
     """
     Get a list of items in a cart
 
-    The default cart can be accessible by using the actual cart UUID or
-    setting the requesting cart ID to "default". If the default cart does
-    not exist, the endpoint will create one automatically.
+    The default cart is accessible under its the actual cart UUID or by passing
+    "default" as the cart ID. If the default cart does not exist, the endpoint
+    will create one automatically.
 
     Returns a 404 error if the cart does not exist or does not belong to the user
 
@@ -1088,9 +1091,9 @@ def add_item_to_cart(cart_id):
     """
     Add cart item to a cart and return the ID of the created item
 
-    The default cart can be accessible by using the actual cart UUID or
-    setting the requesting cart ID to "default". If the default cart does
-    not exist, the endpoint will create one automatically.
+    The default cart is accessible under its the actual cart UUID or by passing
+    "default" as the cart ID. If the default cart does not exist, the endpoint
+    will create one automatically.
 
     Returns a 404 error if the cart does not exist or does not belong to the user
     Returns a 400 error if an invalid item was given
@@ -1142,9 +1145,9 @@ def delete_cart_item(cart_id, item_id):
     """
     Delete an item from the cart
 
-    The default cart can be accessible by using the actual cart UUID or
-    setting the requesting cart ID to "default". If the default cart does
-    not exist, the endpoint will create one automatically.
+    The default cart is accessible under its the actual cart UUID or by passing
+    "default" as the cart ID. If the default cart does not exist, the endpoint
+    will create one automatically.
 
     Returns a 404 error if the cart does not exist or does not belong to the user, or if the item does not exist
 
@@ -1172,9 +1175,9 @@ def add_all_results_to_cart(cart_id):
     """
     Add all entities matching the given filters to a cart
 
-    The default cart can be accessible by using the actual cart UUID or
-    setting the requesting cart ID to "default". If the default cart does
-    not exist, the endpoint will create one automatically.
+    The default cart is accessible under its the actual cart UUID or by passing
+    "default" as the cart ID. If the default cart does not exist, the endpoint
+    will create one automatically.
 
     parameters:
         - name: filters
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -912,7 +912,7 @@ def transform_cart_to_response(cart):
     return {
         'CartId': cart['CartId'],
         'CartName': cart['CartName'],
-        'DefaultCart': bool(cart['DefaultCart'])
+        'IsDefaultCart': bool(cart['DefaultCart'])
     }
 
 
@@ -933,7 +933,7 @@ def create_cart():
         {
             "CartName": str,
             "CartId": str,
-            "DefaultCart": bool
+            "IsDefaultCart": bool
         }
     """
     user_id = get_user_id()
@@ -948,7 +948,7 @@ def create_cart():
     return {
         'CartId': cart_id,
         'CartName': cart_name,
-        "DefaultCart": False
+        "IsDefaultCart": False
     }
 
 
@@ -967,7 +967,7 @@ def get_cart(cart_id):
     :return: {
         "CartName": str,
         "CartId": str,
-        "DefaultCart": bool
+        "IsDefaultCart": bool
     }
     """
     user_id = get_user_id()
@@ -988,7 +988,7 @@ def get_all_carts():
             {
                 "CartName": str,
                 "CartId": str,
-                "DefaultCart": bool
+                "IsDefaultCart": bool
             },
             ...
         ]
@@ -1010,7 +1010,7 @@ def delete_cart(cart_id):
         {
             "CartName": str,
             "CartId": str,
-            "DefaultCart": bool
+            "IsDefaultCart": bool
         }
     """
     user_id = get_user_id()
@@ -1037,7 +1037,7 @@ def update_cart(cart_id):
         {
             "CartName": str,
             "CartId": str,
-            "DefaultCart": bool
+            "IsDefaultCart": bool
         }
     """
     user_id = get_user_id()
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -965,10 +965,10 @@ def get_cart(cart_id):
         "CartId": str
     }
     """
+    cart_id = None if cart_id == 'default' else cart_id
     user_id = get_user_id()
-    manager = CartItemManager()
     try:
-        cart = manager.get_cart(user_id, cart_id)
+        cart = CartItemManager().get_cart(user_id, cart_id)
     except ResourceAccessError:
         raise NotFoundError('Cart does not exist')
     else:
@@ -1051,6 +1051,10 @@ def get_items_in_cart(cart_id):
     """
     Get a list of items in a cart
 
+    The default cart can be accessible by using the actual cart UUID or
+    setting the requesting cart ID to "default". If the default cart does
+    not exist, the endpoint will create one automatically.
+
     Returns a 404 error if the cart does not exist or does not belong to the user
 
     :return: {
@@ -1068,6 +1072,7 @@ def get_items_in_cart(cart_id):
         ]
     }
     """
+    cart_id = None if cart_id == 'default' else cart_id
     user_id = get_user_id()
     try:
         return {
@@ -1082,6 +1087,10 @@ def get_items_in_cart(cart_id):
 def add_item_to_cart(cart_id):
     """
     Add cart item to a cart and return the ID of the created item
+
+    The default cart can be accessible by using the actual cart UUID or
+    setting the requesting cart ID to "default". If the default cart does
+    not exist, the endpoint will create one automatically.
 
     Returns a 404 error if the cart does not exist or does not belong to the user
     Returns a 400 error if an invalid item was given
@@ -1104,6 +1113,7 @@ def add_item_to_cart(cart_id):
         "CartItemId": str
     }
     """
+    cart_id = None if cart_id == 'default' else cart_id
     user_id = get_user_id()
     try:
         request_body = app.current_request.json_body
@@ -1114,7 +1124,12 @@ def add_item_to_cart(cart_id):
     except KeyError:
         raise BadRequestError('EntityId, BundleUuid, BundleVersion, and EntityType must be given')
     try:
-        item_id = CartItemManager().add_cart_item(user_id, cart_id, entity_id, bundle_id, bundle_version, entity_type)
+        item_id = CartItemManager().add_cart_item(user_id,
+                                                  cart_id,
+                                                  entity_id,
+                                                  bundle_id,
+                                                  bundle_version,
+                                                  entity_type)
     except ResourceAccessError as e:
         raise NotFoundError(e.msg)
     return {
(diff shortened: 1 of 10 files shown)


src/azul/service/responseobjects/cart_item_manager.py
```diff
@@ -1,13 +1,16 @@
...
 import json
+from logging import getLogger
 import uuid
 
 from azul import config
 from azul.service.responseobjects.dynamo_data_access import DynamoDataAccessor
 from azul.service.responseobjects.elastic_request_builder import ElasticTransformDump as EsTd
 from azul.service.responseobjects.step_function_helper import StepFunctionHelper
 
+logger = getLogger(__name__)
```
hannes-ucsc:
```suggestion
logger = logging.getLogger(__name__)
```

terraform/dynamo.tf.json.template.py
```diff
@@ -23,6 +23,10 @@
                             {
                                 "name": "CartName",
                                 "type": "S"
+                            },
+                            {
+                                "name": "DefaultCart",
+                                "type": "N"  # Due to Terraform limitation, 0 or 1 will be used instead of boolean.
```
hannes-ucsc: Which limitiation is that, exactly?

lambdas/service/app.py
```diff
@@ -981,7 +992,8 @@ def update_cart(cart_id):
     :return: The updated cart
         {
             "CartName": str,
-            "CartId": str
+            "CartId": str,
+            "DefaultCart": str
```
hannes-ucsc: I see this typed as `int` in other place. This and the fact that we use a `int` as a `bool` makes this variable a bit ambiguous. Should it be named `IsDefaultCart`?

shiroyuki:
@hannes-ucsc, it is certainly possible to consider any cart, whose ID is the same as the user ID, as a default cart as the data service does not assert that IDs must be UUID. However, when someone look into the data directly from DynamoDB, it doesn't seem straight forward that it is a default cart of a particular user unless someone read the code. Also, it may be a bit confusing as the ID of the default cart will not conform with other cart IDs.

Also, we discuss earlier about soft-deleting carts in order to deal with very large carts. If the ID of the default cart is the same as the user and the default cart is marked for deletion, what should we do when a client tries to create a new default cart? Should the default cart get reenabled?

lambdas/service/app.py
shiroyuki: By the way, should the field name (in DynamoDB) remain as "DefaultCart" or do you want me to rename it to `IsDefaultCart` too?

terraform/dynamo.tf.json.template.py
hannes-ucsc: This is not a Terraform limitation, I guess, but a DynamoDB one. That's why I was confused.

```suggestion
                                "type": "N"  # Since DynamoDB lacks support for boolean values we represent them as 0 or 1
```

shiroyuki:
@hannes-ucsc According to [DynamoDB doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.DataTypes.html), `BOOL` is the attribute type for boolean. There are also some other types that we can use as well. However, Terraform's documentation seems to only suggest that it knows only three types. The workaround by declaring the attribute as number (`N`) is kinda working just, IMO, I would like to have it explicit here.

We do have [the code](https://github.com/DataBiosphere/azul/pull/631/files#diff-e030af86124e7abbe73815de9aebd427R49) to prevent the service to create more than one default cart at a time.

With your suggestion to have `CartID == `UserID` as an indicator for the default cart, when we soft-delete the default cart, there are a few scenarios that might confuse the users.

* Suppose the background task starts to hard-delete the data (the associated cart items and eventually the cart) and the client somehow re-enables the default cart. Any newly added cart items may not be deleted but the cart will eventually be deleted.
* If re-enabling before the hard deletion starts, the client may get confused as the old data may resurface.

We can implement versioning for cart object but it is more or less similar to using `IsDefaultCart` approach.

terraform/dynamo.tf.json.template.py
hannes-ucsc: It IS a DynamoDB restriction. See my reply to your comment.

hannes-ucsc:
> According to DynamoDB doc, BOOL is the attribute type for boolean. There are also some other types that we can use as well. However, Terraform's documentation seems to only suggest that it knows only three types. The workaround by declaring the attribute as number (N) is kinda working just, IMO, I would like to have it explicit here.

Don't think so. See

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html#HowItWorks.CoreComponents.PrimaryKey

> Each primary key attribute must be a scalar (meaning that it can hold only a single value). The only data types allowed for primary key attributes are string, number, or binary. There are no such restrictions for other, non-key attributes.

If you want the column to be a key, it has to be N, S or B. Otherwise, you don't even have to specify the type since DynamoDB is schema-less.

> We do have the code to prevent the service to create more than one default cart at a time.

Ok, but you do need to do an extra query, which is racy. In the solution I proposed that wouldn't be necessary. Trust me, if the schema allows for more than one default cart, due to the lack of serialization, that race will occur and more than one default cart will be created. However, your offline deletion argument still stands. I already conceded that the IsDefaultCart solution is simpler in that respect.

Here is another proposal: We store the ID of the default cart in the user object.

shiroyuki:
So, we are having a new table `User` which contains:
* `UserID` (`sub` from JWT)
* `DefaultCartID`

Does that sound like what you think?

src/azul/service/responseobjects/user_service.py
```diff
@@ -0,0 +1,27 @@
+from azul import config
+from azul.service.responseobjects.dynamo_data_access import DynamoDataAccessor
+
+
+class UserService:
```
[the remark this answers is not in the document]
hannes-ucsc: You don't need to act on this comment.

shiroyuki:
@hannes-ucsc That functionality related to adding items to the default cart is not documented in the RFC but I can add that quickly.

shiroyuki:
After taking a moment to reconsider the approach, here is what I think I would do.

To add items to the default cart, the request handling functions, e.g., `add_item_to_cart` (`POST /resources/carts/{cart_id}/items`) and `add_all_results_to_cart` (`POST /resources/carts/{cart_id}/items/batch`), will work with the default cart when `cart_id == "default"`. This would be the same as how the service handles the request to get the default cart (`GET /resources/carts/default`).

One reason to have the logic to find the default cart if cart ID is not specified outside `CartItemManager` is because after the default logic, I plan to submit a smaller PR to address the changes in data structure for Cart Item:

| Current in this PR | Change | Next PR | 
| --- | --- | --- |
| CartId | No change | - |
| EntityId | No change | - |
| EntityType | No change | - | 
| BundleUuid | **To be REMOVED**  | - | 
| BundleVersion | **To be REMOVED** | - |
| - | **To be ADDED** | EntityVersion (`str`) |

src/azul/service/responseobjects/cart_item_manager.py
```diff
@@ -23,19 +23,16 @@ def __init__(self):
         self.dynamo_accessor = DynamoDataAccessor()
         self.user_service = UserService()
 
+    @property
+    def default_cart_id_alias(self):
+        return "default"
```
hannes-ucsc: I do not like this. None is a much better default than a magic string. This just risks that defects in the code create a cart whose ID is actually `default` rather than failing early and obviously with an NPE.

test/service/test_cart_item_manager.py
```diff
@@ -186,6 +211,46 @@ def test_get_cart(self):
...
                                                        keys={'UserId': user_id, 'CartId': cart_id}))
 
+    def test_get_default_cart_with_existed_default_cart(self):
+        """
+        Creating a default cart should create a cart with the ID 'default' and DefaultCart flag set to True
+        """
+        mock_cart_id = 'test_default_cart'
+        user_id = '123'
+        cart_name = 'cart name'
+        with patch('uuid.uuid4') as uuid4_mock:
```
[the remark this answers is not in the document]
shiroyuki: All occurrences of `patch('uuid.uuid4')` is now using `side_effect`.

test/service/test_cart_item_manager.py
```diff
@@ -186,6 +211,46 @@ def test_get_cart(self):
                          self.dynamo_accessor.get_item(config.dynamo_cart_table_name,
                                                        keys={'UserId': user_id, 'CartId': cart_id}))
 
+    def test_get_default_cart_with_existed_default_cart(self):
+        """
+        Creating a default cart should create a cart with the ID 'default' and DefaultCart flag set to True
```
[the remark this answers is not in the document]
shiroyuki: The comment has been updated in 6841dc1.

test/service/test_cart_item_manager.py
```diff
@@ -186,6 +211,46 @@ def test_get_cart(self):
                          self.dynamo_accessor.get_item(config.dynamo_cart_table_name,
                                                        keys={'UserId': user_id, 'CartId': cart_id}))
 
+    def test_get_default_cart_with_existed_default_cart(self):
```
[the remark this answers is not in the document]
shiroyuki: This method will be renamed in the next commit.

src/azul/service/responseobjects/user_service.py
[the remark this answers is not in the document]
shiroyuki: ok

src/azul/service/responseobjects/cart_item_manager.py
shiroyuki: I agree with you on using `None`. I was so under the weather yesterday that I didn't get the chance to discuss with you on this issue.

Currently, as specified in the [RFC](https://github.com/DataBiosphere/azul/blob/develop/rfc/cart-api-v1.md#default-cart---todo), `default` is a magic string representing the default cart. Suppose User A has the default cart with ID `abc123` and he or she wants to get the information on the default cart, User A may send a request by sending either:

```http
GET /resources/carts/abc123
```

or

```http
GET /resources/carts/default
```

While the code in the `develop` branch did mistakenly create a default cart whose ID is `default`, the original consideration on how the service should create the default cart is pretty much the same as what `CartItemManager.create_cart` (in this PR) does. Hence, `CartItemManager.create_cart` in this PR is no longer create a cart whose ID is `default`.

Now, the question is which part of the code should handle `default` (the magic string). By not having `CartItemManager` handle the magic string, we can let request handling methods in `lambdas/service/app.py` to handle the magic string. The requesting handling code should be something like:

```python
@app.route('/resources/carts/{cart_id}', methods=['GET'], cors=True)
def get_cart(cart_id):
    user_id = get_user_id()
    manager = CartItemManager()
    try:
        cart = manager.get_cart(user_id, None if cart_id == "default" else cart_id)
    except ResourceAccessError:
        raise NotFoundError('Cart does not exist')
    else:
        return transform_cart_to_response(cart)
```

shiroyuki:
@hannes-ucsc In our discussion, we agree to have three endpoints to support the default cart ID. However, while I was working on the changes, there are two more endpoints that we should add the same support to allow the client getting items from or adding items in batch to the default cart much easier.

* `GET /resources/carts/{cart_id}/items` (to get cart items)
* `POST /resources/carts/{cart_id}/items/batch` (to add items by filters)

lambdas/service/app.py
```diff
@@ -1055,6 +1065,7 @@ def get_items_in_cart(cart_id):
         ]
     }
     """
+    cart_id = None if cart_id == 'default' else cart_id
```
shiroyuki: This change is not in the discussion but I think it would make getting the list of items of the default cart as easy as adding or removing items from the default cart.

lambdas/service/app.py
```diff
@@ -1155,6 +1185,7 @@ def add_all_results_to_cart(cart_id):
             "statusUrl": "https://status.url/resources/carts/status/{token}"
         }
     """
+    cart_id = None if cart_id == 'default' else cart_id
```
shiroyuki: This change is also not in the discussion but since we are doing this in the method that adding a single item, I think we should do the same for the batch operation.

lambdas/service/app.py
```diff
@@ -1173,14 +1204,7 @@ def add_all_results_to_cart(cart_id):
...
 
-    write_params = {
-        'filters': filters,
-        'entity_type': entity_type,
-        'cart_id': cart_id,
-        'item_count': item_count,
-        'batch_size': 10000
-    }
-    token = CartItemManager().start_batch_cart_item_write(user_id, cart_id, write_params)
+    token = CartItemManager().start_batch_cart_item_write(user_id, cart_id, entity_type, filters, item_count, 10000)
```
shiroyuki: `write_params` seems to be passthrough to the state machine as an execution input. It makes more sense to have `CartItemManager` prepare the execution input instead of this method.

lambdas/service/app.py
```diff
@@ -949,24 +949,30 @@ def create_cart():
...
 
 
-# TODO: implement default cart (may need to change get_all_carts() endpoint)
 @app.route('/resources/carts/{cart_id}', methods=['GET'], cors=True)
 def get_cart(cart_id):
     """
     Get the cart of the given ID belonging to the user
 
-    Returns a 404 error if the cart does not exist or does not belong to the user
+    The default cart can be accessible by using the actual cart UUID or
```
```suggestion
    The default cart is accessible under its actual cart UUID or by
```

lambdas/service/app.py
```diff
@@ -949,24 +949,30 @@ def create_cart():
...
+    setting the requesting cart ID to "default". If the default cart does
```
```suggestion
    setting the requesting cart ID to "default". If the default cart does
```
```suggestion
    passing "default" as the cart ID. If the default cart does
```

lambdas/service/app.py
```diff
@@ -949,24 +949,30 @@ def create_cart():
...
+    not exist, the endpoint will create one automatically.
```
hannes-ucsc: In REST, GET should not have persistent side effects, if at all possible. Would it be difficult to NOT create a default cart during GET requests?

lambdas/service/app.py
```diff
@@ -1045,6 +1051,10 @@ def get_items_in_cart(cart_id):
     """
     Get a list of items in a cart
 
+    The default cart can be accessible by using the actual cart UUID or
```
hannes-ucsc: See above.

src/azul/service/responseobjects/cart_item_manager.py
```diff
@@ -42,29 +42,36 @@ def create_cart(self, user_id, cart_name, default):
...
+            if user['DefaultCartId'] is not None:
                 raise DuplicateItemError('Default cart already exists')
-
-        cart_id = 'default' if default else str(uuid.uuid4())
+        cart_id = str(uuid.uuid4())
         self.dynamo_accessor.insert_item(config.dynamo_cart_table_name,
-                                         item={'CartId': cart_id, 'DefaultCart': default, **query_dict})
+                                         item={'CartId': cart_id, **query_dict})
+        if default:
+            self.user_service.update(user_id, default_cart_id=cart_id)
```
hannes-ucsc: Consider making this update [conditional](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.SimpleComparisons). It would be more efficient (because you can eliminate the user retrieval) and eliminate a race. There is a risk that the actual cart creation below fails but I think that possibility is far more unlikely than that of two concurrent requests for the creation of a default cart. I can elaborate why if needed.

src/azul/service/responseobjects/cart_item_manager.py
```diff
@@ -42,29 +42,36 @@ def create_cart(self, user_id, cart_name, default):
...
         return cart_id
 
     def get_cart(self, user_id, cart_id):
-        return self.dynamo_accessor.get_item(config.dynamo_cart_table_name,
+        if cart_id is None:
+            user = self.user_service.get(user_id)
+            cart_id = user['DefaultCartId'] or self.create_cart(user_id, 'Default Cart', default=True)
```
hannes-ucsc: See above. 

Consider creating three methods: `get_cart` (fails if cart_id is None or cart doesn't exist) `get_default_cart` (doesn't have a `cart_id` argument and fails if default cart is absent)  and `get_or_create_default_cart` (creates the default cart if absent).

shiroyuki: This is now separated into three methods.

src/azul/service/responseobjects/cart_item_manager.py
shiroyuki: Currently, `UserService.get` is technically `get_or_create`. The first user retrieval serves two purposes: create the user object if not already existed, and check whether the user already has the default cart. When there are two concurrent requests, one of them should fail as both tries to create two objects with the same hash key.

I would like to discuss with you on the implementation tomorrow after the standup to see what we can do to eliminate the race condition.

shiroyuki: The refactored code to use `ConditionExpression` and reduce the chance of retrieving user to minimize race condition will be ready by the next review request.

shiroyuki:
To the release managers, this PR requires a new DynamoDB table `azul-user-{deployment}` which will be generated by running `make terraform`.

PR #633, opened 2019-01-03
Service health endpoint throws 503 status code when 'DOWN' (fixes #624)

### LINKED ISSUE #624: Modify /health response to return 503 response when service is Down
Current `/health` request returns with a 200 status response in scenarios when is both 'UP' and 'DOWN' the response should return a '503' status code when is  'DOWN'

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-395)
┆Project Name: azul
┆Issue Number: AZUL-395

achave11-ucsc:
If the service can not ping the Elasticsearch instance, it will respond
with a 503 status code.

### DIFF
--- a/lambdas/indexer/app.py
+++ b/lambdas/indexer/app.py
@@ -9,6 +9,8 @@ import json
 import logging
 import random
 import time
+
+from chalice import Response
 from typing import List, MutableMapping
 import uuid
 
@@ -19,6 +21,7 @@ from dataclasses import asdict, dataclass, replace
 from more_itertools import chunked, partition
 
 from azul import config
+from azul.health import Health
 from azul.indexer import IndexWriter
 from azul.plugin import Plugin
 from azul.reindexer import Reindexer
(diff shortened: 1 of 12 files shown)


hannes-ucsc (changes requested):
This addresses only the service endpoint. The ticket mentions the service only, but it would be silly to only do this for the service health check endpoint but not the indexer health check endpoint. Both should be consistent. 

Also, the way the status is represented as "UP" or "DOWN" is awkward. Let's make the health check responses resemble the examples below. It should simplify the logic and prevent you from sprinkling the "UP" and "DOWN" string over the code base.

https://service.dev.explore.data.humancellatlas.org/health:

```json
{
    "elasticsearch": {
        "up": true
    },
    "up": true
}
```

https://indexer.dev.explore.data.humancellatlas.org/health

```
{
    "elasticsearch": {
        "up": true
    },
    "queues": {
        "fail": {
            "message": "The specified queue does not exist or you do not have access to it.",
            "up": false
        },
        "notify": {
            "messages": {
                "delayed": "0",
                "invisible": "0",
                "queued": "0"
            },
            "up": true
        }
    },
    "up": "false"
}
```

Note that I removed the `domain` element from the ES part of the response. I think it is unnecessary.

Also note that the fail queue is currently reported as being missing. I think that's a problem in the health check itself. If so, please fix that as well in this PR. Preferably a separate commit.

Try to come up with a generic logic for walking over the response body, enumerate all `up` properties in order to aggregate them into a top level `up` property and derive the status code from that top-level `up` property.

achave11:
Superseded by #659

PR #634, opened 2019-01-04
Make service tests use canned bundles (fixes #428)

### LINKED ISSUE #428: Eliminate fake data loader and replace with indexing of canned bundles
Maintaining the fake data templates and the canned ES documents for the service tests is quite tedious. We should settle on a bunch of exemplary bundles, can those and index those as part of the service test setup. For the paging tests we may still need to generate more documents but we could do so by repeating existing documents instead of using a template that's part of the code base.

Related to #426

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-283)
┆Project Name: azul
┆Issue Number: AZUL-283

jessebrennan:
Not ready to merge, just trying to check that my approach is correct. No need for a thorough review, just general stuff is fine.

Files to fix:
- [x] `test_cart_item_manager.py`
- [x] `test_manifest_endpoint.py`
- [x] `test_pagination.py`
- [x] `test_repository_projects.py`
- [x] `test_repository_specimen.py`
- [x] `test_request_validation.py`
- [x] `test_response.py`


test/service/__init__.py
```diff
@@ -1,28 +1,67 @@
...
+        new_doc['contents']['files'][0]['document_id'] = new_id
+        return new_doc
+
+    @classmethod
+    def _fill_index(cls, num_docs=1000):
+        """
+        Makes a bunch of copies of the first doc found in the files index and then bulk uploads them
+        """
+        template_doc = cls._get_documents()[0]['_source']
+        # -2 because the files from the template are already indexed
```
hannes-ucsc:
```suggestion
        # -2 because the files from the canned bundle are already indexed
```

You should assert that there were actually two files in the bundle.

hannes-ucsc: Or, don't hard code the difference but compute it, `- len(_get_documents())`.

test/service/test_response.py
```diff
@@ -1,130 +1,41 @@
...
-                        "organ_part": ["black"],
-                        "total_estimated_cells": 5306
-                    }
-                ]
-            },
-            "bundles": [
-                {
-                    "uuid": "cfc75555-f551-ba6c-2e62-0bf0ee01313c",
-                    "version": "2003-08-12T00:52:21"
+    def _get_index(self, index: str, entity_id: str):
```
```suggestion
    def _get_hits(self, index_name: str, entity_id: str):
```

test/service/test_response.py
```diff
@@ -1,130 +1,41 @@
...
-            ],
-            "entity_id": "08d3440a-7481-41c5-5140-e15ed269ea63"
+            }
         }
-        return [ElasticsearchFakeDataLoader.fix_canned_document(entity_type, template)]
+        # Tests are assumed to only ever run with the azul dev index
+        results = self.es_client.search(index=index, body=body)
+        return [results['hits']['hits'][0]['_source']]
+
+    def get_files_index(self, entity_id: str):
```
```suggestion
    def _get_file_hits(self, entity_id: str):
```

Likewise for the other two.

test/service/test_response.py
```diff
@@ -1,130 +1,41 @@
...
+        return self._get_index('azul_files_aggregate_dev', entity_id)
```
hannes-ucsc: Don't hard-code the index name. `azul.config` has methods for that.

test/service/__init__.py
```diff
@@ -1,28 +1,67 @@
...
+
+    @staticmethod
+    def _duplicate_es_doc(document):
+        """
+        takes in a template Elasticsearch document and returns a duplicate with a new entity_id
+        """
+        new_doc = copy.deepcopy(document)
+        new_id = str(uuid.uuid4())
+        new_doc['entity_id'] = new_id
+        new_doc['contents']['files'][0]['document_id'] = new_id
```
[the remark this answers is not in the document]
jessebrennan: one() is nice! I've always wished there was a prettier way to do this when you always expect a singleton (like responses from database calls).

test/service/__init__.py
jessebrennan: Good idea. Kill the magic number

test/service/test_response.py
jessebrennan: Fixed.

test/service/test_response.py
jessebrennan: I just eliminated these helper methods and make entity_type a parameter to get_hits()

test/service/test_response.py
jessebrennan: Indeed a more accurate name.

test/service/test_response.py
hannes-ucsc: @jessebrennan, looks like this isn't actually fixed. There are still hard-coded index names in your changes. This causes the tests to fail if any deployment other than `dev` is selected.

Issue #638, opened 2019-01-07
Merge dos-azul-lambda


david4096:
~~dos-dss-lambda provides methods for accessing data objects from a data store using a HTTP requests. In working toward a standard method of interchanging data for the Terra handoff, dos-dss-lambda can provide an extensible method for handling data in multiple locations.~~

~~To close this, start dos-dss-lambda against the existing DSS. Currently blocked by https://github.com/DataBiosphere/dos-dss-lambda/issues/30~~

~~https://github.com/DataBiosphere/dos-dss-lambda~~

Opting instead to merge the code from dos-azul-lambda here.

To close this, present the ability to stand up dos azul lambda within the azul codebase and import the existing DAL tests.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-403)
┆Project Name: azul
┆Issue Number: AZUL-403

PR #642, opened 2019-01-09
Add DRS Endpoint

david4096:
Adds features based on the DRS lambda first programmed at https://github.com/DataBiosphere/dos-azul-lambda .

Will address #638 

* A new path at `/ga4gh/dos/v1/dataobjects/{data_object_id}`.
* Made the service a python module so its functions can be unit tested. (2 additional `__init__.py`)
* Tests demonstrating the endpoint usage
* Notebook demonstrating using the DRS endpoint to download a file

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -52,6 +52,17 @@ changes:
       - reindex
     notes: Please append `_set azul_dss_query_prefix '42'` to deployments/*.local/environment.
 
+  - title: Add GA4GH Data Repository Service GET Method
+    issues:
+      - https://github.com/DataBiosphere/azul/issues/638
+      - https://github.com/DataBiosphere/azul/pull/642
+    upgrade:
+      - deploy
+      - clients
+    notes: >
+        Adds a Data Repository Service (DRS) endpoint at `ga4gh/dos/v1/dataobjects`. 
+        Implements the `GetDataObject` method, a subset of the DRS API.
+
   - title: Add endpoint to shorten URLs
     issues:
       - https://github.com/databiosphere/azul/issues/66
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -1270,18 +1270,17 @@ def azul_to_obj(result):
     :param result: the ElasticSearch result dictionary for a single file
     :return: DataObject
     """
-    azul = result
     data_object = {}
-    data_object['id'] = azul['uuid']
+    data_object['id'] = result['uuid']
     replicas = ['aws']
-    urls = [{"url": file_url(azul['uuid'], version=azul['version'], replica=replica)} for replica in replicas]
+    urls = [{"url": file_url(result['uuid'], version=result['version'], replica=replica)} for replica in replicas]
     data_object['urls'] = urls
-    data_object['size'] = str(azul.get('size', ''))
+    data_object['size'] = str(result.get('size', ''))
     data_object['checksums'] = [
-        {'checksum': azul['sha256'], 'type': 'sha256'}]
-    data_object['aliases'] = [azul['name']]
-    data_object['version'] = azul['version']
-    data_object['name'] = azul['name']
+        {'checksum': result['sha256'], 'type': 'sha256'}]
+    data_object['aliases'] = [result['name']]
+    data_object['version'] = result['version']
+    data_object['name'] = result['name']
     return data_object
 
 
@@ -1312,7 +1311,6 @@ def get_data_object(data_object_id):
                                            pagination=pagination,
                                            post_filter=True,
                                            entity_type='files')
-        print(response['hits'][0])
         file_document = response['hits'][0]['files'][0]
         data_obj = azul_to_obj(file_document)
         # Double check to verify identity (since `file_id` is an analyzed field)
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -816,12 +816,8 @@ def file_url(uuid, **params):
 
 
 def self_url(endpoint_path=None):
-    if app.current_request is not None:
-        protocol = app.current_request.headers.get('x-forwarded-proto', 'http')
-        base_url = app.current_request.headers.get('host')
-    else:
-        protocol = 'http'
-        base_url = 'localhost:8000'
+    protocol = app.current_request.headers.get('x-forwarded-proto', 'http')
+    base_url = app.current_request.headers['host']
     if endpoint_path is None:
         endpoint_path = app.current_request.context['path']
     retry_url = f'{protocol}://{base_url}{endpoint_path}'
@@ -1296,7 +1292,7 @@ def get_data_object(data_object_id):
     """
     logger = app.log
     filters = {"file": {"fileId": {"is": [data_object_id]}}}
-    logger.debug('DOS request for Data Object with uuid: {}'.format(data_object_id))
+    logger.debug(f'DOS request for Data Object with uuid: {data_object_id}')
     # We don't care about the query params, only the path
     app.current_request.query_params = {}
     logger.debug("Filters string is: {}".format(filters))
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -816,8 +816,12 @@ def file_url(uuid, **params):
 
 
 def self_url(endpoint_path=None):
-    protocol = app.current_request.headers.get('x-forwarded-proto', 'http')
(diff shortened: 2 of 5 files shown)


hannes-ucsc:
Oh, since this is an externally observable change, an entry to the change log needs to be made.

test/service/test_drs.py
```diff
@@ -0,0 +1,99 @@
...
+        Helper function to get a data object using the
+        repository/files list feature.
+        :return:
+        """
+        url = self.base_url + "/repository/files"
+        response = requests.get(url)
+        response.raise_for_status()
+        response_json = response.json()
+        file_id = response_json['hits'][0]['files'][0]['uuid']
+        get_url = "{}/ga4gh/dos/v1/dataobjects/{}".format(self.base_url, file_id)
```
hannes-ucsc: We use f-strings.

test/service/test_drs.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_data_object_not_found(self):
+        """
+        Ensures that when an unused identifier is requested a 404 status
+        code is returned.
+        :return:
+        """
+        file_id = "NOT_A_GOOD_IDEA"
+        get_url = "{}/ga4gh/dos/v1/dataobjects/{}".format(self.base_url, file_id)
+        # Should cause a 404 error
```
hannes-ucsc: Then it should assert that. Also, `with self.assertRaises(): …; self.raise_for_status()` seems convoluted. Why not simply `assertEquals(404, response.status)` ?

test/service/test_drs.py
```diff
@@ -0,0 +1,99 @@
...
+                   "unskip when the index data is generated dynamically")
+    def test_url_presence(self):
+        """
+        Demonstrates the presence of URLs that can be used to fetch
+        files within the DRS response.
+        :return:
+        """
+        data_object, _ = self.get_data_object()
+        url = data_object['urls'][0]['url']
+        self.assertIn('http', url, "Make sure it is url-like")
```
hannes-ucsc: Any string containing the string "http" is URL-like? This seems too lenient and redundant at the same time, given the fact that `requests` would barf if it weren't a URL.

test/service/test_drs.py
```diff
@@ -0,0 +1,99 @@
...
+        :return:
+        """
+        file_id = "NOT_A_GOOD_IDEA"
+        get_url = "{}/ga4gh/dos/v1/dataobjects/{}".format(self.base_url, file_id)
+        # Should cause a 404 error
+        with self.assertRaises(requests.exceptions.HTTPError):
+            drs_response = requests.get(get_url)
+            drs_response.raise_for_status()
+
+    @unittest.skip("DSS and Azul indices do not match in testing"
```
hannes-ucsc: Reading this, I get the feeling that the tests are actually hitting the DSS. That would disqualify them as unit tests, obviously. If I am correct and if at all possible, try mocking the DSS responses. test_dss_proxy.py does that.

david4096:
Thanks for the comments @hannes-ucsc. I'll remove the format strings. Regarding mocking the DSS response, I believe it is possible but am OK with removing/skipping that test. Is the "fetch" surface tested elsewhere?

Some of the test patterns are copypasta from other tests there and could be less redundant.

I'll also take another look at the import patterns for unit testing so I can match them.

hannes-ucsc:
> Is the "fetch" surface tested elsewhere?

`/fetch/dss/files` is covered by `test_dss_proxy.py` if that's what you mean.

There is not much point in dragging around a test that we don't intend to ever unskip. If this test doesn't add any additional coverage, let's remove it. This is your call, since I haven't sufficiently thought through the cases (and shouldn't have to).

One more thing, I will get an email when you request a review on Github. But additionally, we also assign the PR to the reviewer (me). Only then will it show up when we filter on the reviewer during stand-up. And looking at the PRs in Github, I can tell at a glance which ones need my attention.

david4096:
There's 1 failed and 1 errored.

```
======================================================================
ERROR: test_azul_to_drs (service.test_drs.DataRepositoryServiceEndpointTest)
----------------------------------------------------------------------
  File "/home/travis/build/DataBiosphere/azul/test/service/test_drs.py", line 73, in test_azul_to_drs
    data_object = azul_to_obj(mock_azul_file)
  File "/home/travis/build/DataBiosphere/azul/lambdas/service/app.py", line 1271, in azul_to_obj
    urls = [{"url": file_url(result['uuid'], version=result['version'], replica=replica)} for replica in replicas]
  File "/home/travis/build/DataBiosphere/azul/lambdas/service/app.py", line 1271, in <listcomp>
    urls = [{"url": file_url(result['uuid'], version=result['version'], replica=replica)} for replica in replicas]
  File "/home/travis/build/DataBiosphere/azul/lambdas/service/app.py", line 811, in file_url
    url = self_url(endpoint_path=f'{proxy_endpoint_path}/{uuid}')
  File "/home/travis/build/DataBiosphere/azul/lambdas/service/app.py", line 822, in self_url
    retry_url = f'{protocol}://{base_url}{endpoint_path}'
UnboundLocalError: local variable 'protocol' referenced before assignment

======================================================================
FAIL: test_data_object_not_found (service.test_drs.DataRepositoryServiceEndpointTest)
----------------------------------------------------------------------
  File "/home/travis/build/DataBiosphere/azul/test/service/test_drs.py", line 50, in test_data_object_not_found
    drs_response = requests.get(get_url)
AssertionError: HTTPError not raised

----------------------------------------------------------------------
Ran 140 tests in 762.751s

FAILED (failures=1, errors=1)
```

The reason the first is erroring is because I removed the "request mocking" for when in a test environment I had added here. https://github.com/DataBiosphere/azul/pull/642/commits/48952c849c3730fc482cfe2cc9aebf15b3877258 .

The second is an actual failure that I'll have to diagnose, and which I haven't yet replicated locally.

test/service/test_drs.py
```diff
@@ -0,0 +1,59 @@
+import requests
+import unittest
+import sys
+
+from service import WebServiceTestCase
+
+#from lambdas.service.app import azul_to_obj
```
hannes-ucsc: Remove, please.

lambdas/service/app.py
```diff
@@ -780,7 +780,7 @@ def file_url(uuid, **params):
 
 def self_url(endpoint_path=None):
     protocol = app.current_request.headers.get('x-forwarded-proto', 'http')
-    base_url = app.current_request.headers['host']
+    base_url = app.current_request.headers.get('host')
```
hannes-ucsc: We didn't need this until now. Furthermore, if the `host` header were missing, I would want to learn that here rather than …

lambdas/service/app.py
```diff
@@ -780,7 +780,7 @@ def file_url(uuid, **params):
...
     if endpoint_path is None:
         endpoint_path = app.current_request.context['path']
     retry_url = f'{protocol}://{base_url}{endpoint_path}'
```
hannes-ucsc: … produce a URL that reads `http://None/…` and causes more insidious failures later.

hannes-ucsc:
Remember to assign the PR to me in addition to requesting review from me.

Also: https://github.com/DataBiosphere/azul/pull/642#issuecomment-455395109

lambdas/service/app.py
```diff
@@ -780,7 +780,7 @@ def file_url(uuid, **params):
 
 def self_url(endpoint_path=None):
     protocol = app.current_request.headers.get('x-forwarded-proto', 'http')
-    base_url = app.current_request.headers['host']
+    base_url = app.current_request.headers.get['host']
```
hannes-ucsc: We didn't need this until now. Furthermore, if the host header were missing, I would want to learn that here rather than …

test/service/test_drs.py
hannes-ucsc: You must have accidentally clicked Resolve conversation. This hasn't been addressed.

CHANGELOG.yaml
```diff
@@ -52,6 +52,17 @@ changes:
       - reindex
     notes: Please append `_set azul_dss_query_prefix '42'` to deployments/*.local/environment.
 
+  - title: Add GA4GH Data Repository Service GET Method
+    issues:
+      - https://github.com/DataBiosphere/azul/issues/638
+      - https://github.com/DataBiosphere/azul/pull/642
```
hannes-ucsc: We don't refer to PRs here, only issues.

Issue #621, opened 2018-12-19
Specify alarm actions for metric alarms in cloudwatch.tf


hannes-ucsc:
┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-393)
┆Project Name: azul
┆Issue Number: AZUL-393

hannes-ucsc:
Demoed on 1/18/19

PR #648, opened 2019-01-11
Fix settings that notify slack when alarm goes off (#621)

achave11:


terraform/cloudwatch.tf.json.template.py
```diff
@@ -16,7 +16,8 @@
                     "period": "3600",
                     "statistic": "Average",
                     "threshold": "85",
-                    "alarm_description": "This metric monitors ES CPU utilization",
+                    "alarm_description": f'{{ "slack_channel": "dcp-ops-alerts",'
```
hannes-ucsc: We typically use `json.dumps()` if Terraform wants serialized JSON. Look for other uses of `json.dumps()` in other `*.tf.json.template.py` files for an example.

PR #663, opened 2019-01-18
Allow service lambda to communicate with SQS (#664)

achave11:


lambdas/service/lambda-policy.json.template.py
```diff
@@ -33,6 +33,18 @@
             ],
             "Resource": f"arn:aws:es:{aws.region_name}:{aws.account}:domain/{config.es_domain}"
         },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "sqs:GetQueueAttributes",
+                "sqs:GetQueueUrl",
+                "sqs:ReceiveMessage"
```
hannes-ucsc: I don't think this is necessary.

lambdas/indexer/lambda-policy.json.template.py
```diff
@@ -45,9 +45,7 @@
             ],
             "Resource": [
                 f"arn:aws:sqs:{aws.region_name}:{aws.account}:{name}"
-                for name in (config.notify_queue_name,
-                             config.token_queue_name,
-                             config.document_queue_name)
+                for name in config.all_queue_names
```
hannes-ucsc: I'd rather not give any lambda write permissions to the fail queues.

hannes-ucsc:
>I believe the problem has to do with the lambda health endpoints being unreachable from the outside 

I believe you mean the indexer lambda. This is due to #665.

lambdas/service/lambda-policy.json.template.py
achave11: So I made some research last night and this is what I found.
[events-sqs-permissions](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#events-sqs-permissions) if you check at the bottom under `Execution Role Permissions` [sqs:ReceiveMessage](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html) might be the only one we need. Since we only collect some attributes from the queues. I might have misunderstood the information. @hannes-ucsc

achave11:
@achave11

PR #685, opened 2019-01-24
Prototype authentication

shiroyuki:
* [x] Changed from `AZUL_FUSILLADE_HOSTNAME` (just a hostname) to `AZUL_FUSILLADE_ENDPOINT` (base URL)
* [x] Fixed the merging issue
* [x] Fixed the side effect issue.

### DIFF
--- a/environment
+++ b/environment
@@ -145,6 +145,9 @@ azul_vars=(
     # string, only bundles whose UUID starts with the specified string will be
     # indexed.
     azul_dss_query_prefix
+
+    # The URL Prefix of Fusillade (Authentication Broker Service)
+    AZUL_FUSILLADE_ENDPOINT
 )
 
 unset ${azul_vars[*]}
@@ -225,6 +228,8 @@ _set AZUL_ENABLE_CLOUDWATCH_ALARMS 0
 _set AZUL_CART_API_IP_WHITELIST  # TODO: Remove when authentication is added
 
 _set azul_dss_query_prefix
+
+_set AZUL_FUSILLADE_ENDPOINT "https://auth.dev.data.humancellatlas.org"
 
 unset -f _set
 
--- a/lambdas/indexer/.gitignore
+++ b/lambdas/indexer/.gitignore
@@ -3,4 +3,5 @@
 /.chalice/config.json
 /.chalice/deployed/*.json
 /lambda-policy.json
-/vendor/azul_changes.py
+/vendor/*
+!/vendor/azul
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -5,22 +5,23 @@ import json
 from concurrent.futures import ThreadPoolExecutor
 import hashlib
 import logging.config
+import math
 import os
 import re
 import time
 import urllib.parse
 import uuid
 
-# noinspection PyPackageRequirements
 import boto3
 from botocore.exceptions import ClientError
 # noinspection PyPackageRequirements
-from chalice import BadRequestError, Chalice, ChaliceViewError, NotFoundError, Response, UnauthorizedError
+from chalice import BadRequestError, Chalice, ChaliceViewError, NotFoundError, Response, AuthResponse, UnauthorizedError
 from more_itertools import one
 import requests
 
 from azul import config
 from azul.health import Health
+from azul.security.authenticator import Authenticator, AuthenticationError
 from azul.service import service_config
 from azul.service.responseobjects.cart_item_manager import CartItemManager, DuplicateItemError, ResourceAccessError
 from azul.service.responseobjects.elastic_request_builder import (BadArgumentException,
@@ -43,7 +44,6 @@ app = Chalice(app_name=config.service_name, configure_logs=False)
 app.debug = True
 # FIXME: please use module logger instead (https://github.com/DataBiosphere/azul/issues/419)
 app.log.setLevel(logging.DEBUG)
-
 
 # TODO: Write the docstrings so they can support swagger.
 # Please see https://github.com/rochacbruno/flasgger
@@ -78,6 +78,38 @@ def _get_pagination(current_request, entity_type):
         raise BadArgumentException("Bad arguments, only one of search_after or search_before can be set")
 
     return pagination
+
+
+@app.authorizer()
+def jwt_auth(auth_request) -> AuthResponse:
+    given_token = auth_request.token
+    authenticator = Authenticator()
+
+    try:
+        claims = authenticator.authenticate_bearer_token(given_token)
+    except AuthenticationError:
+        # By specifying no routes, the access is denied.
+        log.warning('Auth: ERROR', stack_info=True)
+        return AuthResponse(routes=[], principal_id='anonymous')
+    except Exception as e:
+        # Unknown error, block all access temporarily.
+        # Note: When the code is running on AWS, this function will become a
+        #       separate lambda function, unlike how to work on a local machine.
+        #       Due to that, this method cannot fail unexpectedly. Otherwise,
+        #       The endpoint will respond with HTTP 500 with no log messages.
+        log.error(f'Auth: AUTHENTICATE: Critical ERROR {e} {given_token}', stack_info=True, exc_info=e)
+        return AuthResponse(routes=[], principal_id='no_access')
+
+    # When this code is running on AWS, the context of the authorizer has to be
+    # a string-to-primitive-type dictionary. When the context is retrieved in
+    # the caller function, the value somehow got casted into string.
+    return AuthResponse(routes=['*'],
+                        principal_id='user',
+                        context={
+                            prop: value
+                            for prop, value in claims.items()
+                            if not isinstance(value, (list, tuple, dict))
+                        })
 
 
 @app.route('/', cors=True)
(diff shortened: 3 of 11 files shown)


shiroyuki:
@hannes-ucsc, in 0ffec77, I'm not sure how you would like me avoid defining side effects at import time. So, I wrap this in a function so that the test can patch the method (`shared_session`).

src/azul/security/auth.py
```diff
@@ -0,0 +1,80 @@
...
+from logging import getLogger
+from typing import Dict, Any
+from urllib.parse import urlencode
+
+from azul import config
+from azul.security.verifier import (verify,
+                                    InvalidTokenError,
+                                    NonDecodableTokenError)
+
+logger = getLogger(__name__)
```
hannes-ucsc:
```suggestion
logger = logging.getLogger(__name__)
```

Like everywhere else.

src/azul/security/auth.py
```diff
@@ -0,0 +1,80 @@
...
+fusillade_callback_url = config.service_endpoint() + '/auth/callback'
```
hannes-ucsc: This seems to be factored into a constant for patching during testing. I don't like it because it essentially caches  the return value of `config.service_endpoint` when the module is imported. Changes to the underlying environment variables from which `config.service_endpoint` is derived will not be reflected. This is why everything in Config is a property and Config is stateless. This type of caching breaks that.

src/azul/security/verifier.py
```diff
@@ -0,0 +1,112 @@
...
+
+from azul import config
+
+logger = logging.getLogger(__name__)
+
+allowed_algorithms = ('RS256',)
+gserviceaccount_domain = "iam.gserviceaccount.com"
+
+# Shared HTTP session
+def shared_session():
```
hannes-ucsc: The name of the method and the (redundant) comment are misleading. There is no sharing going on here. Every invocation of a function in this module will create a new session.

src/azul/security/verifier.py
```diff
@@ -0,0 +1,112 @@
...
+import jwt
+from jwt.exceptions import DecodeError, PyJWTError
+import requests
+
+from azul import config
+
+logger = logging.getLogger(__name__)
+
+allowed_algorithms = ('RS256',)
+gserviceaccount_domain = "iam.gserviceaccount.com"
```
hannes-ucsc: Add to Config or inline.

src/azul/security/verifier.py
```diff
@@ -0,0 +1,112 @@
...
+from cryptography.hazmat.backends import default_backend
+import jwt
+from jwt.exceptions import DecodeError, PyJWTError
+import requests
+
+from azul import config
+
+logger = logging.getLogger(__name__)
+
+allowed_algorithms = ('RS256',)
```
hannes-ucsc: Add to Config or inline.

test/app_test_case.py
```diff
@@ -16,6 +25,37 @@
 from azul_test_case import AzulTestCase
 
 log = logging.getLogger(__name__)
+key_dir = os.path.dirname(__file__)
```
hannes-ucsc: Use tempdir instead.

test/app_test_case.py
```diff
@@ -16,6 +25,37 @@
...
+public_key_path = os.path.abspath(os.path.join(key_dir, 'public.pem'))
```
hannes-ucsc: Shouldn't be a global variable.

test/app_test_case.py
```diff
@@ -16,6 +25,37 @@
...
+private_key_path = os.path.abspath(os.path.join(key_dir, 'private.pem'))
```
hannes-ucsc: Shouldn't be a global variable.

test/app_test_case.py
```diff
@@ -16,6 +25,37 @@
...
+
+
+def generate_test_keys():
```
hannes-ucsc: This should be a method or property in the test case. The property should generate the keypair if necessary and return the path to it.

test/app_test_case.py
```diff
@@ -16,6 +25,37 @@
...
+    if not os.path.exists(private_key_path):
+        call(['openssl', 'genrsa', '-out', private_key_path, '2048'])
+        call(['openssl', 'rsa', '-in', private_key_path, '-outform', 'PEM', '-pubout', '-out', public_key_path])
+
+
+def get_test_keys():
```
hannes-ucsc: Same.

test/app_test_case.py
```diff
@@ -16,6 +25,37 @@
...
+        public_key = serialization.load_pem_public_key(
+            key_file.read(),
+            backend=default_backend()
+        )
+
+    return private_key, public_key
+
+
+def remove_test_keys():
+    os.unlink(public_key_path)
```
hannes-ucsc: Method in a test. I don't understand why this deviates from a well established pattern of test classes and methods to global functions.

lambdas/service/app.py
```diff
@@ -808,13 +842,6 @@ def files_proxy(uuid):
         dss_response.raise_for_status()
 
 
-def file_url(uuid, **params):
```
hannes-ucsc: Inconsequential change. Just changes the order of two methods. Remove.

lambdas/service/app.py
```diff
@@ -824,6 +851,58 @@ def self_url(endpoint_path=None):
     return retry_url
 
 
+def file_url(uuid, **params):
```
hannes-ucsc: Inconsequential change. Just changes the order of two methods. Remove.

src/azul/security/verifier.py
shiroyuki: This is moved to `azul.security.authenticator.Authenticator.session`.

src/azul/security/verifier.py
shiroyuki: Updated (inline)

test/app_test_case.py
shiroyuki: Currently, this is part of a helper class, called `TestKeyManager` in `app_test_case.py`. I am thinking of either leaving it (to keep it isolated from growing `LocalAppTestCase`) as it is, or merging it into `LocalAppTestCase`. Which approach would you prefer, @hannes-ucsc ?

test/app_test_case.py
shiroyuki: Currently, this is part of a helper class, called `TestKeyManager` in `app_test_case.py`.

test/app_test_case.py
shiroyuki: Currently, this is part of a helper class, called `TestKeyManager` in `app_test_case.py`.

lambdas/service/app.py
shiroyuki: It was a rebasing issue. I moved it back.

lambdas/service/app.py
shiroyuki: It was a rebasing issue. It has been moved back to this location.

lambdas/indexer/.gitignore
```diff
@@ -3,4 +3,5 @@
 /.chalice/config.json
 /.chalice/deployed/*.json
 /lambda-policy.json
-/vendor/azul_changes.py
+/vendor/*
```
hannes-ucsc: Next time you push, please push a history with two commits. First commit should have the changes to these `.gitignore` files. Second commit is everything else.

environment
```diff
@@ -145,6 +145,9 @@ azul_vars=(
     # string, only bundles whose UUID starts with the specified string will be
     # indexed.
     azul_dss_query_prefix
+
+    # Hostname of Fusillade (Authentication Broker Service)
```
hannes-ucsc: This isn't actually just a host name. It's a URL prefix.

test/app_test_case.py
```diff
@@ -65,6 +120,7 @@ def base_url(self):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        TestKeyManager.generate_test_keys()
```
hannes-ucsc: Why do we need to do this for every test case? Aren't the keys only needed for the authentication test cases?

test/app_test_case.py
```diff
@@ -108,3 +162,54 @@ def tearDown(self):
         self.server_thread.join(timeout=10)
         if self.server_thread.is_alive():
             self.fail('Thread is still alive after joining')
+
+    def set_up_fake_responses(self):
+        # This requires to be invoked in a method within the context of
+        # ``responses.activate`` or ``responses.RequestsMock()``.
```
hannes-ucsc: Consider making this method a decorator that calls `responses.activate` so as to enforce the dependency.

Also take a look at `test/retorts.py`. It works around a problem that we may also run into again when once we add auth to the cart endpoints AND run tests against those.

Instead of creating another decorator, the best approach may be to extend `ResponsesHelper` in `retorts.py` and to override the `__enter__` method in order to set up the mocked auth endpoint responses. 

If the decorator approach works for now, I'd be happy with that, too.

Also, prefer the term `mock` over `fake`.

test/app_test_case.py
shiroyuki: No, we don't. I'll change this into making the keys on demand.

shiroyuki: In the next set of commits, the test keys will be generated just right before either one of the keys are requested.

test/app_test_case.py
shiroyuki: The setup for mock responses is moved to `AuthResponsesHelper` in `test/retorts.py` and the test case is split up into `LocalAppTestCase` and `AuthLocalAppTestCase` (still in the same file).

lambdas/indexer/.gitignore
hannes-ucsc: This hasn't been addressed.

test/retorts.py
```diff
@@ -51,3 +65,114 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
         self.request_mock.passthru_prefixes = self.passthru_prefixes
+
+
+class AuthResponseHelper(ResponsesHelper):
+    def __init__(self, passthru_url: str = None, request_mock: responses.RequestsMock = None) -> None:
+        super().__init__(request_mock)
+        self.passthru_url = passthru_url
+
+    def __enter__(self):
+        context = super().__enter__()
```
hannes-ucsc: You should return `context`.

shiroyuki: @hannes-ucsc `return context` is already on [line 103](https://github.com/DataBiosphere/azul/pull/685/files/5378ac6a7c9d1ca0b61f711b7d34e0483fdaf9e7#diff-225b710c3efb09eae6db97fdd83cf528R103). If it isn't what it should be, could you clarify what you meant by returning `context`?

lambdas/indexer/.gitignore
shiroyuki: @hannes-ucsc This change was made when we try to streamline `make deploy` to install additional wheels (b7690914). Since then, we identified that the compatibility between two Python versions (3.6 and 3.7) caused the issue and we addressed that by ensuring that everyone must use Python 3.6 for the deployment.

I think this change seems to be unnecessary at this point and I think of reverting this change. What do you think?

test/app_test_case.py
```diff
@@ -78,10 +78,7 @@ def tearDownClass(cls):
 
     def setUp(self):
         super().setUp()
-        log.debug("Setting up tests")
```
hannes-ucsc: Change makes sense, but should have been done in a separate commit.

test/retorts.py
hannes-ucsc: You are right. My mistake.

Issue #604, opened 2018-12-11
Export manifest in BDBag to Terra


cricketsloan:
These metadata must be available for the Firecloud data model. The alternative to including these at index time is to make requests for the requisite metadata as necessary.

To close, ensure file index has participant and specimen metadata available in the exported manifest. These metadata will be used to generated a Firecloud TSV input.

Following [~hannes-ucsc] suggestion, this should be done at indexing time, although it is possible to do at query-time.

The current transformer can be viewed here https://github.com/DataBiosphere/azul/blob/prod/src/azul/project/hca/transformers.py#L81 . 

Specimen metadata are added here at the bundle level: https://github.com/DataBiosphere/azul/blob/prod/src/azul/project/hca/transformers.py#L338

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-377)
┆Project Name: azul
┆Issue Number: AZUL-377
┆Epic: Feature Export from Boardwalk to HCA

mjkrause:
To consolidate tickets, the work needed for #602 (Add participant and specimen metadata to manifest) will be part of this ticket. I will close #602.

mjkrause:
Changed the title to better capture the work it entails.

PR #687, opened 2019-01-26
Export manifest in BDBag to Terra (addresses #604)

mjkrause:
Adds some of the functionality to export the manifest from
the explorer to Broad analysis platform Terra (aka FireCloud).
I used the original manifest as the base, and added four
columns to it, `participant_id`, `sample_id`, and
`cellsuspension_id` (to satisfy current Terra import
requirements), and `drs_uri` (to enable use of DRS URLs
for metadata resolution). 

The manifest is compressed as a zip file, uploaded to 
AWS S3, and a signed URL is generated to that location.

Specific changes are:
* cell_suspension_id -> cellsuspensions.document_id
* participant_id -> specimens.donor_document_id
* sample_id -> specimens.donor_biomaterial_id
* new method `_construct_bdbag` in class `ManifestResponse`
* a second HTTP parameter, `format` was added, which can
assume two values, _tsv_ and _bdbag_.
* currently the BDBag is pushed as a single part
* a new environment variable `AZUL_DRS_ENDPOINT` was added
to file `environment`

### DIFF
--- a/CHANGELOG.yaml
+++ b/CHANGELOG.yaml
@@ -53,6 +53,18 @@ changes:
     notes: Before reinstalling your requirements, delete your virtual environment. Then, use the `python -m venv`
            command instead of `virtualenv` to create your virtual environment.
 
+  - title: Add BDBag as an alternative manifest format
+    issues:
+      - https://github.com/DataBiosphere/azul/issues/604
+    upgrade:
+      - deploy
+      - requirements
+      - clients
+    notes: The query parameter `format` has been added to the '/repository/files/export' endpoint. Possible values are
+           `tsv` (default, retains previous functionality) and `bdbag`. The latter creates a manifest containing the
+           information necessary for importing samples from a DSS instance into Broad's Terra analysis platform. The
+           manifest data will be packaged in a BDBag and persisted as a compressed ZIP file in an S3 bucket.
+
   - title: Include content disposition header in manifest responses
     issues:
       - https://github.com/DataBiosphere/azul/issues/655
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -411,7 +411,9 @@ def get_manifest():
     if app.current_request.query_params is None:
         app.current_request.query_params = {}
     filters = app.current_request.query_params.get('filters', '{"file": {}}')
+    format = app.current_request.query_params.get('format', 'tsv')
     logger.debug("Filters string is: {}".format(filters))
+    logger.debug(f'Format query parameter is: {format}')
     try:
         logger.info("Extracting the filter parameter from the request")
         filters = ast.literal_eval(filters)
@@ -424,7 +426,9 @@ def get_manifest():
     es_td = EsTd()
     # Get the response back
     logger.info("Creating the API response")
-    response = es_td.transform_manifest(filters=filters)
+
+    response = es_td.transform_manifest(filters=filters, format=format)
+
     # Return the excel file
     return response
 
@@ -533,12 +537,14 @@ def handle_manifest_generation_request():
 
     query_params = app.current_request.query_params or {}
     filters = query_params.get('filters')
+    format = query_params.get('format')
     token = query_params.get('token')
     retry_url = self_url()
     manifest_service = ManifestService()
     try:
         return manifest_service.start_or_inspect_manifest_generation(retry_url,
                                                                      token=token,
+                                                                     format=format,
                                                                      filters=filters)
     except ClientError as e:
         if e.response['Error']['Code'] == 'ExecutionDoesNotExist':
@@ -558,10 +564,13 @@ def generate_manifest(event, context):
     :param: event: dict containing function input
         Valid params:
             - filters: dict containing filters to use in ES request
+            - format: str to specify manifest output format, values are
+                      'tsv' (default) or 'bdbag'
     :return: The URL to the generated manifest
     """
     filters = event.get('filters', {'file': {}})
-    response = EsTd().transform_manifest(filters=filters)
+    format = event.get('format', 'tsv')
+    response = EsTd().transform_manifest(format, filters)
     return {'Location': response.headers['Location']}
 
 
--- a/src/azul/service/manifest.py
+++ b/src/azul/service/manifest.py
@@ -38,6 +38,7 @@ class ManifestService(AbstractService):
     def start_or_inspect_manifest_generation(self,
                                              self_url,
                                              token: Optional[str] = None,
+                                             format: Optional[str] = None,
                                              filters: Optional[str] = None
                                              ) -> Tuple[float, str]:
         """
@@ -52,7 +53,7 @@ class ManifestService(AbstractService):
 
         if token is None:
             execution_id = str(uuid.uuid4())
-            self._start_manifest_generation(filters, execution_id)
+            self._start_manifest_generation(format, filters, execution_id)
             token = {'execution_id': execution_id}
         else:
             token = self.decode_token(token)
@@ -66,7 +67,7 @@ class ManifestService(AbstractService):
         else:
             return 0, result
 
-    def _start_manifest_generation(self, filters: dict, execution_id: str):
+    def _start_manifest_generation(self, format: str, filters: dict, execution_id: str) -> None:
         """
         Start the execution of a state machine generating the manifest
 
--- a/src/azul/service/responseobjects/elastic_request_builder.py
+++ b/src/azul/service/responseobjects/elastic_request_builder.py
@@ -531,7 +531,7 @@ class ElasticTransformDump(object):
 
         return final_response
 
-    def transform_manifest(self, request_config_file='request_config.json', filters=None):
+    def transform_manifest(self, format, filters=None, request_config_file='request_config.json'):
         config_folder = os.path.dirname(service_config.__file__)
         request_config_path = "{}/{}".format(config_folder, request_config_file)
         request_config = self.open_and_return_json(request_config_path)
@@ -549,7 +549,7 @@ class ElasticTransformDump(object):
                                         source_filter=source_filter,
                                         enable_aggregation=False)
 
-        manifest = ManifestResponse(es_search, manifest_config, request_config['translation'])
+        manifest = ManifestResponse(es_search, manifest_config, request_config['translation'], format)
 
         return manifest.return_response()
 
--- a/src/azul/service/responseobjects/hca_response_v5.py
+++ b/src/azul/service/responseobjects/hca_response_v5.py
@@ -7,6 +7,7 @@ from tempfile import TemporaryDirectory
 from io import TextIOWrapper, StringIO
 from itertools import chain
 import logging
+from typing import Set, MutableSet
 
 from uuid import uuid4
 
@@ -266,12 +267,14 @@ class ManifestResponse(AbstractResponse):
                                           self.manifest_entries['bundles'].keys(),
                                           self.manifest_entries['contents.files'].keys(),
                                           ['file_url'])))
+        participants = set()
+        for hit in es_search.scan():
+            self._iterate_hit_bdbag(hit, participants, sample_writer)
+
         participant_file_object = StringIO()
         participant_writer = csv.writer(participant_file_object, dialect='excel-tab')
         participant_writer.writerow(['entity:participant_id'])
-        for hit in es_search.scan():
-            self._iterate_hit_bdbag(hit, participant_file_object, participant_writer, sample_writer)
-
+        participant_writer.writerows(zip(participants))
         return self._create_zipped_bdbag(participant_file_object, sample_file_object)
 
     @staticmethod
@@ -281,20 +284,19 @@ class ManifestResponse(AbstractResponse):
         """
 
         with TemporaryDirectory() as bag_path, TemporaryDirectory() as tsv_file_dir:
-            bag = bdbag_api.make_bag(bag_path)
+            # Discard return value since we don't use it, bag is updated below
+            bdbag_api.make_bag(bag_path)
             data_path = os.path.join(bag_path, 'data')
 
             # Write participant and sample data to their respective files.
-            tsv_filenames = ['participant.tsv', 'sample.tsv']
-            for tsv_filename in tsv_filenames:
+            tsvs = [('participant.tsv', participant_file_object),
+                    ('sample.tsv', sample_file_object)]
+            for tsv_filename, file_object in tsvs:
                 with open(os.path.join(tsv_file_dir, tsv_filename), 'w') as f:
-                    if tsv_filename == 'participant.tsv':
-                        f.write(participant_file_object.getvalue())
-                    else:
-                        f.write(sample_file_object.getvalue())
+                    f.write(file_object.getvalue())
                 copy(os.path.join(tsv_file_dir, tsv_filename), data_path)
 
-            assert tsv_filenames == os.listdir(data_path)
+            assert sorted(tsv[0] for tsv in tsvs) == sorted(os.listdir(data_path))
             bag = bdbag_api.make_bag(bag_path, update=True)  # update TSV checksums
             assert bdbag_api.is_bag(bag_path)
             bdbag_api.validate_bag(bag_path)
@@ -313,7 +315,7 @@ class ManifestResponse(AbstractResponse):
             # would download the file twice (https://github.com/DataBiosphere/azul/issues/423).
             writer.writerow(self._translate(bundle, 'bundles') + file_fields)
 
-    def _iterate_hit_bdbag(self, es_search_hit, participant_file_object, participant_writer, sample_writer):
+    def _iterate_hit_bdbag(self, es_search_hit, participants: MutableSet[str], sample_writer):
         hit_dict = es_search_hit.to_dict()
 
         assert len(hit_dict['contents']['specimens']) == 1
@@ -338,9 +340,8 @@ class ManifestResponse(AbstractResponse):
         for bundle in hit_dict['bundles']:
             sample_writer.writerow(chain(specimen_fields[0], specimen_fields[1], cell_suspension_fields[0],
                                          self._translate(bundle, 'bundles'), file_fields + fetch_url))
-            # Keep participant file object as a set.
-            if specimen_fields[1][0] not in participant_file_object.getvalue():
-                participant_writer.writerow(specimen_fields[1])
+            participant_id = specimen_fields[1][0]
+            participants.add(participant_id)
 
     def return_response(self):
         if config.disable_multipart_manifests or self.format == 'bdbag':
@@ -468,6 +469,7 @@ class ProjectSummaryResponse(AbstractResponse):
     """
     Build summary field for each project in projects endpoint
     """
+
     def return_response(self):
         return self.apiResponse
 
--- a/src/azul/service/responseobjects/hca_response_v5.py
+++ b/src/azul/service/responseobjects/hca_response_v5.py
@@ -1,11 +1,13 @@
 #!/usr/bin/python
 import abc
 from collections import OrderedDict, defaultdict
+import os
 import csv
+from tempfile import TemporaryDirectory
 from io import TextIOWrapper, StringIO
 from itertools import chain
 import logging
-import os
+
 from uuid import uuid4
 
 from chalice import Response
@@ -15,7 +17,9 @@ from jsonobject import (FloatProperty,
                         ListProperty,
                         ObjectProperty,
                         StringProperty)
-
+from bdbag import bdbag_api
+from shutil import copy
+from more_itertools import one
 from azul import config
 from azul.service.responseobjects.storage_service import (MultipartUploadHandler,
                                                           StorageService,
@@ -212,7 +216,7 @@ class ManifestResponse(AbstractResponse):
                 writer.writerow(list(self.manifest_entries['bundles'].keys()) +
                                 list(self.manifest_entries['contents.files'].keys()))
                 for hit in self.es_search.scan():
-                    self._iterate_hit(hit, writer)
+                    self._iterate_hit_tsv(hit, writer)
 
         return object_key
 
@@ -222,9 +226,19 @@ class ManifestResponse(AbstractResponse):
 
         :return: S3 object key
         """
-        parameters = dict(object_key=f'manifests/{uuid4()}.tsv',
-                          data=self._construct_tsv_content().encode(),
-                          content_type='text/tab-separated-values')
+        if self.format == 'tsv':
+            data = self._construct_tsv_content().encode()
+            content_type = 'text/tab-separated-values'
+            object_key = f'manifests/{uuid4()}.tsv'
+        elif self.format == 'bdbag':
+            data = self._construct_bdbag()
+            content_type = 'application/zip'
+            object_key = f'manifests/{uuid4()}.zip'
+        else:
+            assert False
+        parameters = dict(object_key=object_key,
+                          data=data,
+                          content_type=content_type)
         return self.storage_service.put(**parameters)
 
     def _construct_tsv_content(self):
@@ -232,33 +246,114 @@ class ManifestResponse(AbstractResponse):
 
         output = StringIO()
         writer = csv.writer(output, dialect='excel-tab')
-
         writer.writerow(list(self.manifest_entries['bundles'].keys()) +
                         list(self.manifest_entries['contents.files'].keys()))
         for hit in es_search.scan():
-            self._iterate_hit(hit, writer)
+            self._iterate_hit_tsv(hit, writer)
 
         return output.getvalue()
 
-    def _iterate_hit(self, es_search_hit, writer):
+    def _construct_bdbag(self) -> str:
+        """
+        Create and return a file object of the sample data.
+        """
+        es_search = self.es_search
+        sample_file_object = StringIO()
+        sample_writer = csv.writer(sample_file_object, dialect='excel-tab')
+        # Add fieldnames.
+        sample_writer.writerow(list(chain(self.manifest_entries['contents.specimens'].keys(),
+                                          self.manifest_entries['contents.cell_suspensions'].keys(),
+                                          self.manifest_entries['bundles'].keys(),
+                                          self.manifest_entries['contents.files'].keys(),
+                                          ['file_url'])))
+        participant_file_object = StringIO()
+        participant_writer = csv.writer(participant_file_object, dialect='excel-tab')
+        participant_writer.writerow(['entity:participant_id'])
+        for hit in es_search.scan():
+            self._iterate_hit_bdbag(hit, participant_file_object, participant_writer, sample_writer)
+
+        return self._create_zipped_bdbag(participant_file_object, sample_file_object)
+
+    @staticmethod
+    def _create_zipped_bdbag(participant_file_object: StringIO, sample_file_object: StringIO) -> str:
+        """
+        Create write participant and sample data files to disk, and create and return zipped BDBag.
+        """
+
+        with TemporaryDirectory() as bag_path, TemporaryDirectory() as tsv_file_dir:
+            bag = bdbag_api.make_bag(bag_path)
+            data_path = os.path.join(bag_path, 'data')
+
+            # Write participant and sample data to their respective files.
+            tsv_filenames = ['participant.tsv', 'sample.tsv']
+            for tsv_filename in tsv_filenames:
+                with open(os.path.join(tsv_file_dir, tsv_filename), 'w') as f:
+                    if tsv_filename == 'participant.tsv':
+                        f.write(participant_file_object.getvalue())
+                    else:
+                        f.write(sample_file_object.getvalue())
+                copy(os.path.join(tsv_file_dir, tsv_filename), data_path)
+
+            assert tsv_filenames == os.listdir(data_path)
+            bag = bdbag_api.make_bag(bag_path, update=True)  # update TSV checksums
+            assert bdbag_api.is_bag(bag_path)
+            bdbag_api.validate_bag(bag_path)
+            assert bdbag_api.check_payload_consistency(bag)
+
+            return bdbag_api.archive_bag(bag_path, 'zip')
+
+    def _iterate_hit_tsv(self, es_search_hit, writer):
         hit_dict = es_search_hit.to_dict()
         assert len(hit_dict['contents']['files']) == 1
         file = hit_dict['contents']['files'][0]
         file_fields = self._translate(file, 'contents.files')
+
         for bundle in hit_dict['bundles']:
             # FIXME: If a file is in multiple bundles, the manifest will list it twice. `hca dss download_manifest`
             # would download the file twice (https://github.com/DataBiosphere/azul/issues/423).
             writer.writerow(self._translate(bundle, 'bundles') + file_fields)
 
+    def _iterate_hit_bdbag(self, es_search_hit, participant_file_object, participant_writer, sample_writer):
+        hit_dict = es_search_hit.to_dict()
+
+        assert len(hit_dict['contents']['specimens']) == 1
+        assert len(hit_dict['contents']['cell_suspensions']) == 1
+        assert len(hit_dict['contents']['files']) == 1
+
+        specimen = one(hit_dict['contents']['specimens'])
+        cell_suspension = one(hit_dict['contents']['cell_suspensions'])
+        file = one(hit_dict['contents']['files'])
+
+        specimen_fields = self._translate(specimen, 'contents.specimens')
+        cell_suspension_fields = self._translate(cell_suspension, 'contents.cell_suspensions')
+        file_fields = self._translate(file, 'contents.files')
+
+        # Construct URL for file.
+        file_id = file['uuid']
+        version = file['version']
+        replica = 'gcp'
+        endpoint = f'files/{file_id}?version={version}&replica={replica}'
+        fetch_url = [config.dss_endpoint + '/' + endpoint]
+
+        for bundle in hit_dict['bundles']:
+            sample_writer.writerow(chain(specimen_fields[0], specimen_fields[1], cell_suspension_fields[0],
+                                         self._translate(bundle, 'bundles'), file_fields + fetch_url))
+            # Keep participant file object as a set.
+            if specimen_fields[1][0] not in participant_file_object.getvalue():
+                participant_writer.writerow(specimen_fields[1])
+
     def return_response(self):
-        object_key = self._push_content_single_part() if config.disable_multipart_manifests else self._push_content()
+        if config.disable_multipart_manifests or self.format == 'bdbag':
+            object_key = self._push_content_single_part()
+        else:
+            object_key = self._push_content()
         file_name = 'hca-manifest-' + object_key.rsplit('/', )[-1]
         presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
         headers = {'Content-Type': 'application/json', 'Location': presigned_url}
 
         return Response(body='', headers=headers, status_code=302)
 
-    def __init__(self, es_search, manifest_entries, mapping):
+    def __init__(self, es_search, manifest_entries, mapping, format):
         """
         The constructor takes the raw response from ElasticSearch and creates
         a csv file based on the columns from the manifest_entries
@@ -270,6 +365,7 @@ class ManifestResponse(AbstractResponse):
         self.manifest_entries = OrderedDict(manifest_entries)
         self.mapping = mapping
         self.storage_service = StorageService()
+        self.format = format
 
 
 class EntryFetcher:
--- a/test/service/test_manifest_endpoints.py
+++ b/test/service/test_manifest_endpoints.py
@@ -51,7 +51,7 @@ class ManifestEndpointTest(WebServiceTestCase):
         mock_uuid.return_value = execution_name
         step_function_helper.describe_execution.return_value = {'status': 'RUNNING'}
         filters = {'file': {'organ': {'is': ['lymph node']}}}
-        current_request.query_params = {'filters': json.dumps(filters)}
+        current_request.query_params = {'filters': json.dumps(filters), 'format': format}
         response = start_manifest_generation()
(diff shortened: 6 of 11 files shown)


src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,9 +224,15 @@ def _push_content_single_part(self) -> str:
 
         :return: S3 object key
         """
+        if self.format == 'tsv':
+            data = self._construct_tsv_content().encode()
+            content_type = 'text/tab-separated-values'
+        else:
```
hannes-ucsc: Every possible value of `self.format` should be explicitly tested for and the else branch should read:

```
else:
    assert False
```

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -233,32 +241,109 @@ def _construct_tsv_content(self):
         output = StringIO()
         writer = csv.writer(output, dialect='excel-tab')
 
-        writer.writerow(list(self.manifest_entries['bundles'].keys()) +
-                        list(self.manifest_entries['contents.files'].keys()))
+        writer.writerow(
```
hannes-ucsc: For the time being, don't format code that you wouldn't otherwise touch. Only people familiar with style conventions should do that and only if they keep those cosmetic changes in a separate commit.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -233,32 +241,109 @@ def _construct_tsv_content(self):
...
-            writer.writerow(self._translate(bundle, 'bundles') + file_fields)
+        if self.format == 'tsv':
+            file_fields = self._translate(file, 'contents.files')
+
+            for bundle in hit_dict['bundles']:
+                # FIXME: If a file is in multiple bundles, the manifest will list it twice. `hca dss download_manifest`
+                # would download the file twice (https://github.com/DataBiosphere/azul/issues/423).
+                writer.writerow(
+                    self._translate(bundle, 'bundles') + file_fields)
+        else:  # construct manifest for BDBag
```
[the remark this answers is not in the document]
hannes-ucsc: In fact, just create two versions of _iterate_hit(), one for BDBag and one for TSV. Don't attempt to have one method handle both cases. I can explain in person why if necessary.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -233,32 +241,109 @@ def _construct_tsv_content(self):
...
+            if participant_writer:
+                file_fields_participant = self._translate(file, 'contents.files')
+                participant_writer.writerow(file_fields_participant)
```
hannes-ucsc: This will create duplicates, won't it? I'd expect there to be one row per participant in this TSV. I may be wrong.

test/service/test_manifest_endpoints.py
```diff
@@ -51,14 +51,17 @@ def test_manifest_endpoint_start_execution(self, mock_uuid, current_request, ste
         mock_uuid.return_value = execution_name
         step_function_helper.describe_execution.return_value = {'status': 'RUNNING'}
         filters = {'file': {'organ': {'is': ['lymph node']}}}
-        current_request.query_params = {'filters': json.dumps(filters)}
+        format = 'tsv'
+        current_request.query_params = {'filters': json.dumps(filters),
+                                        'format': format}
```
hannes-ucsc: Don't pass format here to prove that the default works.

test/service/test_request_validation.py
```diff
@@ -177,3 +179,43 @@ def test_manifest(self):
...
+            url = self.base_url +\
+                  '/repository/files/export?filters={"file":{}}&format=bdbag'
+            response = requests.get(url)
+            self.assertEqual(200, response.status_code,
+                             'Unable to download manifest')
+            zip_fh = ZipFile(response.text, 'r')
+            zip_fh.extractall(zip_dir)
+            zip_fh.close()
+            zip_fname = os.path.basename(os.path.splitext(response.text)[0])
+            fh = open(os.path.join(zip_dir, zip_fname, 'data', 'sample.tsv'))
```
hannes-ucsc: This leaks a file handle when one of the assertions below fails. Always use `with open(…`

test/service/test_request_validation.py
```diff
@@ -177,3 +179,43 @@ def test_manifest(self):
...
+            tsv_file = csv.DictReader(fh, delimiter='\t')
```
hannes-ucsc:
```suggestion
            sample_tsv = csv.DictReader(fh, delimiter='\t')
```

hannes-ucsc: Or
```suggestion
            samples = csv.DictReader(fh, delimiter='\t')
```

test/service/test_request_validation.py
```diff
@@ -177,3 +179,43 @@ def test_manifest(self):
...
+            # 2 because self.bundle has 2 files
+            self.assertEqual(len(list(tsv_file)), 2,
+                             'Wrong number of files were found.')
+            manifest_config = json.load(
+                open('{}/request_config.json'.format(self.service_config_dir),
+                     'r'))['manifest']
+            expected_fieldnames = \
```
hannes-ucsc: Don't make this dynamic, `request_config.json` is going away. Just hard code the expected column names.

hannes-ucsc: The problem with this style of dynamically generated assertion is that it uses exactly the same approach as the code under test. If there is bug in that approach, the assertion will also be wrong and the bug will go undetected.

lambdas/service/app.py
```diff
@@ -548,7 +549,11 @@ def get_manifest():
     es_td = EsTd()
     # Get the response back
     logger.info("Creating the API response")
-    response = es_td.transform_manifest(filters=filters)
+    try:
+        response = es_td.transform_manifest(filters=filters, format=format)
+    except ValueError as e:
+            return 400, str(e)
```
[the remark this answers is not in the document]
hannes-ucsc: Also, this works? Does it actually return an HTTP response with status 400?

hannes-ucsc:
Now that David's DRS PR landed, I'll leave it open to you if you want the URLs to point at the DRS endpoint or directly at `/fetch/files/dss`. Whatever you decide, the BDBag should be fully functional.

hannes-ucsc:
Please don't assign PRs to me until the comments from the previous review cycle are fully addressed.

test/service/test_request_validation.py
mjkrause: fixed

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -549,7 +549,8 @@ def transform_request(self,
 
         return final_response
 
-    def transform_manifest(self, request_config_file='request_config.json', filters=None):
+    def transform_manifest(self, request_config_file='request_config.json',
```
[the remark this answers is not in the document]
mjkrause: I'm implementing this style as I work through your comments.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/python
 import abc
 from collections import OrderedDict, defaultdict
-import csv
+import os, csv
```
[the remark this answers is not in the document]
mjkrause: fixed

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: fixed

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Reverted, sorry was an accident.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Yes, that didn't eliminate duplicates if present. Fixed it.

src/azul/service/responseobjects/hca_response_v5.py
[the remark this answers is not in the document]
mjkrause: I split it into two methods, `_iterate_hit_tsv` and `_iterate_hit_bdbag`. Yes, I'd like to hear your take on why this design is preferred...

test/service/test_manifest_endpoints.py
mjkrause: I tried several variations. It always comes down to the missing format parameter in `start_manifest_generation` in `ManifestService`.

src/azul/__init__.py
```diff
@@ -109,6 +109,10 @@ def dss_checkout_bucket(self):
     def num_dss_workers(self) -> int:
         return int(os.environ['AZUL_DSS_WORKERS'])
 
+    @property
+    def drs_endpoint(self) -> str:
```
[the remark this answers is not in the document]
mjkrause: removed

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +225,146 @@ def _push_content_single_part(self) -> str:
...
+
+            for tsv_file in data.keys():
+                copy(os.path.join(tsv_file_dir, tsv_file), data_path)
+            assert ['participant.tsv', 'sample.tsv'] == os.listdir(data_path)
+            bag = bdbag_api.make_bag(bag_path, update=True)  # update TSV checksums
+            assert bdbag_api.is_bag(bag_path)
+            bdbag_api.validate_bag(bag_path)
+            assert bdbag_api.check_payload_consistency(bag)
+
+            # TODO: in case of concurrency, is it one user = one lambda?
```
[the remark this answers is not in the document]
mjkrause: Agreed. I have some ideas, but before I act on them I would like to discuss it. Some thoughts are:

1. Writing data to the Lambda disk doesn't scale and isn't the right approach in the long-run (as previously discussed), but would that problem persist if we created the bag in memory? (and if so, how?)
2. We might be able to use some form of caller / user info or ID. In that case we can tag the bag by making the ID part of the bag name and prevent that mix-up (that doesn't clarify point 1, though).

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +225,146 @@ def _push_content_single_part(self) -> str:
...
             writer.writerow(self._translate(bundle, 'bundles') + file_fields)
 
+    def _iterate_hit_bdbag(self, es_search_hit, writer, participant_writer=None):
+        hit_dict = es_search_hit.to_dict()
+        assert len(hit_dict['contents']['files']) == 1
+        file = hit_dict['contents']['files'][0]
+
+        if participant_writer:
+            file_fields_participant = self._translate(file, 'contents.files')
+            file_fields_participant = [file_fields_participant[0]] + list(set(file_fields_participant[1:]))
```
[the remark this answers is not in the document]
mjkrause: Sure, although making rows unique was my intent. Here's a snippet:
```
ln [15]: participant = ['participant_id', 'id1', 'id2', 'id1', 'id3']                        

In [16]: [participant[0]] + list(set(participant[1:]))                                       
Out[16]: ['participant_id', 'id1', 'id3', 'id2']
```

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +225,146 @@ def _push_content_single_part(self) -> str:
...
+            participant_writer.writerow(file_fields_participant)
+        file_fields = self._translate(file, 'contents.specimens')
+        file_fields.append(self._translate(file, 'contents.cell_suspensions'))
+        file_fields.append(self._translate(file, 'contents.files'))
+
+        file_id = file['uuid']
+        drs_uri =  [f'drs://{file_id}']
```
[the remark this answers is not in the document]
mjkrause: Changed it to a URL that returns the file location. That solution contains several hard-coded parameters. I'm working on trying to minimize this.

lambdas/service/app.py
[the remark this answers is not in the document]
mjkrause: I just don't know how that code got there in the first place. I suppose I wrote it, but have no recollection. AFIAK, a hard-coded status code 302 (Chalice `Response`) is returned to that method coming from `ManifestResponse` - I'm not sure how it could be anything else. I removed this code for the time being.

src/azul/service/responseobjects/hca_response_v5.py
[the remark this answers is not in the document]
mjkrause: The [latest iteration takes a completely different approach](https://github.com/DataBiosphere/azul/blob/feature/mkrause/604-export-to-terra/src/azul/service/responseobjects/hca_response_v5.py#L671) to this, which I think is both more direct and fast.

src/azul/service/responseobjects/hca_response_v5.py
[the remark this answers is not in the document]
mjkrause: Per our slack convo on Monday I have tried to make use of method `file_url`, but unfortunately ran into the problems which I haven't been able to solve, yet. Stack trace is:
```
     fetch_url = file_url(file_id, version=version, replica=replica)
  File "/home/michael/dev/azul/lambdas/service/app.py", line 827, in file_url
    url = self_url(endpoint_path=f'{proxy_endpoint_path}/{uuid}')
  File "/home/michael/dev/azul/lambdas/service/app.py", line 833, in self_url
    protocol = app.current_request.headers.get('x-forwarded-proto', 'http')
AttributeError: 'NoneType' object has no attribute 'headers'
127.0.0.1 - - [06/Feb/2019 12:59:30] "GET /repository/files/export?filters=%7B%22file%22:%7B%7D%7D&format=bdbag HTTP/1.1" 500 -

Unable to download manifest
500 != 200
```

CHANGELOG.yaml
```diff
@@ -14,6 +14,14 @@ meta:
 
 changes:
 
+  - title: Add export of primary and matrix data to Broad's Terra analysis platform
```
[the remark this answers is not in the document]
mjkrause: done and done

test/service/test_request_validation.py
```diff
@@ -177,3 +179,31 @@ def test_manifest(self):
...
+            with ZipFile(response.text, 'r') as zip_fh:
+                zip_fh.extractall(zip_dir)
+            zip_fname = os.path.basename(os.path.splitext(response.text)[0])
+            with open(os.path.join(zip_dir, zip_fname, 'data', 'sample.tsv')) as fh:
+                samples = csv.DictReader(fh, delimiter='\t')
+                self.assertEqual(len(list(samples)), 2, 'Wrong number of files were found.')  # self.bundle has 2 files
+                expected_fieldnames = ['entity:sample_id', 'participant_id', 'cell_suspension_id', 'bundle_uuid',
+                                       'bundle_version', 'file_content_type', 'file_name', 'file_sha256', 'file_size',
+                                       'file_uuid', 'file_version', 'file_indexed', 'drs_uri']
+                self.assertEqual(expected_fieldnames, samples.fieldnames,
```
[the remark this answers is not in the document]
mjkrause: I would like to discuss this step one more time with you, perhaps next week. Also, pl. see my comment right above re. this.

mjkrause:
I changed the construction of the file URL (in the  primary data BDBag), which uses the `GET /files` endpoint. I hard-coded most of the URL, including the replica.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,135 @@ def _push_content_single_part(self) -> str:
...
+        participant_writer = csv.writer(participant_tsv, dialect='excel-tab')
+        participant_writer.writerow(['entity:participant_id'])
+
+        for hit in es_search.scan():
+            self._iterate_hit_bdbag(hit, sample_writer, participant_writer)
+
+        return self._create_zipped_bdbag({'participant.tsv': participant_tsv, 'sample.tsv': sample_tsv})
+
+    @staticmethod
+    def _create_zipped_bdbag(data: dict):
```
hannes-ucsc: If you annotate the argument type, you should also annotate the return value. Instead of dict, use `typing.Mapping` and parameterize by key and value type.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,135 @@ def _push_content_single_part(self) -> str:
...
+        :returns bdbag: archived (zipped) BDBag with values of data as payloads
+        """
+        with TemporaryDirectory() as bag_path, TemporaryDirectory() as tsv_file_dir:
+            bag = bdbag_api.make_bag(bag_path)
+            data_path = os.path.join(bag_path, 'data')
+            for fname, fobj in data.items():
+                with open(os.path.join(tsv_file_dir, fname), 'w') as f:
+                    if fname == 'participant.tsv':
+                        s = fobj.getvalue()  # remove duplicates from participant.tsv
+                        # TODO: this is rather cryptic, and could be expanded for clarity
```
hannes-ucsc: Address or remove.

src/azul/service/responseobjects/hca_response_v5.py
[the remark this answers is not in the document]
hannes-ucsc: I actually don't understand the concern. Everything happens in temporary directories. Where is the potential for mixup? Please reach out to me in person or on Slack to discuss.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,135 @@ def _push_content_single_part(self) -> str:
...
         for bundle in hit_dict['bundles']:
             # FIXME: If a file is in multiple bundles, the manifest will list it twice. `hca dss download_manifest`
             # would download the file twice (https://github.com/DataBiosphere/azul/issues/423).
             writer.writerow(self._translate(bundle, 'bundles') + file_fields)
 
+    def _iterate_hit_bdbag(self, es_search_hit, sample_writer, participant_writer):
+        hit_dict = es_search_hit.to_dict()
+        assert len(hit_dict['contents']['files']) == 1
+
+        specimen = hit_dict['contents']['specimens'][0]
```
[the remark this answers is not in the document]
mjkrause: Done (nice tool). Also asserted that the list's lengths is 1.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,135 @@ def _push_content_single_part(self) -> str:
...
 
         return output.getvalue()
 
-    def _iterate_hit(self, es_search_hit, writer):
+    def _construct_bdbag(self) -> dict:
+        """Create participant.tsv and sample.tsv and return as zipped BDBag."""
+        es_search = self.es_search
+        sample_tsv = StringIO()
+        sample_writer = csv.writer(sample_tsv, dialect='excel-tab')
+        sample_writer.writerow(
```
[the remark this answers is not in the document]
mjkrause: You addressed this above. As suggested there, I used the `chain` tool to format concatenate all lists.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Added a comment for clarification and removed the TODO.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: I'm still learning about the generic types of the typing module. I tried very `typing.Mapping`, but for now got stuck on how to best handle the iteration in over that structure in `_create_zipped_bag`. I found using `Mapping.Dict` to be the best compromise as (a) it provides finer type granularity compared to `dict`, and (b) retains the ease of iterating over a dictionary. Let me know what you think?

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,135 @@ def _push_content_single_part(self) -> str:
...
+            list(self.manifest_entries['contents.specimens'].keys()) +
+            list(
```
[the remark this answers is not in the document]
mjkrause: fixed

src/azul/service/responseobjects/hca_response_v5.py
hannes-ucsc: > Still puzzled as to how to specify the output type of a `def` if it writes a file disk in the return statement.

This function returns the return value of `archive_bag` which is a path to the file as a `str`, AFAICT. This seem straight-forward, am I missing something?

> I found using `Mapping.Dict` to be the best compromise as (a) it provides finer type granularity compared to `dict`, and (b) retains the ease of iterating over a dictionary. Let me know what you think?

`typing.Mapping` lets you iterate over items. If you need to modify the dict, you should use `typing.MutableMapping`. `typing.Dict` is rarely used. Please show me exactly what prevents you from using `typing.Mapping`. A PyCharm warning? If so which one? Screen shot or Slack call would be appreciated. Replying to comments takes too much of my time.

lambdas/service/app.py
```diff
@@ -705,10 +709,13 @@ def generate_manifest(event, context):
     :param: event: dict containing function input
         Valid params:
             - filters: dict containing filters to use in ES request
+            - format: str to specify manifest output format, values are
+                      'tsv' (default) or 'bdbag'
     :return: The URL to the generated manifest
     """
     filters = event.get('filters', {'file': {}})
-    response = EsTd().transform_manifest(filters=filters)
+    format = event.get('format', 'tsv')
```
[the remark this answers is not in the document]
mjkrause: The specification/invocation of `format` occurs now at the same 2 places as the one for the `files` parameter, [here](https://github.com/DataBiosphere/azul/blob/feature/mkrause/604-export-to-terra/lambdas/service/app.py#L411) and [here](https://github.com/DataBiosphere/azul/blob/feature/mkrause/604-export-to-terra/lambdas/service/app.py#L569), so it is consistent for the two query parameters. Both are in the service app.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: If you've got some time let's discuss the Dict/Mapping (and potentially others). I left it as a Dict for now.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +228,135 @@ def _push_content_single_part(self) -> str:
...
+        specimen = one(hit_dict['contents']['specimens'])
+        cell_suspension = one(hit_dict['contents']['cell_suspensions'])
+        file = hit_dict['contents']['files'][0]
+
+        specimen_fields = self._translate(specimen, 'contents.specimens')
+        cell_suspension_fields = self._translate(cell_suspension, 'contents.cell_suspensions')
+        file_fields = self._translate(file, 'contents.files')
+
+        # Construct URL for file.
+        base_url = 'https://service.staging.explore.data.humancellatlas.org'
```
hannes-ucsc: We can't hard-code the DSS instance, obviously.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +228,135 @@ def _push_content_single_part(self) -> str:
...
+        file_id = file['uuid']
+        version = file['version']
+        replica = 'gcp'
+        endpoint = f'repository/files/{file_id}?version={version}&replica={replica}'
```
hannes-ucsc: The DSS does not have `/repository/…`.  It looks like you never actually tested this.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +228,135 @@ def _push_content_single_part(self) -> str:
...
+        fetch_url = [urljoin(base=base_url, url=endpoint)]
+
+        for bundle in hit_dict['bundles']:
+            sample_writer.writerow(chain(specimen_fields[0], specimen_fields[1], cell_suspension_fields[0],
+                                         self._translate(bundle, 'bundles'), file_fields + fetch_url))
+            participant_writer.writerow(specimen_fields[1])
```
hannes-ucsc: This implies that the participant TSV has as many rows as the sample TSV. That can't be right.

test/service/test_request_validation.py
```diff
@@ -177,3 +179,49 @@ def test_manifest(self):
...
+                self.assertEqual(len(list(samples)), 2, 'Found incorrect number of files.')  # self.bundle has 2 files
+                expected_fieldnames = ['entity:sample_id', 'participant_id', 'cell_suspension_id', 'bundle_uuid',
+                                       'bundle_version', 'file_content_type', 'file_name', 'file_sha256', 'file_size',
+                                       'file_uuid', 'file_version', 'file_indexed', 'file_url']
+                self.assertEqual(expected_fieldnames, samples.fieldnames,
+                                 'Manifest headers are not configured correctly')
+                for line in samples:
+                    self.assertEqual(len(line), 13, 'Incorrect number of entries in row.')
+                    for k, v in line.items():
+                        print(f'{k}:{v}')
```
hannes-ucsc: remove

test/service/test_request_validation.py
```diff
@@ -177,3 +179,49 @@ def test_manifest(self):
...
+                        self.assertFalse(not v, 'Value {v} in sample.tsv is strings.')
```
hannes-ucsc: The error message make no sense. Also, `assertTrue( bool(v), …` is more expressive. Exploiting the equivalence between `not  x` and `not bool(x)` is borderline insidious.

test/service/test_request_validation.py
```diff
@@ -177,3 +179,49 @@ def test_manifest(self):
...
+            with open(os.path.join(zip_dir, zip_fname, 'data', 'sample.tsv'), 'r') as fh:
+                samples = csv.DictReader(fh, delimiter='\t')
+                self.assertEqual(len(list(samples)), 2, 'Found incorrect number of files.')  # self.bundle has 2 files
+                expected_fieldnames = ['entity:sample_id', 'participant_id', 'cell_suspension_id', 'bundle_uuid',
+                                       'bundle_version', 'file_content_type', 'file_name', 'file_sha256', 'file_size',
+                                       'file_uuid', 'file_version', 'file_indexed', 'file_url']
+                self.assertEqual(expected_fieldnames, samples.fieldnames,
+                                 'Manifest headers are not configured correctly')
+                for line in samples:
+                    self.assertEqual(len(line), 13, 'Incorrect number of entries in row.')
```
hannes-ucsc: This needs stricter assertions on the actual values. Especially—like I requested before—the URL to the file. You don't need to show that the URL actually works (the file UUID and version can be fake) but I need proof that it would be a valid DSS URL if the UUID and version were real.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Initially yes, but it's pruned for duplicates in [_create_zipped_bdbag](https://github.com/DataBiosphere/azul/blob/feature/mkrause/604-export-to-terra/src/azul/service/responseobjects/hca_response_v5.py#L297).

test/service/test_request_validation.py
mjkrause: Have re-written assertions around `sample.tsv`. It now includes regex check for empty strings in entire TSV, and asserts entire row of that TSV.

test/service/test_request_validation.py
mjkrause: Implemented, see comment right above.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Base URL is now from `dss_endpoint`, successfully `curl`ed UUID _26dc01c0-8dcb-453e-b53e-24a7b48e2e32_, version _2018-10-23T153112.622014Z_.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +227,143 @@ def _push_content_single_part(self) -> str:
...
+
+        specimen_fields = self._translate(specimen, 'contents.specimens')
+        cell_suspension_fields = self._translate(cell_suspension, 'contents.cell_suspensions')
+        file_fields = self._translate(file, 'contents.files')
+
+        # Construct URL for file.
+        file_id = file['uuid']
+        version = file['version']
+        replica = 'gcp'
+        endpoint = f'v1/files/{file_id}?version={version}&replica={replica}'
```
```suggestion
        endpoint = f'/files/{file_id}?version={version}&replica={replica}'
```

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +227,143 @@ def _push_content_single_part(self) -> str:
...
+        fetch_url = [urljoin(base=config.dss_endpoint, url=endpoint)]
```
hannes-ucsc: It's inconsistent to use urljoin but not use https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode for embedding version and replica. We know that version and replica are urlsafe so I'm ok with the latter. I don't think urljoin() actually buys us anything over just concatenating so

```suggestion
        fetch_url = config.dss_endpoint + '/' + endpoint
```

In fact, you mention `v1` in `endpoint` above which duplicates what's already at the end of `config.dss_endpoint` so your use of urljoin() makes this overly complicated.

test/service/test_request_validation.py
```diff
@@ -177,3 +181,55 @@ def test_manifest(self):
...
+                observed = []
+                for row in participant:
+                    observed.append(row[0])
+                self.assertEqual(len(observed), 2, 'participant.tsv contains duplicates.')
+                def checkListEqual(L1, L2):
+                    return len(L1) == len(L2) and sorted(L1) == sorted(L2)
+                self.assertTrue(checkListEqual(expected, observed), 'participant.tsv contains incorrect data')
+
+            with open(os.path.join(zip_dir, zip_fname, 'data', 'sample.tsv'), 'r') as fh:
+                expected_fieldnames = ['entity:sample_id', 'participant_id', 'cell_suspension_id', 'bundle_uuid',
```
hannes-ucsc: It would help to correlate the column names with the values and to list each field on a separate line.

```suggestion
                expectations = [
                    ('entity:sample_id', 'DID_scRSq06'),
                    ...
                ]
                   
expected_header, expected_row = zip(*expectations)
```

In this case I am fine with hanging indent, BTW.

hannes-ucsc: You may also be able to inline `file_url` in that case.

test/service/test_request_validation.py
```diff
@@ -177,3 +181,55 @@ def test_manifest(self):
...
+                zip_fh.extractall(zip_dir)
+            zip_fname = os.path.basename(os.path.splitext(response.text)[0])
+            with open(os.path.join(zip_dir, zip_fname, 'data', 'participant.tsv'), 'r') as fh:
+                participant = csv.reader(fh, delimiter='\t')
+                expected = ['entity:participant_id', '7b07b9d0-cc0e-4098-9f64-f4a569f7d746']
+                observed = []
+                for row in participant:
+                    observed.append(row[0])
+                self.assertEqual(len(observed), 2, 'participant.tsv contains duplicates.')
+                def checkListEqual(L1, L2):
```
hannes-ucsc: Why does this need to be order-insensitive? The expectation is `['entity:participant_id', '7b07b9d0-cc0e-4098-9f64-f4a569f7d746']`. There is only one other ordering of that: `['7b07b9d0-cc0e-4098-9f64-f4a569f7d746', 'entity:participant_id']` and that would imply we are ok with the header not being the first line in the TSV. That's just insane.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +227,143 @@ def _push_content_single_part(self) -> str:
...
+
+        :parameter sample_file_object: file objects of sample data
+        :returns bdbag: archived (zipped) BDBag with values of participant and sample data in payload
+        """
+
+        with TemporaryDirectory() as bag_path, TemporaryDirectory() as tsv_file_dir:
+            bag = bdbag_api.make_bag(bag_path)
+            data_path = os.path.join(bag_path, 'data')
+
+            # Get sample data from file object and from it create unique participant data.
```
hannes-ucsc: Still not right. I asked you explicitly NOT to parse the finished sample TSV. Maintain a set of participants as you populate `sample_file_object` and pass that set to this function.

test/service/test_request_validation.py
hannes-ucsc: Well, maybe not insane, but incorrect. ;-)

I'm having review fatigue.

src/azul/service/manifest.py
```diff
@@ -66,7 +67,7 @@ def start_or_inspect_manifest_generation(self,
         else:
             return 0, result
 
-    def _start_manifest_generation(self, filters: dict, execution_id: str):
+    def _start_manifest_generation(self, format: str, filters: dict, execution_id: str):
```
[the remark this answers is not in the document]
mjkrause: done

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: done

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Replaced `urljoin` by simple str concatenation.

test/service/test_request_validation.py
mjkrause: Once populated the order of items in list `observed` is inconsistent from run to run. Thus, just using `assertEqual` the test sometimes succeeds, other times fails. That's how I got the idea that we need to sort it. I removed the def and compare now `sorted(...` in the assertion.

test/service/test_request_validation.py
mjkrause: Implemented, looks cleaner and easier to read now.

src/azul/service/responseobjects/hca_response_v5.py
mjkrause: Changed it so that participant is now constructed in the same loop as sample, but as a set.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,134 @@ def _push_content_single_part(self) -> str:
...
+        """
+        Create write participant and sample data files to disk, and create and return zipped BDBag.
+        """
+
+        with TemporaryDirectory() as bag_path, TemporaryDirectory() as tsv_file_dir:
+            bag = bdbag_api.make_bag(bag_path)
+            data_path = os.path.join(bag_path, 'data')
+
+            # Write participant and sample data to their respective files.
+            tsv_filenames = ['participant.tsv', 'sample.tsv']
```
hannes-ucsc: Convoluted. 

```
for x in ['a', 'b']:
    if x == 'a':
        …
    else:
        …
```

is hilarious, wouldn't you agree?

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -222,43 +226,134 @@ def _push_content_single_part(self) -> str:
...
+        file_id = file['uuid']
+        version = file['version']
+        replica = 'gcp'
+        endpoint = f'files/{file_id}?version={version}&replica={replica}'
+        fetch_url = [config.dss_endpoint + '/' + endpoint]
+
+        for bundle in hit_dict['bundles']:
+            sample_writer.writerow(chain(specimen_fields[0], specimen_fields[1], cell_suspension_fields[0],
+                                         self._translate(bundle, 'bundles'), file_fields + fetch_url))
+            # Keep participant file object as a set.
```
hannes-ucsc: This is not using a set and it is horribly O(n^2).

test/service/test_request_validation.py
```diff
@@ -177,3 +181,60 @@ def test_manifest(self):
...
+                                ('file_content_type', 'application/gzip; dcp-type=data'),
+                                ('file_name', 'SRR3562915_1.fastq.gz'),
+                                ('file_sha256', '77337cb51b2e584b5ae1b99db6c163b988cbc5b894dda2f5d22424978c3bfc7a'),
+                                ('file_size', '195142097'),
+                                ('file_uuid', '7b07f99e-4a8a-4ad0-bd4f-db0d7a00c7bb'),
+                                ('file_version', '2018-11-02T113344.698028Z'),
+                                ('file_indexed', 'False'),
+                                ('file_url', config.dss_endpoint + '/files/7b07f99e-4a8a-4ad0-bd4f-db0d7a00c7bb?'
+                                                                   'version=2018-11-02T113344.698028Z&replica=gcp')
+                ]
```
hannes-ucsc: For aligned indent the closing delimiter is at the end of the last line. For hanging indent it's on a line of it's own. You are mixing both styles here.

test/service/test_request_validation.py
```diff
@@ -177,3 +181,60 @@ def test_manifest(self):
...
+            self.assertEqual(200, response.status_code, 'Unable to download manifest')
+            with ZipFile(response.text, 'r') as zip_fh:
+                zip_fh.extractall(zip_dir)
+            zip_fname = os.path.basename(os.path.splitext(response.text)[0])
+            with open(os.path.join(zip_dir, zip_fname, 'data', 'participant.tsv'), 'r') as fh:
+                participant = csv.reader(fh, delimiter='\t')
+                expected = ['entity:participant_id', '7b07b9d0-cc0e-4098-9f64-f4a569f7d746']
+                observed = []
+                for row in participant:
+                    observed.append(row[0])
```
hannes-ucsc: This test would pass if there were extra columns. It can be made simpler and stronger.

test/service/test_request_validation.py
```diff
@@ -177,3 +181,60 @@ def test_manifest(self):
...
+                                ('file_uuid', '7b07f99e-4a8a-4ad0-bd4f-db0d7a00c7bb'),
+                                ('file_version', '2018-11-02T113344.698028Z'),
+                                ('file_indexed', 'False'),
+                                ('file_url', config.dss_endpoint + '/files/7b07f99e-4a8a-4ad0-bd4f-db0d7a00c7bb?'
+                                                                   'version=2018-11-02T113344.698028Z&replica=gcp')
+                ]
+
+                expected_fieldnames, expected_row = map(list, zip(*expectations))
+                samples = csv.reader(fh, delimiter='\t')
+                for row_count, row in enumerate(samples):
```
hannes-ucsc: Consider unrolling this loop. I think it would be simpler.

PR #689, opened 2019-01-29
Set default sort specific to the entity type (fixes #683)

### LINKED ISSUE #683: Default sort of projects tab and files tab is incorrect
The default sort is coming back as specimenId on the first hit for both. 
This should be projectTitle  for project and filename for files.

Thanks!

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-425)
┆Project Name: azul
┆Issue Number: AZUL-425

danielsotirhos:


lambdas/service/app.py
```diff
@@ -51,11 +51,19 @@
...
 
-def _get_pagination(current_request):
+sort_defaults = {
+    'files': ('fileName', 'asc'),
+    'specimens': ('specimenId', 'asc'),
+    'projects': ('projectTitle', 'asc'),
+}
+
+def _get_pagination(current_request, entity_type):
+    default_sort = sort_defaults[entity_type][0]
```
```suggestion
    default_sort, default_order = sort_defaults[entity_type]
```

lambdas/service/app.py
```diff
@@ -51,11 +51,19 @@
...
+    default_order = sort_defaults[entity_type][1]
```
```suggestion
```

lambdas/service/app.py
danielsotirhos: Ah yes, that is cleaner. I'll make that change now.

Issue #628, opened 2018-12-21
Add Authentication to Cart API endpoints


Ben-Wu:
- Get the user id from the JWT tokens
- Remove IP whitelist environment variables

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-398)
┆Project Name: azul
┆Issue Number: AZUL-398
┆Epic: Interactive building of collections

Issue #627, opened 2018-12-21
Export cart to DSS collection API


Ben-Wu:
Cart items should be sent to the `PUT /collections` endpoint at https://dss.data.humancellatlas.org/

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/AZUL-397)
┆Project Name: azul
┆Issue Number: AZUL-397
┆Epic: Interactive building of collections

shiroyuki:
Note that it is difficult to test as we have more than 80k bundles registered in the database (ElasticSearch) but there are just 15 bundles in DSS's S3 bucket (`org-humancellatlas-dss-dev`). Missing information would make development very difficult, especially for very large cart.

PR #703, opened 2019-02-07
Cart Export to DSS Collection API

shiroyuki:
Related to:
* #627 for Cart Export
* #628 for enabling authentication check on the Cart API

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -13,7 +13,7 @@ import urllib.parse
 import boto3
 from botocore.exceptions import ClientError
 # noinspection PyPackageRequirements
-from chalice import BadRequestError, Chalice, ChaliceViewError, NotFoundError, Response, AuthResponse, UnauthorizedError
+from chalice import BadRequestError, Chalice, ChaliceViewError, NotFoundError, Response, AuthResponse
 from more_itertools import one
 import requests
 
@@ -21,7 +21,10 @@ from azul import config
 from azul.health import Health
 from azul.security.authenticator import Authenticator, AuthenticationError
 from azul.service import service_config
+from azul.service.responseobjects.cart_export_job_manager import CartExportJobManager, InvalidExecutionTokenError
 from azul.service.responseobjects.cart_item_manager import CartItemManager, DuplicateItemError, ResourceAccessError
+from azul.service.responseobjects.cart_export_service import CartExportService
+from azul.service.responseobjects.collection_data_access import CollectionDataAccess
 from azul.service.responseobjects.elastic_request_builder import (BadArgumentException,
                                                                   ElasticTransformDump as EsTd,
                                                                   IndexNotFoundError)
@@ -102,13 +105,13 @@ def jwt_auth(auth_request) -> AuthResponse:
     # When this code is running on AWS, the context of the authorizer has to be
     # a string-to-primitive-type dictionary. When the context is retrieved in
     # the caller function, the value somehow got casted into string.
-    return AuthResponse(routes=['*'],
-                        principal_id='user',
-                        context={
-                            prop: value
-                            for prop, value in claims.items()
-                            if not isinstance(value, (list, tuple, dict))
-                        })
+    context = {
+        prop: value
+        for prop, value in claims.items()
+        if not isinstance(value, (list, tuple, dict))
+    }
+    context['token'] = given_token
+    return AuthResponse(routes=['*'], principal_id='user', context=context)
 
 
 @app.route('/', cors=True)
@@ -717,7 +720,7 @@ def access_info():
         k: int(v) if k in ('exp', 'iat') else v
         for k, v in request.context['authorizer'].items()
     }
-    ttl_in_seconds = math.ceil(claims['exp'] - last_updated_timestamp)
+    ttl_in_seconds = math.floor(claims['exp'] - last_updated_timestamp)
     response_body = dict(
         claims=claims,
         ttl=ttl_in_seconds,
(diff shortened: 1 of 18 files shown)


src/azul/service/responseobjects/collection_data_access.py
```diff
@@ -0,0 +1,151 @@
...
+                logger.warning('Request %s: The request fails to respond within the time limit.', request_id)
+                raise ServerTimeoutError(uuid)
+            logger.info('Request %s: Retrying in %d s', request_id, delay)
+            sleep(delay)
+            logger.info('Request %s: Resuming', request_id)
+        request_params = dict(params=params,
+                              headers=dict(Authorization=self.access_token))
+        if payload:
+            request_params['json'] = payload
+        # FIXME "requests" will be replaced with "DSSClient".
```
hannes-ucsc: Didn't we agree we would do this? I can't remember.

src/azul/service/responseobjects/dynamo_data_access.py
```diff
@@ -78,10 +84,20 @@ def query(self, **kwargs):
...
 
+    def iterable_query(self, **kwargs):
+        """
+        Similar to :method:`query` but instead of only yield items, this method
+        will yield result as page.
+
+        :yields: Result page of the query
+        """
+        for items, count, last_evaluated_key in self._make_query(**kwargs):
+            yield ResultPage(last_evaluated_key, iter(items), count)
```
hannes-ucsc: I don't understand why items is eagerly converted into an iterator. If it's an iterable why can't it remain an iterable? Then ResultPage could just be a NamedTuple instead of this weird iterator wrapper.

The only thing that would be different in calling code, is that it would have to do `for item in page.items` instead of `for item in page` but that's well worth the simplification.

src/azul/service/responseobjects/collection_data_access.py
shiroyuki: In our last call, we agree not to use `DSSClient` for this PR.

lambdas/service/app.py
```diff
@@ -1191,6 +1189,141 @@ def get_cart_item_write_progress(token):
...
+        "exportJobId": str,
+        "statusUrl": str
+    }
+    """
+    # FIXME: Check whether JWT is valid for at least an hour (configurable).
+    user_id = get_user_id()
+    access_token = app.current_request.context['authorizer']['token']
+    token = CartExportJobManager().initiate(user_id, cart_id, access_token)
+    status_url = self_url(f'/resources/carts/export/status/{token}')
+    return {'exportJobId': token, 'statusUrl': status_url}
```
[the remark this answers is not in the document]
shiroyuki: It is slightly difficult to mimic the new manifest API.

Unlike the manifest API, the cart export API requires JWTs to initiate the export job. The current implementation requires JWTs for both initiating and checking the cart export job for consistency and security (for the status check). We can remove the JWT requirement for the status check but then we will have the next issue.

Also, while the manifest service will return a pre-signed S3 URL of the manifest file, the cart export API will have problem on choosing what to do with the end product. Here are the samples.

* Option 1: Return the URL of the collection on DSS.
  * No further work required.
  * That URL will require a JWT to access. A normal HTTP request from a web browser won't work in this case.
* Option 2: Use S3 as an intermediate storage (like manifest)
  * Slightly more adjustment to the step function for the export job.
  * The data is accessible by anyone with the pre-signed URL for a limited time.

I don't have an insight in how the generated collection will be used. I need more info on what the expected behaviour of the endpoints should be.

src/azul/service/responseobjects/dynamo_data_access.py
shiroyuki: I agree with you. I'll do that.

shiroyuki: This code was from the experiment I did to find the way to query just one page of results.

lambdas/service/app.py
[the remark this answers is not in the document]
shiroyuki: This endpoint (`export_cart`) and the companion endpoint (`get_cart_export_progress`) have been replaced by new endpoints (`/resources/carts/{cart_id}/export` and `/fetch/resources/carts/{cart_id}/export`).

The new implementation of the endpoints will be consistent with the new Manifest API with a few differences:
* When the job is done, the status code will be `200` and:
  * `/resources/carts/{cart_id}/export` will respond in JSON (e.g., `{"CollectionUrl": str}`,
  * Note: `/fetch/resources/carts/{cart_id}/export` will provide the collection URL in `Location`.
* On job aborted, the status code will be `410` and:
  * `/resources/carts/{cart_id}/export` will return an empty response,
  * `/fetch/resources/carts/{cart_id}/export` will only provide `Status`.

src/azul/service/responseobjects/dynamo_data_access.py
shiroyuki: Summary of changes in related to this:
* `_make_query` is now `make_query`.
* `make_query` now returns `Page`, a named tuple, instead of a normal tuple.
* `iterable_query` has been removed in flavour of `make_query`.

src/azul/service/responseobjects/cart_export_job_manager.py
```diff
@@ -0,0 +1,101 @@
...
+                event_details = event[event_detail_field_name]
+                if 'Entered' in event_type or event_type.endswith('Scheduled'):
+                    current_state = json_loads(event_details['input'])
+                else:
+                    current_state = json_loads(event_details['output'])
+            if event_type == 'ExecutionFailed':
+                event_details = event['executionFailedEventDetails']
+                error = dict(
+                    code=event_details['error'],
+                    cause=event_details['cause']  # json_loads(event_details['cause'])
```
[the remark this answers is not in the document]
shiroyuki: Removed

shiroyuki:
@hannes-ucsc I just finished cleaning up the commit history. It should be ready now. :)

Issue #653, opened 2019-01-15
Ensure that API Gateway logs the requests for service and indexer


jessebrennan:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-412)
┆Project Name: azul
┆Issue Number: AZUL-412

jessebrennan:
I have turned logging on, but at this point it doesn't seem very useful. Also logs are not yet set to expire. Also finding the logs is not straightforward.

PR #721, opened 2019-02-21
Configure API Gateway access logging with CloudWatch (#653)

jessebrennan:
This adds a configuration variable that is set by default to be true for `dev`, `staging`, etc. but not for personal deployments.

If this variable is set, then Terraform will run a script that connects API Gateway for both index and webservice with their own respective CloudWatch log groups.

The log groups are called `/aws/apigateway/azul-<lambda_name>-<deployment_stage>` where `<lambda_name>` is either `index` or `service` and `<deployment_stage>` is the stage (i.e. `dev`, `staging`, etc). This is easy to configure so tell me your preference if you'd like something else.


deployments/.example.local/environment
```diff
@@ -13,6 +13,8 @@
 _set AZUL_DEPLOYMENT_STAGE <yourname>
 _set AZUL_DSS_ENDPOINT "https://dss.staging.data.humancellatlas.org/v1"
 _set AZUL_SUBSCRIBE_TO_DSS 0
+_set AZUL_ENABLE_MONITORING 0
+_set AZUL_LOG_API_GATEWAY 0
```
achave11: Remove this!

environment
```diff
@@ -128,6 +128,11 @@ azul_vars=(
     # Typically only set for main deployments.
     AZUL_ENABLE_MONITORING
 
+    # Enables logging of all API Gateway traffic through Cloudwatch (1 yes, 0
+    # no)
```
achave11: Looks funny, consolidate into one line.

deployments/integration/environment
```diff
@@ -16,3 +16,5 @@ _set AZUL_DSS_ENDPOINT "https://dss.integration.data.humancellatlas.org/v1"
 _set azul_grafana_endpoint "https://metrics.dev.data.humancellatlas.org"
 
 _set AZUL_ENABLE_MONITORING 1
+
+_set AZUL_LOG_API_GATEWAY 1
```
hannes-ucsc: Let's piggy-back this onto AZUL_ENABLE_MONITORING and not add yet another variable.

environment
```diff
@@ -128,6 +128,10 @@ azul_vars=(
     # Typically only set for main deployments.
     AZUL_ENABLE_MONITORING
 
+    # Enables logging of all API Gateway traffic through CloudWatch (1 yes, 0 no)
```
hannes-ucsc: But add this comment to the the one for AZUL_ENABLE_MONITORING.

scripts/log_api_gateway.py
```diff
@@ -0,0 +1,53 @@
...
+    "user": "$context.identity.user",
+    "requestTime": "$context.requestTime",
+    "httpMethod": "$context.httpMethod",
+    "resourcePath": "$context.resourcePath",
+    "status": "$context.status",
+    "protocol": "$context.protocol",
+    "responseLength": "$context.responseLength"
+}
+
+client = boto3.client('apigateway')
```
hannes-ucsc: Avoid side effects at module scope. It's fine to have a throw-away script have side effects at module scope but then it would make sense to have the if `__name__ == __main__` stuff at the bottom. That conditional implies that the script could also be imported. I think it's generally safe practice to always do `if __name__ … ` and never have side-effects at module scope.

terraform/api_gateway_domain_name.tf.json.template.py
```diff
@@ -72,7 +72,25 @@
...
-            }
-        } for lambda_name, gateway_id in gateway_ids.items() if gateway_id
+            },
+        },
+         **({
+             "aws_cloudwatch_log_group": {
+                 lambda_name: {
+                     "name": "/aws/apigateway/azul-%s-%s" % (lambda_name, config.deployment_stage),
+                     "retention_in_days": 90,
+                     "provisioner": [{
```
hannes-ucsc: Very elegant solution.

deployments/integration/environment
jessebrennan: Done.

environment
jessebrennan: :+1:

scripts/log_api_gateway.py
jessebrennan: Good point. I moved it out to avoid redundant calls, but I suppose it could just indirectly add more. Instead I'll put it in the `add_field` function again.

terraform/api_gateway_domain_name.tf.json.template.py
jessebrennan: I'm happy that it works, but I think it's hard to read / parse. I would prefer imperative statements here instead of one massive expression, but I figured that would be rejected since it goes against the style of the rest of the Terraform files. Let me know if you think it's worth changing.

jessebrennan:
@hannes-ucsc. This is ready for another review. I'd like to demo it for you though before it's merged.

terraform/api_gateway_domain_name.tf.json.template.py
hannes-ucsc: Also, `provisioner` can be single dict, too, it doesn't have to be list.

terraform/api_gateway_domain_name.tf.json.template.py
```diff
@@ -72,7 +72,23 @@
...
+                         "local-exec": {
+                             "command": (f"python {config.project_root}/scripts/log_api_gateway.py "
+                                         f"{gateway_id} "
```
hannes-ucsc: For appending or prepending a string, we prefer `+` over f-strings.

terraform/api_gateway_domain_name.tf.json.template.py
```diff
@@ -19,7 +19,7 @@
         }
     ],
     "resource": [
-        {
+        {**{
```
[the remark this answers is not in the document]
jessebrennan: Oh nice! I think this is much better!

terraform/api_gateway_domain_name.tf.json.template.py
```diff
@@ -72,7 +72,23 @@
...
                     }
                 }
-            }
-        } for lambda_name, gateway_id in gateway_ids.items() if gateway_id
+            },
+        },
+         **({
+             "aws_cloudwatch_log_group": {
+                 lambda_name: {
+                     "name": "/aws/apigateway/azul-%s-%s" % (lambda_name, config.deployment_stage),
```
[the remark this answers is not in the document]
jessebrennan: > I would rather inline the contents of `log_api_gateway.py` but I don't think that's possible.
 It could be inlined with a couple calls to the awscli. I didn't realize (until just now) that it's a dependency so I thought it better to use Boto since I know it's installed. If you'd rather me refactor with the cli, let me know.

> The only other issue I have is spaces and other shell meta characters in the interpolated variables.

Using the awscli won't work around this, but I suppose I could encase the interpolated arguments in single quotes. Practically speaking though I don't think it would be an issue.

terraform/api_gateway_domain_name.tf.json.template.py
jessebrennan: :+1:

terraform/api_gateway.tf.json.template.py
```diff
@@ -56,7 +58,7 @@
             "aws_route53_record": {
                 lambda_name + "_domain_validation": {
                     "name": "${aws_acm_certificate.%s.domain_validation_options.0.resource_record_name}" % lambda_name,
-                    "type": "${aws_acm_certificate.%s.domain_validation_options.0.resource_record_type}" % lambda_name,
+                    "type": "api_gateway_domain_name.tf.json.template${aws_acm_certificate.%s.domain_validation_options.0.resource_record_type}" % lambda_name,
```
hannes-ucsc: This broke deployment.

PR #722, opened 2019-02-22
Various deployment and build fixes

hannes-ucsc:


Makefile
```diff
@@ -1,9 +1,9 @@
-include common.mk
+all: hello
 
-all:
+include $(azul_home)/common.mk
```
hannes-ucsc: Chicken-egg. common.mk actually tests for presence of azul_home. Must use relative path.

Issue #680, opened 2019-01-23
Service gives 500 errors with ~11% failure rate at ~10 requests per second under load


jessebrennan:
This is uncovered while using Locust, but capturing the individual errors and the content of the response is more difficult. I will try to update this ticket as I have more info.

#### Manifest error
An error returned while trying to generate a manifest at [this url](https://service.integration.explore.data.humancellatlas.org/manifest/files?token=eyJleGVjdXRpb25faWQiOiAiNmEwMThlOTctZWJhYS00ZTc4LWE0OWItYjIwYThmOThjYmQ4IiwgIndhaXQiOiA0fQ==):
  ```
  {"Code": "ChaliceViewError", "Message": "ChaliceViewError: Failed to generate manifest"}
  ```

#### Specimens error
request info:
```
Request URL: https://service.integration.explore.data.humancellatlas.org/repository/specimens?filters=%7B%7D&size=15&sort=specimenId&order=desc
Request Method: GET
Status Code: 500 
Remote Address: 99.84.194.41:443
Referrer Policy: no-referrer-when-downgrade
access-control-allow-headers: Authorization,Content-Type,X-Amz-Date,X-Amz-Security-Token,X-Api-Key
access-control-allow-origin: *
content-length: 1352
content-type: text/plain
date: Wed, 23 Jan 2019 18:09:49 GMT
status: 500
via: 1.1 8fb273b390f064c411944b3126eb0eb6.cloudfront.net (CloudFront)
x-amz-apigw-id: T-Ac2GtbIAMFt3g=
x-amz-cf-id: cKIQQq1gDenbVafnsQyucvylPhsWBf6uiC4OYwpm7Ziqtx-ylINNOg==
x-amzn-requestid: 11cccfcd-1f3a-11e9-94b1-cb9a07097fa4
x-amzn-trace-id: Root=1-5c48adeb-1129b80e050df564964639da;Sampled=0
x-cache: Error from cloudfront
:authority: service.integration.explore.data.humancellatlas.org
:method: GET
:path: /repository/specimens?filters=%7B%7D&size=15&sort=specimenId&order=desc
:scheme: https
accept: application/json, text/plain, */*
accept-encoding: gzip, deflate, br
accept-language: en-US,en;q=0.9
origin: https://integration.data.humancellatlas.org
referer: https://integration.data.humancellatlas.org/explore/specimens
[...]

PR #730, opened 2019-02-27
Use fewer shards for aggregate index (#680)

jessebrennan:
Using fewer shards speeds up searching since nodes are no longer over-
burdened with so many requests to check each shard just for a single
search. Using only 4 shards on dev should keep search decently fast and
had minimal effect on indexing speed. It also seems to have prevented
the 500 errors that were caused from overwhelming Elasticsearch.

### DIFF
--- a/src/azul/indexer.py
+++ b/src/azul/indexer.py
@@ -34,12 +34,17 @@ class BaseIndexer(ABC):
     def mapping(self) -> JSON:
         raise NotImplementedError()
 
-    def settings(self) -> JSON:
+    def settings(self, index_name) -> JSON:
+        # Setting a large number of shards for the contributions indexes (i.e. not aggregate) greatly speeds up indexing
+        # which is our biggest bottleneck, however doing the same for the aggregate index dramatically limits searching.
+        # This was because every search had to check every shard and nodes became overburdened with so many requests.
+        # Instead we try using one shard per ES node which is optimal for searching since it allows parallelization of
+        # requests (though maybe at the cost of higher contention during indexing).
+        _, aggregate = config.parse_es_index_name(index_name)
+        num_shards = config.es_instance_count if aggregate else config.indexer_concurrency
         return {
             "index": {
-                # This is important. It may slow down searches but it does increase concurrency during indexing,
-                # currently our biggest performance bottleneck.
-                "number_of_shards": config.indexer_concurrency,
+                "number_of_shards": num_shards,
                 "number_of_replicas": 1,
                 "refresh_interval": f"{config.es_refresh_interval}s"
             }
@@ -86,7 +91,7 @@ class BaseIndexer(ABC):
         for index_name in self.index_names():
             es_client.indices.create(index=index_name,
                                      ignore=[400],
-                                     body=dict(settings=self.settings(),
+                                     body=dict(settings=self.settings(index_name),
                                                mappings=dict(doc=self.mapping())))
         contributions = []
         for transformer in self.transformers():
--- a/test/service/test_request_validation.py
+++ b/test/service/test_request_validation.py
@@ -3,7 +3,6 @@ import json
 import logging
 import os
 import sys
-import re
 
 from tempfile import TemporaryDirectory
 from unittest import mock
@@ -13,6 +12,7 @@ import requests
 
 import azul.changelog
 from azul import config
+from azul.json_freeze import freeze
 from azul.service import service_config
 from azul.service.responseobjects.storage_service import StorageService
 from retorts import ResponsesHelper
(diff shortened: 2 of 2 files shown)


src/azul/indexer.py
```diff
@@ -34,12 +34,16 @@ class BaseIndexer(ABC):
     def mapping(self) -> JSON:
         raise NotImplementedError()
 
-    def settings(self) -> JSON:
+    def settings(self, index_name) -> JSON:
+        # Setting a large number of shards for the contributions indexes (i.e. not aggregate) greatly speeds up indexing
+        # which is our biggest bottleneck, however doing the same for the aggregate index dramatically limits searching.
+        # This was because every search had to check every shard and nodes became overburdened with so many requests.
+        # Instead we try using one shard per instance which is optimal for searching since it allows parallelization of
```
hannes-ucsc:
```suggestion
        # Instead we try using one shard per ES node which is optimal for searching since it allows parallelization of
```

src/azul/indexer.py
```diff
@@ -34,12 +34,16 @@ class BaseIndexer(ABC):
...
+        # requests (though maybe at the cost of some speed during indexing).
```
hannes-ucsc:
```suggestion
        # requests (though maybe at the cost of higher contention during indexing).
```

src/azul/indexer.py
```diff
@@ -34,12 +34,16 @@ class BaseIndexer(ABC):
...
+        num_shards = config.es_instance_count if '_aggregate' in index_name else config.indexer_concurrency
```
hannes-ucsc: Use `config.parse_es_index_name`. You always want to keep the implementation of composing and resource decomposing names in one place. In the long run, that makes changing the naming scheme much easier. Also, `'_aggregate' in` is a little weak. The aformentioned method is stricter.

test/service/test_request_validation.py
```diff
@@ -227,9 +227,9 @@ def test_bdbag_manifest(self):
...
-                self.assertEqual(row, expected_row, 'Row in sample.tsv incorrect.')
-                remaining_rows = list(samples)
-                self.assertTrue(all(len(row) == len(expected_row) for row in remaining_rows),
+                samples = list(samples)
+                self.assertEqual(len(samples), 2, 'Found incorrect number of files.')  # self.bundle has 2 files
+                self.assertTrue(all(len(sample) == len(expected_row) for sample in samples),
                                 'Incorrect number of entries in a row.')
-                self.assertEqual(len(remaining_rows), 1, 'Found incorrect number of files.')  # self.bundle has 2 files
+                self.assertIn(expected_row, samples, 'Expected file was not found or had incorrect fields')
+                remaining_rows = list(samples)
```
hannes-ucsc: Unused assignement?

test/service/test_request_validation.py
```diff
@@ -227,9 +227,9 @@ def test_bdbag_manifest(self):
...
-                row = next(samples)
-                self.assertEqual(row, expected_row, 'Row in sample.tsv incorrect.')
-                remaining_rows = list(samples)
-                self.assertTrue(all(len(row) == len(expected_row) for row in remaining_rows),
+                samples = list(samples)
+                self.assertEqual(len(samples), 2, 'Found incorrect number of files.')  # self.bundle has 2 files
+                self.assertTrue(all(len(sample) == len(expected_row) for sample in samples),
                                 'Incorrect number of entries in a row.')
-                self.assertEqual(len(remaining_rows), 1, 'Found incorrect number of files.')  # self.bundle has 2 files
+                self.assertIn(expected_row, samples, 'Expected file was not found or had incorrect fields')
```
hannes-ucsc: Can you think of an easy way to establish that `expected_row` only occurs once in samples.

test/service/test_request_validation.py
```diff
@@ -227,9 +227,9 @@ def test_bdbag_manifest(self):
                 samples = iter(csv.reader(fh, delimiter='\t'))
                 row = next(samples)
                 self.assertEqual(expected_fieldnames, row, 'Manifest headers are not configured correctly')
-                row = next(samples)
-                self.assertEqual(row, expected_row, 'Row in sample.tsv incorrect.')
-                remaining_rows = list(samples)
-                self.assertTrue(all(len(row) == len(expected_row) for row in remaining_rows),
+                samples = list(samples)
```
hannes-ucsc: I like `rows` better.

test/service/test_request_validation.py
```diff
@@ -227,9 +227,9 @@ def test_bdbag_manifest(self):
                 samples = iter(csv.reader(fh, delimiter='\t'))
```
hannes-ucsc: And `rows` instead of `samples` for this, too. They aren't actually samples. The rows represent files.

test/service/test_request_validation.py
jessebrennan: wooops!

src/azul/indexer.py
jessebrennan: Changed.

src/azul/indexer.py
jessebrennan: Done

test/service/test_request_validation.py
jessebrennan: Me too actually...

test/service/test_request_validation.py
jessebrennan: Good point.

test/service/test_request_validation.py
jessebrennan: Not without a loop or something. Let me know if you want me to make it like that.

src/azul/indexer.py
jessebrennan: Yes it is. I've changed to that.

test/service/test_request_validation.py
```diff
@@ -224,12 +224,11 @@ def test_bdbag_manifest(self):
...
-                row = next(samples)
-                self.assertEqual(row, expected_row, 'Row in sample.tsv incorrect.')
-                remaining_rows = list(samples)
-                self.assertTrue(all(len(row) == len(expected_row) for row in remaining_rows),
+                rows = list(rows)
+                self.assertEqual(len(rows), 2, 'Found incorrect number of files.')  # self.bundle has 2 files
+                self.assertTrue(all(len(row) == len(expected_row) for row in rows),
                                 'Incorrect number of entries in a row.')
-                self.assertEqual(len(remaining_rows), 1, 'Found incorrect number of files.')  # self.bundle has 2 files
+                self.assertIn(expected_row, rows, 'Expected file was not found or had incorrect fields')
```
hannes-ucsc: We could freeze `rows`, convert it to a set, assert that the set is of size 2 and assert that `freeze(expected_row)` in that set. Here's how freeze() works:

```python
from azul.json_freeze import freeze
freeze(["a", 1, True]) in set(freeze([["a", 1, True],["b", 2, False]]))
True
freeze(["a", 2, True]) in set(freeze([["a", 1, True],["b", 2, False]]))
False
freeze([1, "a", True]) in set(freeze([["a", 1, True],["b", 2, False]]))
False
```

Issue #699, opened 2019-02-02
Change organ_part to an array


hannes-ucsc:
https://github.com/HumanCellAtlas/metadata-schema/pull/767

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-434)
┆Project Name: azul
┆Issue Number: AZUL-434

theathorn:
metadata-api change required.

hannes-ucsc:
Demoed on 3/1/2019

PR #733, opened 2019-02-28
Change organ_part to an array (#699)

danielsotirhos:


test/indexer/test_hca_indexer.py
```diff
@@ -170,7 +170,7 @@ def _assert_old_bundle(self, num_expected_new_contributions=0, ignore_aggregates
             else:
                 self.assertIn('Farmers Trucks', [c.get('institution') for c in project['contributors']])
             specimen = one(contents['specimens'])
-            self.assertIn('Australopithecus', specimen['genus_species'])
```
hannes-ucsc: This actually renders the test useless. Fix the input metadata so you don't have to do modify the assertion.

Issue #734, opened 2019-02-28
Deletion doesn't work for bundles with more than 10 entities


hannes-ucsc:
… because only the first page of contributions is processed in https://github.com/DataBiosphere/azul/blob/6d7f807bae088cc4e44cb91882b18a255e04e764/src/azul/indexer.py#L277

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-453)
┆Project Name: azul
┆Issue Number: AZUL-453

hannes-ucsc:
Demoed on 3/1/2019

PR #735, opened 2019-02-28
Fix deletion of bundles with > 10 entities (#734)

hannes-ucsc:


hannes-ucsc:
A single failure, test_deletion: https://travis-ci.org/DataBiosphere/azul/builds/499637807#L5933

This proves that the modified deletion test reproduces the issue. The next commit will fix it.

Issue #740, opened 2019-02-28
Copy `manifest` in request_config.json and name it `bdbag`


mjkrause:
Create a new key `bdbag` so that existing key `manifest` can exclusively be used to generate the manifest TSV. This is necessary as Terra has specific requirement for key and value of the first two columns in file `sample.tsv` and `participant.tsv`, which is different from the manifest TSV.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-458)
┆Project Name: azul
┆Issue Number: AZUL-458

PR #741, opened 2019-02-28
Copy `manifest` as `bdbag` in request_config.json (#740)

mjkrause:
* copied `manifest` and named is `bdbag` in request_config.json

* add conditional in elastic_request_builder that decides which key
in request_config.json is read out depending on the format query
parameter

### DIFF
--- a/src/azul/service/responseobjects/elastic_request_builder.py
+++ b/src/azul/service/responseobjects/elastic_request_builder.py
@@ -538,7 +538,10 @@ class ElasticTransformDump(object):
         if not filters:
             filters = {"file": {}}
         filters = filters['file']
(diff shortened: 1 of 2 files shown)


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -538,7 +538,10 @@ def transform_manifest(self, format, filters=None, request_config_file='request_
         if not filters:
             filters = {"file": {}}
         filters = filters['file']
-        manifest_config = request_config['manifest']
+        if format == 'bdbag':
+            manifest_config = request_config['bdbag']
```
```suggestion
            manifest_config = request_config['bdbag' if format == 'bdbag' else 'manifest']
```

would have been more concise

Issue #713, opened 2019-02-16
File Size does not sort correctly


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-443)
┆Project Name: azul
┆Issue Number: AZUL-443

PR #742, opened 2019-02-28
Fixed sorting by fileSize in webservice (#713)

GPelayo:


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -253,18 +254,24 @@ def open_and_return_json(file_path):
...
         It has: 'size', 'sort', 'order', and one of 'search_after', 'search_before', or 'from'.
+        :param entity_type: the string referring to the entity type used to get
+        the ElasticSearch mappings
         :return: An ES Search object where pagination has been applied
         """
-        # Extract the fields for readability (and slight manipulation)
+        # Find the data type of the sort parameter
+        index_client = elasticsearch.client.IndicesClient(self.es_client)
+        index_name = config.es_index_name(entity_type)
+        field_map = index_client.get_field_mapping(fields=[pagination['sort']], index=index_name)
```
hannes-ucsc: It would be sad if we had to do an additional ES round-trip for every page. Let's talk about alternatives tomorrow.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -59,9 +60,9 @@ class PaginationObj(JsonObject):
     count = IntegerProperty()
     total = IntegerProperty()
     size = IntegerProperty()
-    search_after = StringProperty()
```
hannes-ucsc: Can you motivate this change for me?

GPelayo: The `search_after` and `search_before` parameters are always the same data type as the field that is chosen to be the `sort` parameter. Since it's possible for the that field to be an integer or a string, the search parameters must be set as a `DefaultProperty`, the dynamic field/property type of a `JSONObject`.

hannes-ucsc: Makes sense. Thank you.

Issue #743, opened 2019-03-01
Requesting manifest in `bdbag` format returns invalid zip file


hannes-ucsc:
Discovered when demo'ing to Terra folks. The contents is actually the path of a temporary file.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-460)
┆Project Name: azul
┆Issue Number: AZUL-460

PR #756, opened 2019-03-05
Fix bdbag upload to S3 (#743)

mjkrause:
* as BDBag format requires upload of a zipped file it cannot use the
same upload function as a TSV file
* depending on the manifest format, def _push_content_single_part will
call a specific function to upload to S3
* added def `upload` to StorageService to upload the zipped BDBag using
the upload_file function
* changed how zipfile name is obtained

### DIFF
--- a/src/azul/service/responseobjects/storage_service.py
+++ b/src/azul/service/responseobjects/storage_service.py
@@ -38,12 +38,9 @@ class StorageService:
         return object_key
 
     def upload(self, file_name, object_key):
-        aws_profile = os.getenv('AWS_PROFILE')
-        session = boto3.Session(profile_name=aws_profile)
-        s3 = session.resource('s3')
-        s3.meta.client.upload_file(Filename=file_name,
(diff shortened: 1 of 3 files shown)


src/azul/service/responseobjects/storage_service.py
```diff
@@ -36,6 +37,15 @@ def put(self, object_key: str, data: bytes, content_type: Optional[str] = None,
 
         return object_key
 
+    def upload(self, file_name, object_key):
+        aws_profile = os.getenv('AWS_PROFILE')
+        session = boto3.Session(profile_name=aws_profile)
+        s3 = session.resource('s3')
+        s3.meta.client.upload_file(Filename=file_name,
```
hannes-ucsc: Why arent't you using self.client()? Even if `self.client()` wasn't available, this would still be a needlessly complicated way to get to a boto3 client.

mjkrause: fixed

Issue #752, opened 2019-03-05
Integration test occasionally fails against manifest endpoint


hannes-ucsc:
```
======================================================================
ERROR: test_webservice_and_indexer (local_integration_test.IntegrationTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builds/azul/azul/test/local_integration_test.py", line 75, in test_webservice_and_indexer
    self.check_endpoint_is_working(config.service_endpoint(), f'/manifest/files?filters={manifest_filter}')
  File "/builds/azul/azul/test/local_integration_test.py", line 88, in check_endpoint_is_working
    response = requests.get(url)
  File "/build/.venv/lib/python3.6/site-packages/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/build/.venv/lib/python3.6/site-packages/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/build/.venv/lib/python3.6/site-packages/requests/sessions.py", line 524, in request
    resp = self.send(prep, **send_kwargs)
  File "/build/.venv/lib/python3.6/site-packages/requests/sessions.py", line 659, in send
    history = [resp for resp in gen] if allow_redirects else []
  File "/build/.venv/lib/python3.6/site-packages/requests/sessions.py", line 659, in <listcomp>
    history = [resp for resp in gen] if allow_redirects else []
  File "/build/.venv/lib/python3.6/site-packages/requests/sessions.py", line 156, in resolve_redirects
    raise TooManyRedirects('Exceeded %s redirects.' % self.max_redirects, response=resp)
requests.exceptions.TooManyRedirects: Exceeded 30 redirects.

----------------------------------------------------------------------
```

Probably doesn't obey Retry-After when requesting the manifest endpoint.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-465)
┆Project Name: azul
┆Issue Number: AZUL-465

PR #767, opened 2019-03-06
 Integration test honors Retry-After header (#752) 

hannes-ucsc:


src/azul/service/manifest.py
```diff
@@ -40,7 +40,7 @@ def start_or_inspect_manifest_generation(self,
                                              token: Optional[str] = None,
                                              format: Optional[str] = None,
                                              filters: Optional[str] = None
-                                             ) -> Tuple[float, str]:
+                                             ) -> Tuple[int, str]:
```
jessebrennan: You had asked me to make this a float. I'm curious why you're changing it now.

Issue #749, opened 2019-03-04
Include DOS URL in sample BdBag


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-463)
┆Project Name: azul
┆Issue Number: AZUL-463

PR #768, opened 2019-03-06
 Include DOS URL in sample BdBag (#749)

mjkrause:
* added `dos_url` header in def _construct_bdbag
* construct URL of `dos_url` and added it added it to sample.writer
in _iterate_hit_tsv
* added `dos_url` as a column header and row entry to
test_bdbag_manifest


mjkrause:
Rebased on develop and squashed to one commit.

Issue #774, opened 2019-03-08
DOS URL for BDBag export starts with dos://...


mjkrause:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-479)
┆Project Name: azul
┆Issue Number: AZUL-479

theathorn:
I thought we were supposed to use "drs" not "dos" from now on...

hannes-ucsc:
`dos://` is required by DRS v0.1.

hannes-ucsc:
Either way, we should use whatever protocol works. If both work we'll go with `drs://`. Precedence is `dos://` and we should try that first to minimize risk.

hannes-ucsc:
This was demo'ed by @mjkrause.

PR #779, opened 2019-03-11
DRS URLs in BDBag must begin with `dos://` (#774)

hannes-ucsc:
Supersedes https://github.com/databiosphere/azul/issues/777


test/service/test_drs.py
```diff
@@ -43,7 +42,7 @@ def test_data_object_not_found(self):
         :return:
         """
         file_id = "NOT_A_GOOD_IDEA"
-        get_url = "{}/ga4gh/dos/v1/dataobjects/{}".format(self.base_url, file_id)
+        get_url = self.base_url + dos_object_url(file_id)
```
hannes-ucsc: This is still wrong.

PR #812, opened 2019-03-20
Verify authenticity of incoming notifications

achave11:


scripts/delete_bundle.py
```diff
@@ -27,19 +26,8 @@ def main(argv):
...
-        'match': {
-            'bundle_uuid': bundle_uuid,
-            'bundle_version': bundle_version
-        }
-    }
-    response = requests.post(url=deletion_endpoint,
-                             json=notification)
-    response.raise_for_status()
-
+    azul_client = AzulClient()
```
hannes-ucsc: The AzulClient instance should be shared across `delete_bundle` invocations. I think this can be achieved most easily by inlining this function (since it is now pretty anemic) and instantiating AzulCient outside of the bundle loop.

src/azul/__init__.py
```diff
@@ -418,6 +418,18 @@ def access_token_audience_list(self):
             f"{self.access_token_issuer}/userinfo"
         ]
 
+    @property
+    def hmac_key(self):
```
hannes-ucsc: Properties that return a constant value don't need to be properties. A class attribute as in 

```
hmac_key = "…"
```

suffices.

src/azul/__init__.py
```diff
@@ -418,6 +418,18 @@ def access_token_audience_list(self):
...
+        # TODO: Generate subscription HMAC key dynamically and store it securely
+        #  https://github.com/DataBiosphere/azul/issues/794
+        return 'São Paulo number Juan among all the numbers'
+
+    @property
+    def hmac_key_id(self):
```
hannes-ucsc: Ditto.

lambdas/indexer/app.py
```diff
@@ -77,6 +78,7 @@ def post_notification():
     """
     Receive a notification event and either queue it for asynchronous indexing or process it synchronously.
     """
+    validate(current_request=app.current_request)
```
hannes-ucsc: Change the import such that this reads

```suggestion
    hmac.validate(current_request=app.current_request)
```

src/azul/hmac.py
```diff
@@ -0,0 +1,28 @@
+import chalice
+import requests
+from requests_http_signature import HTTPSignatureAuth
+
+from azul import config
+
+
+def validate(current_request):
```
hannes-ucsc:
```suggestion
def verify(current_request):
```

The term *verify* seems to be more commonly used in conjunction with HMAC. For example: https://en.wikipedia.org/wiki/HMAC

src/azul/azulclient.py
```diff
@@ -233,3 +235,16 @@ def delete_all_indices(self):
...
+
+    def delete_bundle(self, bundle_uuid, bundle_version):
+        logger.info('Deleting bundle %s.%s', bundle_uuid, bundle_version)
+        notification = {
+            'match': {
+                'bundle_uuid': bundle_uuid,
+                'bundle_version': bundle_version
+            }
+        }
+        response = requests.post(url=self.indexer_url+'delete',
```
hannes-ucsc: Spaces around operators, please.

hannes-ucsc: Also, could you change the definition of `self.indexer_url` so it doesn't end in a slash? You will have to change all references to it. This should be done in a separate commit, just after the renaming commit. In the end this should read

```suggestion
        response = requests.post(url=self.indexer_url + '/delete',
```

test/indexer/test_hca_indexer.py
```diff
@@ -578,7 +581,7 @@ def lambda_name(cls) -> str:
 
     @mock_sts
     @mock_sqs
-    def test_post_notification_endpoint(self):
+    def test_healthy_post_notification(self):
```
hannes-ucsc: Odd naming choice.

test/indexer/test_hca_indexer.py
```diff
@@ -578,7 +581,7 @@ def lambda_name(cls) -> str:
 
     @mock_sts
     @mock_sqs
-    def test_post_notification_endpoint(self):
+    def test_succesfull_notifications(self):
```
```suggestion
    def test_succesful_notifications(self):
```

src/azul/azulclient.py
```diff
@@ -56,11 +58,9 @@ def post_bundle(self, bundle_fqid, indexer_url):
...
         """
         simulated_event = self._make_notification(bundle_fqid)
-        body = json.dumps(simulated_event).encode('utf-8')
-        request = Request(indexer_url, body)
-        request.add_header("content-type", "application/json")
-        with urlopen(request) as f:
-            return f.read()
+        response = requests.post(indexer_url, json=simulated_event, auth=hmac.prepare())
+        response.raise_for_status()
+        return response
```
hannes-ucsc: Shouldn't this be 

```suggestion
        return response.content
```

to be equivalent to the previous implementation?

PR #826, opened 2019-03-22
Shared ES instance for main deployments

achave11:


requirements.txt
```diff
@@ -13,6 +13,7 @@ jsonobject==0.7.1
 Faker==0.8.11
 future==0.16.0
 hca==4.4.1
+dcplib==1.5.1  # Recent upgrade to 1.6.0 broke the Travis build
```
hannes-ucsc: This should really be in a separate commit titled:

```
Pin dcplib to 1.5.1 as 1.6.0 breaks the build
```

hannes-ucsc:
https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/pipelines/134

Issue #800, opened 2019-03-18
Add cell_suspension.selected_cell_type to the Azul index


theathorn:
See blocked data-browser ticket for details.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-490)
┆Project Name: azul
┆Issue Number: AZUL-490
┆Epic: Improve or Replace Specimens Tab

hannes-ucsc:
https://github.com/HumanCellAtlas/metadata-api/releases/tag/release%2F1.0b12

danielsotirhos:
I demoed this on 04/02/2019

PR #831, opened 2019-03-25
Add cell_suspension.selected_cell_type (#800)

danielsotirhos:


test/service/test_response.py
```diff
@@ -63,6 +63,7 @@ def test_key_search_files_response(self):
                         {
                             "organ": ["pancreas"],
                             "organPart": ["islet of Langerhans"],
+                            "selectedCellType": [],
```
hannes-ucsc: Since `[]` is the default value, this is insufficient proof that the `selectedCellType` property is actually propagated all the way from the contribution, through aggregate to the service response.

hannes-ucsc:
Removing my assignment since builds are currently ongoing.

test/service/test_response.py
```diff
@@ -1149,6 +1156,84 @@ def test_project_accessions_response(self):
...
         self.assertElasticsearchResultsEqual(keyword_response, expected_response)
 
+    def test_cell_suspension_response(self):
+        """
+        This method tests the KeywordSearchResponse object for the projects entity type,
+        specifically making sure the selectedCellType field is present with the correct value
+        """
+        keyword_response = KeywordSearchResponse(
+            hits=self.get_hits('projects', '250aef61-a15b-4d97-b8b4-54bb997c1d7d'),
+            entity_type='cell_suspensions'
```
```suggestion
            entity_type='projects'
```

test/service/test_response.py
```diff
@@ -1149,6 +1156,84 @@ def test_project_accessions_response(self):
...
+                        }
+                    ],
+                    "cellSuspensions": [
+                        {
+                            "organ": ["hemopoietic organ"],
+                            "organPart": [
+                                "bone marrow",
+                                "temporal lobe"
+                            ],
+                            "selectedCellType": ["Plasma cells"],
```
hannes-ucsc: Please assert this more surgically. The rest of the literal is inessential for this particular test case. We don't need to assert the entire response literally if the only thing we're looking for is the presence and value of a single property in the response.

test/service/test_response.py
```diff
@@ -23,6 +23,7 @@ class TestResponse(WebServiceTestCase):
 
     maxDiff = None
     bundles = WebServiceTestCase.bundles + [('fa5be5eb-2d64-49f5-8ed8-bd627ac9bc7a', '2019-02-14T192438.034764Z')]
+    bundles = bundles + [('d0e17014-9a58-4763-9e66-59894efbdaa8', '2018-10-03T144137.044509Z')]
```
hannes-ucsc: Just add the tuple to the list literal in the line above. With proper wrapping, of course.

And you know this: `x = x + y` is the same as `x += y`.

PR #846, opened 2019-03-27
Consolidate boto3 client and resource creation

hannes-ucsc:


theathorn:
@hannes-ucsc says "close it!".

Issue #858, opened 2019-03-27
Add timestamp to logging format for Azul tests


jessebrennan:
When running tests on Gitlab no timestamps are shown. Adding this to the logging format will make debugging a lot easier.

This ticket was originally opened in the wrong repo: humanCellatlas/dcp-cli#273

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-533)
┆Project Name: azul
┆Issue Number: AZUL-533

PR #890, opened 2019-04-09
Add timestamp to integration test logging (#858)

jessebrennan:


jessebrennan:
I ran the test locally and indeed there were log messages.

Issue #867, opened 2019-04-03
Add facet for sequencing_protocol.paired_end


NoopDog:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-540)
┆Project Name: azul
┆Issue Number: AZUL-540

hannes-ucsc:
@danielsotirhos, consider creating a blocking metadata-api ticket if needed.

PR #899, opened 2019-04-12
Add facet for sequencing_protocol.paired_end (#867)

danielsotirhos:


src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -646,7 +646,8 @@ def make_protocols(self, entry):
         for protocol in entry["contents"]["protocols"]:
             translated_process = {
                 "libraryConstructionApproach": protocol.get("library_construction_approach", []),
-                "instrumentManufacturerModel": protocol.get("instrument_manufacturer_model", [])
+                "instrumentManufacturerModel": protocol.get("instrument_manufacturer_model", []),
+                "pairedEnd": protocol.get("paired_end", None),
```
hannes-ucsc: Why is the default different from the above two properties?

danielsotirhos: Ah, I think i was looking at make_specimen() as a guide which uses None as the default for all the properties.

Issue #794, opened 2019-03-15
Generate subscription HMAC key dynamically and store it securely


achave11:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-496)
┆Project Name: azul
┆Issue Number: AZUL-496

theathorn:
Relates to #96.

hannes-ucsc:
Demo-ed on 5/13/2019

PR #900, opened 2019-04-12
Securely store HMAC key and google creds (#794)

jessebrennan:
This creates a new provisioner for the Google service account that takes care of creating (and destroying) the Account credentials, generating an HMAC key, and storing them in AWS Secrets Manager secrets.

Destroying resources has some limitations. From [Terraform docs](https://www.terraform.io/docs/provisioners/index.html#destroy-time-provisioners)

> Destroy-time provisioners can only run if they remain in the configuration at the time a resource is destroyed.

This is not the case with our code. This means secrets, etc. won't be destroyed if `AZUL_SUBSCRIBE_TO_DSS` **was** set to true and then is set to false.

### DIFF
--- a/scripts/provision_credentials.py
+++ b/scripts/provision_credentials.py
@@ -1,0 +1,169 @@
+import argparse
+import base64
+import json
+import logging
+import os
+import uuid
+from functools import partial
+
+import boto3
+import googleapiclient.discovery
+from botocore.exceptions import ClientError
+from google.oauth2 import service_account
+from googleapiclient.errors import HttpError
+
+from azul import config
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def parse_google_key(response):
+    return base64.decodebytes(bytes(response['privateKeyData'], 'ascii')).decode()
+
+
+def get_google_service():
+    credentials = service_account.Credentials.from_service_account_file(
+        filename=os.environ['GOOGLE_APPLICATION_CREDENTIALS'],
+        scopes=['https://www.googleapis.com/auth/cloud-platform'])
+    return googleapiclient.discovery.build('iam', 'v1', credentials=credentials)
+
+
+class CredentialsProvisioner:
+
+    def __init__(self):
+        self.secrets_manager = boto3.client('secretsmanager')
+
+    def provision_google_from_args(self, args):
+        self.provision_google(args.build, args.email)
+
+    def provision_hmac_from_args(self, args):
+        self.provision_hmac(args.build)
+
+    def provision_hmac(self, build):
+        secret_name = config.secrets_manager_secret_name('indexer', 'hmac')
+        if build:
+            self._create_secret(secret_name)
+            if not self._secret_is_stored(secret_name):
+                self._write_secret_value(secret_name, self._random_hmac_key())
+        else:
+            self._destroy_aws_secrets_manager_secret(secret_name)
+
+    def provision_google(self, build, email):
+        secret_name = config.secrets_manager_secret_name('indexer', 'google_service_account')
+        if build:
+            self._create_secret(secret_name)
+            if not self._secret_is_stored(secret_name):
+                google_key = self._create_service_account_creds(email)
+                self._write_secret_value(secret_name, google_key)
+        else:
+            self._destroy_service_account_creds(email)
+            self._destroy_aws_secrets_manager_secret(secret_name)
+
+    @classmethod
+    def _random_hmac_key(cls):
+        # Even though an HMAC key can be any sequence of bytes, we restrict to base64 in order to encode as string
+        key = base64.encodebytes(os.urandom(48)).decode().replace('=', '').replace('\n', '')
+        assert len(key) == 64
+        return json.dumps({'key': key, 'key_id': str(uuid.uuid4())})
+
+    def _write_secret_value(self, name, value):
+        self.secrets_manager.put_secret_value(
+            SecretId=name,
+            SecretString=value
+        )
+        logger.info("Successfully wrote value to AWS secret '%s'.", name)
+
+    def _create_secret(self, name):
+        try:
+            self.secrets_manager.create_secret(Name=name)
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'ResourceExistsException':
+                logger.info('AWS secret %s already exists.', name)
+            else:
+                raise
+        else:
+            logger.info('AWS secret %s created.', name)
+
+    def _secret_is_stored(self, name):
+        try:
+            response = self.secrets_manager.get_secret_value(SecretId=name)
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'ResourceNotFoundException':
+                return False
+            else:
+                raise
+        try:
+            return response['SecretString'] != ''
+        except KeyError:
+            return False
+
+    def _create_service_account_creds(self, service_account_email):
+        service = get_google_service()
+        key = service.projects().serviceAccounts().keys().create(
+            name='projects/-/serviceAccounts/' + service_account_email, body={}
+        ).execute()
+        logger.info("Successfully created service account key for user '%s'", service_account_email)
+        return parse_google_key(key)
+
+    def _destroy_aws_secrets_manager_secret(self, secret_name):
+        try:
+            response = self.secrets_manager.delete_secret(
+                SecretId=secret_name,
+                ForceDeleteWithoutRecovery=True
+            )
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'ResourceNotFoundException':
+                logger.info('AWS secret %s does not exist. No changes will be made.', secret_name)
+            else:
+                raise
+        else:
+            assert response['Name'] == secret_name
+            logger.info('Successfully deleted AWS secret %s.', secret_name)
+
+    def _destroy_service_account_creds(self, service_account_email):
+        try:
+            creds = self.secrets_manager.get_secret_value(
+                SecretId=config.secrets_manager_secret_name('indexer', 'google_service_account')
+            )
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'ResourceNotFoundException':
+                logger.info('Secret already deleted, cannot get key_id for %s', service_account_email)
+                return
+            else:
+                raise
+        else:
+            key_id = json.loads(creds['SecretString'])['private_key_id']
+            service = get_google_service()
+            try:
+                service.projects().serviceAccounts().keys().delete(
+                    name='projects/-/serviceAccounts/' + service_account_email + '/keys/' + key_id).execute()
+            except HttpError as e:
+                if e.resp.reason != 'Not Found':
+                    raise
+            logger.info("Successfully deleted service account key with id '%s' for user '%s'",
+                        key_id, service_account_email)
+
+
+if __name__ == "__main__":
+    provision_parser = argparse.ArgumentParser(add_help=False)
+    group = provision_parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('--build', '-b', action='store_true', dest='build',
+                       help='Create credentials instead of destroying them. This action is idempotent.')
+    group.add_argument('--destroy', '-d', action='store_false', dest='build',
+                       help='Destroy credentials instead of building them. This action is idempotent.')
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(help='Specify action', dest='action')
+    subparsers.required = True
+    google_parser = subparsers.add_parser('google-key', parents=[provision_parser],
+                                          help='Create Google service account key and store in an AWS secret.')
+    google_parser.set_defaults(func=CredentialsProvisioner.provision_google_from_args)
+    google_parser.add_argument('email', type=str,
+                               help='Email address for the Google service account '
+                                    'for which to provision credentials')
+    hmac_parser = subparsers.add_parser('hmac-key', parents=[provision_parser],
+                                        help='Create a random HMAC key and store in an AWS secret.')
+    hmac_parser.set_defaults(func=CredentialsProvisioner.provision_hmac_from_args)
+    args = parser.parse_args()
+    credentials_provisioner = CredentialsProvisioner()
+    args.func(credentials_provisioner, args)
--- a/scripts/subscribe.py
+++ b/scripts/subscribe.py
@@ -24,7 +24,7 @@ def main(argv):
 
     if options.shared:
         sm = boto3.client('secretsmanager')
-        creds = sm.get_secret_value(SecretId=config.google_service_account('indexer'))
+        creds = sm.get_secret_value(SecretId=config.secrets_manager_secret_name('indexer', 'google_service_account'))
         with tempfile.NamedTemporaryFile(mode='w+') as f:
             f.write(creds['SecretString'])
             f.flush()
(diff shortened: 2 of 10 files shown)


jessebrennan:
@hannes-ucsc tests aren't passing, but this is because this test needs to be fixed. I need input on what the best approach would be. The code can otherwise be reviewed (though probably it would better to talk about the test. I've put it in review so that it is on your radar.

requirements.txt
```diff
@@ -12,6 +12,7 @@ elasticsearch-dsl>=5.0.0,<6.0.0
 jsonobject==0.7.1
 Faker==0.8.11
 future==0.16.0
+google-api-python-client==1.7.8
```
hannes-ucsc: Shouldn't this be a dev requirement?

scripts/provision_credentials.py
```diff
@@ -0,0 +1,141 @@
...
+                name='projects/-/serviceAccounts/' + service_account_email + '/keys/' + key_id).execute()
+        except HttpError as e:
+            if e.resp.reason != 'Not Found':
+                raise
+        logger.info("Successfully deleted service account key with id '%s' for user '%s'",
+                    key_id, service_account_email)
+
+
+def random_hmac_key():
+    # Even though an hmac key can be any sequence of bytes, we restrict to base64 in order to encode as string
```
hannes-ucsc:
```suggestion
    # Even though an HMAC key can be any sequence of bytes, we restrict to base64 in order to avoid the complexities of embedding a binary HMAC key in JSON.
```

scripts/provision_credentials.py
```diff
@@ -0,0 +1,141 @@
...
+    # 48 because this works out a string of length 64
+    key = base64.encodebytes(os.urandom(48)).replace(b'=', b'').replace(b'\n', b'').decode()
```
hannes-ucsc:
```suggestion
    key = base64.encodebytes(os.urandom(48)).decode().replace('=', '').replace('\n', '')
```

Assert the expected length afterwards. That would also eliminate/reduce the need for the comment above.

scripts/provision_credentials.py
```diff
@@ -0,0 +1,141 @@
...
+    return json.dumps({'key': key, 'key_id': str(uuid.uuid4())})
+
+
+def create_aws_secrets_manager_secret(client, secret_name, value_callback):
+    """
+    Only creates a secret if one with the same name does not currently exist.
```
hannes-ucsc:
```suggestion
    Create a secret unless a secret of the given name already exists.
```

scripts/provision_credentials.py
```diff
@@ -0,0 +1,141 @@
...
+
+def destroy_credentials(account_email):
+    client = boto3.client('secretsmanager')
+    destroy_service_account_creds(client,  account_email)
+    destroy_aws_secrets_manager_secret(client, 'google_service_account')
+    destroy_aws_secrets_manager_secret(client, 'hmac')
+
+
+if __name__ == "__main__":
+    action = sys.argv[1]
```
hannes-ucsc: Consider spending 10min to add an argparse parser with help string etc.

src/azul/hmac.py
```diff
@@ -16,13 +17,16 @@ def verify(current_request):
     endpoint = f'{base_url}{path}'
     method = current_request.context['httpMethod']
     headers = current_request.headers
-    hmac_secret_key = config.hmac_key.encode()
+
+    def key_resolver(key_id, algorithm):
+        return aws.get_hmac_key_and_id_cached(key_id)['key']
```
hannes-ucsc: Uhh, this is may be a bit of a vulnerability. We definitely don't want to pass a user-supplied value to a function that extracts a secret using the user-supplied value.

hannes-ucsc: Ignore  this.

requirements.txt
jessebrennan: Yes, I'm moving it.

scripts/provision_credentials.py
jessebrennan: Good idea.

scripts/provision_credentials.py
jessebrennan: Changed.

scripts/provision_credentials.py
jessebrennan: Argparser made.

scripts/subscribe.py
```diff
@@ -24,7 +24,7 @@ def main(argv):
 
     if options.shared:
         sm = boto3.client('secretsmanager')
-        creds = sm.get_secret_value(SecretId=config.google_service_account('indexer'))
+        creds = sm.get_secret_value(SecretId=config.secrets_manager_secret_name('google_service_account'))
```
jessebrennan: @hannes-ucsc here is the changed to the subscription code.

jessebrennan: Ahh. but I see I should have changed more. There were some lines I overlooked.

src/azul/__init__.py
```diff
@@ -309,8 +309,8 @@ def _validate_term(self, term: str, name: str = 'Term'):
         require(self.term_re.fullmatch(term) is not None,
                 f"{name} is either too short, too long or contains invalid characters: '{term}'")
 
-    def google_service_account(self, lambda_name):
-        return f"dcp/azul/{self.deployment_stage}/{lambda_name}/google_service_account"
+    def secrets_manager_secret_name(self, *args):
+        return f"dcp/azul/{self.deployment_stage}/{'/'.join(args)}"
```
```suggestion
        return '/'.join(['dcp', 'azul', self.deployment_stage, *args])
```

Is more expressive, less brittle (doesn't duplicate the separator) and as long (not counting whitespace, which increases readability).

scripts/subscribe.py
```diff
@@ -24,7 +24,7 @@ def main(argv):
 
     if options.shared:
         sm = boto3.client('secretsmanager')
-        creds = sm.get_secret_value(SecretId=config.google_service_account('indexer'))
+        creds = sm.get_secret_value(SecretId=config.secrets_manager_secret_name('google_service_account', 'indexer'))
```
hannes-ucsc:
```suggestion
        creds = sm.get_secret_value(SecretId=config.secrets_manager_secret_name('indexer', 'google_service_account'))
```

I don't know why I asked for the reverse order yesterday. It doesn't make sense.

src/azul/deployment.py
```diff
@@ -106,6 +110,19 @@ def permissions_boundary_tf(self) -> Mapping[str, str]:
             'permissions_boundary': self.permissions_boundary['Arn']
         }
 
+    def get_hmac_key_and_id(self):
+        # Note: dict contains 'key' and 'key_id' as keys and is provisioned in scripts/provision_credentials.py
+        response = self.secretsmanager.get_secret_value(SecretId=config.secrets_manager_secret_name('hmac', 'indexer'))
```
hannes-ucsc:
```suggestion
        response = self.secretsmanager.get_secret_value(SecretId=config.secrets_manager_secret_name('indexer', 'hmac_key'))
```

src/azul/__init__.py
jessebrennan: I like it better too.

jessebrennan:
Change the way the Terraform is configured so that the HMAC key is always generated, not just when subscribing to DSS.

jessebrennan:
Also let's not merge this until I've done a live demo for you (since it's not really covered by the tests).

hannes-ucsc:
Gitlab sandbox build failed with insufficient permissions to delete service account credentials. This is expected since Gitlab has read-only permissions to Google Cloud. I ran `make terraforn` manually and retriggered the build.

jessebrennan:
@hannes-ucsc this issue should be fixed. I tested make subscribe locally. I think it is ready to be run in the sandbox again.

scripts/provision_credentials.py
```diff
@@ -0,0 +1,157 @@
...
+        filename=os.environ['GOOGLE_APPLICATION_CREDENTIALS'],
+        scopes=['https://www.googleapis.com/auth/cloud-platform'])
+    return googleapiclient.discovery.build('iam', 'v1', credentials=credentials)
+
+
+def create_service_account_creds(service_account_email):
+    service = get_google_service()
+    key = service.projects().serviceAccounts().keys().create(
+        name='projects/-/serviceAccounts/' + service_account_email, body={}
+        ).execute()
```
hannes-ucsc:
```suggestion
    ).execute()
```

scripts/provision_credentials.py
```diff
@@ -0,0 +1,157 @@
...
+            raise
+    else:
+        logger.info('AWS secret %s already exists. No changes will be made.', qualified_name)
+        return
+    try:
+        response = client.create_secret(Name=qualified_name, SecretString=value_callback())
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'ResourceExistsException':
+            logger.info('AWS secret %s already exists. No changes will be made.', qualified_name)
+            # FIXME: could add a cleanup callback
```
hannes-ucsc: I don't understand this or the warning below.

scripts/provision_credentials.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def destroy_aws_secrets_manager_secret(client, secret_name):
+    qualified_name = config.secrets_manager_secret_name('indexer', secret_name)
+    try:
+        response = client.delete_secret(SecretId=qualified_name,
+                                        ForceDeleteWithoutRecovery=True)
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'ResourceNotFoundException':
+            logger.info('AWS secret %s doesn\'t exist. No changes will be made.', qualified_name)
```
hannes-ucsc:
```suggestion
            logger.info("AWS secret %s doesn't exist. No changes will be made.", qualified_name)
```

jessebrennan:
@hannes-ucsc  could this be a Gitlab permissions error? I don't get it when I run locally.

scripts/provision_credentials.py
```diff
@@ -0,0 +1,169 @@
...
+                    name='projects/-/serviceAccounts/' + service_account_email + '/keys/' + key_id).execute()
+            except HttpError as e:
+                if e.resp.reason != 'Not Found':
+                    raise
+            logger.info("Successfully deleted service account key with id '%s' for user '%s'",
+                        key_id, service_account_email)
+
+
+if __name__ == "__main__":
+    credentials_provisioner = CredentialsProvisioner()
```
hannes-ucsc: This means that AWS credential errors would be reported before command line usage errors. Generally, the instantiation of the helper class could also have other side-effects that require it to occur after the CLI argumemt checking. Most importantly, the instance may need to parameterized with CLI arguments in which case it can only be instantiated after the arguments have been parsed. See below.

scripts/provision_credentials.py
```diff
@@ -0,0 +1,169 @@
...
+    group.add_argument('--build', '-b', action='store_true', dest='build',
+                       help='Create credentials instead of destroying them. This action is idempotent.')
+    group.add_argument('--destroy', '-d', action='store_false', dest='build',
+                       help='Destroy credentials instead of building them. This action is idempotent.')
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(help='Specify action', dest='action')
+    subparsers.required = True
+    email_parser = subparsers.add_parser('google-key', parents=[provision_parser],
+                                         help='Create Google service account key and store in an AWS secret.')
+    email_parser.set_defaults(func=credentials_provisioner.provision_google_from_args)
```
hannes-ucsc:
```suggestion
    email_parser.set_defaults(func=CredentialsProvisioner.provision_google_from_args)
```

scripts/provision_credentials.py
```diff
@@ -0,0 +1,169 @@
...
+    email_parser.add_argument('email', type=str,
+                              help='Email address for the Google service account '
+                                   'for which to provision credentials')
+    subparsers.add_parser('hmac-key', parents=[provision_parser],
+                          help='Create a random HMAC key and store in an AWS secret.'
+                          ).set_defaults(func=credentials_provisioner.provision_hmac_from_args)
```
hannes-ucsc:
```suggestion
                          ).set_defaults(func=CredentialsProvisioner.provision_hmac_from_args)
```

scripts/provision_credentials.py
```diff
@@ -0,0 +1,169 @@
...
+    args = parser.parse_args()
+    args.func(args)
```
hannes-ucsc:
```suggestion
    credentials_provisioner = CredentialsProvisioner()
    args.func(credentials_provisioner, args)
```

scripts/provision_credentials.py
```diff
@@ -0,0 +1,169 @@
...
+    provision_parser = argparse.ArgumentParser(add_help=False)
+    group = provision_parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('--build', '-b', action='store_true', dest='build',
+                       help='Create credentials instead of destroying them. This action is idempotent.')
+    group.add_argument('--destroy', '-d', action='store_false', dest='build',
+                       help='Destroy credentials instead of building them. This action is idempotent.')
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(help='Specify action', dest='action')
+    subparsers.required = True
+    email_parser = subparsers.add_parser('google-key', parents=[provision_parser],
```
hannes-ucsc: I find the `email_parser` name to be counter-intuitive. `google_parser` maybe?

scripts/provision_credentials.py
hannes-ucsc: Not a  fan of the chaining btw because it is asymmetric with the other subparser.

scripts/provision_credentials.py
jessebrennan: This is cool, I didn't realize you could do this. You're passing the 'self' parameter directly to the class method? I'll try it.

scripts/provision_credentials.py
jessebrennan: Good point, I've refactored with your suggestions.

scripts/provision_credentials.py
jessebrennan: Ok I removed it for symmetry

jessebrennan:
Note that switching to a different branch that has not integrated these changes may cause the following error:
```
Error: Error applying plan:

1 error(s) occurred:

* aws_secretsmanager_secret.indexer_google_service_account: 1 error(s) occurred:

* aws_secretsmanager_secret.indexer_google_service_account: error creating Secrets Manager Secret: ResourceExistsException: The operation failed because the secret dcp/azul/jesse/indexer/google_service_account already exists.
        status code: 400, request id: 4324f7f6-7026-4a4d-ab84-fb615877d578

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.


Makefile:17: recipe for target 'apply' failed
make[1]: *** [apply] Error 1
make[1]: Leaving directory '/home/jesse/gi/azul/terraform'
Makefile:9: recipe for target 'terraform' failed
make: *** [terraform] Error 2
```

To resolve this you will need to run:
```
git checkout feature/jessebrennan/794-provision-credentials
python -c 'from scripts.provision_credentials import CredentialsProvisioner as P; P()._destroy_aws_secrets_manager_secret("SECRET_NAME")'
# where SECRET_NAME is something like dcp/azul/jesse/indexer/google_service_account
```

Issue #839, opened 2019-03-26
Validate `format` parameter for manifest endpoints 


hannes-ucsc:
https://github.com/HumanCellAtlas/data-browser/issues/531

Ensure that only `tsv` or `bdbag` are passed.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-519)
┆Project Name: azul
┆Issue Number: AZUL-519

PR #917, opened 2019-04-17
Added validation for manifest types (#839)

GPelayo:


lambdas/service/app.py
```diff
@@ -421,12 +421,18 @@ def get_manifest():
...
-    format = params.get('format', 'tsv')
     es_td = EsTd()
-    response = es_td.transform_manifest(format, filters)
+    response = es_td.transform_manifest(get_format(params), filters)
     return response
 
 
+def get_format(params):
+    format_ = params.get('format', 'tsv')
+    if format_ not in {'tsv', 'bdbag'}:
```
hannes-ucsc: Eliminate `not`, use `return` in `if` branch and `raise` in `else` branch.

hannes-ucsc: Nit: 

```suggestion
    if format_ not in ('tsv', 'bdbag'):
```

lambdas/service/app.py
```diff
@@ -457,6 +463,9 @@ def start_manifest_generation():
     If the manifest generation is done and the manifest is ready to be downloaded, the response will
     have a 302 status and will redirect to the URL of the manifest.
     """
+    query_params = app.current_request.query_params
```
hannes-ucsc: Inline.

test/service/test_request_validation.py
```diff
@@ -487,3 +487,10 @@ def test_bdbag_manifest(self):
                     'fastq[read2].dos_url',
                     'fastq[read2].file_url',
                 ], reader.fieldnames)
+
+    def test_manifest_format_validation(self):
+        for manifest_endpoint in '/manifest/files', '/repository/files/export':
+            with self.subTest(manifest_endpoint=manifest_endpoint):
+                url = self.base_url + f'{manifest_endpoint}?format=invalid-type'
+                response = requests.get(url)
+                self.assertEqual(400, response.status_code, response.content)
```
hannes-ucsc: Why do you use the `response.content` as the assertion error message? Please explain.

GPelayo: While coding this test, this assert kept getting 500 response errors. Adding `response.content` as the assertion error message helped me diagnose the source of those errors.

hannes-ucsc:
Sandbox was poisoned by a previous PR.

hannes-ucsc:
Fixed sandbox, now retrying build.

Issue #896, opened 2019-04-12
Add secondary analysis workflow type to index


theathorn:
The browser needs to be able to distinguish the workflow type for secondary analysis files (e.g. Cellranger vs SS2 vs Optimus, etc.)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-555)
┆Project Name: azul
┆Issue Number: AZUL-555
┆Epic: Improved Matrix Service Integration

hannes-ucsc:
Demo-ed on 5/13/2019

PR #924, opened 2019-04-18
Add AnalysisProtocol.protocol_id to index (#896)

danielsotirhos:


src/azul/project/hca/transformers.py
```diff
@@ -106,6 +106,7 @@ def _file_dict(file: api.File) -> JSON:
 
 def _protocol_dict(protocol: api.Protocol) -> JSON:
     protocol_dict = {"document_id": protocol.document_id}
+    protocol_dict['protocol_id'] = protocol.protocol_id
```
hannes-ucsc: I have the feeling we should only include the `protocol_id` from `AnalyisProtocol` instances. 

If we don't do this, we'll get all those garbage-y `library_preparation_protocol_1` that would clutter up the facet drop down in the browser.

We may also want to introduce two keys by splitting `AnalyisProtocol.protocol_id` at `'_'` and assigning to `protocol_dict['workflow_name']` and `protocol_dict['workflow_version']`. Not sure about this at all. It would introduce another "nested facet". cc:@NoopDog

danielsotirhos: Modified so that only AnalysisProtocol objects will have their `protocol_id` value split up on the last underscore and stored in `protocol_dict['workflow']` and `protocol_dict['workflow_version']`.

src/azul/service/service_config/request_config.json
```diff
@@ -6,9 +6,12 @@
     "fileId": "contents.files.uuid",
     "fileVersion": "contents.files.version",
 
+    "protocolId": "contents.protocols.protocol_id",
```
hannes-ucsc: This should be removed, right?

test/indexer/test_hca_indexer.py
```diff
@@ -615,10 +615,24 @@ def test_paired_end_field(self):
                 self.assertEqual(0, len(contents['protocols']))
             else:
                 if aggregate:
-                    self.assertEqual(one(contents['protocols'])['paired_end'], [True])
+                    self.assertEqual(set(one(contents['protocols'])['paired_end']), set([True]))
                 else:
-                    values = [p['paired_end'] for p in contents['protocols'] if 'paired_end' in p]
-                    self.assertEqual(set(values), set([True]))
+                    paired_end_values = [p['paired_end'] for p in contents['protocols'] if 'paired_end' in p]
```
```suggestion
                    paired_end_values = {p.get('paired_end') for p in contents['protocols']}
```

test/indexer/test_hca_indexer.py
```diff
@@ -615,10 +615,24 @@ def test_paired_end_field(self):
...
+                    self.assertEqual(set(paired_end_values), set([True]))
```
```suggestion
                    self.assertEqual(paired_end_values, {True, None}))
```

Aside from the use of set literals/comprehensions, my version is also a stronger assertion.

test/indexer/test_hca_indexer.py
```diff
@@ -615,10 +615,24 @@ def test_paired_end_field(self):
...
+        hits = self._get_hits()
+        for hit in hits:
+            contents = hit['_source']['contents']
+            entity_type, aggregate = config.parse_es_index_name(hit['_index'])
+            if aggregate:
+                self.assertEqual(one(contents['protocols'][0]['workflow']), 'smartseq2')
+                self.assertEqual(one(contents['protocols'][0]['workflow_version']), 'v2.1.0')
+            else:
+                protocols = contents['protocols']
+                self.assertEqual({p.get('workflow', None) for p in protocols}, set(['smartseq2', None]))
```
hannes-ucsc: See above wrt set literals.

src/azul/service/service_config/request_config.json
danielsotirhos: Yes, good catch

test/indexer/test_hca_indexer.py
```diff
@@ -617,8 +617,20 @@ def test_paired_end_field(self):
...
+                    self.assertEqual({p.get('paired_end') for p in contents['protocols']}, {True, None})
+
+    def test_analysis_protocol_fields(self):
+        self._index_canned_bundle(('3f8176ff-61a7-4504-a57c-fc70f38d5b13', '2018-10-24T234431.820615Z'))
+        hits = self._get_hits()
+        for hit in hits:
+            contents = hit['_source']['contents']
+            entity_type, aggregate = config.parse_es_index_name(hit['_index'])
+            if aggregate:
+                self.assertEqual(one(contents['protocols'][0]['workflow']), 'smartseq2')
```
```suggestion
                self.assertEqual(one(one(contents['protocols'])['workflow']), 'smartseq2')
```

?

test/indexer/test_hca_indexer.py
```diff
@@ -617,8 +617,20 @@ def test_paired_end_field(self):
...
+                self.assertEqual(one(contents['protocols'][0]['workflow_version']), 'v2.1.0')
```
hannes-ucsc: Ditto.

danielsotirhos:
Rebased on top of develop 4/30 Tuesday morning

danielsotirhos:
- `test_sequencing_protocol_fields()` was merged into `test_plate_bundle()`
- `test_analysis_protocol_fields()` was merged into `test_well_bundles()` (couldn't merge into `test_plate_bundle()` as that test doesn't index a bundle containing an analysis_protocol)

hannes-ucsc:
Note the [improved commit message](https://github.com/DataBiosphere/azul/commit/a09ff818826f81b1b3fdd5edd1dcea9c680a4df2), @danielsotirhos.

Issue #894, opened 2019-04-10
Correlate project short name and UUID in term facets response


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-553)
┆Project Name: azul
┆Issue Number: AZUL-553

hannes-ucsc:
Looks like we can use a nested `term` aggregation to associate each project short name term with the set of corresponding project UUID terms at very little cost. Note that there could be more than one UUID per short name.

PR #929, opened 2019-04-19
Nest projectId under project termFacet (#894)

danielsotirhos:


src/azul/service/service_config/request_config.json
```diff
@@ -175,6 +175,7 @@
     "instrumentManufacturerModel",
     "pairedEnd",
     "project",
+    "projectId",
```
hannes-ucsc: Doesn't this have the effect of adding yet another aggregation?

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -799,9 +799,16 @@ def choose_entry(_term):
...
-                     for term in contents['myTerms']['buckets']]
+        term_list = []
+        for term in contents['myTerms']['buckets']:
+            if 'myProjectIds' in term:
+                project_ids = [bucket['key'] for bucket in term['myProjectIds']['buckets']]
+                term_list.append(TermObj(**{'term': choose_entry(term),
+                                            'count': term['doc_count'],
+                                            'projectId': project_ids}))
+            else:
+                term_list.append(TermObj(**{'term': choose_entry(term),
```
hannes-ucsc: `f(**{"foo":bar})` <=> `f(foo=bar)`

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -799,9 +799,16 @@ def choose_entry(_term):
...
 
-        term_list = [TermObj(**{'term': choose_entry(term),
-                                'count': term['doc_count']})
-                     for term in contents['myTerms']['buckets']]
+        term_list = []
+        for term in contents['myTerms']['buckets']:
+            if 'myProjectIds' in term:
+                project_ids = [bucket['key'] for bucket in term['myProjectIds']['buckets']]
+                term_list.append(TermObj(**{'term': choose_entry(term),
+                                            'count': term['doc_count'],
```
hannes-ucsc: These two lines are identical in both branches. See if you can reduce duplication. `**` is your friend, but not in the way you currently use it.

src/azul/service/service_config/request_config.json
danielsotirhos: Unnecessary new facet "projectId" removed

src/azul/service/responseobjects/hca_response_v5.py
danielsotirhos: Refactored the construction of the term_list list to avoid code duplication.

test/service/test_response.py
```diff
@@ -1205,6 +1205,9 @@ def test_filter_by_projectId(self):
                                 self.assertEqual(test_data['title'], project['projectTitle'])
                             else:
                                 self.assertIn(test_data['title'], project['projectTitle'])
+                    for term in response_json['termFacets']['project']['terms']:
+                        self.assertIsInstance(term['projectId'], list)
+                        self.assertEqual(one(term['projectId']), test_data['id'])
```
```suggestion
                        self.assertEqual(term['projectId'], [test_data['id']])
```

test/service/test_response.py
```diff
@@ -1205,6 +1205,9 @@ def test_filter_by_projectId(self):
                                 self.assertEqual(test_data['title'], project['projectTitle'])
                             else:
                                 self.assertIn(test_data['title'], project['projectTitle'])
+                    for term in response_json['termFacets']['project']['terms']:
+                        self.assertIsInstance(term['projectId'], list)
```
```suggestion
```

test/service/test_response.py
danielsotirhos: Good idea, testing the singleton list value like this (instead of using one()) makes it unnecessary to first test if the value is a list.

Issue #927, opened 2019-04-19
Upgrade urllib3


hannes-ucsc:
https://github.com/DataBiosphere/azul/network/alert/requirements.txt/urllib3/open

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-577)
┆Project Name: azul
┆Issue Number: AZUL-577

hannes-ucsc:
<img width="1046" alt="Screen Shot 2019-04-23 at 1 03 12 PM" src="https://user-images.githubusercontent.com/5143256/56612062-30e9bd00-65c8-11e9-82ab-1f5c6513be6d.png">

PR #935, opened 2019-04-23
Updated urllib3 to 1.24.2 (#927)

GPelayo:


requirements.txt
```diff
@@ -22,5 +22,5 @@ PyJWT==1.6.4
 requests==2.20.0
 requests-http-signature==0.0.3
 rsa==4.0
-urllib3==1.23
+urllib3>=1.24.2, <1.25
```
hannes-ucsc: Do you see version ranges anywhere in this file? Stick with precedent.

Please pin to latest 1.24.x release.

GPelayo: There are version ranges here:
https://github.com/DataBiosphere/azul/blob/0fd69736402df94283afb4d2f70eb69f2065cb12/requirements.txt#L10-L11

Issue #838, opened 2019-03-26
Add bundle deletion to integration test


hannes-ucsc:
This will also have the added benefit of leaving test projects around.

Ideally the test would leave the index as it originally was. This may require careful thought. I am not 100% certain that deletion is the right way to undo the indexing of bundles with the project name patch in place. But it may well be.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-518)
┆Project Name: azul
┆Issue Number: AZUL-518

jessebrennan:
while doing the integration test no genuine notifications should be processed because they may collide with the test set.

hannes-ucsc:
Demo-ed on 5/13/2019

but @jessebrennan to file follow-up ticket about clean-up.

PR #946, opened 2019-04-26
Add deletion to integration test (#838)

jessebrennan:
The fix works by changing the test reindexing to also assign a new
uuid for the test bundles. These new uuids are then searched in the
index to make sure that indexing was successful. Finally deletions for
all of the relevant bundles that were found in the index are sent to
Azul

### DIFF
--- a/src/azul/azulclient.py
+++ b/src/azul/azulclient.py
@@ -57,13 +57,6 @@ class AzulClient(object):
         response.raise_for_status()
         return response.content
 
-    def _make_test_notification(self, bundle_fqid, new_bundle_uuid, test_name, test_uuid):
-        notification = self._make_notification(bundle_fqid)
-        notification["test_bundle_uuid"] = new_bundle_uuid
-        notification["test_name"] = test_name
-        notification["test_uuid"] = test_uuid
-        return notification
-
     def _make_notification(self, bundle_fqid):
         bundle_uuid, _, bundle_version = bundle_fqid.partition('.')
         return {
@@ -93,7 +86,11 @@ class AzulClient(object):
             new_bundle_uuid = str(uuid.uuid4())
             _, _, version = bundle_fqid.partition('.')
             effective_bundle_fqids.add(new_bundle_uuid + '.' + version)
-            notifications.append(self._make_test_notification(bundle_fqid, new_bundle_uuid, test_name, test_uuid))
+            notification = dict(self._make_notification(bundle_fqid),
+                                test_bundle_uuid=new_bundle_uuid,
+                                test_name=test_name,
+                                test_uuid=test_uuid)
+            notifications.append(notification)
         self._reindex(notifications, sync=sync)
         return effective_bundle_fqids
 
--- a/src/azul/azulclient.py
+++ b/src/azul/azulclient.py
@@ -1,3 +1,4 @@
+import uuid
 from collections import defaultdict
 from concurrent.futures import Future, ThreadPoolExecutor
 from functools import partial, lru_cache
@@ -7,7 +8,7 @@ import logging
 from pprint import PrettyPrinter
 
 import requests
-from typing import List, Optional, Iterable
+from typing import List, Optional, Iterable, Set
 from urllib.error import HTTPError
 from urllib.parse import parse_qs, urlencode, urlparse
 
@@ -29,7 +30,6 @@ class AzulClient(object):
                  query: Optional[JSON] = None,
                  prefix: str = config.dss_query_prefix,
                  num_workers: int = 16,
-                 test_name: str = None,
                  dryrun: bool = False):
         assert query is None or not prefix, "Cannot mix `prefix` and `query`"
         self.num_workers = num_workers
@@ -37,7 +37,6 @@ class AzulClient(object):
         self.prefix = prefix
         self.dss_url = dss_url
         self.indexer_url = indexer_url
-        self.test_name = test_name
         self.dryrun = dryrun
 
     def query(self, prefix=None):
@@ -50,14 +49,20 @@ class AzulClient(object):
             assert not prefix
             return self._query
 
-    def post_bundle(self, bundle_fqid, indexer_url):
+    def post_bundle(self, indexer_url, notification):
         """
         Send a mock DSS notification to the indexer
         """
-        simulated_event = self._make_notification(bundle_fqid)
-        response = requests.post(indexer_url, json=simulated_event, auth=hmac.prepare())
+        response = requests.post(indexer_url, json=notification, auth=hmac.prepare())
         response.raise_for_status()
         return response.content
+
+    def _make_test_notification(self, bundle_fqid, new_bundle_uuid, test_name, test_uuid):
+        notification = self._make_notification(bundle_fqid)
+        notification["test_bundle_uuid"] = new_bundle_uuid
+        notification["test_name"] = test_name
+        notification["test_uuid"] = test_uuid
+        return notification
 
     def _make_notification(self, bundle_fqid):
         bundle_uuid, _, bundle_version = bundle_fqid.partition('.')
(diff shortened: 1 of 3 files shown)


test/local_integration_test.py
```diff
@@ -44,32 +65,36 @@ def setUp(self):
         super().setUp()
         self.set_lambda_test_mode(True)
         self.bundle_uuid_prefix = ''.join([str(random.choice('abcdef0123456789')) for _ in range(self.prefix_length)])
+        self.indexed_uuids_to_version = {}
```
hannes-ucsc: I have the feeling that returning fqids from the reindex method will allow this to be a set instead.

src/azul/azulclient.py
```diff
@@ -50,29 +53,33 @@ def query(self, prefix=None):
...
             "query": self.query(),
             "subscription_id": 'feeb0f2b-c16f-48f8-bac0-e6fd09b92320',
             "transaction_id": 'feeb0f2b-c16f-48f8-bac0-e6fd09b92320',
             "match": {
                 "bundle_uuid": bundle_uuid,
                 "bundle_version": bundle_version
             },
-            "test_name": self.test_name
         }
+        if new_bundle_fqid:
```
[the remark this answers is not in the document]
jessebrennan: I do not prefer this. It uses special syntax and is unintuitive. Considering the second suggestion though.

src/azul/azulclient.py
```diff
@@ -137,7 +150,7 @@ def handle_future(future):
         logger.warning("Total number of errors by code:\n%s", printer.pformat(dict(errors)))
         logger.warning("Missing bundle_fqids and their error code:\n%s", printer.pformat(missing))
 
-        return bundle_fqids
+        return new_bundle_uuids if test else bundle_fqids
```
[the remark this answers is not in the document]
jessebrennan: Good point. I didn't do this at first because I didn't think it was possible, but now i see that I can just use the version from the _real_ bundle.

test/local_integration_test.py
```diff
@@ -33,6 +33,27 @@ def setUpModule():
 
 
 class IntegrationTest(unittest.TestCase):
+    """
+    The integration tests works by first setting the lambdas in test mode (this happens in setUp). This
+    Sets some environment variable in the indexer lambda that changes it's behavior slightly, allowing
+    it to process modified test notifications notifications differently.
+
+    Next the test querying Azul for all of the bundles that match a certain prefix. In
```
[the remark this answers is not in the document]
jessebrennan: Whoops... thanks.

test/local_integration_test.py
```diff
@@ -196,15 +225,15 @@ def wait_for_queue_level(self, empty: bool = True, timeout: int = 60):
...
         service_check_timeout = 600
         delay_between_retries = 5
         page_size = 100
 
-        num_bundles = len(indexed_bundle_fqids)
+        num_bundles = len(expected_bundle_uuids)
         logger.info('Starting integration test %s with the prefix %s for the entity type %s. Expected %i bundle(s).',
                     test_name, self.bundle_uuid_prefix, entity_type, num_bundles)
-        logger.debug('Expected bundles %s ', sorted(indexed_bundle_fqids))
+        logger.debug('Expected bundles %s ', sorted(expected_bundle_uuids))
```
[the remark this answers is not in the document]
jessebrennan: I didn't write this line, but would happily delete it. Empircally it seems to be in the low 20s.

test/local_integration_test.py
jessebrennan: This is true.

src/azul/azulclient.py
```diff
@@ -50,15 +49,21 @@ def query(self, prefix=None):
...
         Send a mock DSS notification to the indexer
         """
-        simulated_event = self._make_notification(bundle_fqid)
-        response = requests.post(indexer_url, json=simulated_event, auth=hmac.prepare())
+        response = requests.post(indexer_url, json=notification, auth=hmac.prepare())
         response.raise_for_status()
         return response.content
 
+    def _make_test_notification(self, bundle_fqid, new_bundle_uuid, test_name, test_uuid):
+        notification = self._make_notification(bundle_fqid)
```
hannes-ucsc: Consider 

```suggestion
        return dict(self._make_notification(bundle_fqid), test_bundle_uuid=…, …)
```

Also consider inlining this call as its pretty anemic considering the long signature.

test/local_integration_test.py
```diff
@@ -33,6 +33,27 @@ def setUpModule():
...
+    """
+    The integration tests work by first setting the lambdas in test mode (this happens in setUp). This
+    Sets some environment variable in the indexer lambda that change its behavior slightly, allowing
+    it to process modified test notifications differently.
+
+    Next, the test queries Azul for all of the bundles that match a certain prefix. In
+    AzulClient.reindex() these bundle are used to construct fake notifications. These fake notifications
+    Have an extra information embedded. They include:
+     - the name of the integration test run,
+     - a new uuid for the project, and
```
hannes-ucsc:
```suggestion
     - a new UUID for the project, and
```

hannes-ucsc: Here and everywhere else in prose.

test/local_integration_test.py
```diff
@@ -33,6 +33,27 @@ def setUpModule():
...
+     - a new unique bundle UUID.
+    When the bundle is processed by the indexer lambda the original bundle uuid is used to fetch the
+    metadata, etc., but these new fields are what's actually stored in Azul. This effectively creates a
+    copy of the bundles, but with a new uuid and shared project.
```
hannes-ucsc:
```suggestion
    copy of the bundles, but each copy with a new and different bundle UUID and all copies sharing a synthesized project with a different project UUID. Note that unlike real projects, the shared project will contain a mix of metadata graph shapes since the bundles that contribute to it are selected randomly.
```

test/local_integration_test.py
```diff
@@ -33,6 +33,27 @@ def setUpModule():
...
+
+    The UUIDs for these new bundles are used within the integration tests to assert that everything is
+    indexed. Health endpoints etc. are also checked.
+
+    Finally, the tests schedule deletion for these new UUIDs which should remove all trace of the
```
hannes-ucsc:
```suggestion
    Finally, the tests send deletion notifications for these new bundle UUIDs which should remove all trace of the
```

src/azul/azulclient.py
hannes-ucsc: I don't think the additional function simplifies anything. The signature is long and brittle and the body is anemic. This doesn't increase readability.

src/azul/azulclient.py
```diff
@@ -69,41 +74,57 @@ def _make_notification(self, bundle_fqid):
...
+        logger.info('Querying DSS using %s', json.dumps(self.query(), indent=4))
+        real_bundle_fqids = self._post_dss_search()
+        logger.info("Bundle FQIDs to index: %i", len(real_bundle_fqids))
+        notifications = []
+        effective_bundle_fqids = set()
+        for bundle_fqid in real_bundle_fqids:
+            new_bundle_uuid = str(uuid.uuid4())
+            _, _, version = bundle_fqid.partition('.')
+            effective_bundle_fqids.add(new_bundle_uuid + '.' + version)
+            notifications.append(self._make_test_notification(bundle_fqid, new_bundle_uuid, test_name, test_uuid))
```
hannes-ucsc: Inlining the method would brake this long, harder to read line into multiple lines.

test/local_integration_test.py
```diff
@@ -222,25 +255,26 @@ def check_bundles_are_indexed(self, test_name: str, entity_type: str, indexed_bu
...
         while True:
             response_json = self.requests.get(url).json()
             hits = response_json.get('hits', [])
-            found_bundle_fqids.update(f"{entity['bundleUuid']}.{entity['bundleVersion']}"
-                                      for bundle in hits for entity in bundle.get('bundles', [])
-                                      if entity['bundleUuid'].startswith(self.bundle_uuid_prefix))
+            self.indexed_fqids.update({f"{entity['bundleUuid']}.{entity['bundleVersion']}"
+                                       for hit in hits
+                                       for entity in hit.get('bundles', [])
+                                       # FIXME: make sure this is right
```
hannes-ucsc: ?

test/local_integration_test.py
```diff
@@ -264,7 +298,32 @@ def check_bundles_are_indexed(self, test_name: str, entity_type: str, indexed_bu
                             f' {hit["entryId"]}. Bundle(s) ({",".join(bundle_fqids)})'
                             f' have been indexed without the debug project name. Contains {project}')
 
-    @classmethod
-    def _get_file_uuid(cls, response: bytes) -> str:
-        """Returns file UUID of first FASTQ file."""
-
+    def project_removed_from_azul(self):
+        results_empty = [len(self._get_entities_by_project(entity, self.test_name)) == 0
+                         for entity in ['files', 'projects', 'samples']]
```
hannes-ucsc: Include `'bundles'`.

test/local_integration_test.py
```diff
@@ -264,7 +298,32 @@ def check_bundles_are_indexed(self, test_name: str, entity_type: str, indexed_bu
...
+        logger.info("Project removed from index files: {}, projects: {}, specimens: {}".format(*results_empty))
+        return all(results_empty)
+
+    def _get_entities_by_project(self, entity_type, project_shortname):
+        """
+        Returns all entities in a given project.
+        """
+        filters = {'file': {'project': {'is': [project_shortname]}}}
+        files = []
```
hannes-ucsc:
```suggestion
        entities = []
```

test/local_integration_test.py
```diff
@@ -264,7 +298,32 @@ def check_bundles_are_indexed(self, test_name: str, entity_type: str, indexed_bu
...
+        size = 100
+        params = dict(filters=str(filters), size=str(size))
+        while True:
```
hannes-ucsc: There is similar code in `check_bundles_are_indexed` and I think we can't afford to maintain them alongside each other. Could you please consolidate these two? `check_bundles_are_indexed` also uses a weird timeout which I think is unnecessary.

On a general note: As long as I am project lead, I will pursue keeping entropy and duplication low in the code base unless there is a good reason not do do so, like time pressure. While I understand the desire to churn thru tickets and check boxes, our main concern should be long-term maintainability and minimization of technical debt.

test/local_integration_test.py
```diff
@@ -264,7 +298,32 @@ def check_bundles_are_indexed(self, test_name: str, entity_type: str, indexed_bu
...
-
+    def project_removed_from_azul(self):
+        results_empty = [len(self._get_entities_by_project(entity, self.test_name)) == 0
+                         for entity in ['files', 'projects', 'samples']]
+        logger.info("Project removed from index files: {}, projects: {}, specimens: {}".format(*results_empty))
+        return all(results_empty)
+
+    def _get_entities_by_project(self, entity_type, project_shortname):
+        """
+        Returns all entities in a given project.
```
hannes-ucsc:
```suggestion
        Returns all entities of a given type in a given project.
```

hannes-ucsc:
The good news is that the sandbox build passed.

src/azul/azulclient.py
jessebrennan: Done. Sorry I missed this review comment before.

test/local_integration_test.py
jessebrennan: Development artifact... removed, thanks.

test/local_integration_test.py
jessebrennan: I appreciate the desire to refactor. Will do.

src/azul/azulclient.py
```diff
@@ -50,15 +49,20 @@ def query(self, prefix=None):
...
         """
         Send a mock DSS notification to the indexer
         """
-        simulated_event = self._make_notification(bundle_fqid)
-        response = requests.post(indexer_url, json=simulated_event, auth=hmac.prepare())
+        response = requests.post(indexer_url, json=notification, auth=hmac.prepare())
         response.raise_for_status()
         return response.content
 
+    def _make_test_notification(self, bundle_fqid, new_bundle_uuid, test_name, test_uuid):
```
hannes-ucsc: This can be removed now, yes?

jessebrennan: Whoops! Should be good now.

PR #949, opened 2019-04-30
Backport exclusion of BM_PC and neuron_diff projects in prod

achave11:


achave11:
@hannes-ucsc merge instead of rebasing.

Issue #686, opened 2019-01-25
Download of small files is too slow


theathorn:
After selecting a very small file (e.g. 32KB pdf) for single file download, the download takes approx. 10 seconds when it should be almost instantaneous. @NoopDog theorized that Azul is polling the file creation in S3 at a fixed interval and that we could speed up the process by polling at a very short interval initially and then backing off exponentially.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-426)
┆Project Name: azul
┆Issue Number: AZUL-426

theathorn:
10s delay if checkout required. Fast if checkout already happened.
data-store issue.

hannes-ucsc:
In our DSS proxy, we could ignore the Retry-After header sent back by DSS and implement exponential back-off: 1, 2, 4, 10 or so like we do for TSVs. But still file a ticket against DSS.

hannes-ucsc:
Demo-ed on 5/13/2019

PR #963, opened 2019-05-07
Ignore Retry-After from DSS for small file downloads (#686)

achave11:


lambdas/service/app.py
```diff
@@ -771,9 +771,16 @@ def _dss_files(uuid, fetch=True):
     url = config.dss_endpoint + '/files/' + urllib.parse.quote(uuid, safe='')
     file_name = params.pop('fileName', None)
     wait = params.pop('wait', None)
+    request_number = params.pop('requestNumber', None)
```
```suggestion
    request_index = params.pop('requestIndex', 0)
```

There is precedent for this naming in the manifest generation code.

We prefer 0-based.

lambdas/service/app.py
```diff
@@ -771,9 +771,16 @@ def _dss_files(uuid, fetch=True):
...
     dss_response = requests.get(url, params=params, allow_redirects=False)
     if dss_response.status_code == 301:
-        retry_after = int(dss_response.headers.get('Retry-After'))
+        if request_number is None:
```
hannes-ucsc: Simplify with above use of 0 as the default.

test/service/test_dss_proxy.py
```diff
@@ -72,7 +72,7 @@ def test_dss_files_proxy(self):
         s3.upload_fileobj(Bucket=bucket_name, Fileobj=io.BytesIO(b'foo'), Key=key)
         file_uuid = '701c9a63-23da-4978-946b-7576b6ad088a'
         file_version = '2018-09-12T121154.054628Z'
-        retry_after = 3
+        retry_after = 1
```
hannes-ucsc: This makes the mock DSS return a retry after of 1 which means the test doesn't prove that your changes override the retry after specified by the DSS.
```suggestion
        retry_after = 10
```

lambdas/service/app.py
```diff
@@ -771,9 +771,13 @@ def _dss_files(uuid, fetch=True):
     url = config.dss_endpoint + '/files/' + urllib.parse.quote(uuid, safe='')
     file_name = params.pop('fileName', None)
     wait = params.pop('wait', None)
+    request_index = params.pop('requestIndex', 0)
     dss_response = requests.get(url, params=params, allow_redirects=False)
     if dss_response.status_code == 301:
-        retry_after = int(dss_response.headers.get('Retry-After'))
+        params['requestIndex'] = request_index + 1
```
hannes-ucsc: Swap these two statements.

lambdas/service/app.py
```diff
@@ -771,9 +771,13 @@ def _dss_files(uuid, fetch=True):
...
+        retry_after = min(int(dss_response.headers.get('Retry-After')),
+                          int(1.3 ** params['requestIndex']))
```
```suggestion
                          int(1.3 ** request_index))
```

test/service/test_dss_proxy.py
```diff
@@ -106,7 +105,7 @@ def test_dss_files_proxy(self):
                                                           url=dss_url.url,
                                                           status=301,
                                                           headers={'Location': dss_url_with_token.url,
-                                                                   'Retry-After': str(retry_after)}))
+                                                                   'Retry-After': str(10)}))  # Default value from DSS
```
hannes-ucsc:
```suggestion
                                                                   'Retry-After': '10'}))  # Default value from DSS
```

hannes-ucsc:
```suggestion
                                                                   'Retry-After': str(10)}))
```

What the DSS actually returns might change in the future, in which case the comment becomes false.

Issue #930, opened 2019-04-20
Index cell line fields


hannes-ucsc:
… most prominently `model_organ`. Don't forget to include cell lines and their model organ in the secondary visitation, the one that starts from cell suspensions.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-579)
┆Project Name: azul
┆Issue Number: AZUL-579

hannes-ucsc:
Demo-ed on 5/13/2019

PR #974, opened 2019-05-08
Index cell_line fields (#930)

danielsotirhos:


test/indexer/test_hca_indexer.py
```diff
@@ -651,10 +652,22 @@ def test_cell_line_sample(self):
...
+                    self.assertEqual(one(contents['specimens'])['organ'], ['blood'])
+                    self.assertEqual(one(contents['specimens'])['organ_part'], ['venous blood'])
+                    self.assertEqual(one(contents['cell_lines'])['model_organ'], ['blood (cell_line)'])
+                    self.assertEqual(one(contents['cell_suspensions'])['organ'], ['blood (cell_line)'])
+                    self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [None])
                 else:
                     document_ids = [d['document_id'] for d in contents[sample_entity_type]]
                     entity = one([d for d in contents[sample_entity_type] if d['document_id'] == sample['document_id']])
                     self.assertEqual('cell_line_Day7_hiPSC-CM_BioRep2', entity['biomaterial_id'])
+                    self.assertEqual(one(contents['specimens'])['organ'], 'blood')
```
hannes-ucsc: Almost the same as the line in the if branch. Only difference is that RHS of comparison is `["…"]` vs `"…"`

test/indexer/test_hca_indexer.py
```diff
@@ -651,10 +652,22 @@ def test_cell_line_sample(self):
...
+                    self.assertEqual(one(contents['specimens'])['organ_part'], ['venous blood'])
```
hannes-ucsc: Identical.

test/indexer/test_hca_indexer.py
```diff
@@ -651,10 +652,22 @@ def test_cell_line_sample(self):
...
+                    self.assertEqual(len(contents['cell_lines']), 2)
```
hannes-ucsc: Similar, RHS is 1 or 2 in either branch.

test/indexer/test_hca_indexer.py
```diff
@@ -651,10 +652,22 @@ def test_cell_line_sample(self):
...
+                    cell_lines_model_organ = {cl['model_organ'] for cl in contents['cell_lines']}
+                    self.assertEqual(cell_lines_model_organ, {'blood (cell_line)'})
+                    self.assertEqual(one(contents['cell_suspensions'])['organ'], ['blood (cell_line)'])
```
hannes-ucsc: Identical

test/indexer/test_hca_indexer.py
```diff
@@ -651,10 +652,22 @@ def test_cell_line_sample(self):
...
+                    self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [None])
```
hannes-ucsc: Identical.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -725,6 +725,8 @@ def make_cell_suspensions(self, entry):
     def make_cell_line(self, cell_line):
         return {
             "id": cell_line["biomaterial_id"],
+            "cellLineType": cell_line.get("cell_line_type", None),
```
hannes-ucsc: I'm surprised that these two additional keys don't require some change in a service test assertion.

test/indexer/test_hca_indexer.py
danielsotirhos: Refactored test case to minimize duplication

src/azul/service/responseobjects/hca_response_v5.py
danielsotirhos: Thought this was covered by existing test cases however turns out they are checking the response from projects with no cell_lines. Added a new test case to fix this missing coverage.

PR #1004, opened 2019-05-15
Renamed temporary subfolder in BDbag to `manifest` (#991)

GPelayo:


src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -293,14 +293,18 @@ def _write_tsv(self, output: IO[str]) -> None:
...
                 self._write_bdbag_samples_tsv(samples_tsv)
-            bag = bdbag_api.make_bag(bag_path, update=True)  # update TSV checksums
-            assert bdbag_api.is_bag(bag_path)
-            bdbag_api.validate_bag(bag_path)
+            bag = bdbag_api.make_bag(manifest_folder, update=True)  # update TSV checksums
+            assert bdbag_api.is_bag(manifest_folder)
+            bdbag_api.validate_bag(manifest_folder)
             assert bdbag_api.check_payload_consistency(bag)
-            return bdbag_api.archive_bag(bag_path, 'zip')
+            new_zip_path = os.path.join(os.path.dirname(bag_path), 'manifest.zip')
```
hannes-ucsc: #991 does not specify this change. What's the reason for changing the ZIP file's name?

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -293,14 +293,18 @@ def _write_tsv(self, output: IO[str]) -> None:
 
     def _create_bdbag_archive(self) -> str:
         with TemporaryDirectory() as bag_path:
-            bdbag_api.make_bag(bag_path)
-            with open(os.path.join(bag_path, 'data', 'samples.tsv'), 'w') as samples_tsv:
+            manifest_folder = os.path.join(bag_path, 'manifest')
```
hannes-ucsc: Rename `bag_path` to `temp_path` and `manifest_folder` to `bag_path`. I want to avoid the overloading the term "manifest" with yet another term.

src/azul/service/responseobjects/hca_response_v5.py
GPelayo: The ZIP file isn't being renamed. I'm using the `os.rename()` to move the ZIP file out of the temporary folder. `bdbag_api.archive_bag()` creates the ZIP file in the temporary directory because `bag_path` refers to a subfolder within the temporary folder. Since the temporary folder is deleted before anything is uploaded into the S3 bucket, I need to move that ZIP file out of the temporary folder.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -292,15 +292,19 @@ def _write_tsv(self, output: IO[str]) -> None:
...
+            os.makedirs(bag_path)
             bdbag_api.make_bag(bag_path)
             with open(os.path.join(bag_path, 'data', 'samples.tsv'), 'w') as samples_tsv:
                 self._write_bdbag_samples_tsv(samples_tsv)
             bag = bdbag_api.make_bag(bag_path, update=True)  # update TSV checksums
             assert bdbag_api.is_bag(bag_path)
             bdbag_api.validate_bag(bag_path)
             assert bdbag_api.check_payload_consistency(bag)
-            return bdbag_api.archive_bag(bag_path, 'zip')
+            _, temp_path = mkstemp()
```
hannes-ucsc: You are leaking a file handle.

```suggestion
            temp, temp_path = mkstemp()
            temp.close()
```

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -292,15 +292,19 @@ def _write_tsv(self, output: IO[str]) -> None:
...
+            os.rename(bdbag_api.archive_bag(bag_path, 'zip'), temp_path)
```
```suggestion
            archive_path = bdbag_api.archive_bag(bag_path, 'zip')
            os.rename(archive_path, temp_path)
```

hannes-ucsc: And add brief comment as to why were moving the file.

test/service/test_request_validation.py
```diff
@@ -282,6 +282,7 @@ def test_bdbag_manifest(self):
             self.assertEqual(200, response.status_code, 'Unable to download manifest')
             with ZipFile(BytesIO(response.content), 'r') as zip_fh:
                 zip_fh.extractall(zip_dir)
+                self.assertTrue(['manifest' == first(name.split(os.path.sep)) for name in zip_fh.namelist()])
```
hannes-ucsc: I don't think the separator in a ZIP entry is OS-dependent. It's always going to be `/`.

hannes-ucsc: And your assertTrue() doesn't work as intended:

```
from unittest import TestCase
TestCase().assertTrue([False])
```
No exception is raised.

src/azul/service/responseobjects/hca_response_v5.py
GPelayo: I don't think that file handle is leaking. mkstemp() returns an (integer, str).

Issue #920, opened 2019-04-18
DRS endpoint returns cloud-native URL; involves Retry-After


mikebaumann:
Revise the HCA Azul DRS endpoint (/ga4gh/dos/v1/dataobjects/{uuid}) to behave as follows:

Perform a DSS GET /files/{uuid} operation:
- If the DSS response code is 301, wait and retry DSS operation.
Monitor Lambda remaining lifetime, and if getting close to expiration, return a 5xx to Martha with a message indicating the file was not yet available from the DSS
- Else if the DSS response code is 302, obtain the gs:// URL from the DSS response and compose a DRS format response body including the gs://URL and respond with code 200.

For more information, see:
[Terra/Cromwell Access to Files Identified by DRS URIs for the HCA MVP](https://docs.google.com/document/d/18WDYr3uS82Eu1GhHmeH6PX7tFY6Zk8piKd3ixOoE8Bg/edit#)

The final implementation of this should obtain the cloud native URLs from the DSS directly using an enhanced version of `GET /files/{uuid}` with `directurls=true` (https://github.com/HumanCellAtlas/data-store/issues/2045), but as a temporary workaround, a cloud native URL could be derived from the presigned URL currently returned by the DSS.

Done when:
The HCA DRS resolver behaves as described above for all HCA deployments.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-571)
┆Project Name: azul
┆Issue Number: AZUL-571
┆Epic: Terra hand-off

hannes-ucsc:
> The final implementation of this should obtain the cloud native URLs from the DSS directly using an enhanced version of GET /files/{uuid} with directurls=true (HumanCellAtlas/data-store#2045), but as a temporary workaround a cloud native URL could be derived from the presigned URL currently returned by the DSS.

We should decide which alternative to implement. If we go for the final implementation, HumanCellAtlas/data-store#2045 should be marked as blocker.

mikebaumann:
Agreed, I previously tried to configure HumanCellAtlas/data-store#2045 as blocker for this issue, yet ZenHub didn't seem to accept it. Perhaps you know the trick to do so?

Kevin originally thought HumanCellAtlas/data-store#2045 could be done this sprint, yet yesterday he said any new work for his team would need to be done next sprint.
[...]

PR #1028, opened 2019-05-20
Return gs url with DRS endpoint (#920)

jessebrennan:


test/service/test_drs.py
```diff
@@ -1,16 +1,28 @@
...
+from azul import drs, config
 from service import WebServiceTestCase
 
 
 class DataRepositoryServiceEndpointTest(WebServiceTestCase):
 
+    def chalice_config(self):
+        return ChaliceConfig.create(lambda_timeout=15)
+
+    @responses.activate
```
hannes-ucsc: While this works here because you're not using moto or other libraries that mock `requests`, I would still want to use `ResponsesHelper` for consistency's sake, to reinforce the pattern and to make future changes to this test easier. It's the same amount of code, just a CM instead of a decorator.

lambdas/service/app.py
```diff
@@ -1611,3 +1616,35 @@ def get_data_object(file_uuid):
...
+        if app.lambda_context.get_remaining_time_in_millis() / 1000 > 3:
+            dss_response = requests.get(url, params=params, allow_redirects=False)
+            if dss_response.status_code == 302:
+                url = dss_response.next.url
+                assert url.startswith('gs')
+                return url
+            elif dss_response.status_code == 301:
+                url = dss_response.next.url
+                remaining_lambda_seconds = app.lambda_context.get_remaining_time_in_millis() / 1000
+                server_side_sleep = min(
```
hannes-ucsc: Rather ugly wrapping IMO

lambdas/service/app.py
```diff
@@ -1611,3 +1616,35 @@ def get_data_object(file_uuid):
...
+                    10, max(remaining_lambda_seconds - config.api_gateway_timeout_padding - 3, 0))
```
mikebaumann: The default minimum sleep of 10 seconds before retrying should be 1 second instead.
Given this endpoint primarily exists for the HCA/Terra handoff MVP, and this is all on Google, and the DSS checkout time for Google is typically very fast (<= 1 second) this endpoint should retry the request very quickly (sleep time of 1 second).
We had discussed this with the Broad and verbally agreed on this as the plan, yet it had not yet been incorporated into our documentation. Sorry about that, I am working on incorporating that now.

lambdas/service/app.py
```diff
@@ -1557,19 +1558,23 @@ def file_to_drs(doc):
     """
     Converts an aggregate file document to a DRS data object response.
     """
-    replicas = ['aws']
+    urls = [
+        file_url(uuid=doc['uuid'],
+                 version=doc['version'],
+                 replica='aws',
```
mikebaumann: Given the intended use of this endpoint, and that are trying to get a `gs://` URL, I think the `replica` value here be changed from `aws` to `gcp`.

PR #1041, opened 2019-05-23
Speed up indexing tests with `refresh=True`

hannes-ucsc:


hannes-ucsc:
Test time is down from 19min to 8min on Gitlab.

Issue #994, opened 2019-05-13
Ensure deletion of bundles during integration test


hannes-ucsc:
As soon as the mock bundle UUIDs are known, the test knows everything it needs to clean up after itself. According to @jessebrennan the test could be improved so that clean-up is more likely to occur in case of test failure. Currently, clean-up does not always occur after certain test failures.

PR #1043, opened 2019-05-23
Ensure integration tests always cleanup bundles (#994)

jessebrennan:


jessebrennan:
@hannes-ucsc it looks like this was rebased after #1028 was merged but
before it was reverted. That's what's causing the failure. I just rebased.
This is now fine to test again.

Issue #1047, opened 2019-05-29
terms aggregated fields in /summary limited to 10 results


danielsotirhos:
The terms aggregated fields `organTypes` and `cellCountSummaries` are being limited to a result set of 10 due to a default size parameter.

From Elasticsearch documentation:
By default, the terms aggregation will return the buckets for the top ten terms ordered by the doc_count. One can change this default behaviour by setting the size parameter.

![Screen Shot 2019-05-29 at 12 39 46 PM](https://user-images.githubusercontent.com/46407464/58586025-0adeba80-820f-11e9-9b42-bf63b09fa01b.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-653)
┆Project Name: azul
┆Issue Number: AZUL-653

PR #1048, opened 2019-05-29
Increase terms aggregation size limit (#1047)

danielsotirhos:


danielsotirhos:
No new test added for this fix. I was able to confirm the 'size' property's effect on the terms aggregation by running tests with a small size value of 2 and verifying the aggregation output was limited to 2 results.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -373,7 +373,7 @@ def transform_summary(
 
         # Add a per_organ aggregate to the ElasticSearch request
         es_search.aggs.bucket(
-            'group_by_organ', 'terms', field='contents.cell_suspensions.organ.keyword'
+            'group_by_organ', 'terms', field='contents.cell_suspensions.organ.keyword', size=99999
```
hannes-ucsc: Can you make this an attribute in `azul.Config`. You'd have to hunt down every `terms` aggregation but I just did a search for `terms` and it came up with 48 matches half of which are obviously not aggregations.

Issue #1038, opened 2019-05-22
Summary totalCellCount calculation incorrectly inflated


danielsotirhos:
The `totalCellCount` calculated by the '/summary' response is reported incorrectly if a project has a cell suspension that has more than 1 input sample.

![cell_counts](https://user-images.githubusercontent.com/46407464/58214708-a15d2a00-7cab-11e9-8a4e-9893d695d900.jpg)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-649)
┆Project Name: azul
┆Issue Number: AZUL-649

hannes-ucsc:
The fix is to not use the `samples` index to compute the cell count summary but instead use an index in which each document represents exactly one cell suspension. The samples index is not suitable for aggregating cell suspensions because one cell suspension could occur in more than one document therefore inflating the sums. 

Adding a new index should not be that hard, since we practiced it for bundles and samples. Especially bundles straight forward to add.

Note that the recent change of making `contents.cell_suspensions.organ` an array potentially holding more than one element is another cause for inflated cell counts due to how ES bucket aggregations work. We should consider using a [value script](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-script). The script would concatenate the array elements into a single string using a separator that does not naturally occur in organs.

hannes-ucsc:
@NoopDog, we didn't create a browser bug for this but you should be aware of it since users may notice.

PR #1051, opened 2019-05-30
Add Cell Suspension index (#1038)

danielsotirhos:


src/azul/project/hca/transformers.py
```diff
@@ -437,6 +437,44 @@ def transform(self,
...
+        samples: MutableMapping[str, Sample] = dict()
+        for file in bundle.files.values():
+            SampleTransformer.get_ancestor_samples(file, samples)
+        for cell_suspension in bundle.biomaterials.values():
+            if not isinstance(cell_suspension, api.CellSuspension):
+                continue
+            visitor = TransformerVisitor()
+            cell_suspension.accept(visitor)
+            cell_suspension.ancestors(visitor)
+            contents = dict(samples=[_sample_dict(s) for s in samples.values()],
```
hannes-ucsc: We don't want to associate every cell suspension with every sample. Each CS should only be associated with those samples that are reachable from it via the graph.

src/azul/project/hca/indexer.py
```diff
@@ -56,7 +56,7 @@ def mapping(self) -> JSON:
         }
 
     def transformers(self) -> Iterable[Transformer]:
-        return FileTransformer(), SampleTransformer(), ProjectTransformer(), BundleTransformer()
+        return FileTransformer(), CellSuspensionTransformer(), SampleTransformer(), ProjectTransformer(), BundleTransformer()
 
     def entities(self) -> Iterable[str]:
-        return ['files', 'samples', 'projects', 'bundles']
+        return ['files', 'cellsuspensions', 'samples', 'projects', 'bundles']
```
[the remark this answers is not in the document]
danielsotirhos: Unfortunately the Config function es_index_name() uses underscore as a separator character when the "full" index name is constructed (eg. "azul_cellsuspension_aggregate_dev")
```
return f"{self._index_prefix}_{entity_type}{'_aggregate' if aggregate else ''}_{self.deployment_stage}"
```
and underscore is used to split this string in the Config function es_index_name()
```
index_name = index_name.split('_')
```

src/azul/project/hca/transformers.py
danielsotirhos: Ah yes, this makes sense.

Issue #1022, opened 2019-05-16
Expose effective organ of sample in response and as facet


hannes-ucsc:
Expose `content.samples.effective_organ` in service response under both `hits[*].samples[*]` and `termFacets` of any `/repository/{projects|samples|files|bundles}` response. This will allow users to search for entities that associated with organ X OR modelOrgan X, something the user can't currently do.

We planned to do this in #977 (see first bullet in the description) but forgot to implement that aspect of it.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-638)
┆Project Name: azul
┆Issue Number: AZUL-638

PR #1063, opened 2019-06-06
Add effectiveOrgan facet (#1022)

danielsotirhos:


test/service/test_response.py
```diff
@@ -1225,13 +1225,12 @@ def test_sample(self):
...
                             sample_entity_type = sample['sampleEntityType']
                             for key, val in sample.items():
-                                if key in ['sampleEntityType','effectiveOrgan']:
-                                    continue
-                                if isinstance(val, list):
-                                    for one_val in val:
-                                        self.assertIn(one_val, hit[sample_entity_type][0][key])
-                                else:
-                                    self.assertIn(val, hit[sample_entity_type][0][key])
+                                if key not in ['sampleEntityType', 'effectiveOrgan']:
```
hannes-ucsc: I'd default to using a tuple. It may not make a significant difference performance-wise but it emphasizes the immutability of the literal value.

Issue #537, opened 2018-11-08
Convert `filters` parameter from Python literal to JSON


hannes-ucsc:
We currently use `ast.literal_eval()` to parse the `filters` request parameter. This is a constant cause for confusion, because Python literals resemble JSON but are nevertheless different: `True` is `true`, `None` is `null` and

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-346)
┆Project Name: azul
┆Issue Number: AZUL-346

Issue #279, opened 2018-08-31
Remove the `file` key from the `filters` parameter


hannes-ucsc:
Filters are currently passed to the endpoints that accept filters using the following structure:

```
{'file':{'organPart':{'is':['foo']},'otherFacet':{'is':['bar']}}}
```

The `file` indirection is redundant and was probably introduced to ensure backwards compatibility to existing UIs. These lines gives that away:

https://github.com/DataBiosphere/azul/blob/01c5fc89bce32e672af63add58b7b584f7fb9cd2/src/azul/service/responseobjects/elastic_request_builder.py#L334

https://github.com/DataBiosphere/azul/blob/01c5fc89bce32e672af63add58b7b584f7fb9cd2/src/azul/service/responseobjects/elastic_request_builder.py#L438

The filters parameter should be structured without the `file` indirection

```
{'organPart':{'is':['foo']},'otherFacet':{'is':['bar']}}
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-184)
┆Project Name: azul
┆Issue Number: AZUL-184

PR #1065, opened 2019-06-06
Use JSON for filters (#537) (#279)

danielsotirhos:
- Convert filters parameter from Python literal to JSON https://github.com/databiosphere/azul/issues/537
- Remove 'file' key from the 'filters' parameter https://github.com/databiosphere/azul/issues/279

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -469,9 +469,9 @@ def get_manifest():
           https://software.broadinstitute.org/firecloud/documentation/article?id=10954
     :return: A manifest that the user can use to download the files in there
     """
-    params = app.current_request.query_params
-    if params is None:  # FIXME: is this necessary?
+    if app.current_request.query_params is None:
         app.current_request.query_params = {}
+    params = app.current_request.query_params
     from azul.service import AbstractService
     filters = AbstractService.parse_filters(params.get('filters'))
     es_td = EsTd()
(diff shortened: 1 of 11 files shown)


test/local_integration_test.py
```diff
@@ -313,12 +313,12 @@ def _get_entities_by_project(self, entity_type, project_shortname):
...
         """
-        filters = {'file': {'project': {'is': [project_shortname]}}}
+        filters = '{"project": {"is": ["' + project_shortname + '"]}}'
         entities = []
         size = 100
-        params = dict(filters=str(filters), size=str(size))
+        params = dict(filters=filters, size=str(size))
         while True:
-            query = urllib.parse.urlencode(params, safe="{}/'")
+            query = urllib.parse.urlencode(params, safe='{}/"')
```
hannes-ucsc: Both, `'` and `"` should be considered safe WRT URL-encoding URL params.

scripts/count-bundles.py
```diff
@@ -21,7 +21,7 @@ def get_project_name(document_id):
 def get_project_bundle_count(project_name):
     page_size = 1000
 
-    filter_dict = urllib.parse.quote('{"file":{"project":{"is":["' + project_name + '"]}}}')
+    filter_dict = urllib.parse.quote('{"project":{"is":["' + project_name + '"]}}')
```
hannes-ucsc: What if `project_name` contained `"`? Use json.dumps here as well.

lambdas/service/app.py
```diff
@@ -1601,10 +1601,8 @@ def get_data_object(file_uuid):
     params = app.current_request.query_params or {}
     file_version = params.get('version')
     filters = {
-        "file": {
-            "fileId": {"is": [file_uuid]},
-            **({"fileVersion": {"is": [file_version]}} if file_version else {})
-        }
+        "fileId": {"is": [file_uuid]},
+        **({"fileVersion": {"is": [file_version]}} if file_version else {})
```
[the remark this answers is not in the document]
danielsotirhos: I think as written this meets style guide specification for "wrap all": dictionary `filters` has 2 possible keys `fileId` and optional `fileVersion`, and these keys are each on a new line

PR #1067, opened 2019-06-07
Isolate test fixtures for service tests (#1064)

### LINKED ISSUE #1064: Manifest tests contaminate each other's index
The manifest tests in `test_request_validation.py` don't clean up the bundles they index which can lead to failure. They testing infrastructure should be cleaned up in such a way that tests only have what they need and clean up when they're done.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-662)
┆Project Name: azul
┆Issue Number: AZUL-662

jessebrennan:
Now test classes that inherit from WebServiceTestCase all need to
explicitly provision the documents that they need to be indexed, instead
of just always indexing a document in WebServiceTestCase.

Also TestManifestContent class was created which now contains all of
the tests for the Manifest's content. Some of these tests were
previously located in test_request_validation. Moving them to their own
class is what ultimately fixes #1064 since they are now correctly
provisioning and de-provisioning the fixtures between tests.


test/indexer/test_hca_indexer.py
```diff
@@ -754,7 +754,7 @@ def test_metadata_field_exclusion(self):
         self.assertEqual(1, hits["hits"]["total"])
 
 
-class TestValidNotificationRequests(LocalAppTestCase):
+class TestValidNotificationRequests(WebServiceTestCase):
```
hannes-ucsc: Why?

test/service/__init__.py
```diff
@@ -21,16 +21,19 @@ def lambda_name(cls) -> str:
...
-        super().setUpClass()
-        for bundle in cls.bundles:
-            cls._index_canned_bundle(bundle)
+    def tearDownClass(cls):
+        cls.clear_indices()
+        super().tearDownClass()
 
     @classmethod
-    def tearDownClass(cls):
+    def clear_indices(cls):
```
hannes-ucsc:
```suggestion
    def _teardown_indices(cls):
```

test/service/__init__.py
```diff
@@ -21,16 +21,19 @@ def lambda_name(cls) -> str:
...
         for index_name in cls.get_hca_indexer().index_names():
             cls.es_client.indices.delete(index=index_name, ignore=[400, 404])
-        super().tearDownClass()
+
+    @classmethod
+    def setup_class_bundles(cls):
```
hannes-ucsc:
```suggestion
    def _setup_indices(cls):
```

test/service/__init__.py
```diff
@@ -21,16 +21,19 @@ def lambda_name(cls) -> str:
         return "service"
 
     @classmethod
-    def setUpClass(cls):
-        super().setUpClass()
-        for bundle in cls.bundles:
-            cls._index_canned_bundle(bundle)
+    def tearDownClass(cls):
```
hannes-ucsc: I understand that `clear_indices()` is idempotent but the asymmetry makes my head hurt. How many `tearDownClass` overrides do we save by doing this? And do those savings outweigh the potential confusion of seeing the effects of two consecutive invocations of `clear_indices()`. The pattern is "What you create in setUp, you destroy in tearDown. That symmetry should be maintained across all tests. Following that pattern slavishly adds sanity and predictability, if at the expense of mild increase in verbosity.

test/indexer/test_hca_indexer.py
jessebrennan: I think this was accidental. I'll revert and see if there aren't any problems.

test/service/__init__.py
jessebrennan: I expected some naming suggestions here ;)

test/service/__init__.py
jessebrennan: Changed. I agree, this is more clear.

test/service/__init__.py
jessebrennan: I just counted 7 places where tearDownClass would need to be overridden but isn't because tearDownClass is is a part of WebServiceTestCase.

jessebrennan: To me that justifies keeping it there. What do you think?

hannes-ucsc: To me it doesn't. If suddenly that number dropped to five would we then remove the method? Three? We probably wouldn't because we'd simply forget. The surprising side-effect of performing the index deletion twice would remain. Also, say we left `tearDownClass()` here and had to override it in a subclass for some reason, would we rely on `super().tearDownClass()` calling `_teardown_indices()` or would we call it explicitly? In a `setupClass()` override we couldn't rely on `super.setupClass()` to call `_setup_indices()` for us. So we'd have to be asymmetric to be efficient (indices deleted once) or symmetric and inefficient (one redundant index deletion). I prefer to have symmetry and efficiency at the expense of some code duplication.

test/service/__init__.py
```diff
@@ -21,16 +21,19 @@ def lambda_name(cls) -> str:
...
 
     @classmethod
-    def tearDownClass(cls):
+    def _teardown_indices(cls):
         for index_name in cls.get_hca_indexer().index_names():
             cls.es_client.indices.delete(index=index_name, ignore=[400, 404])
-        super().tearDownClass()
+
+    @classmethod
+    def _setup_indices(cls):
```
hannes-ucsc: `_setup_indices` should precede `_teardown_indices`.

test/service/__init__.py
jessebrennan: OK. I'm convinced.

jessebrennan: Slightly embarrassed to admit it, but it probably doesn't need to be duplicated at all.. it looks like it can just be removed.

jessebrennan:
@hannes-ucsc I force pushed so now there are just two commits that contain slightly different implementations. Please look at both and if one is to your preference, let's go with that one. Otherwise I will squash the commits down and make whatever review changes you suggest.

jessebrennan:
@hannes-ucsc sorry about the force pushes and double implementations. I know it's not standard procedure and I should have been more clear with what my intentions were. I rebased and reset the branch to B' so I think we are on the same page and this should be good to go (once tests pass).

Issue #1084, opened 2019-06-13
/health/progress is slow


hannes-ucsc:
The computation of the top-level `up` element of the response defeats the lazy approach because it eagerly computes `up` for nested keys.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-677)
┆Project Name: azul
┆Issue Number: AZUL-677

hannes-ucsc:
Also see https://github.com/DataBiosphere/azul/issues/1085

hannes-ucsc:
I can't demo the speed-up because there is no environment available with the previous, slow implementation. Dev, integration and staging currently have the fix while `prod` does not have /health/progress at all yet.

PR #1087, opened 2019-06-13
Fix: /health/progress is not lazy and therefore slow (#1084) 

hannes-ucsc:
The added test is expected to fail.


hannes-ucsc:
Integration tests are covered by running #1088 (which is based on this one) in the sandbox.

Issue #1085, opened 2019-06-13
/health often times out with 504 in API gateway


hannes-ucsc:
Since `/health` now recursively hits several API endpoints it has a tendency to timeout, especially with an ongoing reindexing in any deployment that the current deployment shares and ES domain with.

Consider parallelizing the recursive API endpoint requests.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-678)
┆Project Name: azul
┆Issue Number: AZUL-678

hannes-ucsc:
Demo is blocked on https://github.com/DataBiosphere/azul/pull/1143 landing.

hannes-ucsc:
Demoed 7/29

PR #1088, opened 2019-06-13
Parallelize API endpoint health checks (#1085)

hannes-ucsc:


hannes-ucsc:
Integration tests are covered by running #1088 (which is based on this one) in the sandbox.

Issue #512, opened 2018-10-30
Expose min and max age range on service side


hannes-ucsc:
Need to make sure that min and max aggregation works as desired and that the service exposes the minimum and maximum age range fields.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-332)
┆Project Name: azul
┆Issue Number: AZUL-332

theathorn:
@hannes-ucsc : From @NoopDog : Please combine age and age unit in the index. I started on this but don’t want to over specify it. The representation needs to support searching by one or more age ranges.

NoopDog:
@GPelayo Thanks, can you add examples to the ticket on how this changes the API? Also is this a breaking change or just an extra field?

GPelayo:
@NoopDog The change only exposes the age range field in the indexer, so it shouldn't be breaking change.

GPelayo:
These `donorOrganisms. organismAgeRange` is now exposed as a double range datatype. Double range types have the subfields `gte` (greater than or equal) and `lte` (lesser than or equal) representing min and max ages, respectively. Also, the ages are only in seconds.

```
"organismAgeRange": [
        {
            "gte": 1734480000.0,
            "lte": 1860624000.0
        },
        {
            "gte": 1261440000.0,
            "lte": 1387584000.0
        }
],
```

This change also makes it possible to find documents that have number values that fall within a certain range.

For example, if you want to find entities with a `donorAgeRange` that falls between 6-14, then the filters param needs to be:
```
filters={“donorAgeRange”: {“within”: [[6, 14]]}}
```
`donorAgeRange` is the field name that is queried. 
`within` is the comparison operator (Explained in-depth below)
`[[6, 14]]` is the desired range. When querying number ranges, the value is always an array of number ranges arrays. In this case, `[[6, 14]]` means  a `donorAgeRange` between 6-14.

In the place of the `is` comparison modifier, there are also new comparison operators specifically for querying number ranges.

The `within` keyword will return results with a number range field that is within the queried range. For example, a user has `within` in their query to find donors within the age range 5-10. Then, Azul will return donors with ages between 7-8 or 6-9, since these age ranges are within the queried age range of 5-10.

The `contain` keyword is for returning results with a number range field that could contain the queried range. For example, a user has `contain` in their query to find donors containing the age range 5-10 years. Azul will then return donors with age ranges 4-11 or 1-18, since these age ranges could contain the queried age range 5-10.

[...]

PR #1107, opened 2019-07-02
Expose min and max ages (#512)

GPelayo:


test/service/test_response.py
```diff
@@ -1261,6 +1270,230 @@ def test_bundles_outer_entity(self):
         self.assertEqual(len(response['hits']), len(hits_uuids))
         self.assertSetEqual(indexed_uuids, hits_uuids)
 
+    expected_response = []
+
+    def test_ranged_values(self):
+        test_hits = [{'cellLines': [{'cellLineType': ['induced pluripotent'],
```
danielsotirhos: The formatting of this data structure seems a bit random, maybe dump it into an online formatter?

test/service/test_response.py
```diff
@@ -90,6 +92,7 @@ def test_key_search_files_response(self):
                             "id": ["DID_scRSq06"],
                             "organismAge": ["38"],
                             "organismAgeUnit": ["year"],
+                            "organismAgeRange": [{"gte": 1198368000.0, "lte": 1198368000.0}]
```
danielsotirhos: Do we have a test case that uses a range of values (min != max)?

test/service/test_response.py
achave11: @GPelayo ensure that we have consistency in the structure as do the other test in this series.

test/service/test_response.py
GPelayo: Yes, test.response.TestResponse.test_ranged_values() test scenarios where  organismAgeRange's min and max are different values.

src/azul/project/hca/indexer.py
```diff
@@ -11,6 +11,14 @@ class Indexer(BaseIndexer):
     def mapping(self) -> JSON:
         return {
             "dynamic_templates": [
+                {
+                    "donor_age_range": {
+                        "path_match": "contents.donors.organism_age_range_in_seconds",
+                        "mapping": {
+                            "type": "double_range"
```
hannes-ucsc: Can you document the reasoning behind our choice of `double` instead of `float` in a comment here?

src/azul/project/hca/transformers.py
```diff
@@ -182,8 +183,10 @@ def _donor(self, donor: api.DonorOrganism) -> JSON:
             'organism_age_unit': donor.organism_age_unit,
             **(
                 {
-                    'min_organism_age_in_seconds': donor.organism_age_in_seconds.min,
-                    'max_organism_age_in_seconds': donor.organism_age_in_seconds.max,
+                    'organism_age_range_in_seconds': {
```
hannes-ucsc:
```suggestion
                    'organism_age_range': {
```

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -83,13 +84,21 @@ def create_query(filters):
...
         filter_list = []
         for facet, values in filters.items():
-            value = values.get('is', {})
-            if value is None:
-                # If filter is {"is": None}, search for values where field does not exist
-                f = Q('bool', must_not=Q('exists', field=f'{facet}.keyword'))
-            else:
-                f = Q('terms', **{f'{facet.replace(".", "__")}__keyword': value})
-            filter_list.append(f)
+            relation, value = one(values.items())
```
hannes-ucsc: +1

test/service/test_response.py
```diff
@@ -1261,6 +1270,98 @@ def test_bundles_outer_entity(self):
...
+
+            with self.subTest(relation=relation, value=range_value):
+                params = {
+                    'filters': json.dumps({'organismAgeRange': {relation: [range_value]}}),
+                    'order': 'desc',
+                    'sort': 'entryId'
+                }
+            response = requests.get(url, params=params)
+            actual_value = [hit['donorOrganisms'] for hit in response.json()['hits']]
+            self.assertElasticsearchResultsEqual(expected_hits, actual_value)
```
hannes-ucsc: These three lines need to be indented so they are part of the body of the subTest context manager.

src/azul/project/hca/indexer.py
hannes-ucsc: The "float with 2 decimal precision" phrase is misleading and irrelevant. Please change to:

> A `float` (single precision IEEE-754) can represent all integers up to 16,777,216. If we used `float` values for organism ages in seconds, we would not be able to accurately represent an organism age of 16,777,217 seconds. That is 194 days and 15617 seconds. A double precision IEEE-754 representation loses accuracy at 9,007,199,254,740,993 which is more than 285616415 years.
>
> Note that Python's `float` uses double precision IEEE-754. 
>
> (https://docs.python.org/3/tutorial/floatingpoint.html#representation-error)

Issue #1068, opened 2019-06-07
Add `file_uuid` and `file_version` columns to full metadata TSV


hannes-ucsc:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-664)
┆Project Name: azul
┆Issue Number: AZUL-664

hannes-ucsc:
See https://github.com/HumanCellAtlas/hca_bundles_to_csv/issues/12

theathorn:
Automation is not required for General Availability. Removing from epic.

PR #1114, opened 2019-07-03
Add `file_uuid` and `file_version` columns to full metadata TSV (#1068)

achave11:


src/azul/project/hca/metadata_generator.py
```diff
@@ -108,17 +114,21 @@ def _get_schema_name_from_object(obj: JSON):
...
                    bundle_uuid: str,
                    bundle_version: str,
+                   manifest,
                    metadata_files: List[JSON]) -> None:
 
-        file_uuids = self._get_file_uuids_from_objects(metadata_files)
+        file_info = self._get_file_uuids_from_objects(manifest, metadata_files)
 
-        for file, content in file_uuids.items():
+        for file, content in file_info.items():
```
danielsotirhos: If the key isn't needed then no need to fetch the key
```suggestion
        for content in file_info.values():
```

src/azul/project/hca/metadata_generator.py
```diff
@@ -51,20 +51,26 @@ def __init__(self, order=None, ignore=None, format_filter=None):
...
                 raise MissingSchemaTypeError("JSON objects must declare a schema type")
 
-            if _object["schema_type"] == "file" and self._deep_get(_object, ["provenance", "document_id"]):
-                file_uuid = self._deep_get(_object, ["provenance", "document_id"])
-                file_uuids[file_uuid] = _object
+            if _object["schema_type"] == "file":
+                file_name = self._deep_get(_object, ["file_core", "file_name"])
+                if file_name is None:
+                    raise MissingFileNameError("expecting file_core.file_name")
+                file_manifest = file_manifests[file_name]
```
GPelayo: Apparently, there are some bundles in production with sequence_files JSONs reffering to files that are not in the manifest. This causes a `KeyError` to be thrown if the generator encounters any of these bundles.
Surround this line with try-except like the following.
```suggestion
                try:
                    file_manifest = file_manifests[file_name]
                except KeyError:
                    continue
```
I'm doing this change too in my own version of the generator.

src/azul/project/hca/metadata_generator.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, order=None, ignore=None, format_filter=None):
             "^analysis_process.*",
             "^process.*",
             "^bundle_.*",
-            "^\\*\\.provenance\\.update_date"
```
hannes-ucsc: This deletion seems like an unrelated change. What does it do? And shouldn't it be committed separately?

src/azul/project/hca/metadata_generator.py
```diff
@@ -51,20 +51,29 @@ def __init__(self, order=None, ignore=None, format_filter=None):
...
-                file_uuid = self._deep_get(_object, ["provenance", "document_id"])
-                file_uuids[file_uuid] = _object
+            if _object["schema_type"] == "file":
+                file_name = self._deep_get(_object, ["file_core", "file_name"])
+                if file_name is None:
+                    raise MissingFileNameError("expecting file_core.file_name")
+                try:
+                    file_manifest = file_manifests[file_name]
+                except KeyError:
+                    continue
```
hannes-ucsc:
```suggestion
                    pass
```

src/azul/project/hca/metadata_generator.py
```diff
@@ -51,20 +51,29 @@ def __init__(self, order=None, ignore=None, format_filter=None):
...
+                file_info[file_manifest['uuid']] = {'metadata': _object, 'manifest': file_manifest}
```
hannes-ucsc:
```suggestion
                else:
                    file_info[file_manifest['uuid']] = {'metadata': _object, 'manifest': file_manifest}
```

src/azul/project/hca/metadata_generator.py
achave11: Hey @GPelayo, I could not remember the explanation you gave me for removing this field. Could you elaborate to @hannes-ucsc on the reason for the removal? Thank you in advance.

achave11:
@hannes-ucsc so the new fields are not explicitly tested in the `test_full_metadata`, but they are tested literally on the `test_indexing` when a suit of expected results is loaded for comparison. 
Adding them to the `test_full_metadata` suit will add an extra layer of complexity parsing through the hits in order to obtain the specific `file_uuid` and `file_version` and compare them with the expected ones.

hannes-ucsc:
@achave11, 

> @hannes-ucsc so the new fields are not explicitly tested in the test_full_metadata, but they are tested literally on the test_indexing when a suit of expected results is loaded for comparison.

test_indexing tests the indexing, which is only half way to the final TSV.

> Adding them to the test_full_metadata suit will add an extra layer of complexity parsing through the hits in order to obtain the specific file_uuid and file_version and compare them with the expected ones.

That's a problem you need to solve. If you claim that your changes add two columns to the full TSV, then you will need to prove in a test that those two columns are actually part of the full TSV.

Why is `test_full_metadata` currently passing? You didn't add those two columns to the assertions in that test so I would expect it to fail, complaining about two extra columns. Why does it not do that?

hannes-ucsc:
One more thing:

> > Adding them to the test_full_metadata suit will add an extra layer of complexity parsing through the hits in order to obtain the specific file_uuid and file_version and compare them with the expected ones.

> That's a problem you need to solve. If you claim that your changes add two columns to the full TSV, then you will need to prove in a test that those two columns are actually part of the full TSV.

I don't think it is that complicated. The version and UUID will always be the same since the bundle is canned and the UUIDs and versions of the files in it are constant.

However, `test_full_metadata` was passing in this PR without you having modified its assertions. That still makes me wonder if there is a bug hiding somewhere.

achave11:
@hannes-ucsc checkout https://api.travis-ci.org/v3/job/561247671/log.txt shows `test_full_metadata` not passing. 
I will add a check for the two new fields in the mentioned test.

hannes-ucsc:
> @hannes-ucsc checkout https://api.travis-ci.org/v3/job/561247671/log.txt shows test_full_metadata not passing.

Yes, I noticed that earlier and edited my comment.

> I notice indexing did not have the full implementation of using the manifest for fetching file data. Which skew the test results.

So in the absence of "the full implementation" the generation of the two columns is silently skipped? We should fail eagerly and not be lenient. Does that leniency originate in the `except KeyError: continue` section that I commented on  a few days ago?

src/azul/project/hca/metadata_generator.py
GPelayo: @hannes-ucsc  The `*.provenance.update_date` column was added in the full metadata TSV because originally  I thought that field was synonymous with the `file_version`. Since we know now that the `*.provenance.update_date` is not the `file_version` and, we are extracting the file_version from `bundle.json`, I advised @achave11 to remove it.

src/azul/project/hca/metadata_generator.py
```diff
@@ -51,20 +51,30 @@ def __init__(self, order=None, ignore=None, format_filter=None):
...
-                file_uuid = self._deep_get(_object, ["provenance", "document_id"])
-                file_uuids[file_uuid] = _object
+            if _object["schema_type"] == "file":
+                file_name = self._deep_get(_object, ["file_core", "file_name"])
+                if file_name is None:
+                    raise MissingFileNameError("expecting file_core.file_name")
+                try:
+                    file_manifest = file_manifests[file_name]
+                except KeyError:
+                    pass
```
hannes-ucsc: Isn't this the undesirable leniency that caused `test_full_metadata` to pass? I think this should raise if the manifest entry can't be found.

src/azul/project/hca/metadata_generator.py
hannes-ucsc: Got it. Thank you, @GPelayo

src/azul/project/hca/metadata_generator.py
```diff
@@ -51,20 +51,30 @@ def __init__(self, order=None, ignore=None, format_filter=None):
...
-            if _object["schema_type"] == "file" and self._deep_get(_object, ["provenance", "document_id"]):
-                file_uuid = self._deep_get(_object, ["provenance", "document_id"])
-                file_uuids[file_uuid] = _object
+            if _object["schema_type"] == "file":
+                file_name = self._deep_get(_object, ["file_core", "file_name"])
+                if file_name is None:
+                    raise MissingFileNameError("expecting file_core.file_name")
+                try:
+                    file_manifest = file_manifests[file_name]
+                except KeyError as error:
```
hannes-ucsc: There is no point in catching the error and then only re-raising it.

hannes-ucsc: At this point you can simply remove the entire surrounding try - except - else statements and only leave the bodies of the try and else statements in place.

Issue #1021, opened 2019-05-16
Eliminate `hits[*].projectSummary` element from `/projects` response


hannes-ucsc:
All but `donorCount` can be gotten from other elements of the same hit.

TODO: fill in details, add ticket for replacing `donorCount`.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-637)
┆Project Name: azul
┆Issue Number: AZUL-637

danielsotirhos:
Where to find values previously in /repository/projects `hit['projectSummary']`

```
/repository/projects

hit['projectSummary']['donorCount']         => (length of array) len(hit['donorOrganisms'][0]['id'])
hit['projectSummary']['totalCellCount']     => hit['cellSuspensions'][0]['totalCells']
hit['projectSummary']['genusSpecies']       => hit['donorOrganisms'][0]['genusSpecies']
hit['projectSummary']['libraryConstructionApproach'] => hit['protocols'][0]['libraryConstructionApproach']
hit['projectSummary']['disease']            => hit['specimens'][0]['disease']
hit['projectSummary']['organTypes']         => hit['samples'][0]['effectiveOrgan']
hit['projectSummary']['cellCountSummaries'] => (/repository/summary) hit['cellCountSummaries']
```

NoopDog:
The updates for this for data browser are deployed through to staging. Please feel free to deploy this update to dev and we can validate that all still is ok. Thanks!

hannes-ucsc:
Demoed on 7/22

PR #1118, opened 2019-07-05
Remove projectSummary from response (#1021)

danielsotirhos:


test/service/test_repository_projects.py
```diff
@@ -53,7 +53,6 @@ def assert_file_type_summaries(hit):
             assert_file_type_summaries(hit)
             self.assertTrue('projects' in hit)
             self.assertTrue('specimens' in hit)
-            self.assertTrue('projectSummary' in hit)
```
jessebrennan: Why not make assertFalse instead, to test that it is truly removed?

test/service/test_repository_projects.py
```diff
@@ -82,7 +81,6 @@ def _test_detail_response(self, uuid):
         self.assertTrue('entryId' in hit)
         self.assertTrue('projects' in hit)
         self.assertTrue('specimens' in hit)
-        self.assertTrue('projectSummary' in hit)
```
jessebrennan: Same as above.

test/service/test_repository_specimen.py
```diff
@@ -46,7 +46,6 @@ def assert_file_type_summaries(hit):
             self.assertTrue('donorOrganisms' in hit)
             self.assertTrue('organoids' in hit)
             self.assertTrue('cellSuspensions' in hit)
-            self.assertFalse('projectSummary' in hit)
```
jessebrennan: Is there any harm in leaving this line in? Maybe a comment could explain that it's been deprecated.

NoopDog:
Please deploy to dev.

test/service/test_repository_projects.py
hannes-ucsc: there is also `assertIn` which would yield a better failure message.

hannes-ucsc: And yes, inverting rather than removing makes sense. So use `assertNotIn`

danielsotirhos: Implemented check for non-existence of the 'projectSummary' object with `assertNotIn`

Issue #612, opened 2018-12-13
Group fields should be simple values, not singleton lists


hannes-ucsc:
<img width="525" alt="screen shot 2018-12-12 at 10 28 37 pm" src="https://user-images.githubusercontent.com/5143256/49919974-5bc74600-fe5d-11e8-90ff-8327c70f1464.png">

This can be 

```
file_format: "fastq.gz"
```

there is no reason a group field should be a list.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-383)
┆Project Name: azul
┆Issue Number: AZUL-383

hannes-ucsc:
Demoed on 7/22

PR #1120, opened 2019-07-09
Make `file_format` a single value (#612)

danielsotirhos:


danielsotirhos:
@NoopDog FYI: Internal change only, no change to `fileType` property in response given to browser which already was a single string value.

PR #1133, opened 2019-07-13
Idempotent DSS subscription by validating with hmac_secret_key 

achave11:
This change is not verifiable until the DSS promotes dev onto staging.

### DIFF
--- a/src/azul/subscription.py
+++ b/src/azul/subscription.py
@@ -21,8 +21,7 @@ def manage_subscriptions(dss_client, subscribe=True):
         new_subscriptions = [freeze(dict(replica='aws',
                                          es_query=query,
                                          callback_url=base_url + path,
-                                         hmac_key_id=key_id,
-                                         hmac_secret_key=key))
+                                         hmac_key_id=key_id))
                              for query, path in [(plugin.dss_subscription_query(prefix), '/'),
                                                  (plugin.dss_deletion_subscription_query(prefix), '/delete')]]
     else:
@@ -46,7 +45,7 @@ def manage_subscriptions(dss_client, subscribe=True):
 
     for subscription in new_subscriptions:
         subscription = thaw(subscription)
-        response = dss_client.put_subscription(**subscription)
+        response = dss_client.put_subscription(**subscription, hmac_secret_key=key)
         subscription['uuid'] = response['uuid']
         logging.info('Registered subscription %r.', subscription)
 


hannes-ucsc:
Hey @amarjandu, what's the best way for us to determine when your `hmac_key_id` fixes land on integration, staging and prod?

amarjandu:
@hannes-ucsc `git branch -r --contains {commit-hash}` should do the job
```git branch -r --contains b539e6084c8e71be102c9a90e654277d581c88db
  origin/HEAD -> origin/master
  origin/amar-opsCheckDel
  origin/amar-subv2Sec
  **origin/integration
  origin/master
  origin/staging**
  origin/tsmith-authz
  origin/xbrianh-foo
  origin/xbrianh-test
```
next time we promote to prod it should be there. I let Abraham know that we percolated it up to staging, i can ping you when this occurs.

hannes-ucsc:
@amarjandu, I'd rather not repeatedly poll the repository to wait for the change to appear. We rotate the release engineer (RE) role which means each of us would have to maintain a DSS clone and poll it that way. At some point I proposed (as part of a RFC review, cc: @brianraymor) that we add Zenhub pipelines for each of the deployments but I think that RFC was abandoned. But I also recognize that asking the DSS team to notify us is not a solution either, as your RE on duty would have to remember to do so, which I think is too unreliable.

Your example output lists the staging the branch but the example command line does not mention the commit hash. Does that mean that the DSS change we depend on is already on staging or did you just run the command with a different hash?

achave11:
@hannes-ucsc I should have mention that @amarjandu did notify me about the changes in staging on Wednesday. That is how I was able to expedite this ticket and verify it against current subscriptions.

hannes-ucsc:
@achave11, @amarjandu I see.

I'm starting to realize that we may want to prefer a mention (`@achave`) on a channel (`#azul-dev`) over a direct message.

amarjandu:
so strange, i placed the hash in the markdown section, but it did not get presented into the comment, 
Screen Shot 2019-07-19 at 10 17 36 AM

`git branch -r --contains b539e6084c8e71be102c9a90e654277d581c88db`

I would not be opposed to creating a ZH ticket (or GH issue) to inform other services about commit propagation, we also place notes about commits and releases inside Allsparks Release Notes, these are checked by the RE, and can be used to ensure notifications/pings are sent out.

hannes-ucsc:
@achave11, note the [commit title](https://github.com/DataBiosphere/azul/commit/438dade5b6b674afcf1e2b36c38863c9104ae0ae) change.

hannes-ucsc:
https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/2370

Nicely, done. @achave11

Issue #840, opened 2019-03-26
Ensure that only valid URL parameters are passed


hannes-ucsc:
I just wasted 30min diagnosing a test failure because I mistyped `&filter=` instead of `&filters=`. The service return a 400 if the client passes unknown URL parameters.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-520)
┆Project Name: azul
┆Issue Number: AZUL-520

PR #1135, opened 2019-07-16
Validate request parameters in URL (#840)

achave11:


lambdas/service/app.py
```diff
@@ -139,8 +139,39 @@ def version():
...
+        'search_before': str,
+        'search_before_uid': str,
+        'size': int,
+        'sort': str,
+        **kwargs
+    }
+    validate_params(incoming_params, **valid_params)
+
+
+def validate_params(query_params, allow_extra_params=False, **kwargs):
```
jessebrennan: At first I was going to complain about using `kwags`, but I think that the usage looks better than passing in a dict. Nice :+1:

jessebrennan: On the other hand, a docstring _could_ be helpful here.

lambdas/service/app.py
```diff
@@ -741,6 +775,7 @@ def fetch_dss_files(uuid):
 
 def _dss_files(uuid, fetch=True):
     query_params = app.current_request.query_params
+    validate_params(query_params, True, fileName=str, wait=int, requestIndex=int)
     url = config.dss_endpoint + '/files/' + urllib.parse.quote(uuid, safe='')
     file_name = query_params.pop('fileName', None)
     wait = query_params.pop('wait', None)
```
jessebrennan: Also, this is not your change, but it looks like https://github.com/DataBiosphere/azul/pull/1135/commits/efa0c70195a4bf89f62acee3dce6bbeb9a0cc4f2#diff-a3dec4f6b35a08e7cb29086cb8766c3aR793 and the following line do nothing and can be removed.

test/service/test_request_validation.py
```diff
@@ -29,6 +29,14 @@ def lambda_name(cls) -> str:
                           "Message": "BadRequestError: Unable to sort by undefined facet bad-facet."}
     service_config_dir = os.path.dirname(service_config.__file__)
 
+    def _bad_query_message(self, key, type_=None):
```
jessebrennan: _Consider_ making a `@classmethod`.

jessebrennan: Also, I know @hannes-ucsc often prefers verbatim assertions in tests instead of computed results, though for this case I doubt he'd object.

lambdas/service/app.py
```diff
@@ -744,16 +744,16 @@ def _dss_files(uuid, fetch=True):
...
     dss_response = requests.get(url, params=query_params, allow_redirects=False)
     if dss_response.status_code == 301:
         retry_after = min(int(dss_response.headers.get('Retry-After')),
                           int(1.3 ** request_index))
-        query_params['requestIndex'] = request_index + 1
         location = dss_response.headers['Location']
         location = urllib.parse.urlparse(location)
         query = urllib.parse.parse_qs(location.query, strict_parsing=True)
         query_params = {k: one(v) for k, v in query.items()}
+        query_params['requestIndex'] = request_index + 1
```
achave11: @jessebrennan plus the reassignment of this line composes the fix.

lambdas/service/app.py
```diff
@@ -744,16 +744,17 @@ def _dss_files(uuid, fetch=True):
     url = config.dss_endpoint + '/files/' + urllib.parse.quote(uuid, safe='')
     file_name = query_params.pop('fileName', None)
     wait = query_params.pop('wait', None)
-    request_index = query_params.pop('requestIndex', 0)
+    request_index = int(query_params.pop('requestIndex', 0))
```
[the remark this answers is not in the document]
achave11: That is correct, plus comment below.

lambdas/service/app.py
achave11: Added docstrings, please let me know if we can improve the description of the params.

lambdas/service/app.py
```diff
@@ -139,8 +139,47 @@ def version():
...
+        **kwargs
+    }
+    validate_params(incoming_params, **valid_params)
+
+
+def validate_params(query_params, allow_extra_params=False, **kwargs):
+    """
+    Validates incoming request query parameters for web-service API.
+    :param query_params: Incoming parameters to be validated.
+    :param allow_extra_params: Regards the strictness by which the incoming parameters should be validated. When False,
```
jessebrennan: Regards isn't the right word here. Maybe "configure"?
```suggestion
    :param allow_extra_params: When False,
```

lambdas/service/app.py
```diff
@@ -139,8 +139,47 @@ def version():
...
+                               only fields in '**kwards' are accepted, and fails validation if anything else is given.
```
jessebrennan:
```suggestion
                               only fields in '**kwards' are accepted, and validation fails if anything else is given.
```

lambdas/service/app.py
```diff
@@ -139,8 +139,46 @@ def version():
...
+    """
+    Validates incoming request query parameters for web-service API.
+    :param query_params: Incoming parameters to be validated.
+    :param allow_extra_params: When False, only fields in '**kwards' are accepted, and validation fails if anything else
+                               is given. When True, allows extra fields to be present without being properly validated.
+    :param kwargs: Parameters 'query_params' validates against.
+    """
+    for key in query_params.keys():
+        try:
+            type_ = kwargs[key] if not allow_extra_params else None
```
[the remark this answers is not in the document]
hannes-ucsc: Original reviewer should have caught this. :-\

test/service/test_request_validation.py
```diff
@@ -229,3 +232,22 @@ def test_file_order(self):
...
+            with self.subTest(entity_name=f'{entity_type} field'):
+                params = {
+                    'some_nonexistent_filter': 1,
+                }
+                response = requests.get(url, params=params)
+                self.assertEqual(400, response.status_code, response.json())
+                self.assertEqual(self._bad_query_message('some_nonexistent_filter'), response.json())
+            with self.subTest(entity_name=f'{entity_type} type'):
+                params = {
+                    'size': 'wrong_type',
```
hannes-ucsc:
```suggestion
                    'size': 'foo',
```

test/service/test_request_validation.py
```diff
@@ -29,6 +30,15 @@ def lambda_name(cls) -> str:
                           "Message": "BadRequestError: Unable to sort by undefined facet bad-facet."}
     service_config_dir = os.path.dirname(service_config.__file__)
 
+    @classmethod
```
hannes-ucsc: Not a good idea to duplicate production code this way. Try to make the assertions that use this method less strict by asserting a fragment of the response. That will reduce the brittleness of the tests that make those assertions.

test/service/test_dss_proxy.py
```diff
@@ -81,6 +81,7 @@ def test_dss_files_proxy(self):
                 'version': file_version
             }).add(path=file_uuid)
         dss_token = 'some_token'
+        request_to_index = '1'
```
hannes-ucsc: Inline. Used only once, doesn't add clarity. No need to extract variable. Also what's up with the "to"? I don't get why these weird inconsistencies pop up occasionally. It's almost like there's intent to trip the reader up.

achave11:
Had to force push because I rebased my branch on top of develop to capture some updates. That change the shas of the existing commits.

hannes-ucsc:
Force pushing by the author is fine as long as the PR is not assigned to someone else.

hannes-ucsc:
You just shouldn't squash unless requested by a reviewer. (See bullets 6 and 8 of https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#5pull-requests)

lambdas/service/app.py
```diff
@@ -554,7 +671,8 @@ def handle_manifest_generation_request():
     the view function to handle
     """
     query_params = app.current_request.query_params or {}
-    filters = query_params.get('filters')
+    validate_params(query_params, filters=str, format=str, token=str)
+    filters = query_params.get('filters') or {}
```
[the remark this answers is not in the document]
achave11: I included it by mistake. The purpose was the same as the line above for `query_params`

test/test_doctests.py
```diff
@@ -12,7 +12,7 @@
 import azul.time
 import azul.transformer
 import azul.vendored.frozendict
-
+import lambdas.service
```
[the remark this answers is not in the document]
achave11: Will block on the PR that includes changes to address this issues.

lambdas/service/app.py
```diff
@@ -187,8 +187,133 @@ def version():
     }
 
 
+def validate_repository_search(incoming_params, **validators):
```
```suggestion
def validate_repository_search(params, **validators):
```

Typically, parameters are by definition "incoming". Distinguishing between incoming and outgoing parameters would require very special circumstances, that don't apply here. This is run-of-the-mill validation.

For that reason, please remove all other mentions of that word from this PR.

lambdas/service/app.py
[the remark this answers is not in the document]
hannes-ucsc: Then why is it still there?

test/service/test_request_validation.py
```diff
@@ -225,3 +216,22 @@ def test_file_order(self):
...
+                response = requests.get(url, params=params)
+                self.assertEqual(400, response.status_code, response.json())
+                self.assertEqual('BadRequestError', response.json()['Code'])
+            with self.subTest(test='malformed parameter', entity_type=entity_type):
+                params = {
+                    'size': 'foo',
+                }
+                response = requests.get(url, params=params)
+                self.assertEqual(400, response.status_code, response.json())
+                self.assertEqual('BadRequestError', response.json()['Code'])
```
hannes-ucsc: missing NL at EOL

lambdas/service/app.py
```diff
@@ -187,8 +187,133 @@ def version():
...
+    for param_name, param_value in query_params.items():
+        try:
+            validator = validators[param_name]
+        except KeyError:
+            if not allow_extra_params:
+                raise BadRequestError(msg=f'Invalid query parameter `{param_name}`')
+        else:
+            try:
+                validator(param_value)
+            except (TypeError, ValueError):
```
hannes-ucsc: This is inconsistent with the docstring. I'm ok with restricting the set of exceptions that validators can raise for invalid input but that should be documented and it should be documented what happens if the validator raises an exception that is of neither of those types.

lambdas/service/app.py
```diff
@@ -187,8 +186,138 @@ def version():
...
+
+    :param query_params: Parameters to be validated.
+
+    :param allow_extra_params: When False, only parameters specified via '**validators' are accepted, and validation
+                               fails if additional parameters are present. When True, additional parameters are allowed
+                               but their value is not validated.
+
+    :param validators: A dictionary mapping the name of a parameter to a function that will be used to validate the
+                       parameter. The callable will be called with a single argument, the parameter value to be
+                       validated, and is expected to raise an exception if the value is invalid or of the wrong type.
```
```suggestion
                       validated, and is expected to raise ValueError or TypeError if the value is invalid. Only these exception will yield a 4xx status response, all other exceptions will yield a 500 status response.
```

lambdas/service/app.py
```diff
@@ -187,8 +186,138 @@ def version():
...
+                       Any other exception raised would cause a 500 status code with an internal server error message.
```
```suggestion
```

hannes-ucsc:
The history needed to be cleaned up. Please review the mergeing of the split commit and the shortened commit messages.

Last but not least, somehow something went wrong with commits

https://github.com/DataBiosphere/azul/commit/6ece222e479349aaea86e3d679a83f92d9d17511
https://github.com/DataBiosphere/azul/commit/d0c46e205242916a8c63a0068acd6bcfaaf5b80d

Which I had introduced into your branch. The latter is an improved version of the former. I remember you asking me a few days ago about a build failure that looked like it was caused by the fact that the *older* of the two commits was in your branch. That seems to suggest that you rolled back my force push from 6ece222e479349aaea86e3d679a83f92d9d17511 to d0c46e205242916a8c63a0068acd6bcfaaf5b80d. Later you removed these changes completely based on our decision that we would reintroduce them as a separate PR. Not sure you would have remembered that aspect either. It was a mere coincidence that I found this, otherwise we would have lost some of my work.

achave11:
@hannes-ucsc 
Yes, I did make you aware of the failures introduced by your changes. And we reason we would rid of those changes to disregard the failures. So I shelved the changes and was planning to create a PR but did not get around to it, I provably would have forgotten. And there would not have been a retribution to you other than a mention in the commit message, since I did a soft reset on the second to last commit and removed the annotations to your contributions. 
Thank you for the improved commit messages.

PR #1142, opened 2019-07-17
Step function creates full TSV using multipart upload

achave11:
* Include full TSV in suit of multipart uploads.
* Hard check the format of generated object in multipart upload.

### DIFF
--- a/src/azul/service/responseobjects/storage_service.py
+++ b/src/azul/service/responseobjects/storage_service.py
@@ -158,6 +158,7 @@ class MultipartUploadHandler:
         upload_part = self.mp_upload.Part(part.part_number)
         result = upload_part.upload(Body=part.content)
         part.etag = result['ETag']
+        part.content = None
 
 
 @dataclass
--- a/test/local_integration_test.py
+++ b/test/local_integration_test.py
@@ -123,7 +123,7 @@ class IntegrationTest(unittest.TestCase):
         for format_, validator in [
             (None, self.check_manifest),
             ('tsv', self.check_manifest),
-            ('bdbag', self.check_bdbag)
+            ('bdbag', self.check_bdbag)  # TODO: Add 'full' format https://github.com/databiosphere/azul/issues/1171
         ]:
             with self.subTest(format=format_, filter=manifest_filter):
                 query = {'filters': manifest_filter}
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -449,14 +449,6 @@ def get_order():
     return {'order': order_list}
 
 
-def get_format(params):
-    format_ = params.get('format', 'tsv')
-    if format_ in ('tsv', 'bdbag', 'full'):
-        return format_
-    else:
-        raise BadRequestError(f'{format_} is not a valid manifest format.')
-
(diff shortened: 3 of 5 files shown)


src/azul/service/responseobjects/storage_service.py
```diff
@@ -158,6 +158,7 @@ def _upload_part(self, part):
         upload_part = self.mp_upload.Part(part.part_number)
         result = upload_part.upload(Body=part.content)
         part.etag = result['ETag']
+        part.content = None
```
hannes-ucsc: This should be in a separate commit titled: "Fix: multipart TSV upload leaks memory"

test/local_integration_test.py
```diff
@@ -121,9 +121,8 @@ def _test_other_endpoints(self):
     def _test_manifest(self):
         manifest_filter = json.dumps({'project': {'is': [self.test_name]}})
         for format_, validator in [
-            (None, self.check_manifest),
```
hannes-ucsc: This can't be right. The default value for the format parameter is `tsv` and the integration test should cover that. The title of the commit that introduces this deletion (`Fix: None format is not recognize for generating a manifest`) is also misleading as it claims to be a fix which it is not. At best, it is a work-around, albeit a incorrect one, if I am not mistaken.

test/local_integration_test.py
```diff
@@ -121,9 +121,8 @@ def _test_other_endpoints(self):
...
-            ('tsv', self.check_manifest),
-            ('bdbag', self.check_bdbag)
+            ('tsv', self.check_manifest),  # TODO: Validate `full` format when
```
hannes-ucsc: Right. Please create ticket so we can track this TODO and reference that ticket here.

test/test_doctests.py
```diff
@@ -12,6 +12,7 @@
 import azul.time
 import azul.transformer
 import azul.vendored.frozendict
+import azul.service.responseobjects.elastic_request_builder
```
hannes-ucsc: Remove, duplicate import.

hannes-ucsc:
I merged this locally and pushed the merge but unlike a previous PR where I did the same and GH detected the merge and closed the PR this didn't happen here. Unsure what I should be trying next. I still get the green merge button which makes zero sense because the changes are already included and I would expect conflicts in any attempt to merge it again.

hannes-ucsc:
So I did click Merge and the branch was merged again, resulting in the empty commit aee2898. I see what happened now: @achave11 rebased and force-pushed this branch after assigning the PR to me. This prevented GH from detecting the local merge and kept the PR open.

Screen Shot 2019-08-02 at 9 21 38 AM

To be very clear: if a PR is assigned to someone, only the assignee may touch the branch. If a PR is assigned to no one, only the author may touch the PR branch. This isn't explicitly mentioned in the contribution guidelines (and I will add it) but has been stated before, verbally and on PR comments.

To fix this, I force pushed my rebase again. I could have force pushed the PR branch again to undo @achave11's force push but I didn't realize this until after I clicked Merge here. I also reset and force pushed develop to undo aee2898.

Issue #1124, opened 2019-07-11
Health endpoint overloading ES when doing all API endpoints


achave11:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-705)
┆Project Name: azul
┆Issue Number: AZUL-705

achave11:
- Need to fix test that were changed from 503 to 200 in order to pass API endpoints tests.

achave11:
Revert expected when fixed
```
    def _expected_api_endpoints(self, endpoint_states: Mapping[str, bool]) -> JSON:
        # FIXME https://github.com/DataBiosphere/azul/issues/1124
        return {
            'api_endpoints': {
                'up': all(up for endpoint, up in endpoint_states.items()),
                **({
                    config.service_endpoint() + endpoint: {
                        'up': up
                    } if up else {
                        'up': up,
                        'error': f'503 Server Error: Service Unavailable for url: {config.service_endpoint()}{endpoint}'
                    } for endpoint, up in endpoint_states.items()
                })
            }
        }
```

jessebrennan:
When hitting the original `/health` endpoint with the Route 53 health check the request interval was 30 seconds. According to AWS docs, there are ~15 health checkers running which means that a request is sent about every 2-3 seconds (closer to 2 unless the request takes a long time). The health check for the service sends a request about once ever 2 seconds. The health check for the indexer also hits `/health/basic` so the indexer receives about 1 request per second.

@GPelayo's change to the health endpoint which added further requests to 5 other endpoint caused the webservice to receive ~6 requests per second. As this change propagated to the higher environments the shared Elasticsearch instance was being overwhelmed with ~18 requests per second.

The first and simplest fix was to revert Geryl's change. This helped.

Also we plan to reduce the frequency of health check polling to about once every 20 seconds.

jessebrennan:
At first we thought that @achave11's metadata generation changes were causing the slow indexing so these changes were reverted in an attempt to fix Azul. These changes will be reinstated.

PR #1143, opened 2019-07-18
Reduce load from health endpoint (#1124)

jessebrennan:
Several changes are in here that reduce the load on Azul caused by the health endpoint.

Restricting the health checker regions will probably make the biggest impact since this will reduce the number of request made by each health check by more than half. Removing the summary endpoint should also reduce the effect on Elasticsearch of calling health by a factor of two. Finally, removing the redundancy between the /health endpoints for the service and the indexer will reduce the number of Elasticsearch requests by 2 again.

My back of the napkin math says at least 8 times in reduction of Elasticsearch load from the health checker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,8 @@ WORKDIR /build
 
 RUN mkdir terraform \
     && (cd terraform \
-        && wget --quiet https://releases.hashicorp.com/terraform/0.11.11/terraform_0.11.11_linux_amd64.zip \
-        && unzip terraform_0.11.11_linux_amd64.zip \
+        && wget --quiet https://releases.hashicorp.com/terraform/0.12.5/terraform_0.12.5_linux_amd64.zip \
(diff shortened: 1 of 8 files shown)


jessebrennan:
@hannes-ucsc the history has been consolidated.

hannes-ucsc:
My request

> "Work around Terraform health check bug" should precede or be squashed with "Restrict health checker regions (#1124)".

was not realized.

> Please make sure that the change in the health check regions triggers an in-place update in TF, not a replacement.

@jessebrennan confirmed that he had verified this to be "in-place". The sandbox build confirms this:  https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/2387

jessebrennan:
@hannes-ucsc, sorry about that. I forgot to respond in regards to that comment. I'll try and be more thorough.

I didn't understand the reason for the commit to be moved, but I improved the commit message to better explain its role. I thought this would satisfy your request.

Issue #637, opened 2019-01-06
Imports from app.py interfering with test logging


hannes-ucsc:
Both `app.py` files configure logging at module scope. Certain tests import functions from `app.py` which causes that logging configuration in `app.py` to take precedence over the logging configuration in `setupModule()`.

To work around this, I've localized the imports for now but that is overly verbose since the import has to be repeated in every test method.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-402)
┆Project Name: azul
┆Issue Number: AZUL-402

PR #1157, opened 2019-07-25
Importing app.py in setUpModule() to fix test logging (#637)

GPelayo:


test/service/test_cart_item_manager.py
```diff
@@ -11,6 +11,8 @@
 
 def setUpModule():
     logging.basicConfig(level=logging.INFO)
+    global app
```
hannes-ucsc:
```suggestion
    # Late import of Chalice app, otherwise it would interfere with test logging
    global app
```

hannes-ucsc: Here and in the other places where this is done.

GPelayo:
> If you check the Travis job log, you'll see that there is still DEBUG level logging and I think that's because of `logging.getLogger(top_level_pkg).setLevel(logging.DEBUG)` in `app.py`
> 
> ```diff
> Index: lambdas/indexer/app.py
> IDEA additional info:
> ...
> ```
> 
> Please incorporate the above and see if that eliminates the DEBUG logs.

I checked Travis after apply your patch and it looks like it removed all the DEBUG level Travis logs.

Issue #608, opened 2018-12-12
Installing `envhook.py` doesn't work on Ubuntu


jessebrennan:
envhooks.py installs a `sitecustomize.py` but on ubuntu this is overwritten by [Apport](https://wiki.ubuntu.com/Apport) from their docs:

> For intercepting Python crashes it installs a /etc/python*/sitecustomize.py to call apport on unhandled exceptions.

If any `sitecustomize.py` is found, the others that would be found later are ignored.

Solution: import sitecustomize. If it fails, install! else if it succeeds, confirm that it is envhook.py. if it's not envhook and it succeeds, warn that sitecustomize is already in use.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-381)
┆Project Name: azul
┆Issue Number: AZUL-381

hannes-ucsc:
`envhook.py install` should check if there already is a `sitecustomize.py` in another place on sys.path before installing its own copy.

hannes-ucsc:
… and if there is, complain and refuse to install.

PR #1158, opened 2019-07-25
Check for existing sitecustomize.py when installing envhook (#608)

jessebrennan:


scripts/envhook.py
```diff
@@ -41,6 +41,18 @@ def main(argv):
         else:
             raise
     if options.action == 'install':
+        try:
+            import sitecustomize
+        except ImportError:
+            pass
+        else:
+            try:
+                from sitecustomize import setenv
```
hannes-ucsc: I'd prefer a stronger signal

hannes-ucsc: Like comparing `__file__`, if necessary with os.realpath.

jessebrennan: Good point! I forgot that one could do this.

scripts/envhook.py
```diff
@@ -41,6 +41,16 @@ def main(argv):
         else:
             raise
     if options.action == 'install':
+        try:
+            import sitecustomize
+        except ImportError:
+            pass
+        else:
+            if os.path.realpath(sitecustomize.__file__) == os.path.abspath(__file__):
```
hannes-ucsc: `os.path.abspath(__file__)` is alrady available as `dst`. And I don't think we should mix realpath and abspath, we should use either one.

Also, not the [existing logic](https://github.com/DataBiosphere/azul/pull/1158/files#diff-446a633c96640bbd37d927917bf990ceL40) for checking. We should probably move this section up so it is run both when we install and remove. I think it should precede the code in line 20.

scripts/envhook.py
```diff
@@ -41,6 +41,16 @@ def main(argv):
...
+                raise RuntimeError('envhook is already installed')
```
hannes-ucsc:
```suggestion
                pass
```

jessebrennan: I'm happy to change this, but isn't it better to be explicit?

jessebrennan: I see, this matches the pattern already set in the code.

scripts/envhook.py
jessebrennan: `os.path.abspath(__file__)` is not already available as dst because it is turned into a relative path in the next line. I will change it so that a separate variable is used for the absolute and relative paths and the code isn't duplicated.

realpath and abspath do different things and both are necessary. realpath here resolves the link in the path to the `sitecustomize.__file__` and abspath gives the absolute path to the script that's running. Using realpath in both places would technically work, but it seems to me to be more explicit to use abspath for the current file.

jessebrennan: I see your last point, the code should be moved. I think it should be moved after line 30 so that you are warned about not having a virtualenv active before you are warned about the already installed sitecustomize.py. Feel free to message back on Slack for faster back and forth here.

scripts/envhook.py
```diff
@@ -28,9 +28,16 @@ def main(argv):
...
         raise RuntimeError('Need to be run from within a virtualenv')
-    link = os.path.join(link_dir, 'sitecustomize.py')
     dst = os.path.abspath(__file__)
-    dst = os.path.relpath(dst, link_dir)
+    try:
+        import sitecustomize
+    except ImportError:
+        pass
+    else:
+        if os.path.realpath(sitecustomize.__file__) != dst:
```
hannes-ucsc: `realpath` calls `abspath` on Posix and Darwin, the two platforms we care about. I think I can easily create a situation in which `realpath(p1) == realpath(p2) and realpath(p1) != abspath(p2)`. Just make one of the directories in `p2` a symlink. Maybe I'm not getting it but my default would be to only ever compare realpath's. I think my original code was already broken in this way. So what I am proposing is to switch out every use of abspath and replace it with realpath.

scripts/envhook.py
```diff
@@ -28,9 +28,16 @@ def main(argv):
...
+            raise RuntimeError(f'A different sitecustomize.py already exists at {sitecustomize.__file__}')
```
hannes-ucsc:
```suggestion
            raise RuntimeError(f'A different sitecustomize module already exists at {sitecustomize.__file__}')
```

Both `sitecustomize.py` and `sitecustomize/__init__.py` can be imported using `import sitecustomize`, IIRC.

scripts/envhook.py
hannes-ucsc: > os.path.abspath(__file__) is not already available as dst because it is turned into a relative path in the next line.

I missed that.

> I think it should be moved after line 30

Yes, that's better.

> realpath and abspath do different things

Realpath calls abspath on Darwin and Posix. I therefore think of one of them to be a superset of the other and it would be misleading to treat them as doing "different things".

> Using realpath in both places would technically work, but it seems to me to be more explicit to use abspath for the current file.

See my most recent review comment

scripts/envhook.py
jessebrennan: Good point.

scripts/envhook.py
jessebrennan: I see. Ok I'll think this through and make the change.

jessebrennan:
@hannes-ucsc I changed one usage of abspath to realpath (the one used for the comparison), but I left the others. To me it made sense to leave them as abspath because:
- it's functionally the same and
- to me it seems more explicit to call abspath when you want the absolute path instead of realpath.

I hope my reasoning makes sense. If I interpret you comment literally, then I'm disregarding part of it, but I hope my actions best capture the intent of what you were saying.

Issue #1147, opened 2019-07-22
Eliminate empty columns in full metadata TSV


achave11:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-718)
┆Project Name: azul
┆Issue Number: AZUL-718

hannes-ucsc:
```
GET azul_bundles_aggregate_dev/_search
{
  "query": {
    "term": {
      "contents.projects.document_id.keyword": {
        "value": "2cd14cf5-f8e0-4c97-91a2-9e8957f41ea8"
      }
    }
  },
  "aggs": {
    "fields": {
      "scripted_metric": {
        "init_script": "params._agg.fields = new HashSet()",
        "map_script": "for (row in params._source.contents.metadata) { for (f in row.keySet()) { params._agg.fields.add(f)}}",
        "combine_script": "return new ArrayList(params._agg.fields)",
        "reduce_script": "Set fields = new HashSet(); for (agg in params._aggs) { fields.addAll(agg) } return new ArrayList(fields)"
      }
    }
  }
}
```

Is a search request with an aggregation that computes the content.metadata.* fields that are actually present in the response.

The problem is that the search request that is currently used to get the metadata out of the bundles index is consumed with the `scan()` helper which I think that does let you access aggregation results in the response. We'd need to issue the request twice, once with the above script aggregation (and no other aggregations) and `"size": 0` to suppress any actual hits in the response, and a second time with scan() but without the aggregation. The first request will tell you all the field names in the response and therefore the column headers for the full TSV, the second request will yield all the rows for that TSV.

PR #1160, opened 2019-07-25
Eliminate empty columns in in full metadata TSV (#1147)

achave11:


hannes-ucsc:
We only assign PR to requested reviewers, not when a PR changes authors.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -571,7 +571,28 @@ def transform_manifest(self, format: str, filters: JSON):
...
+            es_search = self.create_request(filters,
+                                            self.es_client,
+                                            request_config,
+                                            post_filter=False,
+                                            source_filter=source_filter,
+                                            enable_aggregation=False,
+                                            entity_type=entity_type)
+            es_search.aggs.metric('fields', 'scripted_metric',
+                                  init_script='params._agg.fields = new HashSet()',
+                                  map_script='for (row in params._source.contents.metadata) { '
```
hannes-ucsc: Triple quoted string with newlines and indented Painless code would be better. And extract into separate variable.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -571,7 +571,28 @@ def transform_manifest(self, format: str, filters: JSON):
...
+                                             'for (f in row.keySet()) {'
+                                             'params._agg.fields.add(f)}}',
+                                  combine_script='return new ArrayList(params._agg.fields)',
+                                  reduce_script='Set fields = new HashSet();'
```
hannes-ucsc: Same here

test/service/test_manifest.py
```diff
@@ -755,6 +755,90 @@ def test_full_metadata(self):
...
+
+            params = {'filters': json.dumps({'project': {'is': ['Mouse Melanoma']}}), 'format': 'full'}
+            response = self.get_manifest(params)
+            self.assertEqual(200, response.status_code, 'Unable to download manifest')
+            tsv_file2 = csv.reader(response.iter_lines(decode_unicode=True), delimiter='\t')
+            fieldnames2 = next(tsv_file2)
+
+            symmetric_diff = set(fieldnames1) ^ set(fieldnames2)
+
+            expected_diff = {
```
hannes-ucsc: This is a nifty test. Since there is already a test that asserts the presence of individual columns in the full TSV, could we change this to `assertNotEmpty(symmetric_diff)`? And maybe also add `assertNotEmpty(union)`?

danielsotirhos: Added assertion of the sets intersection along with the symmetric diff. No `assertNotEmpty()` available so I used `assertGreater(len(diff), 0)` instead.

danielsotirhos:
Rebased & resolved conflicts

PR #1052, opened 2019-05-30
Preserve metadata field in aggregate (#966)

achave11:

PR #1166, opened 2019-07-26
Fix #1052: Disable ES dynamic type setting for metadata fields

achave11:
This reverts commit 162deaba

### DIFF
--- a/lambdas/indexer/app.py
+++ b/lambdas/indexer/app.py
@@ -30,14 +30,17 @@ from azul.time import RemainingLambdaContextTime
 from azul.transformer import EntityReference
 from azul.types import JSON
 
-logging.basicConfig(level=logging.WARNING)
 log = logging.getLogger(__name__)
-for top_level_pkg in (__name__, 'azul'):
-    logging.getLogger(top_level_pkg).setLevel(logging.DEBUG)
+if len(logging.root.handlers) == 0:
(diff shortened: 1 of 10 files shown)


hannes-ucsc:
You haven't requested a review so you may already be aware of it: We agreed that there would be one commit that adds the test and that the build for that commit should fail in that test with the same error that we got in `prod`, as captured in the ticket.

achave11:
Correct, notice the commit `8214d89` that introduces the test which reproduces the error seen in production. It is a rather interesting issue, remind me to explain if you are interested in the details. 
Also notice that the previously discussed solution of deactivating the mappings for all fields in metadata is not possible with out breaking the dynamic TSV generation mechanisms. Reason being dynamic mapping of those fields, in the present, allows us to generate the header row of the TSV. 
Until the update provided by #1147 we need to preserve the mapping.

hannes-ucsc:
> notice the commit TBD that

I don't get it. Notice what commit? I also see no Travis build that fails with the same error we get in production.

I had to force push develop to fix a mistake I made on Friday.

hannes-ucsc:
Given the confusing exchange of comments on this PR, I'm closing it. Please create another PR when you have a branch with two commits: one that reverts the hotfix and another commit that adds a test reproducing the problem exactly as it happened on prod.

PR #1180, opened 2019-07-31
Fix translated None values

danielsotirhos:


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -402,14 +419,14 @@ def transform_summary(
         ).bucket(
             'cell_count',
             'sum',
-            field='contents.cell_suspensions.total_estimated_cells'
+            field='contents.cell_suspensions.total_estimated_cells_ORIGINAL'
```
hannes-ucsc: I'd prefer 

```suggestion
            field='contents.cell_suspensions.total_estimated_cells_'
```

unless ES objects.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -391,25 +404,25 @@ def transform_summary(
...
             'sum',
-            field='contents.files.size')
+            field='contents.files.size_')
 
         # Add a per_organ aggregate to the ElasticSearch request
         es_search.aggs.bucket(
             'group_by_organ',
             'terms',
-            field='contents.cell_suspensions.organ.keyword',
+            field='contents.cell_suspensions.organ.keyword',    # FIXME contents.samples.organ?
```
hannes-ucsc: Let's talk about this.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -420,7 +433,7 @@ def transform_summary(
             ('contents.specimens.document_id', 'specimenCount'),
             ('contents.files.uuid', 'fileCount'),
             ('contents.donors.document_id', 'donorCount'),
-            ('contents.projects.laboratory', 'labCount'),
+            ('contents.projects.laboratory', 'labCount'),   # FIXME Is this now counting '__null__' as a value?
```
hannes-ucsc: Probably, but I am just going to say that that's ok.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -817,8 +827,14 @@ def choose_entry(_term):
                 term_object_params['projectId'] = [bucket['key'] for bucket in term['myProjectIds']['buckets']]
             term_list.append(TermObj(**term_object_params))
 
-        # Add 'unspecified' term if there is at least one unlabelled document
         untagged_count = contents['untagged']['doc_count']
+
+        # Update or add a termObj for a None value if the aggregation on missing (untagged) fields returned a count > 0
+        if untagged_count > 0:
```
hannes-ucsc: The next conditional has the same condition

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -805,8 +815,8 @@ def create_facet(contents):
         def choose_entry(_term):
             if 'key_as_string' in _term:
                 return _term['key_as_string']
-            elif not isinstance(_term['key'], str):
-                return str(_term['key'])
+            elif isinstance(_term['key'], bool):
```
hannes-ucsc: This changes what happens if `_term['key']` is an int. Before it would be converted to a string. Now it remains an int. Is that your intention?

src/azul/transformer.py
```diff
@@ -125,7 +125,16 @@ def translate_fields(cls, source: JSON, path: Tuple[str, ...] = (), forward: boo
         :return: A copy of the original document with values translated according to their type
         """
         if isinstance(source, dict):
-            return {key: cls.translate_fields(val, path=path + (key,), forward=forward) for key, val in source.items()}
+            new_dict = {}
+            for key, val in source.items():
+                # Skip any shadow copy fields to remove them from the dictionary when reverse translating
+                if not key.endswith('_'):
```
```suggestion
                if key.endswith('_'):
                    assert not forward
                else:
```

The usual strictness rule.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -559,6 +559,16 @@ def __init__(self, raw_response):
             fileTypeSummaries=[FileTypeSummary.for_bucket(bucket) for bucket in _sum['buckets']],
             cellCountSummaries=[OrganCellCountSummary.for_bucket(bucket) for bucket in _organ_group['buckets']])
 
+        # Un-translate values that had been translated with translate_fields() to handle Nones in Elasticsearch
+        # The path for each translation directly relates to the field used in transform_summary() for the aggregation
+        for summary in kwargs['cellCountSummaries']:
```
hannes-ucsc: This entire section looks brittle. If the dependent code is in this method, I'd be ok with it. Otherwise we may want to think about alternatives.

src/azul/service/responseobjects/elastic_request_builder.py
danielsotirhos: Created new ticket #1188 to address this issue

hannes-ucsc:
Noticed that I was still assigned and unassigned myself. Since you're still pushing changes, feel free to assign back to me when this is ready to be looked at again.

src/azul/service/responseobjects/hca_response_v5.py
danielsotirhos: Modified function to handle all cases, now will return a string or None. Boolean values converted and lowercased to match previous behavior. Numeric values converted to string.

src/azul/service/responseobjects/hca_response_v5.py
danielsotirhos: Moved translation of values in `cellCountSummaries` into its constructor method `for_bucket()`.

danielsotirhos:
Ready for re-evaluation.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -559,6 +562,13 @@ def __init__(self, raw_response):
             fileTypeSummaries=[FileTypeSummary.for_bucket(bucket) for bucket in _sum['buckets']],
             cellCountSummaries=[OrganCellCountSummary.for_bucket(bucket) for bucket in _organ_group['buckets']])
 
+        # Un-translate values that had been translated with translate_fields() to handle Nones in Elasticsearch
```
hannes-ucsc: Check out what I did here in the next commit. It's odd that a list is generated with a comprehension and then patched in place 5 lines down. My changes will reduce the churn in state and be more efficient.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -100,6 +100,9 @@ class OrganCellCountSummary(JsonObject):
 
     @classmethod
     def for_bucket(cls, bucket):
+        bucket['key'] = Document.translate_field(bucket['key'],
```
hannes-ucsc: Try not to patch argument data structures in place. Modifying arguments is a nasty source of bugs. It's more defensive to make a copy. I fix this in my commit.

Issue #1138, opened 2019-07-16
Cache full metadata TSVs for each project


hannes-ucsc:
The links to the full, per-project TSVs in the Data Browser will have a project filter set that point to a single project by UUID. When determining the name of TSV file in the S3 bucket we currently allocate a simple v4 UUID. If many users click the same link, we would generate a copy of the TSV for each of them. Each user would have to wait a few minutes and we would waste a lot of space.

For full per-project TSVs we should hash the project UUID, the current Azul commit SHA and the current date. We should use the hash to name the TSV file and skip the generation if the file already exists.  Hashing the commit means that redeploying Azul invalidates the cache. Hashing the date means that the cache is invalidated at midnight.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-711)
┆Project Name: azul
┆Issue Number: AZUL-711

hannes-ucsc:
A slight tweak on this: We should use a deterministic file name for all `?format=full` manifests and rely on the storage bucket's life cycle policy to evict them. The expiration is 1 day which seems appropriate. We may want to extend the determinism to other manifests, but for now I'd like to keep that limited to `full` manifests.

PR #1215, opened 2019-08-09
Cache full metadata TSV (#1138)

danielsotirhos:


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -573,6 +576,18 @@ def transform_request(self,
...
         return final_response
 
+    def _generate_manifest_uuid_name(self, filters: JSON) -> str:
+        """
+        Generate and return a UUID string generated using the latest git commit and filters
+        :param filters: The filters used in the manifest request
+        :return: String representation of a UUID
+        """
+        uuid_namespace = uuid.UUID('ca1df635-b42c-4671-9322-b0a7209f0235')
+        git_status = config._git_status
```
hannes-ucsc: There is a reason this property is protected. This would fail when run in a lambda.

Use lambda_git_status instead and mock as necessary during the test.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -238,15 +241,53 @@ def validate(s: str) -> str:
...
+                raise e
+        return header
+
+    def _can_use_cached_manifest(self, object_key):
+        """
+        Check if a s3 object was previously created and still exists in the bucket and won't be expiring soon.
+        :param object_key: S3 object key (eg. 'manifests/e0fabf97-7abb-5111-af97-810f1e736c71.tsv'
+        :return: boolean
+        """
+        minimum_remaining_threshold = 60 * 15
```
hannes-ucsc: Config, next to the other property you added.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -238,15 +241,53 @@ def validate(s: str) -> str:
...
+        header = self._get_header(object_key)
+        if header is not None:
+            object_age = datetime.now(timezone.utc) - header['LastModified']
+            seconds_until_expire = config.s3_object_lifetime_seconds - object_age.total_seconds()
+            # if 'x-amz-expiration' in header:
```
hannes-ucsc: This is an RFC 2616 HTTP date,  it seems, for which we have `parse_http_date` in `time.py`

hannes-ucsc:
Also, lets still consider the verification of the expiration.

And if we could test both the single and multipart case that would be great. There is a config property and corresponding environment variable to force single part. See if using mock.patch.dict on `os.environ` can be used for that.

danielsotirhos:
Requesting code review. Note: Still missing is a test/verification of the the code comparing the expiration date in the header 'x-amz-expiration' against the calculated expiration date .

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -238,15 +242,58 @@ def validate(s: str) -> str:
...
+            object_age = date_time_now - last_modified
+            seconds_until_expire = config.s3_object_lifetime_seconds - object_age.total_seconds()
+            if 'x-amz-expiration' in header:
+                # Verify x-amz-expiration matches the expiration date/time we expect
+                # example value: 'expiry-date="Fri, 21 Dec 2012 00:00:00 GMT", rule-id="Rule for testfile.txt"'
+                x_amz_expiration = parse_dict_header(header['x-amz-expiration'])
+                expiry_date = parse_date(x_amz_expiration['expiry-date'])
+                expected_expiry_date = date_time_now + timedelta(seconds=seconds_until_expire)
+                date_diff = expiry_date - expected_expiry_date
+                assert(abs(date_diff.total_seconds()) < config.s3_object_min_allowed_life_remaining)
```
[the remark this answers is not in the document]
hannes-ucsc: I think we want to use a different epsilon here. The purpose of s3_object_min_allowed_life_remaining is to ensure that the replacement is ready before the replaced object is deleted by the lifecycle rule. The epsilon for this comparison is not related to that concern, AFAICT. We should account for clock skew and maybe leap seconds, but that's it. Reading, https://aws.amazon.com/blogs/aws/keeping-time-with-amazon-time-sync-service/ I think 10s are a good epsilon for this.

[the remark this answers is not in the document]
hannes-ucsc: I'm a bit scared to make this an assert(). Note that typically there is a delay between `make deploy` and `make terraform` and during that time window the S3 expiration could differ dramatically from what the code expects it to be. I think we should change this to `log.error`.

src/azul/__init__.py
```diff
@@ -71,6 +71,16 @@ def disable_multipart_manifests(self) -> bool:
     def s3_bucket(self) -> str:
         return os.environ['AZUL_S3_BUCKET']
 
+    @property
+    def s3_object_lifetime_seconds(self) -> int:
```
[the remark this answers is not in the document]
danielsotirhos: Changed this value to # of days (instead # of seconds) as the `expiration` config in terraform is specified in # of days.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -573,6 +576,17 @@ def transform_request(self,
...
+    def _generate_manifest_uuid_name(self, filters: JSON) -> str:
+        """
+        Generate and return a UUID string generated using the latest git commit and filters
+        :param filters: The filters used in the manifest request
+        :return: String representation of a UUID
+        """
+        git_status = config.lambda_git_status
+        git_commit = git_status['commit']
+        filter_string = repr(sort_frozen(freeze(filters)))
+        return str(uuid.uuid5(uuid.NAMESPACE_DNS, git_commit + filter_string))
```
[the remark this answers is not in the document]
danielsotirhos: A custom namespace seemed unnecessary given that we're generating a UUID to get a unique file name, and the git commit and filter string are the variable seed to the UUID generation. The namespace UUID just needs to be constant, and since the uuid module already has a constant available for use it seemed like a good fit.
I see how it's not that variable's intended use however and will re-add the custom namespace value

src/azul/service/responseobjects/hca_response_v5.py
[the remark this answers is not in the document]
danielsotirhos: Question: if the expiration date from the `x-amz-expiration` header and the calculated value don't match, should we use the header value as the truth?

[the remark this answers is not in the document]
hannes-ucsc: Good point. Maybe we should. `x-amz-expiration` is the truth and that's the earliest time, the object could disappear. If the goal is to prevent a gap, we should use `x-amz-expiration`.

src/azul/__init__.py
```diff
@@ -71,6 +71,20 @@ def disable_multipart_manifests(self) -> bool:
     def s3_bucket(self) -> str:
         return os.environ['AZUL_S3_BUCKET']
 
+    @property
+    def manifest_expiration(self) -> float:
+        """
+        Length of time (in days) before an S3 object will be deleted
```
```suggestion
        Number of days before a manifest will be deleted from the storage bucket
```

src/azul/__init__.py
```diff
@@ -71,6 +71,20 @@ def disable_multipart_manifests(self) -> bool:
...
+        """
+        return 1
+
+    @property
+    def manifest_expiration_margin(self) -> float:
+        """
+        Minimum length of time (in seconds) before an object is considered too close to expiration for use
```
```suggestion
        Minimum duration (in seconds) before a manifest in the storage bucket is considered too close to expiration for use
```

test/service/test_manifest.py
```diff
@@ -784,6 +793,87 @@ def test_full_metadata_missing_fields(self):
...
+                'format': 'full'
+            }
+            # Run get_manifest() with various values returned by mocked _get_seconds_until_expire() to verify log msg
+            for seconds_until_expire in None, 3600, 10:
+                with self.subTest(seconds_until_expire=seconds_until_expire):
+                    get_seconds.return_value = seconds_until_expire
+                    with self.assertLogs(level='INFO') as logs:
+                        response = self.get_manifest(params)
+                        self.assertEqual(200, response.status_code, 'Unable to download manifest')
+                    if seconds_until_expire is None:    # on this first request the cached manifest doesn't exist yet
```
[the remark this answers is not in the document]
danielsotirhos: Refactored testcase to use an inner function

test/service/test_manifest.py
```diff
@@ -784,6 +793,87 @@ def test_full_metadata_missing_fields(self):
...
+                            file_name = urlparse(response.url).path
+                            file_names[project_id].append(file_name)
+
+                        self.assertEqual(file_names.keys(), set(project_ids))
+                        self.assertEqual([2, 2], list(map(len, file_names.values())))
+                        self.assertEqual([1, 1], list(map(len, map(set, file_names.values()))))
+
+    def test_get_seconds_until_expire(self):
+        self.assertEqual(config.manifest_expiration, 1) # number of days until expiration
+        for h_expire, h_modified in (12, 12), (1, 23), (23, 1):
```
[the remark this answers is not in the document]
danielsotirhos: Modified test case to mock time.time() and verify the results of the `_get_seconds_until_expire()` function for a range of `LastModified` values.
Also added a new test case that tests for the presence of an error log message if the header properties `x-amz-expiration` and `LastModified` don't line up as expected.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -238,15 +242,72 @@ def validate(s: str) -> str:
...
+        last_modified = header['LastModified']
+        assert isinstance(last_modified, datetime)
+        expected_expiry_date = last_modified + timedelta(days=config.manifest_expiration)
+        expected_value = (expected_expiry_date - date_time_now).total_seconds()
+        date_diff_margin = 10  # seconds
+        if abs(seconds_until_expire - expected_value) > date_diff_margin:
+            expected_expiry_date = date_time_now + timedelta(seconds=seconds_until_expire)
+            logger.error('x-amz-expiration %s != calculated %s', expiry_date, expected_expiry_date)
+        else:
+            pass
```
hannes-ucsc: Hmm, I think this is taking it too far. Unless this is a oversight that you intended to remove, I will probably have to specify more clearly where I think redundant `else` branches should be added.

test/service/test_manifest.py
```diff
@@ -784,6 +797,127 @@ def test_full_metadata_missing_fields(self):
...
+
+            def log_messages_from_get_seconds(expire_date: datetime, last_modified: datetime) -> List[str]:
+                expire_date_string = expire_date.strftime("%a, %d %b %Y %H:%M:%S %Z")
+                header = {
+                    'x-amz-expiration': f'expiry-date="{expire_date_string}", rule-id="Test Rule"',
+                    'LastModified': last_modified
+                }
+                with self.assertLogs(level='ERROR') as logs:
+                    ManifestResponse._get_seconds_until_expire(header)
+                    logger.error('Dummy error message to have at least one error message logged')
```
hannes-ucsc: What would happen if this line didn't exist?

test/service/test_manifest.py
```diff
@@ -784,6 +797,127 @@ def test_full_metadata_missing_fields(self):
...
+        with mock.patch('time.time', new=lambda: 1547691253.07010):
+            dt_now = datetime.now(timezone.utc)
+
+            def log_messages_from_get_seconds(expire_date: datetime, last_modified: datetime) -> List[str]:
+                expire_date_string = expire_date.strftime("%a, %d %b %Y %H:%M:%S %Z")
+                header = {
+                    'x-amz-expiration': f'expiry-date="{expire_date_string}", rule-id="Test Rule"',
+                    'LastModified': last_modified
+                }
+                with self.assertLogs(level='ERROR') as logs:
```
hannes-ucsc:
```suggestion
                from azul.service.responseobjects.hca_response import logger as logger_
                with self.assertLogs(logger=logger_, level='ERROR') as logs:
```

test/service/test_manifest.py
```diff
@@ -784,6 +797,127 @@ def test_full_metadata_missing_fields(self):
...
+                    logger.error('Dummy error message to have at least one error message logged')
+                    return logs.output
+
+            hours_until_expire = 8
+            expire_date = dt_now + timedelta(hours=hours_until_expire)
+            last_modified = dt_now - timedelta(hours=((config.manifest_expiration * 24) - hours_until_expire))
+            msg_prefix = 'ERROR:azul.service.responseobjects.hca_response_v5:x-amz-expiration'
+            # Verify no error message is logged when expire_date and last_modified match
+            logs_output = log_messages_from_get_seconds(expire_date, last_modified)
+            self.assertEqual(set([message.startswith(msg_prefix) for message in logs_output]), {False})
```
hannes-ucsc:
```suggestion
            self.assertEqual(set(['x-amz-expiration' in message for message in logs_output]), {False})
```

In combination with the `logger` argument to `assertLogs` this is less sensitive to log format changes.

test/service/test_manifest.py
danielsotirhos: As part of the test case I verify that a log message is added when the two date fields (expire date & last modified) don't add up correctly, and that no log message is added when they do. Without a dummy log message, `assertLogs` throws an exception when no message is logged.
I could add a comment or just make the dummy log message more explanatory
```
logger.error('Dummy log message so assertLogs() does not fail if no other error log is generated')
```

test/service/test_manifest.py
danielsotirhos: I will also apply this style of log verification to the other test case `test_metadata_cache_expiration` where i'm verifying the presence and absence of log messages.

test/service/test_manifest.py
```diff
@@ -161,9 +169,14 @@ class ManifestGenerationTest(WebServiceTestCase):
     def setUp(self):
         super().setUp()
         self._setup_indices()
+        # A git commit hash is required to generate the file name of format='full' type manifests
+        os.environ['azul_git_commit'] = '9347432ab0da43c73409ac7fd3edfe29cf3ae678'
```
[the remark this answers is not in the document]
danielsotirhos: Adding assert and refactoring into `_setup_git_commit()` and `_teardown_git_commit()` methods to keep `setUp()` and `tearDown()` clean and easier to read.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -238,15 +242,71 @@ def validate(s: str) -> str:
...
+        """
+        Get the number of seconds before a cached manifest is past its expiration.
+
+        :param header: A storage service object header dict
+        :return: time to expiration in seconds
+        """
+        # example x-amz-expiration = 'expiry-date="Fri, 21 Dec 2012 00:00:00 GMT", rule-id="Rule for testfile.txt"'
+        date_time_now = datetime.now(timezone.utc)
+        x_amz_expiration = parse_dict_header(header['x-amz-expiration'])
+        expiry_date = parse_date(x_amz_expiration['expiry-date']).replace(tzinfo=date_time_now.tzinfo)
```
hannes-ucsc: Forcing the timezone is not a good idea since we don't know for sure that the TZ offset in header is UTC. This is actually compensating for a bug in `parse_date` (which somehow parses out the TZ offset but then returns a offset-naive datetime object). See my commit for a safer solution.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -238,15 +242,71 @@ def validate(s: str) -> str:
...
+        seconds_until_expire = (expiry_date - date_time_now).total_seconds()
+        # Verify value calculated from 'x-amz-expiration' matches value calculated from 'LastModified'
+        last_modified = header['LastModified']
+        assert isinstance(last_modified, datetime)
```
hannes-ucsc: This is not a bad thing per se but it makes mocking harder so I took it out.

test/service/test_manifest.py
```diff
@@ -25,6 +31,8 @@
 from service import WebServiceTestCase
 from lambdas.service import app
 
+logger = logging.getLogger(__name__)
```
hannes-ucsc: When a logger is added, it needs to be passed to configure_test_loggging()

test/service/test_manifest.py
```diff
@@ -161,11 +169,26 @@ class ManifestGenerationTest(WebServiceTestCase):
...
+        """
+        Set git variables required to generate the object key of format='full' type manifests
+        """
+        assert 'azul_git_commit' not in os.environ
+        assert 'azul_git_dirty' not in os.environ
+        os.environ['azul_git_commit'] = '9347432ab0da43c73409ac7fd3edfe29cf3ae678'
+        os.environ['azul_git_dirty'] = 'False'
+
+    def _teardown_git_commit(self):
+        os.environ.pop('azul_git_commit', None)
```
hannes-ucsc: The default is unnecessary since we know that the variable must be present.

test/service/test_manifest.py
```diff
@@ -784,6 +807,125 @@ def test_full_metadata_missing_fields(self):
...
+                    hours_till_expiration = (config.manifest_expiration * 24) - hours_last_modified
+                    expire_date = (dt_now + timedelta(hours=hours_till_expiration)).strftime("%a, %d %b %Y %H:%M:%S %Z")
+                    header = {
+                        'x-amz-expiration': f'expiry-date="{expire_date}", rule-id="Test Rule"',
+                        'LastModified': dt_now - timedelta(hours=hours_last_modified)
+                    }
+                    seconds_remaining = ManifestResponse._get_seconds_until_expire(header)
+                    self.assertAlmostEqual(seconds_remaining, 60 * 60 * hours_till_expiration, delta=1)
+
+    def test_get_seconds_until_expire_logs(self):
```
hannes-ucsc: This test can be combined with the previous one.

test/service/test_manifest.py
```diff
@@ -784,6 +807,125 @@ def test_full_metadata_missing_fields(self):
...
+                            response = self.get_manifest(params)
+                            self.assertEqual(200, response.status_code, 'Unable to download manifest')
+                            file_name = urlparse(response.url).path
+                            file_names[project_id].append(file_name)
+
+                        self.assertEqual(file_names.keys(), set(project_ids))
+                        self.assertEqual([2, 2], list(map(len, file_names.values())))
+                        self.assertEqual([1, 1], list(map(len, map(set, file_names.values()))))
+
+    def test_get_seconds_until_expire(self):
```
hannes-ucsc: This test doesn't need ES so it should be separate.

danielsotirhos:
All suggested changes accounted for. I've reviewed the the new `test_get_seconds_until_expire()` and concur it does a much better job at verifying the behavior of `ManifestResponse._get_seconds_until_expire()` for a variety of fail situations in a less verbose way

hannes-ucsc:
It was a bit of an effort to get this actually working. The lambda policy needed to be updated and the `head_object` response element is actually called `Expiration`, not `x-amz-expiration`. Boto3 apparently translates the HTTP response header names (x-amz-foo) to JSON response keys (Foo).

In the future, @danielsotirhos, we can't just rely on unit tests for validation. The author of a PR for functionality that involves cloud resources should verify that the new functionality actually works when deployed.

Issue #1139, opened 2019-07-16
Set content disposition header on full manifests


hannes-ucsc:
When creating the manifest in the bucket, set the content disposition header so the user gets a friendly local file name on download. @GPelayo has code for this in his ad-hoc script for pregenerating the TSVs.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-712)
┆Project Name: azul
┆Issue Number: AZUL-712

hannes-ucsc:
I removed blockage of https://github.com/HumanCellAtlas/data-browser/issues/712 because this will take a little longer and is cosmetic only (the file name will be a UUID, which after #1215 will be deterministic). 

cc: @NoopDog, @theathorn

PR #1216, opened 2019-08-09
Add Content-Disposition header to the full manifest (#1139)

GPelayo:


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -614,6 +616,12 @@ def transform_manifest(self, format: str, filters: JSON):
             response = es_search.execute()
             aggregate = response.aggregations
             manifest_config = self.generate_full_manifest_config(aggregate)
+            hit = response.hits.hits[0]
```
hannes-ucsc: This only works by accident, see https://github.com/DataBiosphere/azul/issues/1219. We may have to apply the header to the object in a separate request at the end, after the object has been created. Also, the full manifest can be generated for any filter. The response could contain any number of projects. Taking the one in the first hit isn't going to work.

hannes-ucsc:
Also keep an eye on https://github.com/DataBiosphere/azul/pull/1215, @GPelayo.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -300,16 +304,20 @@ def _write_tsv(self, output: IO[str]) -> None:
                     self._extract_fields(entities, column_mapping, row)
                 writer.writerow(row)
 
-    def _write_full(self, output: IO[str]) -> None:
+    def _write_full(self, output: IO[str]) -> str:
```
```suggestion
    def _write_full(self, output: IO[str]) -> Optional[str]:
```

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -300,16 +304,20 @@ def _write_tsv(self, output: IO[str]) -> None:
...
+        project_short_names = set()
         for hit in self.es_search.scan():
             doc = hit['contents'].to_dict()
             for metadata in list(doc['metadata']):
+                if len(project_short_names) < 2:
+                    project_short_names.add(metadata['project.project_core.project_short_name'])
                 row = dict.fromkeys(sources)
                 row.update(metadata)
                 writer.writerow(row)
+        return project_short_names.pop() if len(project_short_names) == 1 else ''
```
```suggestion
        return project_short_names.pop() if len(project_short_names) == 1 else None
```

src/azul/service/responseobjects/storage_service.py
```diff
@@ -133,6 +134,11 @@ def __complete(self):
             self.__abort()
             raise MultipartUploadError(self.bucket_name, self.object_key) from exception
 
+        if self.content_disposition:
+            s3_object = self.mp_upload.Object()
+            s3_object.copy_from(CopySource={'Bucket': self.bucket_name, 'Key': self.object_key},
```
hannes-ucsc: Apologies, I had forgotten that one can't update metadata on an object. Unless you can show me that no actual copy is made (I found no evidence online that S3 is optimizing this case), this is a show-stopper because it's too expensive for a convenience feature like this. 

Instead, can we create a shadow object (same key but append `.name` or something) that stores the name and then use it when creating the signed URL to the actual manifest?

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -253,12 +255,22 @@ def _push_content(self) -> str:
...
                     raise NotImplementedError(f'Multipart upload not implemented for `{self.format}`')
 
         return object_key
 
+    def _push_name_file(self, object_key, project_short_name):
+        project_short_name = project_short_name or ''
+        file_name_prefix = unicodedata.normalize('NFKD', project_short_name)
+        file_name_prefix = re.sub(r'[^\w ,.@%&\-_()\\[\]/{}]', '_', file_name_prefix).strip()
+        timestamp = datetime.now().strftime("%Y-%m-%d %H.%M")
+        filename = f'{file_name_prefix} {timestamp}.tsv'.strip()
```
hannes-ucsc: It looks like if `project_short_name` is None or False or empty etc, that we'd end up with a file name that starts with a space. I think that we only want to create a `.name` object if we have a project name.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -482,6 +500,10 @@ def return_response(self):
         else:
             object_key = self._push_content()
             file_name = 'hca-manifest-' + object_key.rsplit('/', )[-1]
+
+        if self.format == 'full':
+            file_name = self.storage_service.get(object_key + '.name').decode('utf-8')
```
hannes-ucsc: This should gracefully handle the absence of a `.name` file.

hannes-ucsc: If you decode UTF8 you should specify that at encoding time as well. But 'utf-8' is the default so

```suggestion
            file_name = self.storage_service.get(object_key + '.name').decode()
```

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -253,12 +255,22 @@ def _push_content(self) -> str:
...
 
         return object_key
 
+    def _push_name_file(self, object_key, project_short_name):
+        project_short_name = project_short_name or ''
+        file_name_prefix = unicodedata.normalize('NFKD', project_short_name)
+        file_name_prefix = re.sub(r'[^\w ,.@%&\-_()\\[\]/{}]', '_', file_name_prefix).strip()
+        timestamp = datetime.now().strftime("%Y-%m-%d %H.%M")
+        filename = f'{file_name_prefix} {timestamp}.tsv'.strip()
+        self.storage_service.put(object_key=f'{object_key}.name',
```
```suggestion
        self.storage_service.put(object_key=object_key + '.name',
```

hannes-ucsc: And make `'.name'` a constant class attribute.

src/azul/service/responseobjects/hca_response_v5.py
GPelayo: So @hannes-ucsc if a project_short_name is None or False or empty etc, does the tsv filename default to `hca-manifest-5d67606f-472a-4bb9-9dac-5135ef5c164c.tsv` (if multiple part manifest are enabled)?

hannes-ucsc: The manifest object's key should be the same as it is now, prior to this change. Only if all rows in the TSV refer to exactly one `project_short_name` AND if that `project_short_name` is not empty, will the `.name` object be created. If a `.name` object exists when the signed URL is created, the CD header's fileName element will be set to the contents of the `.name` object. This will happen independently of whether single or multipart upload was enabled.

hannes-ucsc: Rather than silently stripping the leading space that would result from an empty `project_short_name`, assert that it is not empty.

Geryl, I beg you, please break this habit of writing lenient code that masks errors. Be strict.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -482,6 +503,16 @@ def return_response(self):
         else:
             object_key = self._push_content()
             file_name = 'hca-manifest-' + object_key.rsplit('/', )[-1]
+
+        if self.format == 'full':
+            try:
+                file_name = self.storage_service.get(object_key + self.name_file_suffix).decode()
+            except ClientError as e:
+                if e.response['Error']['Code'] == 'NoSuchKey':
+                    logger.warning(f"{object_key}{self.name_file_suffix} doesn't exist. Renaming the filename of the"
```
hannes-ucsc: The warning message is crap. Please prefer % over f string in log messages. This leaves file_name unassigned.

test/service/test_manifest.py
```diff
@@ -788,3 +791,41 @@ def test_manifest_format_validation(self):
...
+                        params = {
+                            'filters': json.dumps(filters),
+                            'format': 'full'
+                        }
+                        filters = AbstractService.parse_filters(params.get('filters'))
+                        estd = ElasticTransformDump()
+                        url = estd.transform_manifest(params.get('format', 'full'), filters).headers['Location']
+                        query = urlparse(url).query
+                        content_dispositions = parse_qs(query).get('response-content-disposition')
+                        if content_dispositions:
```
hannes-ucsc: There is the potential that a bug in production code is never true. The condition should not be driven by actual state, but by expected state. Feel free to talk to me if that's not clear.

test/service/test_manifest.py
```diff
@@ -788,3 +791,41 @@ def test_manifest_format_validation(self):
...
+                            content_disposition = one(content_dispositions)
+                            expected_prefix = 'attachment;filename=' + expected_name
+                            expected_suffix = '.tsv'
+                            actual_prefix = content_disposition[:len(expected_prefix)]
+                            actual_date = content_disposition[len(expected_prefix):-len(expected_suffix)]
+                            # Asserts if the date in the filename is correctly formatted
+                            if name_file:
+                                datetime.strptime(actual_date, "%Y-%m-%d %H.%M")
```
hannes-ucsc: Use mocks to pin the time so you can hard-code the expected file name. Whenever expectations are computed using logic from code under test, you risk duplicating a bug such that the bug in the expectation cancels out the bug in code under test.

https://github.com/DataBiosphere/azul/blob/ec89044d1ef8100749aa4459662d6d0f1969542e/test/service/test_manifest.py#L899

Issue #1173, opened 2019-07-30
Re-enable date detection, unmap contents.metadata in ES (#1173)


achave11:
* Indexer file {project.home}/azul/src/azul/project/hca/indexer.py requires removal of `date_detection` mapping setting. 
* Disable mapping for metadata fields so no ES analysis is done on them.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-732)
┆Project Name: azul
┆Issue Number: AZUL-732

PR #1221, opened 2019-08-13
Remove date detection condition and disable mapping for metadata (#1173)

achave11:


test/indexer/test_hca_indexer.py
```diff
@@ -818,13 +818,11 @@ def test_metadata_generator(self):
     def test_metadata_field_exclusion(self):
         self._index_canned_bundle(self.old_bundle)
 
-        # Check that the dynamic mapping for the field is present
+        # Check that the dynamic mapping has the field disabled
```
danielsotirhos:
```suggestion
        # Check that the dynamic mapping has the metadata field disabled
```

test/indexer/test_hca_indexer.py
```diff
@@ -818,13 +818,11 @@ def test_metadata_generator(self):
...
         bundles_index = config.es_index_name('bundles')
         mapping = self.es_client.indices.get_mapping(index=bundles_index)
-        self.assertIn('bundle_uuid', mapping[bundles_index]
-        ['mappings']['doc']['properties']
-        ['contents']['properties']
-        ['metadata']['properties'])
+        self.assertFalse(
+            mapping[bundles_index]['mappings']['doc']['properties']['contents']['properties']['metadata']['enabled'])
```
danielsotirhos:
```suggestion
        contents = mapping[bundles_index]['mappings']['doc']['properties']['contents']
        self.assertFalse(contents['properties']['metadata']['enabled'])
        self.assertEqual(contents['properties']['metadata']['type'], 'object')
```

Issue #1163, opened 2019-07-26
Generate service API documentation using OpenAPI


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-725)
┆Project Name: azul
┆Issue Number: AZUL-725

jessebrennan:
Report:

Swagger/OpenAPI is sufficient for generating documentation for the API. There are several steps to accomplishing this:
- **Serve the Swagger UI from Azul.** This is mostly complete as part of my investigation.
- **Generate the Swagger/OpenAPI spec for the service.** This can be done with Chalice during build time or pulled from APIGateway once deployed. There are pros and cons to each approach, but currently I'm leaning towards the using Chalice.
- **Load the documentation from Azul's docstrings.** It's possible that updating the version of Chalice my be sufficient to accomplish this. Otherwise, if the Swagger spec is generated during build time, the annotations could be added in (with some work). If the Swagger spec is not deployed at build time, then the doc strings will have to be added in during run time which could be a hit to efficiency.

PR #1227, opened 2019-08-17
Display Swagger UI with docs from app.py (#1163)

jessebrennan:
Things outstanding:
- The docs are very much incomplete. This need to be done in a follow-up ticket.
- We need to decide how to deal with/display API versioning

I have a script that somewhat improves the development process for writing the docs by rendering them, unfortunately I cannot figure out how to display validation. I'm unsure whether or not to include it.

For an example of what this looks like, check out https://service.jesse0.dev.explore.data.humancellatlas.org/

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -4,6 +4,7 @@ import hashlib
 import json
 import logging.config
 import math
+import os
 import re
 import time
 from typing import Optional
@@ -21,6 +22,7 @@ from azul import config, drs, RequirementError
 from azul.chalice import AzulChaliceApp
 from azul.health import Health
 from azul.logging import configure_app_logging
+from azul.openapi import annotated_specs
 from azul.plugin import Plugin
 from azul.security.authenticator import AuthenticationError, Authenticator
 from azul.service.manifest import ManifestService
@@ -41,6 +43,21 @@ app = AzulChaliceApp(app_name=config.service_name)
 
 configure_app_logging(app, log)
 
+
+openapi_spec = {
+    'info': {
+        'description': 'This should probably be really long',
+        # Version should be updated in any PR tagged API with a major version
+        # update for breaking changes, and a minor version otherwise
+        'version': '1.0'
+},
+    'tags': [
+        {
+            'name': 'Health',
+            'description': 'For checking on service health'
+        }
+    ]
+}
 
 sort_defaults = {
     'files': ('fileName', 'asc'),
@@ -105,12 +122,41 @@ def jwt_auth(auth_request) -> AuthResponse:
     return AuthResponse(routes=['*'], principal_id='user', context=context)
 
 
+pkg_root = os.path.dirname(os.path.abspath(__file__))
+
+
 @app.route('/', cors=True)
-def hello():
-    return {'Hello': 'World!'}
+def swagger_ui():
+    local_path = os.path.join(pkg_root, 'vendor')
+    dir_name = local_path if os.path.exists(local_path) else pkg_root
+    with open(os.path.join(dir_name, 'static', 'swagger-ui.html')) as f:
+        openapi_ui_html = f.read()
+    return Response(status_code=200,
+                    headers={"Content-Type": "text/html"},
+                    body=openapi_ui_html)
 
 
-@app.route('/health', methods=['GET'], cors=True)
+@app.route('/openapi', methods=['GET'], cors=True)
+def openapi():
+    gateway_id = os.environ['api_gateway_id']
+    spec = annotated_specs(gateway_id, app, openapi_spec)
+    return Response(status_code=200,
+                    headers={"content-type": "application/json"},
+                    body=spec)
+
+
+@app.route('/health', methods=['GET'], cors=True, spec={
+    'parameters': [
+        {
+            'in': 'query',
+            'name': 'foo',
+            'description': 'This is not a real parameter',
+            'required': False,
+            'schema': {'$ref': "#/components/schemas/Empty"}
+        }
+    ],
+    'tags': ['Health']
+})
 @app.route('/health/{keys}', methods=['GET'], cors=True)
 def health(keys: Optional[str] = None):
     if keys == 'basic':
--- a/lambdas/service/lambda-policy.json.template.py
+++ b/lambdas/service/lambda-policy.json.template.py
@@ -136,6 +136,16 @@ emit({
                 f"arn:aws:states:{aws.region_name}:{aws.account}:execution:{config.cart_item_state_machine_name}:*",
                 f"arn:aws:states:{aws.region_name}:{aws.account}:execution:{config.cart_export_state_machine_name}"
             ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "apigateway:GET"
+            ],
+            "Resource": [
+                f"arn:aws:apigateway:{aws.region_name}::"
+                f"/restapis/{aws.api_gateway_id(config.service_name)}/stages/{config.deployment_stage}/exports/oas30"
+            ]
         }
     ]
 })
--- a/scripts/apidev.py
+++ b/scripts/apidev.py
@@ -1,0 +1,61 @@
+import json
+import os
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+
+from watchdog.events import FileSystemEventHandler
+from watchdog.observers import Observer
+
+from azul import config
+from azul.deployment import aws
+from azul.modules import load_app_module
+from azul.openapi import annotated_specs
+
+spec_file = 'openapi.json'
+parent_dir = os.path.realpath(os.path.dirname(__file__))
+
+
+def write_specs(gateway_id, app, openapi_spec):
+    specs = annotated_specs(gateway_id, app, openapi_spec)
+    with open(os.path.join(parent_dir, spec_file), 'w') as f:
+        json.dump(specs, f)
+
+
+def main():
+    static_dir = os.path.join(parent_dir, 'apidev_static')
+    web_page = os.path.join(static_dir, 'swagger-editor.html')
+
+    host, port = 'localhost', 8787
+    server_url = f"http://{host}:{port}"
+    httpd = HTTPServer((host, port), SimpleHTTPRequestHandler)
+    address = f"{server_url}/{os.path.relpath(web_page)}?url={server_url}/{os.path.relpath(parent_dir)}/{spec_file}"
+    print(f'Open {address} in browser to validate changes.')
+    print('If you changed/added any routes, make sure to deploy afterwords (also update API version)!')
+    print('This is because the specs are pulled from API Gateway which needs '
+          'to be up to date with any new routes')
+
+    service = load_app_module('service')
+    gateway_id = aws.api_gateway_id(config.service_name, validate=True)
+    event_handler = UpdateHandler(service, gateway_id)
+    observer = Observer()
+    observer.schedule(event_handler, path=os.path.dirname(service.__file__), recursive=False)
+    observer.start()
+
+    write_specs(gateway_id, service.app, service.openapi_spec)
+    httpd.serve_forever()
+
+
+class UpdateHandler(FileSystemEventHandler):
+    def __init__(self, service, gateway_id):
+        self.service = service
+        self.gateway_id = gateway_id
+        self.tracked_file = os.path.join(os.path.dirname(service.__file__), 'app.py')
+
+    def on_modified(self, event):
+        if event.src_path == self.tracked_file:
+            self.service = load_app_module('service')
+            write_specs(self.gateway_id, self.service.app, self.service.api_info)
+            print('Spec updated')
+
+
+if __name__ == "__main__":
+    main()
--- a/src/azul/chalice.py
+++ b/src/azul/chalice.py
@@ -1,5 +1,6 @@
 from chalice import Chalice
 
+from azul.openapi import openapi_spec
 from azul.types import LambdaContext
 
 
@@ -13,9 +14,12 @@ class AzulChaliceApp(Chalice):
         Same as method in supper class but stashes URL path a view function is bound to as an attribute of the
         function itself.
         """
+        spec = kwargs.pop('spec', None)
         decorator = super().route(path, **kwargs)
 
         def _decorator(view_func):
+            if spec is not None:
+                view_func = openapi_spec(spec)(view_func)
             view_func.path = path
             return decorator(view_func)
 
--- a/src/azul/deployment.py
+++ b/src/azul/deployment.py
@@ -57,6 +57,13 @@ class AWS:
     def dynamo(self, endpoint_url, region_name):
         return boto3.resource('dynamodb', endpoint_url=endpoint_url, region_name=region_name)
 
+    def api_gateway_export(self, gateway_id):
+        response = self.apigateway.get_export(restApiId=gateway_id,
+                                              stageName=config.deployment_stage,
+                                              exportType='oas30',
+                                              accepts='application/json')
+        return json.load(response['body'])
+
     def api_gateway_id(self, function_name: str, validate=True) -> Optional[str]:
         try:
             response = self.lambda_.get_policy(FunctionName=function_name)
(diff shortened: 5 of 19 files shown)


lambdas/service/lambda-policy.json.template.py
```diff
@@ -122,6 +122,15 @@
...
                 f"arn:aws:states:{aws.region_name}:{aws.account}:execution:{config.cart_export_state_machine_name}"
             ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "apigateway:GET"
+            ],
+            "Resource": [
+                f"arn:aws:apigateway:{aws.region_name}::/restapis/{aws.api_gateway_id('service')}/stages/{config.deployment_stage}/exports/oas30"
```
danielsotirhos: Line longer than 120 char, however I'm not certain the "fix" below makes things better. Possibly?
```suggestion
                f"arn:aws:apigateway:{aws.region_name}::"
                f"/restapis/{aws.api_gateway_id('service')}/stages/{config.deployment_stage}/exports/oas30"
```

jessebrennan: I'll leave it and see if Hannes complains.

jessebrennan:
@hannes-ucsc I assigned you for review. One of the builds is failing, but I think it's a Travis error; I would appreciate your insight on the issue.

src/azul/deployment.py
```diff
@@ -57,8 +57,18 @@ def secretsmanager(self):
     def dynamo(self, endpoint_url, region_name):
         return boto3.resource('dynamodb', endpoint_url=endpoint_url, region_name=region_name)
 
-    def api_gateway_id(self, function_name: str, validate=True) -> Optional[str]:
+    def api_gateway_export(self, gateway_id):
+        response = self.apigateway.get_export(
+            restApiId=gateway_id,
```
hannes-ucsc: We're still doing aligned indent for function arguments.

lambdas/service/lambda-policy.json.template.py
hannes-ucsc: I would wrap it there, too.

scripts/apidev.py
```diff
@@ -0,0 +1,59 @@
...
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+
+from watchdog.events import FileSystemEventHandler
+from watchdog.observers import Observer
+
+from azul.deployment import aws
+from azul.modules import load_app_module
+from azul.openapi import annotated_specs
+
+SPEC_FILE = 'openapi.json'
```
hannes-ucsc: We don't do upper case constants.

lambdas/service/Makefile
```diff
@@ -12,6 +12,7 @@ wheels:
 	unzip $(azul_home)/lambdas/.wheels/jsonobject-0.9.9-cp36-cp36m-linux_x86_64.whl -d vendor
 
 deploy: state wheels
+	cp static/openapi.html vendor/
```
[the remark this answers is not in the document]
hannes-ucsc: In fact I think the whole `static` directory should be in `vendor/` then we can add more files later without having to add inclusions rules to .gitignore.

lambdas/service/app.py
```diff
@@ -106,11 +119,43 @@ def jwt_auth(auth_request) -> AuthResponse:
...
+@app.route('/openapi', methods=['GET'], cors=True)
+def openapi():
+    gateway_id = os.environ['AZUL_SERVICE_API_GATEWAY_ID']
+    specs = annotated_specs(gateway_id, app, api_info)
+    return Response(status_code=200,
+                    headers={"content-type": "application/json"},
+                    body=specs)
+
+
+@spec(
```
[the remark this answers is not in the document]
hannes-ucsc: Also, to eliminate one indention level, how about this?

```suggestion
@spec({
```

lambdas/service/app.py
```diff
@@ -106,11 +119,43 @@ def jwt_auth(auth_request) -> AuthResponse:
...
+                'name': 'foo',
+                'description': 'This is not a real parameter',
+                'required': False,
+                'schema': {'$ref': "#/components/schemas/Empty"}
+            }
+        ],
+        'tags': ['Health']
+    }
+)
 @app.route('/health/basic', methods=['GET'], cors=True)
```
hannes-ucsc: And lets consider this 

```suggestion
@app.route('/health/basic', methods=['GET'], cors=True, openapi={
})
```

src/azul/deployment.py
```diff
@@ -88,10 +98,13 @@ def es_endpoint(self) -> Netloc:
...
     @property
     def lambda_env(self) -> Mapping[str, str]:
-        return config.lambda_env(self.es_endpoint)
+        return {
+            **config.lambda_env(self.es_endpoint),
+            'AZUL_SERVICE_API_GATEWAY_ID': aws.api_gateway_id('service', validate=True)
+        }
 
     def get_lambda_arn(self, function_name, suffix):
-        return f"arn:aws:lambda:{aws.region_name}:{aws.account}:function:{function_name}-{suffix}"
```
hannes-ucsc: There may be more. Fix them all and move to separate commit if you haven't done so already.

src/azul/openapi.py
```diff
@@ -0,0 +1,113 @@
+import copy
+from typing import MutableMapping, List
+
+from azul.deployment import aws
+from azul.types import JSON
+
+
+def spec(description: JSON):
```
hannes-ucsc: In the `@app.route(…, openapi=…` approach, our `AzulChalicApp.route` override would call this function. The reason I like tying this to `app.route` is that `@spec` doesn't make sense by itself, it should only be used on routed functions.

src/azul/openapi.py
```diff
@@ -0,0 +1,113 @@
...
+        func = next(iter(entries.values())).view_function
+        if hasattr(func, 'api_spec'):
+            docs[route] = func.api_spec
+    return docs
+
+
+def merge_dicts(d1: MutableMapping, d2: MutableMapping, mutate: bool = False, override: bool = False):
+    """
+    Merge two dictionaries deeply such that:
+    - collisions raise an error,
```
hannes-ucsc:
```suggestion

    - collisions raise an error,
```

src/azul/openapi.py
```diff
@@ -0,0 +1,113 @@
...
+    - lists are appended, and
+    - child dictionaries are recursively merged
+    If override is true, d2 overrides d1 on collisions.
```
hannes-ucsc:
```suggestion

    If override is true, d2 overrides d1 on collisions.
```

hannes-ucsc: And consider adding empty lines before each case (after all but the last assertion).

src/azul/openapi.py
```diff
@@ -0,0 +1,113 @@
...
+    >>> merge_dicts({1: {'a': 'z'}}, {1: {'a': 'y'}}, override=True)
+    {1: {'a': 'y'}}
+    """
+    if mutate:
+        d1 = copy.deepcopy(d1)
+        d2 = copy.deepcopy(d2)
+    return recursive_merge_dicts(d1, d2, [], override)
+
+
+def recursive_merge_dicts(merged: MutableMapping, d2: MutableMapping, path: List[str], override: bool):
```
hannes-ucsc: If a method is solely for internal use, prefix its name with underscore. It helps communicating what's intended to be used from the outside and what should be considered internal.

test/local_integration_test.py
```diff
@@ -340,6 +342,19 @@ def _get_entities_by_project(self, entity_type, project_short_name):
         return entities
 
 
+class OpenAPIIntegrationTest(unittest.TestCase):
+
+    def test_openapi(self):
+        service = config.service_endpoint()
+        response = requests.get(service + '/')
+        self.assertEqual(response.status_code, 200)
```
hannes-ucsc: Maybe assert the content type (should be html) and length (should be non-zero).

src/azul/openapi.py
```diff
@@ -0,0 +1,113 @@
...
+    >>> merge_dicts({1: {'a': 'z', 'c': [1]}}, {1: {'b': 'y', 'c': [2]}})
+    {1: {'a': 'z', 'c': [1, 2], 'b': 'y'}}
+    >>> merge_dicts({1: {'a': 'z'}}, {1: {'a': 'y'}})
+    Traceback (most recent call last):
+    ...
+    ValueError: Cannot merge path '1.a' with values: z, y
+    >>> merge_dicts({1: {'a': 'z'}}, {1: {'a': 'y'}}, override=True)
+    {1: {'a': 'y'}}
+    """
+    if mutate:
```
hannes-ucsc: I don't see this ever being used. If not, just eliminate it. If it is, the naming is odd. mutate=True cause a copy to be made? I'd expect the opposite, mutate=False should make the copy. Also if the function modifies the argument, it should not return anything. So: eliminate if not used or replace this modal argument with a make a wrapper function that copies first.

src/azul/openapi.py
hannes-ucsc: And explain the parameters.

src/azul/openapi.py
hannes-ucsc: BTW, `inplace` is a more common name for this behavior.

scripts/apidev.py
```diff
@@ -0,0 +1,59 @@
+import json
+import os
+import shutil
```
hannes-ucsc: unused

scripts/apidev.py
hannes-ucsc:
```diff
Index: scripts/apidev.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- scripts/apidev.py	(revision 7068a20c2848c16d2304c25ffcbd187d239aa565)
+++ scripts/apidev.py	(date 1566351360490)
@@ -1,6 +1,5 @@
 import json
 import os
-import shutil
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 
 from watchdog.events import FileSystemEventHandler
@@ -12,8 +11,6 @@
 
 SPEC_FILE = 'openapi.json'
 parent_dir = os.path.realpath(os.path.dirname(__file__))
-static_dir = os.path.join(parent_dir, 'apidev_static')
-web_page = os.path.join(static_dir, 'swagger-editor.html')
 
 
 def write_specs(gateway_id, app, api_info):
@@ -23,14 +20,17 @@
 
 
 def main():
+    static_dir = os.path.join(parent_dir, 'apidev_static')
+    web_page = os.path.join(static_dir, 'swagger-editor.html')
     service = load_app_module('service')
-
-    server_address = ('localhost', 8787)
-    server_url = f"http://{':'.join(map(str, server_address))}"
-    httpd = HTTPServer(server_address, SimpleHTTPRequestHandler)
+    # That's what made me chuckle earlier today
+    host, port = 'localhost', 8787
+    server_url = f"http://{host}:{port}"
+    httpd = HTTPServer((host, port), SimpleHTTPRequestHandler)
     address = f"{server_url}/{os.path.relpath(web_page)}?url={server_url}/{os.path.relpath(parent_dir)}/{SPEC_FILE}"
     print(f'Open {address} in browser to validate changes.')
-    print('If you changed/added any routes, make sure to deploy!')
+    print('If you changed/added any routes, make sure to `make deploy` afterwards!')
+    # TODO: explain why
 
     gateway_id = aws.api_gateway_id('service', validate=True)
     event_handler = UpdateHandler(service, gateway_id)
```

src/azul/deployment.py
```diff
@@ -88,10 +98,13 @@ def es_endpoint(self) -> Netloc:
 
     @property
     def lambda_env(self) -> Mapping[str, str]:
-        return config.lambda_env(self.es_endpoint)
+        return {
+            **config.lambda_env(self.es_endpoint),
+            'AZUL_SERVICE_API_GATEWAY_ID': aws.api_gateway_id('service', validate=True)
```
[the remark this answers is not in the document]
hannes-ucsc: Evidence for `inplace`:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html

Note the return value of None if inplace=True

src/azul/openapi.py
jessebrennan: Woops! The opposite behavior was what was intended. I'll get rid of it then... I don't believe it's covered in doctests. I made it originally because I wasn't sure if I'd need it or not.

lambdas/service/lambda-policy.json.template.py
jessebrennan: Changed.

lambdas/service/Makefile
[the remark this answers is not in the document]
jessebrennan: Ok great. I made this change.

lambdas/service/app.py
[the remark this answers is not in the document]
jessebrennan: Done.

src/azul/deployment.py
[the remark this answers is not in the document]
jessebrennan: In response to the first comment here, I made the change.

scripts/apidev.py
jessebrennan: Ahh... Yeah I had just copied that straight out of the HTTPServer docs. Didn't even think to simplify.

lambdas/service/app.py
jessebrennan: I dig it. The only con I can think of is that it takes some of the OpenAPI logic out of `openapi.py`.

src/azul/openapi.py
jessebrennan: I see. OK I'll give it a whirl.

src/azul/deployment.py
jessebrennan: That was the only one I found. I will more to a separate commit.

lambdas/service/app.py
jessebrennan: I would still prefer the parameter name `spec` over `openapi` however.

src/azul/openapi.py
jessebrennan: Done.

src/azul/openapi.py
jessebrennan: Good point.

scripts/apidev.py
jessebrennan: Removed.

src/azul/deployment.py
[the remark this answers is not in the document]
hannes-ucsc: The `inplace` comment was in reference to  https://github.com/DataBiosphere/azul/pull/1227#discussion_r315967458

src/azul/chalice.py
```diff
@@ -13,9 +14,12 @@ def route(self, path, **kwargs):
         Same as method in supper class but stashes URL path a view function is bound to as an attribute of the
         function itself.
         """
+        description = kwargs.pop('spec', None)
```
hannes-ucsc:
```suggestion
        spec = kwargs.pop('spec', None)
```

src/azul/chalice.py
```diff
@@ -13,9 +14,12 @@ def route(self, path, **kwargs):
...
         decorator = super().route(path, **kwargs)
 
         def _decorator(view_func):
+            if description:
+                spec(description)(view_func)
```
hannes-ucsc:
```suggestion
                view_func = spec(description)(view_func)
```

This is the decorator protocol. If you didn't assign, you'd be relying on an implementation detail of `spec`.

lambdas/service/app.py
```diff
@@ -43,6 +44,18 @@
 configure_app_logging(app, log)
 
 
+api_info = {
```
hannes-ucsc: Can we consolidate the terminology a bit? If it is `spec` for the decorator and keyword argument, it should be `spec` here as well. OTOH, `spec` is a bit too generic to be used at the module scope. That's why I'd prefer `openapi` for everything.

src/azul/chalice.py
jessebrennan: This will cause a name collision. I will import the spec decorator so that it's use is qualified (to avoid the collision).

src/azul/chalice.py
jessebrennan: That's what I had originally, but I realized it was redundant. I changed it back however.

lambdas/service/app.py
jessebrennan: What about something like `top_level_spec` or `global_spec`? `openapi` just parses kind of funny for me.

hannes-ucsc: `top_level_spec` is long and awkward. Both "global" and "toplevel" are implied because the variable is defined at the module scope. I like `openapi` because it is both short and unique. Can you explain what you mean by "parses funny"?

hannes-ucsc:
I am absolutely baffled as to why the builds are failing. One is failing in a health checker test the other one in trufflehog.

jessebrennan:
@hannes-ucsc I could try making a new branch and opening a new PR?

hannes-ucsc:
I'll look into the test_laziness failure because it's happening on other branches, too.

hannes-ucsc:
I did a binary search over the Trufflehog patterns. It' s the Github pattern that triggers this:

https://github.com/DataBiosphere/azul/blob/b82b32b035118a367c63e7fa4d109f298ad539e7/.trufflehog.json#L12

The pattern is a bit odd. In fact, I think it's just dumb. Github API keys are just hex strings of a fixed length so I don't know what that pattern even attempts to match.

jessebrennan:
@hannes-ucsc, I'm not sure what you mean by "without (!) rebasing". I think you mean without rebasing onto the latest develop. I squashed commits properly, but didn't rebase onto develop.

PR #1230, opened 2019-08-19
Various fixes

hannes-ucsc:


test/service/test_response.py
```diff
@@ -38,7 +38,9 @@ def tearDownClass(cls):
         super().tearDownClass()
 
     def get_hits(self, entity_type: str, entity_id: str):
-        """Fetches hits from es instance searching for a particular entity ID"""
+        """
+        Fetches hits from es instance searching for a particular entity ID
```
danielsotirhos:
```suggestion
        Fetches hits from ES instance searching for a particular entity ID
```
or `Elasticsearch`

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -542,11 +538,10 @@ def transform_request(self,
...
             list_adjustment = 1 if len(es_hits) > pagination['size'] else 0
             if 'search_before' in pagination:
-                hits = [x['_source'] for x in
-                        reversed(es_hits[0:len(es_hits) - list_adjustment])]
+                hits = reversed(es_hits[0:len(es_hits) - list_adjustment])
             else:
-                hits = [x['_source'] for x in es_hits[0:len(es_hits) - list_adjustment]]
-
+                hits = es_hits[0:len(es_hits) - list_adjustment]
+            hits = [hit['_source'] for hit in hits]
```
danielsotirhos:
```suggestion
            hits = [hit['_source'] for hit in es_hits[0:len(es_hits) - list_adjustment]]
            if 'search_before' in pagination:
                hits = reversed(hits)
```

test/service/test_response.py
hannes-ucsc: Fixed.

src/azul/service/responseobjects/elastic_request_builder.py
hannes-ucsc: As mentioned in stand-up, while I think distinguishing between hits and docs would made sense but that would be a rabbit hole I am not prepared to take on right now.

Issue #1191, opened 2019-08-05
Generalise field translation of aggregations in ES responses


hannes-ucsc:
The terms in a term aggregation response need to be translated back from the internal representation (`__null__, `-min_int`, etc) back to `None`. Currently those backward translations are generic for term facets in a normal response, but hard-coded for a summary response. 

When creating an aggregation for a summary request, stash the field path in the [meta](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/agg-metadata.html#agg-metadata) key and use it to generically translate back each resulting term from a `terms` aggregation.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-741)
┆Project Name: azul
┆Issue Number: AZUL-741

hannes-ucsc:
This is a follow-up on #1179.

PR #1246, opened 2019-08-29
Generalise aggregation field translations (#1191)

danielsotirhos:


danielsotirhos:
Rebased branch on latest develop

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -393,6 +393,7 @@ def transform_summary(self,
             'group_by_organ',
             'terms',
             field='contents.cell_suspensions.organ.keyword',
+            meta={'path': ('contents', 'cell_suspensions', 'organ')},
```
hannes-ucsc: I had something even more generic in mind. Notice how `field` and `meta.path` are covariant. Why do I have to hard-code the derived value if it can be computed?

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -626,9 +620,13 @@ def __init__(self, raw_response):
             donorCount=self.agg_contents(self.aggregates, 'donorCount', agg_form='value'),
             labCount=self.agg_contents(self.aggregates, 'labCount', agg_form='value'),
             totalCellCount=self.agg_contents(self.aggregates, 'total_cell_count', agg_form='value'),
-            organTypes=list(map(OrganType.for_bucket, _organ_types['buckets'])),
+            organTypes=list(map(OrganType.for_bucket,
+                                _organ_types['buckets'],
+                                repeat(tuple(_organ_types['meta']['path'])))),
```
hannes-ucsc: This assumes that certain aggregation have the `meta.path` entry. That is still too brittle. The rule of thumb is that if you find yourself adding code that mentions concrete document fields or their path, you're doing it wrong. :-)

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -659,7 +652,19 @@ def __init__(self, raw_response):
             totalCellCount=self.agg_contents(self.aggregates, 'total_cell_count', agg_form='value'),
             organTypes=list(map(OrganType.for_bucket, _organ_types['buckets'])),
             fileTypeSummaries=list(map(FileTypeSummary.for_bucket, _file_types['buckets'])),
-            cellCountSummaries=list(map(OrganCellCountSummary.for_bucket, _cell_counts['buckets'])))
+            cellCountSummaries=list(map(OrganCellCountSummary.for_bucket, _cell_counts['buckets']))
+        )
+
+    @classmethod
+    def _translate_aggregate(cls, aggregate: MutableMapping[str, MutableMapping]):
+        if 'field' in aggregate.get('meta', {}):
```
hannes-ucsc: This is always True at the moment. You know the `meta.field` entry is present for both call sites. And that is the bad smell. You need to refactor the to make this condition count. 

I would like this translation problem solved generically for any `terms` aggregation, not just specific ones, for individual fields. If someone adds another terms aggregation I don't want them to have to worry about the translation aspect.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -154,8 +142,33 @@ def _create_aggregate(self, filters, facet_config, agg):
...
         if excluded_filter is not None:
             filters[facet_config[agg]] = excluded_filter
+        # Store a copy of the ES field name aggregated in each Terms aggregate's 'meta' field
+        self.add_meta_field_to_aggregate(aggregate)
         return aggregate
 
+    @classmethod
+    def add_meta_field_to_aggregate(cls, aggregate):
+        """
+        Update the aggregate to add a meta field containing the full name of the ES field being aggregated
```
```suggestion
        Update the aggregate to add a meta field containing the path of the ES field being aggregated
```

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -154,8 +142,33 @@ def _create_aggregate(self, filters, facet_config, agg):
...
+        """
+        # If we have a Filter object iterate over the contents to find any Terms objects
+        if isinstance(aggregate, Filter):
+            for agg_name in aggregate:
+                agg = aggregate[agg_name]
+                if isinstance(agg, Terms):
+                    cls.add_meta_field_to_aggregate(agg)
```
hannes-ucsc: This recursion is unnecessary. Note how the type is checked twice for `Terms`.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -154,8 +142,33 @@ def _create_aggregate(self, filters, facet_config, agg):
...
+        # Store a copy of the ES field name aggregated in each Terms aggregate's 'meta' field
+        self.add_meta_field_to_aggregate(aggregate)
         return aggregate
 
+    @classmethod
+    def add_meta_field_to_aggregate(cls, aggregate):
+        """
+        Update the aggregate to add a meta field containing the full name of the ES field being aggregated
+        """
+        # If we have a Filter object iterate over the contents to find any Terms objects
```
hannes-ucsc: This is obvious. Remove.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -154,8 +142,33 @@ def _create_aggregate(self, filters, facet_config, agg):
...
+        if isinstance(aggregate, Filter):
+            for agg_name in aggregate:
+                agg = aggregate[agg_name]
+                if isinstance(agg, Terms):
+                    cls.add_meta_field_to_aggregate(agg)
+        # If we have a Terms object copy out the 'field' value and stuff it in the 'meta' field
```
hannes-ucsc: Also obvious, IMO.

hannes-ucsc: Consider this a compliment. Your code is self-explanatory.

src/azul/transformer.py
```diff
@@ -141,6 +141,19 @@ def translate_fields(cls, source: AnyJSON, path: Tuple[str, ...] = (), forward:
         else:
             return cls.translate_field(source, path=path, forward=forward)
 
+    @classmethod
+    def untranslate_aggregates(cls, es_response):
+        """
+        Untranslate aggregated field values using the ES field name stored in the 'meta' field
+        """
+        for aggregation in es_response.aggregations:
+            if 'myTerms' in aggregation:
```
hannes-ucsc: Why?

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -154,8 +142,33 @@ def _create_aggregate(self, filters, facet_config, agg):
...
+        elif isinstance(aggregate, Terms):
+            field = aggregate.field
+            suffix = '.keyword'
```
hannes-ucsc: Try splitting here and stashing the tuple like we talked about. I'd like to be convinced that that is not easier.

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -369,9 +386,16 @@ def transform_summary(self,
                 agg_name, 'cardinality',
                 field='{}.keyword'.format(cardinality),
                 precision_threshold="40000")
+
+        # Store a copy of the ES field name aggregated in each Terms aggregate's 'meta' field
+        # We can't iterate over the values of an AggsProxy directly but we can iterate over the keys
+        for agg_name in es_search.aggs:
```
hannes-ucsc: Can't this loop be pushed down into add_meta_field_to_aggregate?

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -441,6 +466,8 @@ def transform_request(self,
             except elasticsearch.NotFoundError as e:
                 raise IndexNotFoundError(e.info["error"]["index"])
 
+            Document.untranslate_aggregates(es_response)
+
```
```suggestion
```

src/azul/transformer.py
```diff
@@ -141,6 +141,19 @@ def translate_fields(cls, source: AnyJSON, path: Tuple[str, ...] = (), forward:
         else:
             return cls.translate_field(source, path=path, forward=forward)
 
+    @classmethod
+    def untranslate_aggregates(cls, es_response):
+        """
+        Untranslate aggregated field values using the ES field name stored in the 'meta' field
```
hannes-ucsc: I think "untranslate" is not a word.

src/azul/transformer.py
danielsotirhos: It is necessary to specify the 'myTerms' property by name here due to the class BucketData having an iter function that iterates over the object's 'bucket' property only.

hannes-ucsc: Hmm, I doubt that there is no less brittle way to iterate over the aggregation. Remember how we  assumed we had to use protected members which turned out to be false upon closer inspection of the code? If you really think you can't work around this, please prepare a debugging session during which you and I can verify together if what you assert is correct.

src/azul/transformer.py
danielsotirhos: Reverse translate?

src/azul/transformer.py
danielsotirhos: Digging through the class hierarchy I was able to find `dir()` can be used to get a list of keys

PR #1254, opened 2019-09-04
Require fixup! for PR commit titles

hannes-ucsc:


CONTRIBUTING.rst
```diff
@@ -539,19 +543,19 @@ Pull Requests
   
 * At times it may be necessary to temporarily add a commit to a PR branch e.g.,
   to facilitate testing. These commits should be removed prior to landing the
-  PR and their title is prefixed with ``DELETE ME:``.
+  PR and their title is prefixed with ``drop! ``.
```
jessebrennan: Inconsistent use of spaces after commit prefixes.

.travis.yml
```diff
@@ -25,6 +25,8 @@ script:
   - export BOTO_CONFIG=/dev/null
   - make trufflehog
   - make test
+  - make check_clean
```
jessebrennan: Consider moving these before `make test` since it will give quicker feedback in the event of a failure.

Makefile
```diff
@@ -46,4 +46,19 @@ check_trufflehog:
...
-.PHONY: all hello terraform deploy subscribe everything reindex clean test travis integration_test trufflehog check_trufflehog delete
+check_clean:
+	git diff --exit-code  && git diff --cached --exit-code
+
+check_autosquash:
+	_azul_target_branch="$${TRAVIS_BRANCH:=develop}" \
+	; _azul_merge_base=$$(git merge-base HEAD "$${_azul_target_branch}") \
+	; if GIT_SEQUENCE_EDITOR=: git rebase -i --autosquash "$${_azul_merge_base}"; then \
+	    git reset --hard ORIG_HEAD \
+	    ; true \
```
jessebrennan: Consider echoing a string in either case explaining the conclusion.

hannes-ucsc: I'll add one for the negative case. The positive case should not produce output, following the Unix Mantra of not producing extraneous output.

.travis.yml
hannes-ucsc: The main intent behind `check_clean` is to detect tests that dirty the working copy. So that's why it follows `test`. A failing test is a stronger signal about the quality of a PR than lack of squashability. I wouldn't want to mask a test failure with behind non-squashability. I don't usually even look at PRs with failing tests. I might review one that's not squashable.

CONTRIBUTING.rst
hannes-ucsc: Will do.

hannes-ucsc: Wait, the new version is correct. The prefix must end end in a space. But when I later refer to these types of commits the space can be omitted.

Makefile
hannes-ucsc: I'll add one for the negative case. The positive case should not produce output, following the Unix Mantra of not producing extraneous output.

hannes-ucsc: I added the message for the positive case and silenced the echoing of the make rule.

PR #1259, opened 2019-09-05
Provision Azul health checks in DCP wide composite health check #(1150)

achave11:
Automatically link and unlink route53 health check resources managed by DCP-Monitoring repo. Giving Azul a smoother platform to create and destroy its health checks.

### DIFF
--- a/scripts/manage_health_checks.py
+++ b/scripts/manage_health_checks.py
@@ -1,0 +1,85 @@
+from itertools import chain
+
+import argparse
+from typing import Mapping
+
+import boto3
+import logging
+import json
+import sys
+
+import more_itertools
+
+from azul import config
+from azul.logging import configure_script_logging
+from azul.types import JSON
+
+log = logging.getLogger(__name__)
+
+HealthCheck = JSON
+
+
+class HealthCheckManager:
+
+    def __init__(self):
+        self.route53 = boto3.client('route53')
+
+    def get_health_checks(self) -> Mapping[str, HealthCheck]:
+        paginator = self.route53.get_paginator('list_health_checks')
+        hcs = chain.from_iterable(page['HealthChecks'] for page in paginator.paginate())
+        batch_size = 10  # Route53 lets us get tags for at most ten resources at a time
+        hc_batches = more_itertools.chunked(hcs, batch_size)
+        dcp_hcs = {}
+        for hc_batch in hc_batches:
+            hc_batch = {hc['Id']: hc for hc in hc_batch}
+            response = self.route53.list_tags_for_resources(ResourceType='healthcheck',
+                                                            ResourceIds=list(hc_batch.keys()))
+            for tag_set in response['ResourceTagSets']:
+                assert tag_set['ResourceType'] == 'healthcheck'
+                for tag in tag_set['Tags']:
+                    if tag['Key'] == 'Name':
+                        hc_name = tag['Value']
+                        hc_id = tag_set['ResourceId']
+                        dcp_hcs[hc_name] = hc_batch[hc_id]
+        return dcp_hcs
+
+    def provision_health_check(self, all_hcs: Mapping[str, HealthCheck], dcp_hc: HealthCheck, link: bool):
+        dcp_child_hcs = set(dcp_hc['HealthCheckConfig']['ChildHealthChecks'])
+        azul_child_hcs = {all_hcs[hc_name]['Id'] for hc_name in (config.service_name,
+                                                                 config.indexer_name,
+                                                                 config.data_browser_name,
+                                                                 config.data_portal_name)}
+
+        new_dcp_child_hcs = dcp_child_hcs - azul_child_hcs if not link else dcp_child_hcs | azul_child_hcs
+        if new_dcp_child_hcs == dcp_child_hcs:
+            log.info('DCP-wide composite health check up to date, no changes required!')
+        else:
+            response = self.route53.update_health_check(HealthCheckId=dcp_hc['Id'],
+                                                        HealthCheckVersion=dcp_hc['HealthCheckVersion'],
+                                                        ChildHealthChecks=list(new_dcp_child_hcs),
+                                                        Inverted=dcp_hc['HealthCheckConfig']['Inverted'],
+                                                        HealthThreshold=len(dcp_child_hcs))
+            log.info("Update response: " + json.dumps(response))
+
+
+def main(argv):
+    manager = HealthCheckManager()
+    parser = argparse.ArgumentParser(description='Dynamically reference and dereference health checks '
+                                                 'not managed by Azul.')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--link', '-L', dest='link', action='store_true', default=False,
+                       help='Link Azul-managed health check resources to DCP-wide composite health check.')
+    group.add_argument('--unlink', '-U', dest='link', action='store_false', default=False,
+                       help='Unlink Azul-managed health check resources to DCP-wide composite health check.')
+    options = parser.parse_args(argv)
+    hcs = manager.get_health_checks()
+    dcp_health_deployment_name = f'dcp-health-check-{config.deployment_stage}'
+    if dcp_health_deployment_name in hcs:
+        manager.provision_health_check(hcs, hcs[dcp_health_deployment_name], options.link)
+    else:
+        log.info('DCP wide health check does not exist for ' + config.deployment_stage)
+
+
+if __name__ == '__main__':
+    configure_script_logging(log)
+    main(sys.argv[1:])
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -14,12 +14,22 @@ plan: validate
 	terraform plan
 
 apply: validate
+	$(MAKE) unlink_health_checks
 	terraform apply
+	$(MAKE) link_health_checks
 	$(MAKE) grafana
 
 auto_apply: validate
+	$(MAKE) unlink_health_checks
 	terraform apply -auto-approve
+	$(MAKE) link_health_checks
 	$(MAKE) grafana
+
+link_health_checks:
+	python $(azul_home)/scripts/manage_health_checks.py --link
+
+unlink_health_checks:
+	python $(azul_home)/scripts/manage_health_checks.py --unlink
 
 grafana:
 	python $(azul_home)/scripts/update_grafana.py
@@ -34,4 +44,5 @@ auto_destroy: validate
 clean:
 	rm -f *.tf.json
 
-.PHONY: all config init validate plan apply auto_apply grafana destroy auto_destroy clean
+.PHONY: all config init validate plan apply auto_apply link_health_checks unlink_health_checks grafana \
+        destroy auto_destroy clean
--- a/terraform/route53_health_check.tf.json.template.py
+++ b/terraform/route53_health_check.tf.json.template.py
@@ -59,6 +59,11 @@ emit(None if not config.enable_monitoring else {
                         "request_interval": "30",
                         "tags": {
                             "Name": full_name
+                        },
+                        # This is necessary only because of a Terraform bug:
+                        # https://github.com/hashicorp/terraform/issues/22171
+                        "lifecycle": {
+                            "create_before_destroy": True
                         }
                     }
                 }


scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+
+def health_check_ops():
+    client = boto3.client('route53')
+    paginator = client.get_paginator('list_health_checks')
+    return client, chain.from_iterable(page['HealthChecks'] for page in paginator.paginate())
+
+
+def provision_health_check(health_checks, dcp_health_check, reference):
+    dcp_health_check_config = dcp_health_check['HealthCheckConfig']
+    azul_service = health_checks[config.service_name]
```
jessebrennan: If you're spinning up a deployment for the first time, the Azul health checks won't exist. The script will fail here. :hatching_chick: :egg:. Make sure the script can handle this case.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    azul_indexer = health_checks[config.indexer_name]
+    azul_data_browser = health_checks[config.data_browser_name]
```
jessebrennan: In any case, I think the script should be able to handle the case where the Data Browser / Portal health checkers aren't up.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+        updated_child_health_checks.append(azul_service['Id'])
+        updated_child_health_checks.append(azul_indexer['Id'])
+        updated_child_health_checks.append(azul_data_browser['Id'])
+        updated_child_health_checks.append(azul_data_portal['Id'])
+    response = update_dcp_wide_health_check(dcp_health_check, updated_child_health_checks)
+    logger.info(json.dumps(response))
+
+
+def update_dcp_wide_health_check(dcp_wide_health_check, updated_child_health_checks):
+    client = boto3.client('route53')
```
jessebrennan: I _think_ the Boto client is cached, but still I think it's more elegant to only instantiate it once. One way to do this is to put all of these functions into a class and keep the client as a class attribute. Actually I recommend doing this.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def health_check_ops():
+    client = boto3.client('route53')
+    paginator = client.get_paginator('list_health_checks')
+    return client, chain.from_iterable(page['HealthChecks'] for page in paginator.paginate())
+
+
+def provision_health_check(health_checks, dcp_health_check, reference):
+    dcp_health_check_config = dcp_health_check['HealthCheckConfig']
```
jessebrennan: I think this variable is confusing and unhelpful. Instead consider
```suggestion
    dcp_health_check_children = dcp_health_check['HealthCheckConfig']['ChildHealthChecks']
```
And then removing the indexing where the variable is used.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    azul_service = health_checks[config.service_name]
+    azul_indexer = health_checks[config.indexer_name]
+    azul_data_browser = health_checks[config.data_browser_name]
+    azul_data_portal = health_checks[config.data_portal_name]
+    updated_child_health_checks = [health_check for health_check in dcp_health_check_config['ChildHealthChecks']
+                                   if health_check not in (azul_service['Id'],
```
jessebrennan: Save the tuple `(azul_service['Id'], ..., azul_data_portal['Id'])` as a list instead. Then the code at line 71 becomes much more concise---you just join the lists.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    return client, chain.from_iterable(page['HealthChecks'] for page in paginator.paginate())
+
+
+def provision_health_check(health_checks, dcp_health_check, reference):
+    dcp_health_check_config = dcp_health_check['HealthCheckConfig']
+    azul_service = health_checks[config.service_name]
+    azul_indexer = health_checks[config.indexer_name]
+    azul_data_browser = health_checks[config.data_browser_name]
+    azul_data_portal = health_checks[config.data_portal_name]
+    updated_child_health_checks = [health_check for health_check in dcp_health_check_config['ChildHealthChecks']
```
jessebrennan: The prefix `updated_` for this variable is confusing to me.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    if reference:
+        updated_child_health_checks.append(azul_service['Id'])
+        updated_child_health_checks.append(azul_indexer['Id'])
+        updated_child_health_checks.append(azul_data_browser['Id'])
+        updated_child_health_checks.append(azul_data_portal['Id'])
+    response = update_dcp_wide_health_check(dcp_health_check, updated_child_health_checks)
+    logger.info(json.dumps(response))
+
+
+def update_dcp_wide_health_check(dcp_wide_health_check, updated_child_health_checks):
```
jessebrennan: Again, I think it would make more sense to call `updated_child_health_checks` just `child_health_checks`.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                                   if health_check not in (azul_service['Id'],
+                                                           azul_indexer['Id'],
+                                                           azul_data_browser['Id'],
+                                                           azul_data_portal['Id'])]
+    if reference:
+        updated_child_health_checks.append(azul_service['Id'])
+        updated_child_health_checks.append(azul_indexer['Id'])
+        updated_child_health_checks.append(azul_data_browser['Id'])
+        updated_child_health_checks.append(azul_data_portal['Id'])
+    response = update_dcp_wide_health_check(dcp_health_check, updated_child_health_checks)
```
jessebrennan: The way this is written, the health check will be updated even if there are no changes being made. Consider changing it so that `child_health_checks` is compared with what is there already and only updated in the case that it needs updating.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                                                 'not managed by Azul.')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--reference', '-R', dest='reference', action='store_true', default=False,
+                       help='Link Azul managed Route53 health check resources to DCP wide composite health check.')
+    group.add_argument('--dereference', '-D', dest='reference', action='store_false', default=False,
+                       help='Unlink Azul managed Route53 health check resources to DCP wide composite health check.')
+    options = parser.parse_args(argv)
+    health_checks = get_health_checks()
+    try:
+        dcp_wide_health_check = health_checks[f'dcp-health-check-{config.deployment_stage}']
```
danielsotirhos: A try/except block seems unnecessary to check if a key exists in a dict. Why not simply:
```suggestion
        if key in health_checks
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+            assert tag_sets['ResourceType'] == 'healthcheck'
+            for tag in tag_sets['Tags']:
+                if tag['Key'] == 'Name':
+                    health_check_name = tag['Value']
+                    health_check_id = tag_sets['ResourceId']
+                    dcp_health_checks[health_check_name] = health_check_batch[health_check_id]
+    return dcp_health_checks
+
+
+def health_check_ops():
```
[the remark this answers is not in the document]
danielsotirhos: Or possibly remove `health_check_ops` and move that code into `get_health_checks`. Having it as a separate function buries the param literals 'route53' and 'list_health_checks' further down in the code.

scripts/provision_health_checks.py
achave11: There is precedent for instantiating boto3 a few times, since it is an inexpensive call, contrary to passing it around across methods.

scripts/provision_health_checks.py
achave11: Updated the list name, I think for a method name it is descriptive. Let me know if you disagree.

scripts/provision_health_checks.py
achave11: I decided to keep it as is, I think calling the dic with the known key has a look up of O(1) and just falling if it is not found contrary to iterating through all the keys in the dic O(n). Let me know if you disagree.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    except KeyError:
+        logger.info(f'DCP wide health check does not exist for {config.deployment_stage}')
+    else:
+        logger.info('Referencing HealthChecks')
+        provision_health_check(health_checks, dcp_wide_health_check, options.reference)
+
+
+def get_health_checks():
+    client, health_checks = health_check_ops()
+    health_check_batches = more_itertools.chunked(health_checks, 10)
```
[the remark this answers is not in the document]
achave11: I believe that the comment is unnecessary. There is other boto3 calls in our codebase that
have a similar batched structure and do not contain comments. I do not think that this line obscures what is happening to the `health_checks` var.

scripts/provision_health_checks.py
achave11: I think that the scenario you describe is highly unlikely. Since the Azul repo is the one managing the portal's monitoring resources. If there are no Azul resources there should not be any browser resources either.

scripts/provision_health_checks.py
achave11: Correct I had a similar concern and addressed it with Hannes. There is no way to actuate a "as a per needed" logic with out using the terraform provisioning.

scripts/provision_health_checks.py
jessebrennan: Yeah, I just meant the parameter, The function name is good.

scripts/provision_health_checks.py
jessebrennan: Checking membership in a dictionary is O(1). Either way works well, though I've heard it argued "more Pythonic" to do the try-except.

scripts/provision_health_checks.py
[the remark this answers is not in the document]
jessebrennan: I was perplexed by the number 10, until I remember overhearing you and @hannes-ucsc talking about the batch size. Other's reading the code won't have that context. Also it's generally a good practice to document "magic numbers" like this.

scripts/provision_health_checks.py
jessebrennan: It doesn't look like the name was actually updated.

scripts/provision_health_checks.py
```diff
@@ -51,30 +52,22 @@ def get_health_checks():
...
-    response = update_dcp_wide_health_check(dcp_health_check, updated_child_health_checks)
-    logger.info(json.dumps(response))
+    dcp_health_check_children = dcp_health_check['HealthCheckConfig']['ChildHealthChecks']
+    try:
+        azul_health_checks = [health_checks[resource_name]['Id'] for resource_name in (config.service_name,
+                                                                                       config.indexer_name,
+                                                                                       config.data_browser_name,
+                                                                                       config.data_portal_name)]
+    except KeyError:
+        logger.info(f'No Azul health checks found for {config.deployment_stage} deployment!')
```
jessebrennan: It's technically possible that some of the health checks are not provisioned, but the others are. Consider if you want to explicitly handle this situation.

scripts/provision_health_checks.py
jessebrennan: Yes there is... You have the variable `dcp_health_check_children = dcp_health_check['HealthCheckConfig']['ChildHealthChecks']` which tells you exactly which health checks are being provisioned. If this is the same as the health checks passed into `update_dcp_wide_health_check`, then you don't need to call update.

In any case, this is an optional optimization, but you should consider it.

scripts/provision_health_checks.py
```diff
@@ -54,20 +54,19 @@ def get_health_checks():
...
-        if reference:
-            dcp_health_check_children.extend(azul_health_checks)
-        response = update_dcp_wide_health_check(dcp_health_check, dcp_health_check_children)
-        logger.info(json.dumps(response))
+    azul_health_checks = []
+    for resource_name in (config.service_name, config.indexer_name, config.data_browser_name, config.data_portal_name):
+        try:
+            azul_health_checks.append(health_checks[resource_name]['Id'])
+        except KeyError:
+            logger.info(f'No Azul health checks found for {resource_name}')
```
jessebrennan: I wonder if this should be `logger.warning`. Oh! also use `log` instead of `logger` (as per style guide https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#logging).

scripts/provision_health_checks.py
```diff
@@ -14,71 +14,76 @@
...
-                    health_check_id = tag_sets['ResourceId']
-                    dcp_health_checks[health_check_name] = health_check_batch[health_check_id]
-    return dcp_health_checks
-
-
-def provision_health_check(health_checks, dcp_health_check, reference):
-    dcp_health_check_children = dcp_health_check['HealthCheckConfig']['ChildHealthChecks']
-    azul_health_checks = []
-    for resource_name in (config.service_name, config.indexer_name, config.data_browser_name, config.data_portal_name):
+    def main(self, argv):
```
jessebrennan: I would suggest leaving `main()` as a function instead of a member of the class. That leaves the class as a self contained Python interface to the provisioning, and `main` exists separately to add a CLI.

This pattern is used in https://github.com/DataBiosphere/azul/blob/develop/scripts/provision_credentials.py and https://github.com/DataBiosphere/azul/blob/develop/scripts/upload-tsv-files.py.

In those two particular cases, `main` is not a separate function, but I think it's more than practical to keep it separate.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                                               HealthCheckVersion=dcp_wide_health_check['HealthCheckVersion'],
+                                               ChildHealthChecks=updated_child_health_checks,
+                                               Inverted=dcp_wide_health_check['HealthCheckConfig']['Inverted'],
+                                               HealthThreshold=len(updated_child_health_checks))
+
+
+def main(argv):
+    provision_health_checks = ProvisionHealthChecks()
+    parser = argparse.ArgumentParser(
+        description='Dynamically reference or dereference Azul\'s Route53 health checks '
```
hannes-ucsc: Bad indention.

hannes-ucsc: Backslash can be avoided.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                    'to/from composite health checkers not managed by Azul.')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--reference', '-R', dest='reference', action='store_true', default=False,
```
hannes-ucsc:
```suggestion
    group.add_argument('--link', '-R', dest='reference', action='store_true', default=False,
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                       help='Link Azul managed Route53 health check resources to DCP wide composite health check.')
+    group.add_argument('--dereference', '-D', dest='reference', action='store_false', default=False,
```
hannes-ucsc:
```suggestion
    group.add_argument('--unlink', '-D', dest='reference', action='store_false', default=False,
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def main(argv):
+    provision_health_checks = ProvisionHealthChecks()
+    parser = argparse.ArgumentParser(
+        description='Dynamically reference or dereference Azul\'s Route53 health checks '
+                    'to/from composite health checkers not managed by Azul.')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--reference', '-R', dest='reference', action='store_true', default=False,
+                       help='Link Azul managed Route53 health check resources to DCP wide composite health check.')
```
hannes-ucsc:
```suggestion
                       help='Link Azul-managed health check resources to DCP-wide composite health check.')
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    group.add_argument('--dereference', '-D', dest='reference', action='store_false', default=False,
+                       help='Unlink Azul managed Route53 health check resources to DCP wide composite health '
```
hannes-ucsc: See abvoe

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                            'check.')
+    options = parser.parse_args(argv)
+    health_checks = provision_health_checks.get_health_checks
+    try:
+        dcp_wide_health_check = health_checks[f'dcp-health-check-{config.deployment_stage}']
+    except KeyError:
+        log.info(f'DCP wide health check does not exist for {config.deployment_stage}')
```
hannes-ucsc:
```suggestion
        log.info(f'DCP-wide health check does not exist for {config.deployment_stage}')
```

hannes-ucsc: Should this be a warning?

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+from azul import config
+from azul.logging import configure_script_logging
+
+log = logging.getLogger(__name__)
+
+
+class ProvisionHealthChecks:
+
+    def __init__(self):
+        self.client = boto3.client('route53')
```
hannes-ucsc:
```suggestion
        self.route53 = boto3.client('route53')
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+    def get_health_checks(self):
+        paginator = self.client.get_paginator('list_health_checks')
+        health_checks = chain.from_iterable(page['HealthChecks'] for page in paginator.paginate())
+        health_check_batches = more_itertools.chunked(health_checks, 10)
+        dcp_health_checks = {}
+        for health_check_batch in health_check_batches:
+            health_check_batch = {health_check['Id']: health_check for health_check in health_check_batch}
+            response = self.client.list_tags_for_resources(ResourceType='healthcheck',
+                                                           ResourceIds=list(health_check_batch.keys()))
+            for tag_sets in response['ResourceTagSets']:
```
hannes-ucsc:
```suggestion
            for tag_set in response['ResourceTagSets']:
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ProvisionHealthChecks:
+
+    def __init__(self):
+        self.client = boto3.client('route53')
+
+    @property
```
hannes-ucsc:
```suggestion
```

Add type hints for clarity.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                    if tag['Key'] == 'Name':
+                        health_check_name = tag['Value']
+                        health_check_id = tag_sets['ResourceId']
+                        dcp_health_checks[health_check_name] = health_check_batch[health_check_id]
+        return dcp_health_checks
+
+    def provision_health_check(self, health_checks, dcp_health_check, reference):
+        dcp_health_check_children = dcp_health_check['HealthCheckConfig']['ChildHealthChecks']
+        azul_health_checks = []
+        for resource_name in (config.service_name,
```
hannes-ucsc:
```suggestion
        for health_check_name in (config.service_name,
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                              config.indexer_name,
+                              config.data_browser_name,
+                              config.data_portal_name):
+            try:
+                azul_health_checks.append(health_checks[resource_name]['Id'])
```
hannes-ucsc:
```
azul_health_checks.append(health_checks[resource_name]['Id'])
                  ^      ^             ^              ^
```

are the places where technically a KeyError could occur. Only one should be caught and handled in this way. The other ones should cause the program to exit with a Traceback and a non-zero status code (the default behavior for unhandled exceptions).

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+
+import more_itertools
+
+from azul import config
+from azul.logging import configure_script_logging
+
+log = logging.getLogger(__name__)
+
+
+class ProvisionHealthChecks:
```
hannes-ucsc:
```suggestion
class HealthCheckManager:
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+        dcp_wide_health_check['HealthCheckConfig'].pop('ChildHealthChecks')
+        return self.client.update_health_check(HealthCheckId=dcp_wide_health_check['Id'],
+                                               HealthCheckVersion=dcp_wide_health_check['HealthCheckVersion'],
+                                               ChildHealthChecks=updated_child_health_checks,
+                                               Inverted=dcp_wide_health_check['HealthCheckConfig']['Inverted'],
+                                               HealthThreshold=len(updated_child_health_checks))
+
+
+def main(argv):
+    provision_health_checks = ProvisionHealthChecks()
```
hannes-ucsc:
```suggestion
    manager = HealthCheckManager()
```

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                       help='Unlink Azul managed Route53 health check resources to DCP wide composite health '
+                            'check.')
+    options = parser.parse_args(argv)
+    health_checks = provision_health_checks.get_health_checks
+    try:
+        dcp_wide_health_check = health_checks[f'dcp-health-check-{config.deployment_stage}']
+    except KeyError:
+        log.info(f'DCP wide health check does not exist for {config.deployment_stage}')
+    else:
+        provision_health_checks.provision_health_check(health_checks, dcp_wide_health_check, options.reference)
```
hannes-ucsc: Your naming choices lead to very unwieldy lines.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+                log.info(f'No Azul health checks found for {resource_name}')
+
+        dcp_health_check_children = [health_check for health_check in dcp_health_check_children
+                                     if health_check not in azul_health_checks]
+        if reference:
+            dcp_health_check_children.extend(azul_health_checks)
+        response = self.update_dcp_wide_health_check(dcp_health_check, dcp_health_check_children)
+        log.info(json.dumps(response))
+
+    def update_dcp_wide_health_check(self, dcp_wide_health_check, updated_child_health_checks):
```
hannes-ucsc: Inline.

scripts/provision_health_checks.py
```diff
@@ -0,0 +1,92 @@
...
+        dcp_wide_health_check['HealthCheckConfig'].pop('ChildHealthChecks')
```
hannes-ucsc: I don't think this is necessary.

terraform/Makefile
```diff
@@ -14,16 +14,26 @@ plan: validate
 	terraform plan
 
 apply: validate
+	$(MAKE) dereference_health_checks
 	terraform apply
+	$(MAKE) reference_health_checks
```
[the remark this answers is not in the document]
hannes-ucsc: The original solution was better, @jessebrennan.

terraform/Makefile
```diff
@@ -14,16 +14,26 @@ plan: validate
...
 auto_apply: validate
+	$(MAKE) dereference_health_checks
 	terraform apply -auto-approve
+	$(MAKE) reference_health_checks
 	$(MAKE) grafana
 
 grafana:
 	python $(azul_home)/scripts/update_grafana.py
 
+reference_health_checks:
```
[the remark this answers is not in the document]
hannes-ucsc: It's more direct, but not more concise. The Python invocation is complicated enough to warrant a convenience rule around it.

scripts/provision_health_checks.py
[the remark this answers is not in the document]
hannes-ucsc: I agree with @jessebrennan.

```
batch_size = 10  # Route53 let's us get tags for at most ten resources at a time
```

scripts/manage_health_checks.py
```diff
@@ -0,0 +1,82 @@
...
+        return dcp_hcs
+
+    def provision_health_check(self, hcs, dcp_hc, link):
+        dcp_hcs = dcp_hc['HealthCheckConfig']['ChildHealthChecks']
+        azul_hcs = []
+        for hc_name in (config.service_name,
+                        config.indexer_name,
+                        config.data_browser_name,
+                        config.data_portal_name):
+            azul_hcs.append(hcs[hc_name]['Id'])
```
hannes-ucsc: Set comprehension would make this more concise.

scripts/manage_health_checks.py
```diff
@@ -0,0 +1,82 @@
...
+
+        dcp_hcs = list(set(dcp_hcs).difference(set(azul_hcs)) if not link else set(dcp_hcs).union(set(azul_hcs)))
+        if dcp_hcs == dcp_hc['HealthCheckConfig']['ChildHealthChecks']:
+            log.info('Azul health checks are already linked to DCP-wide composite health check!')
```
hannes-ucsc: This message will be confusing in the case of `--unlink`.

scripts/manage_health_checks.py
```diff
@@ -0,0 +1,82 @@
...
+    def provision_health_check(self, hcs, dcp_hc, link):
+        dcp_hcs = dcp_hc['HealthCheckConfig']['ChildHealthChecks']
+        azul_hcs = []
+        for hc_name in (config.service_name,
+                        config.indexer_name,
+                        config.data_browser_name,
+                        config.data_portal_name):
+            azul_hcs.append(hcs[hc_name]['Id'])
+
+        dcp_hcs = list(set(dcp_hcs).difference(set(azul_hcs)) if not link else set(dcp_hcs).union(set(azul_hcs)))
```
hannes-ucsc: See if you can simplify this by making the variables sets instead of converting to sets. 

Also see if you can use operators instead of methods to perform set operations:

```
{1,2}-{2,3}
{1}
```

scripts/manage_health_checks.py
```diff
@@ -0,0 +1,82 @@
...
+        if dcp_hcs == dcp_hc['HealthCheckConfig']['ChildHealthChecks']:
+            log.info('Azul health checks are already linked to DCP-wide composite health check!')
+        else:
+            response = self.route53.update_health_check(HealthCheckId=dcp_hc['Id'],
+                                                        HealthCheckVersion=dcp_hc['HealthCheckVersion'],
+                                                        ChildHealthChecks=dcp_hcs,
+                                                        Inverted=dcp_hc['HealthCheckConfig']['Inverted'],
+                                                        HealthThreshold=len(dcp_hcs))
+            log.info(json.dumps(response))
```
hannes-ucsc: There should always be some English at the beginning of a log message.

scripts/manage_health_checks.py
```diff
@@ -0,0 +1,80 @@
...
+                                                        HealthThreshold=len(dcp_hcs))
+            log.info("Update response: " + json.dumps(response))
+
+
+def main(argv):
+    manager = HealthCheckManager()
+    parser = argparse.ArgumentParser(description='Dynamically reference and dereference health checks '
+                                                 'not managed by Azul.')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--link', '-R', dest='link', action='store_true', default=False,
```
hannes-ucsc:
```suggestion
    group.add_argument('--link', '-L', dest='link', action='store_true', default=False,
```

scripts/manage_health_checks.py
```diff
@@ -0,0 +1,80 @@
...
+                       help='Link Azul-managed health check resources to DCP-wide composite health check.')
+    group.add_argument('--unlink', '-D', dest='link', action='store_false', default=False,
```
hannes-ucsc:
```suggestion
    group.add_argument('--unlink', '-U', dest='link', action='store_false', default=False,
```

achave11:
Will await review until preliminary demo.

PR #1277, opened 2019-09-13
Azul Integration deployment to point at DSS staging instance

achave11:
Since DSS has recently deleted most of its bundles from their integration DSS instance. Azul is unable to obtain results back for a prefix query, due to the small set of bundles available. Most of the bundles available are test bundles. That is why I suggest we direct our integration environment to the DSS staging instance. 
We could also solve the issue by having special logic in our integration test when dealing with the integration DSS instance. 

@hannes-ucsc for further evaluation of this issue.

### DIFF
--- a/deployments/integration/environment
+++ b/deployments/integration/environment
@@ -11,7 +11,7 @@
 # at the project root.
 
 _set AZUL_DEPLOYMENT_STAGE integration
-_set AZUL_DSS_ENDPOINT "https://dss.integration.data.humancellatlas.org/v1"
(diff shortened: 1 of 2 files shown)


hannes-ucsc:
It's a good idea but the point of the `integration` deployment is that all components prove in this deployment that they work together before they are promoted in lock-step to the staging deployment.

hannes-ucsc:
We should consider shortening the suffix for `integration` or copy some bundles from staging to integration.

Issue #1282, opened 2019-09-16
Local integration tests fail with no bundles


jessebrennan:
Since the integration environment has been cleaned up, there are not enough bundles in the DSS for the query Azul makes to consistently return anything.

Recommended quick fix is to remove the prefix if the environment is integration.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-794)
┆Project Name: azul
┆Issue Number: AZUL-794

PR #1285, opened 2019-09-16
Remove prefix for 'integration' integration test (#1282)

jessebrennan:
Since there are few bundles in the integration environment, Azul's integration tests should without a prefix.

This passed the integration tests with the integration environment when I can them locally.


hannes-ucsc:
Taking this on.

hannes-ucsc:
This was merged into integration and the merge commit was pushed. I don't understand why Github doesn't get that.

Issue #1224, opened 2019-08-13
Service health checker times out during reindexing


hannes-ucsc:
It looks like the the ES load incurred by reindexing slows down the service's `/health` endpoint above 4s+2s timeout enforced by AWS Route53 health checkers of HTTP(S) endpoints.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-763)
┆Project Name: azul
┆Issue Number: AZUL-763

hannes-ucsc:
@jessebrennan similar to what you suggested, my current thinking is that we'll add a lambda that's triggered periodically, say, every minute by a CloudWatch event rule. That lambda updates writes the health object to the Azul storage bucket. We'll also add an `/health/cached` endpoint to the service that returns the contents of that object provided that it is no older than 2min. Along those lines.

PR #1297, opened 2019-09-18
Cache health check response to reduce service load (#1224)

achave11:


lambdas/service/app.py
```diff
@@ -162,6 +165,13 @@ def health(keys: Optional[str] = None):
         return {
             'up': True,
         }
+    elif keys == 'cached':
+        health_object = json.loads(storage_service.get('health/service'))
+        two_minutes = 120
```
GPelayo:
```suggestion
        cached_object_lifespan = 60 * 2
```

lambdas/indexer/app.py
```diff
@@ -54,6 +57,13 @@ def health(keys: Optional[str] = None):
         return {
             'up': True,
         }
+    elif keys == 'cached':
+        health_object = json.loads(storage_service.get('health/indexer'))
+        two_minutes = 120
```
GPelayo:
```suggestion
        cached_object_lifespan = 60 * 2
```

lambdas/indexer/app.py
```diff
@@ -54,6 +57,13 @@ def health(keys: Optional[str] = None):
         return {
             'up': True,
         }
+    elif keys == 'cached':
+        health_object = json.loads(storage_service.get('health/indexer'))
+        allowed_cached_lifespan = 2 * 60
```
```suggestion
        max_age = 2 * 60
```

lambdas/indexer/app.py
```diff
@@ -54,6 +57,13 @@ def health(keys: Optional[str] = None):
...
+        if (time.time() - health_object['time']) > allowed_cached_lifespan:
```
hannes-ucsc: redundant parentheses

lambdas/indexer/app.py
```diff
@@ -54,6 +57,13 @@ def health(keys: Optional[str] = None):
...
+            raise RuntimeError('Cached health object is stale!')
```
```suggestion
            raise RuntimeError('Cached health object is stale')
```

lambdas/indexer/app.py
```diff
@@ -54,6 +57,13 @@ def health(keys: Optional[str] = None):
...
+        else:
+            body = health_object['indexer_health']
```
```suggestion
            body = health_object['health']
```

lambdas/indexer/app.py
```diff
@@ -65,8 +75,17 @@ def health(keys: Optional[str] = None):
...
-        return Response(body=json.dumps(body),
-                        status_code=200 if body['up'] else 503)
+    return Response(body=json.dumps(body),
+                    status_code=200 if body['up'] else 503)
+
+
+@app.schedule('rate(1 minute)', name=config.indexer_cache_health_lambda_basename)
+def generate_health_object(event: chalice.app.CloudWatchEvent):
+    assert event is not None  # avoid linter warning
+    indexer_health = Health('indexer')
```
```suggestion
    health = Health('indexer')
```

in order to reduce the differences between the two copies of this method. That will make extracting a single method easier in the future.

lambdas/indexer/app.py
```diff
@@ -65,8 +75,17 @@ def health(keys: Optional[str] = None):
...
             raise BadRequestError('Invalid health keys')
-        return Response(body=json.dumps(body),
-                        status_code=200 if body['up'] else 503)
+    return Response(body=json.dumps(body),
+                    status_code=200 if body['up'] else 503)
+
+
+@app.schedule('rate(1 minute)', name=config.indexer_cache_health_lambda_basename)
+def generate_health_object(event: chalice.app.CloudWatchEvent):
+    assert event is not None  # avoid linter warning
```
hannes-ucsc: This should not be necessary. Can you post screen shot with the warning? I assume by "linter" you mean PyCharm, right?

lambdas/indexer/app.py
```diff
@@ -65,8 +75,17 @@ def health(keys: Optional[str] = None):
...
+    indexer_health = Health('indexer')
+    indexer_health = indexer_health.as_json()
+    health_object = {'time': time.time(), 'indexer_health': indexer_health}
```
hannes-ucsc: If you do a one-liner , I would use `dict()` instead of `{}`

lambdas/indexer/lambda-policy.json.template.py
```diff
@@ -86,5 +86,15 @@
...
             ]
         },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObject",
+                "s3:PutObject",
+            ],
+            "Resource": [
+                f"arn:aws:s3:::{config.s3_bucket}/health/indexer",
```
hannes-ucsc:
```suggestion
                f"arn:aws:s3:::{config.s3_bucket}/health/*",
```

test/health_check_test_case.py
```diff
@@ -84,6 +89,35 @@ def test_health_endpoint_keys(self):
...
+                app.generate_health_object(MagicMock(), MagicMock())
+                response = requests.get(self.base_url + '/health/cached')
+                self.assertEqual(200, response.status_code)
+
+        future_time = time.time() + 3 * 60
+        with ResponsesHelper() as helper:
+            helper.add_passthru(self.base_url)
+            with mock.patch('time.time', new=lambda: future_time):
+                response = requests.get(self.base_url + '/health/cached')
+                self.assertEqual(500, response.status_code)
```
hannes-ucsc: Can you assert the body here? It should contain "Cached health object is stale". I don't remember if raising an exception produces a Chalice response whose body is the message from the exception. But if that doesn't work, you can always `return Response(body="Cached health object is stale", status_code=500)` instead of raising.

test/health_check_test_case.py
```diff
@@ -84,6 +89,35 @@ def test_health_endpoint_keys(self):
...
+            response = requests.get(self.base_url + '/health/cached')
+            self.assertEqual(500, response.status_code)
+
+        self._create_mock_queues()
+        endpoint_states = self._make_endpoint_states(self.endpoints)
+        app = load_app_module(self.lambda_name())
+        with ResponsesHelper() as helper:
+            helper.add_passthru(self.base_url)
+            self._mock_service_endpoints(helper, endpoint_states)
+            with mock.patch.dict(os.environ, AWS_DEFAULT_REGION='us-east-1'):
```
hannes-ucsc: The need for this isn't immediately apparent to me. Can you explain in a comment? Or, if it should be ovbious, let me know by replying to this comment.

test/health_check_test_case.py
```diff
@@ -84,6 +89,35 @@ def test_health_endpoint_keys(self):
...
+    @mock_s3
+    @mock_sts
+    @mock_sqs
+    def test_cached_health(self):
+        storage_service = StorageService()
+        storage_service.create_bucket()
+        with ResponsesHelper() as helper:
+            helper.add_passthru(self.base_url)
+            response = requests.get(self.base_url + '/health/cached')
+            self.assertEqual(500, response.status_code)
```
hannes-ucsc: If possible, this section should assert that the error is because the object is missing. That would help with clarifying the intent of the test, too. At the very least there should be a comment explaining each of the three sections of this test.

lambdas/indexer/app.py
achave11: Screen Shot 2019-09-25 at 11 18 28 AM

achave11: @hannes-ucsc 
Replicating [nudge lambda](https://github.com/DataBiosphere/azul/blob/45eb2228fe4c648cec06de27c3f30d2ee057d5b2/lambdas/indexer/app.py#L452) work.

test/health_check_test_case.py
achave11: Checkout: https://api.travis-ci.org/v3/job/587629946/log.txt
The patch is required for all the health test that access `api_endpoints`, in most cases test that involve the service.

lambdas/indexer/app.py
hannes-ucsc: Well, the person who committed that must have been intellectually challenged.

I've since learned that naming the parameter with a leading underscore also suppresses that warning.

lambdas/indexer/app.py
```diff
@@ -65,8 +75,17 @@ def health(keys: Optional[str] = None):
...
+    return Response(body=json.dumps(body),
+                    status_code=200 if body['up'] else 503)
+
+
+@app.schedule('rate(1 minute)', name=config.indexer_cache_health_lambda_basename)
+def generate_health_object(event: chalice.app.CloudWatchEvent):
+    assert event is not None  # avoid linter warning
+    health = Health('indexer')
+    health = health.as_json()
+    health_object = dict([('time', time.time()), ('health', health)])
```
hannes-ucsc: I meant

```suggestion
    health_object = dict(time=time.time(), health=health])
```

test/health_check_test_case.py
hannes-ucsc: I guess the primary error is not exposed.

test/health_check_test_case.py
hannes-ucsc: I see. I'm glad I asked. In that case, let's make _mock_service_endpoints an @contextmanager and move the region patching into that method …

test/health_check_test_case.py
```diff
@@ -84,6 +89,41 @@ def test_health_endpoint_keys(self):
...
+            self.assertEqual({'Code': 'InternalServerError',
+                              'Message': 'An internal server error occurred.'}, response.json())
+
+        # A successful response is obtained when all the systems are functional
+        self._create_mock_queues()
+        endpoint_states = self._make_endpoint_states(self.endpoints)
+        app = load_app_module(self.lambda_name())
+        with ResponsesHelper() as helper:
+            helper.add_passthru(self.base_url)
+            self._mock_service_endpoints(helper, endpoint_states)
```
hannes-ucsc: … so this line would read 

```suggestion
            with self._mock_service_endpoints(helper, endpoint_states):
```

lambdas/service/app.py
```diff
@@ -162,6 +165,13 @@ def health(keys: Optional[str] = None):
         return {
             'up': True,
         }
+    elif keys == 'cached':
+        health_object = json.loads(storage_service.get('health/service'))
+        max_age = 2 * 60
+        if time.time() - health_object['time'] > max_age:
+            return Response(body=json.dumps({'Message': 'Cached health object is stale'}), status_code=500)
```
```suggestion
            return Response(status_code=500, 
                            body=json.dumps({'Message': 'Cached health object is stale'}))
```

Just looks better, IMO

test/health_check_test_case.py
hannes-ucsc: Also, it looks as if `boto3.resource('sqs')` is the culprit that requires the region to be set. In _mock_service_endpoints I would mention that fact so the next person isn't puzzled like I was.

test/health_check_test_case.py
```diff
@@ -176,16 +215,19 @@ def _test(self, endpoint_states: Mapping[str, bool], lambdas_up: bool, path: str
...
                 return requests.get(self.base_url + path)
 
+    @contextmanager
     def _mock_service_endpoints(self, helper: ResponsesHelper, endpoint_states: Mapping[str, bool]) -> None:
-        for endpoint, endpoint_up in endpoint_states.items():
-            helper.add(responses.Response(method='HEAD',
-                                          url=config.service_endpoint() + endpoint,
-                                          status=200 if endpoint_up else 503,
-                                          json={}))
+        with mock.patch.dict(os.environ, AWS_DEFAULT_REGION='us-east-1'):
```
hannes-ucsc: This could move down to right before the yield. Then you wouldn't have to indent the loop.

lambdas/indexer/app.py
```diff
@@ -33,6 +34,8 @@
 
 app = AzulChaliceApp(app_name=config.indexer_name)
 
+storage_service = StorageService()
```
hannes-ucsc: While this probably makes sense in the long run, it is inconsistent with where this service is currently instantiated, which is within the individual methods that use the service. I think we should stick with this pattern for now. Sorry for not noticing this earlier.

lambdas/service/app.py
```diff
@@ -42,6 +43,8 @@
 
 app = AzulChaliceApp(app_name=config.service_name)
 
+storage_service = StorageService()
```
hannes-ucsc: Same here.

lambdas/service/app.py
```diff
@@ -41,7 +42,6 @@
 log = logging.getLogger(__name__)
 
 app = AzulChaliceApp(app_name=config.service_name)
-
```
hannes-ucsc: Superfluous change.

Issue #1298, opened 2019-09-18
Ensure teardown runs when tests are cancelled


jessebrennan:
This indicates that it may be possible https://docs.python.org/3.6/library/unittest.html#signal-handling

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-801)
┆Project Name: azul
┆Issue Number: AZUL-801

PR #1300, opened 2019-09-19
Ensure tearDown runs when tests are cancelled (#1298)

jessebrennan:


test/azul_test_case.py
```diff
@@ -1,10 +1,49 @@
...
+from unittest import TestCase
 
 import boto3.session
 
 
-class AzulTestCase(TestCase):
+class SafeTestCase(TestCase):
+    """
+    SafeTestCase makes sure that tearDown / cleanup methods are always run when
+    They should be.
```
hannes-ucsc:
```suggestion
    they should be.
```

test/azul_test_case.py
```diff
@@ -1,10 +1,49 @@
...
+
+    This means that
```
hannes-ucsc:
```suggestion
    This means that
    
```

hannes-ucsc: here and everywhere

test/azul_test_case.py
```diff
@@ -1,10 +1,49 @@
-import os
-from unittest import TestCase, mock
+from unittest import TestCase
 
 import boto3.session
 
 
-class AzulTestCase(TestCase):
+class SafeTestCase(TestCase):
```
hannes-ucsc:
```suggestion
class AlwaysTearDownTestCase(TestCase):
```

I find "Safe" a little bit presumptuous. Whether one approach is "safer" than the other is debatable and depends on the circumstances. I prefer to be more specific because the verbosity of the name isn't hurting in this case.

test/azul_test_case.py
```diff
@@ -1,10 +1,49 @@
...
+      not be run. (Working around this involves making a custom TestSuite which
+      will be ignored if you run tests through PyCharm)
+    """
+
+    def run(self, result=None):
+        test_method = getattr(self, self._testMethodName)
+        wrapped_test = self._cleanup_wrapper(test_method, KeyboardInterrupt)
+        setattr(self, self._testMethodName, wrapped_test)
+
+        self.setUp = self._cleanup_wrapper(self.setUp, BaseException)
```
hannes-ucsc: What about exceptions in setupClass()? Should an exception in that method still lead to a call of tearDownClass()?

test/azul_test_case.py
```diff
@@ -1,10 +1,49 @@
...
+    They should be.
+
+    This means that
+    - if a KeyboardInterrupt is raised in a test, then tearDown, tearDownClass,
+      and tearDownModule will all still run
+    - if any exception is raised in a setUp, then tearDown, tearDownClass,
+      and tearDownModule will all still run
+
+    Caveats:
+    - All tearDown methods MUST be idempotent. It's possible a step of tearDown
```
hannes-ucsc: I don't think that's accurate. They only need to handle the case that the corresponding setUp() hasn't finished. An idempotent tearDown is likely to have that property but I'm not sure they are equivalent.

test/azul_test_case.py
jessebrennan: @hannes-ucsc see the second caveat on line 20.

jessebrennan:
> Is there a way we can test/demonstrate this?

@hannes-ucsc it can be demonstrated easily, but I had trouble thinking of how to test it. Maybe a doc test? Maybe with mocking

test/azul_test_case.py
jessebrennan: Ahh, I see. It's a subtle distinction, but I think yes, you're correct.

test/azul_test_case.py
```diff
@@ -1,9 +1,50 @@
-import os
-from unittest import TestCase, mock
+from unittest import TestCase
 
 import boto3.session
 
 
+class AlwaysTearDownTestCase(TestCase):
+    """
+    SafeTestCase makes sure that tearDown / cleanup methods are always run when
```
hannes-ucsc:
```suggestion
    AlwaysTearDownTestCase makes sure that tearDown / cleanup methods are always run when
```

That's why I avoid duplicating class and method names in doc strings. So really,

```suggestion
    Makes sure that tearDown / cleanup methods are always run when
```

test/azul_test_case.py
```diff
@@ -1,9 +1,50 @@
...
+    - if a KeyboardInterrupt is raised in a test, then tearDown, tearDownClass,
+      and tearDownModule will all still run
+    - if any exception is raised in a setUp, then tearDown, tearDownClass,
+      and tearDownModule will all still run
+
+    Caveats:
+
+    - All tearDown methods should pass even if their corresponding setUps don't
+      run at all, as in the case of a KeyboardInterrupt or other exception.
+    - If an exception is raised in SetUpClass or SetUpModule, tearDowns will
```
hannes-ucsc:
```suggestion
    - If an exception is raised in setUpClass or setUpModule, the corresponding tear-down method will
```

test/azul_test_case.py
```diff
@@ -1,9 +1,50 @@
...
+      not be run. (Working around this involves making a custom TestSuite which
```
hannes-ucsc: The part in parentheses is a little bit hypothetical. I'd remove it.

test/local_integration_test.py
```diff
@@ -343,7 +343,7 @@ def _get_entities_by_project(self, entity_type, project_short_name):
         return entities
 
 
-class OpenAPIIntegrationTest(unittest.TestCase):
+class OpenAPIIntegrationTest(AlwaysTearDownTestCase):
```
hannes-ucsc: I wouldn't even do this one …

test/local_integration_test.py
```diff
@@ -358,7 +358,7 @@ def test_openapi(self):
         validate_spec(spec)
 
 
-class DSSIntegrationTest(unittest.TestCase):
+class DSSIntegrationTest(AlwaysTearDownTestCase):
```
hannes-ucsc: … or this one.

test/security/test_authenticator.py
```diff
@@ -15,10 +12,11 @@
                                          InvalidTokenError,
                                          NonDecodableTokenError,
                                          InvalidRedirectUriError)
+from azul_test_case import AlwaysTearDownTestCase
 from retorts import AuthResponseHelper, TestKeyManager
 
 
-class AuthenticatorTestCase(TestCase):
+class AuthenticatorTestCase(AlwaysTearDownTestCase):
```
hannes-ucsc: And why this one?

test/azul_test_case.py
```diff
@@ -1,9 +1,50 @@
...
+    def run(self, result=None):
+        test_method = getattr(self, self._testMethodName)
+        wrapped_test = self._cleanup_wrapper(test_method, KeyboardInterrupt)
+        setattr(self, self._testMethodName, wrapped_test)
+
+        self.setUp = self._cleanup_wrapper(self.setUp, BaseException)
+
+        return super().run(result)
+
+    def _cleanup_wrapper(self, method, exception):
```
hannes-ucsc: `exception` indicates an instance. But this is a class. So
```suggestion
    def _cleanup_wrapper(self, method, exception_cls):
```

test/local_integration_test.py
jessebrennan: My thought was to just always use `AlwaysTearDownTestCase`. It seemed to me like the best practice since anyone could cancel any test at any time and we don't want to risk leaving around any resources.

I'm curious why you disagree.

jessebrennan:
@hannes-ucsc I had to squash down my changes in order to simplify rebasing on to develop.

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
     def tearDownClass(cls) -> None:
         boto3.session.Session.get_credentials = cls.get_credentials_orig
         super().tearDownClass()
+
+
+class Test:
```
hannes-ucsc:
```suggestion
class Hidden:
```

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
...
+    # Keep this test case out of the main namespace
+
+    class LoggingTestCase(TestCase):
```
hannes-ucsc:
```suggestion
    class TracingTestCase(TestCase):
```

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
...
+        """A test case which logs its calls."""
```
hannes-ucsc:
```suggestion
        """A test case which traces its calls."""
```

hannes-ucsc: I'm ok with leaving this in but just know that `mock` does this magically.

See https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.method_calls

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
...
+            self.events.append('tearDown')
+
+
+class TestAlwaysTearDownTestCase(TestCase):
+
+    def test_regular_execution_order(self):
+        for always_tear_down in (False, True):
+            events = []
+            tear_down_parent = [AlwaysTearDownTestCase] if always_tear_down else []
+            with self.subTest(always_tear_down=always_tear_down):
```
hannes-ucsc: I'd move this up to right after the `for` loop. That way it doesn't disrupt the flow.

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
...
+
+                class TC(*([Test.LoggingTestCase] + tear_down_parent)):
```
hannes-ucsc:
```suggestion
                class TC(Test.LoggingTestCase, *tear_down_parent):
```

???

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
...
+            self.events.append('test')
+
+        def tearDown(self):
+            self.events.append('tearDown')
+
+
+class TestAlwaysTearDownTestCase(TestCase):
+
+    def test_regular_execution_order(self):
+        for always_tear_down in (False, True):
```
hannes-ucsc:
```suggestion
        for always_tear_down in False, True:
```

test/local_integration_test.py
hannes-ucsc: We don't know which of the existing tearDown methods satisfy the requirement of not failing after a partial setup. Did that requirement disappear?

We should extend this to all tests at some point. Feel free to create a ticket. But we are stirring up the guts of unittest with this so I just want to introduce this slowly and cautiously.

Note that cancelling the `test` stage in the Gitlab pipeline crashes the runner. I wonder if this would help with that. But again, let's move slowly unless I'm mistaken about the requirement.

test/azul_test_case.py
```diff
@@ -20,3 +60,86 @@ def setUpClass(cls) -> None:
...
+
+                class TC(*([Test.LoggingTestCase] + tear_down_parent)):
+                    pass
+
+                TC(events).run()
+                expected = ['setUp', 'test', 'tearDown']
+                self.assertEqual(events, expected)
+
+    def test_keyboard_interrupt_in_test(self):
+        for always_tear_down in (False, True):
```
hannes-ucsc: Consider 

```suggestion
        for extra_parents, expected_events in [([], […]), ([AlwaysTearDownTestCase], […])]:
```

But I am not sure.

test/azul_test_case.py
hannes-ucsc: Missed this one.

test/azul_test_case.py
jessebrennan: I think I will leave it as it is since this works, and I can avoid the complications of having to integrate a Mock class into a TestCase.

test/azul_test_case.py
jessebrennan: Lol, I'm not sure what I was thinking here. Glad there's a neater alternative.

test/azul_test_case.py
jessebrennan: I like this better. Have a look at my take.

test/security/test_authenticator.py
hannes-ucsc: I assume you overlooked this. I'm just gonna remove it.

jessebrennan: Woops.. Yes I did.

Issue #1079, opened 2019-06-11
Add logging to Azul DRS endpoint that includes key request/response values 


mikebaumann:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-675)
┆Project Name: azul
┆Issue Number: AZUL-675

hannes-ucsc:
> Add logging to Azul DRS endpoint that includes key request/response values

While it's obvious which request elements are key (uuid and version), the same determination is not as obvious for the response. My commit adds logging of the first 1KiB of the response, but only if `AZUL_DEBUG` is set to 1.

PR #1301, opened 2019-09-19
Add tests of request/response logging (#1079)

danielsotirhos:


danielsotirhos:
> The base branch should be my branch.

I thought it was, no?

Screen Shot 2019-09-20 at 9 45 24 AM

test/service/test_logging.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def _get_request_logs(self, level, endpoint, params={}):
+        with self.assertLogs(logger=log, level=level) as logs:
+            url = self.base_url + endpoint
+            requests.get(url, params=params)
+            return logs.output
+
+    def test_request_logs_debug_off(self):
+        logs = self._get_request_logs(level='INFO', endpoint='/health')
+        self.assertTrue(any(re.search(r'^INFO.*Received GET request to \'/health\'', message) for message in logs))
```
hannes-ucsc:
```suggestion
        self.assertTrue(any(re.search(r"^INFO.*Received GET request to '/health'", message) for message in logs))
```

test/service/test_logging.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertTrue(any(re.search(r'^INFO.*Returning 200 response.*set AZUL_DEBUG', message) for message in logs))
+
+    def test_request_logs_debug_on(self):
+        logs = self._get_request_logs(level='DEBUG', endpoint='/health')
+        self.assertTrue(any(re.search(r'^INFO.*Received GET request to \'/health\'', message) for message in logs))
+        self.assertTrue(any(re.search(r'^DEBUG.*Returning 200 response.*See next line', message) for message in logs))
```
hannes-ucsc: I'd like to assert that there is actually JSON in the next line.

test/service/test_logging.py
hannes-ucsc: Can this be tightened to assert that there is exactly one such log line?

danielsotirhos:
Tried to title last commit with `fixup!` and Travis tests failed with:
```
Serving on localhost:0
Serving on localhost:0
The command "make test" exited with 0.
0.58s$ make check_clean
common.mk:26: Looks like TerraForm is not installed. This is ok as long as you're not trying to create a new deployment. Deploying new lambdas is still possible with `make deploy` but `make terraform` will not work.
git diff --exit-code  && git diff --cached --exit-code
The command "make check_clean" exited with 0.
0.63s$ make check_autosquash
common.mk:26: Looks like TerraForm is not installed. This is ok as long as you're not trying to create a new deployment. Deploying new lambdas is still possible with `make deploy` but `make terraform` will not work.
fatal: Not a valid object name issues/hannes/1079-generic-request-logging
fatal: Needed a single revision
invalid upstream 
No rebase in progress?
The current branch doesn't appear to be automatically squashable
make: *** [check_autosquash] Error 1
The command "make check_autosquash" exited with 2.
cache.2
store build cache
Done. Your build exited with 1.
```

test/service/test_logging.py
```diff
@@ -0,0 +1,65 @@
...
+    def _valid_json(cls, json_string):
+        try:
+            json.loads(json_string)
+        except ValueError:
+            return False
+        return True
+
+    def test_request_logs_debug_off(self):
+        logs = self._get_request_logs(level='INFO', endpoint='/health')
+        log_started_count = sum([1 for msg in logs if re.search(r'^INFO.*Received GET request to \'/health\'', msg)])
```
hannes-ucsc: Whenever you enter the backslash key, an alarm bell should go off and you should ask yourself if you can avoid the backslash. In this case, I think you can use `r""`.

hannes-ucsc: Rather convoluted. Please learn the distinction between a generator expression and a list comprehension. Study the syntactic shortcut that allows you to omit parentheses around a generator expression when it is the sole argument to another function. Also, `sum([1 for x in xs if x …]) is the same as `len([x in xs if x…])` if I'm not mistaken. The sum(1 for x in xs) is only needed for iterables, like generator expressions, that don't have an inherent length.

test/service/test_logging.py
```diff
@@ -0,0 +1,65 @@
...
+        for msg in logs:
+            if re.search(r'^INFO.*Received GET request to \'/health\'', msg):
+                log_started_count += 1
+            if re.search(r'^DEBUG.*Returning 200 response.*See next line', msg):
+                log_ended_count += 1
+                match = re.search(r'See next line.*\n({.*})$', msg)
+                self.assertIsNotNone(match)
+                json_string = match.group(1)
+                self.assertGreater(len(json_string), 2)
+                self.assertTrue(self._valid_json(json_string))
```
hannes-ucsc: Can you assert the literal content of the response, not just that it is valid JSON. You don't have to use the /health endpoint. And also, if you stick with health, I want you to run this test with network cable pulled and wifi off. I have the feeling it is hitting actual cloud resources.

test/service/test_logging.py
danielsotirhos: Between the two options
```
count = sum(1 for log in logs if 'foo' in log)
count = len([log for log in logs if 'foo' in log])
```
the `sum` approach seems more readable IMHO for a one liner list item counter, and has (as you pointed out) has the benefit of being able to use a generator saving memory needed to build a temporary list for `len`.
If you don't approve of stuffing all this into a one liner however I can unpack it to a `for` loop

test/service/test_logging.py
danielsotirhos: `/health` returns a 500 with no internet access. Changing tests to use `/health/basic` which returns a simple `{"up": true}`

test/service/test_logging.py
```diff
@@ -0,0 +1,54 @@
...
+def setUpModule():
+    configure_test_logging()
+
+
+class TestRequestResponse(WebServiceTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls._setup_indices()
```
hannes-ucsc: Same as in the other PR. Not necessary.

test/service/test_logging.py
```diff
@@ -0,0 +1,54 @@
...
+
+from azul.chalice import log
+from azul.logging import configure_test_logging
+
+
+def setUpModule():
+    configure_test_logging()
+
+
+class TestRequestResponse(WebServiceTestCase):
```
hannes-ucsc: Same as in the other PR. The ES container that `WebServiceTestCase` provides is not necessary for this test.

Issue #1302, opened 2019-09-19
Modify work-around for zarray files


jessebrennan:
Filter by filename instead of file_format.

This will follow up on these changes: https://github.com/HumanCellAtlas/pipeline-tools/commit/9bc31f3a8fc692e4058ce64880375bfcc8117940

see also
- https://github.com/HumanCellAtlas/metadata-schema/issues/623
- https://github.com/HumanCellAtlas/metadata-schema/issues/579 (superseded by ^)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-803)
┆Project Name: azul
┆Issue Number: AZUL-803

PR #1304, opened 2019-09-20
Modify work-around for zarray files (#1302)

jessebrennan:


jessebrennan:
Sounds good. I chose my ordering to match the previous and to keep the diff
minimal, But I see your point.

Issue #1260, opened 2019-09-05
Web service endpoint to get failed bundles queue


theathorn:
Expose the Azul failed-to-index-bundles queue so that the Data Browser can provide details (mainly for Data Ops) via a bundles status page.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-781)
┆Project Name: azul
┆Issue Number: AZUL-781
┆Epic: Broken bundles

GPelayo:
The endpoint will implemented under the `health/failure` endpoint. It should return the uuid and version of bundle notifications in the fail queues and the count of messages that are not bundle notifications. The following is the response that will be returned by `health/failure` endpoint.

```
{
	failed_bundle_notifications: [
		{
			bundle_uuid: <uuid value>,
			bundle_version: <version value>
		}			
	],
	other_failed_messages: {
		count: 0
	}
}
```

melainalegaspi:
This was relevant for DCP1, not DCP2.

PR #1305, opened 2019-09-20
Implemented /health/failures endpoint (#1260)

GPelayo:


lambdas/indexer/app.py
```diff
@@ -458,3 +459,23 @@ def nudge(event: chalice.app.CloudWatchEvent):
         log.info('Nudging queue with %i token(s) for a total value of %i', len(tokens), num_documents)
         for batch in chunked(tokens, 10):
             _send_tokens(batch)
+
+
+@app.schedule('rate(5 minutes)')
+def retrieve_fail_messages(event: chalice.app.CloudWatchEvent):
```
achave11: I followed the same pattern, but it seems like you can rid off the warning by setting the parameter to `event_`

lambdas/indexer/app.py
```diff
@@ -458,3 +459,23 @@ def nudge(event: chalice.app.CloudWatchEvent):
...
+    """
+    Get all the messages from the fail queue and save them in the the DynamoDB failure message table.
+    """
+    assert event is not None  # avoid linter warning
```
achave11: And you can remove this line.

lambdas/indexer/app.py
```diff
@@ -458,3 +459,23 @@ def nudge(event: chalice.app.CloudWatchEvent):
...
+    messages = Queues().receive_all_messages(config.fail_queue_name)
```
achave11: I would suggest something like the following. But is up to you.  
```suggestion
    queues = Queues()
    messages = queues.receive_all_messages(config.fail_queue_name)
```

lambdas/service/lambda-policy.json.template.py
```diff
@@ -146,6 +146,15 @@
...
             ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
+                "dynamodb:Query"
+            ],
+            "Resource": [
+                f"arn:aws:dynamodb:{aws.region_name}:{aws.account}:table/{config.dynamo_failure_message_table_name}",
+            ]
```
danielsotirhos: Instead of adding a new Statement dict here, could the Resource be added to the existing `"Effect": "Allow"` Statement on line 108?

src/azul/queues.py
```diff
@@ -51,6 +51,16 @@ def dump_all(self):
 
     def _dump(self, queue, path):
         logger.info('Writing messages from queue "%s" to file "%s"', queue.url, path)
+        messages = self._receive_all_messages(queue)
+        self._dump_messages(messages, queue.url, path)
+        logger.info(f'Finished writing {path !r}')
```
danielsotirhos: Looks like it works with or without a space however I see more instances in the codebase without space 
```suggestion
        logger.info(f'Finished writing {path!r}')
```

lambdas/service/lambda-policy.json.template.py
GPelayo: I just tested your change and it doesn't mess up the deployment. Plus its cuts a few lines in the commit.

lambdas/service/app.py
```diff
@@ -15,7 +15,7 @@
 import boto3
 from botocore.exceptions import ClientError
 # noinspection PyPackageRequirements
-from chalice import AuthResponse, BadRequestError, ChaliceViewError, NotFoundError, Response
+from chalice import AuthResponse, BadRequestError, ChaliceViewError, NotFoundError, Response, app
```
achave11: `app` Not required

lambdas/indexer/app.py
```diff
@@ -463,3 +464,23 @@ def nudge(event: chalice.app.CloudWatchEvent):
         log.info('Nudging queue with %i token(s) for a total value of %i', len(tokens), num_documents)
         for batch in chunked(tokens, 10):
             _send_tokens(batch)
+
+
+@app.schedule('rate(5 minutes)')
+def retrieve_fail_messages(event_: chalice.app.CloudWatchEvent):
```
achave11: I apologize for this again. Correction, is actually `_event` and you should see the linter warning disappear.

test/service/test_health_check.py
```diff
@@ -64,6 +68,78 @@ def test_elasticsearch_down(self):
...
+                }
+                bundle_notifications = [{
+                    "action": "add",
+                    "notification": {
+                        "query": '{}',
+                        "subscription_id": i['bundle_uuid'],
+                        "transaction_id": i['bundle_uuid'],
+                        "match": i,
+                    }
+                } for i in expected_failed_bundles]
```
danielsotirhos:
```suggestion
                } for bundle in expected_failed_bundles]
```

src/azul/health.py
```diff
@@ -93,6 +94,25 @@ def queues(self):
                 }
         return response
 
+    @memoized_property
+    def failures(self):
+        failed_bundle_notifications = []
+        database = boto3.resource('dynamodb')
+        table = database.Table(config.dynamo_failure_message_table_name)
+        bundle_key_condition = {'MessageType': {'AttributeValueList': ['bundle_notification'],
```
hannes-ucsc: Read contrib guide on formatting collection literals.

src/azul/health.py
```diff
@@ -93,6 +94,25 @@ def queues(self):
...
+                                                'ComparisonOperator': 'EQ'}}
+        bundle_notification_items = table.query(KeyConditions=bundle_key_condition, Limit=100)['Items']
+        for item in bundle_notification_items:
+            body = json.loads(item['Body'])
+            failed_bundle_notifications.append({'bundle_uuid': body['notification']['match']['bundle_uuid'],
```
hannes-ucsc: This is too much knowledge about the notification format in a place where it doesn't belong. I think we should just include the entire message verbatim in the response.

lambdas/indexer/app.py
```diff
@@ -484,3 +484,43 @@ def nudge(event: chalice.app.CloudWatchEvent):
...
+                    batch_requests.append({
+                        'Id': message.message_id,
+                        'ReceiptHandle': message.receipt_handle,
+                        'VisibilityTimeout': 0
+                    })
+                fail_queue.change_message_visibility_batch(Entries=batch_requests)
+                for message in batch_requests:
+                    del message['VisibilityTimeout']
+                fail_queue.delete_messages(Entries=batch_requests)
+                del messages[:10]
```
hannes-ucsc: I think this is overly complicated. You can deal with each batch separately there is no need to put message in a global list and then remove at the end. If you get only three messages back from SQS its ok to only write three  messages to DynanmoDb.

lambdas/indexer/app.py
```diff
@@ -484,3 +484,39 @@ def nudge(event: chalice.app.CloudWatchEvent):
...
+                        'MessageType': 'bundle_notification' if 'notification' in body.keys() else 'other',
+                        'SentTimeMessageId': f"{message.attributes['SentTimestamp']}-{message.message_id}",
+                        'Body': message.body
+                    })
+                    batch_requests.append({
+                        'Id': message.message_id,
+                        'ReceiptHandle': message.receipt_handle,
+                        'VisibilityTimeout': 0
+                    })
+                fail_queue.change_message_visibility_batch(Entries=batch_requests)
```
hannes-ucsc: This make no sense. Why would you first return the messages to the queue and then delete them?

lambdas/indexer/app.py
```diff
@@ -484,3 +484,39 @@ def nudge(event: chalice.app.CloudWatchEvent):
...
+                                                            MaxNumberOfMessages=10,
+                                                            WaitTimeSeconds=20,
+                                                            VisibilityTimeout=300)
+            if received_messages:
+                batch_requests = []
+                for message in received_messages:
+                    body = json.loads(message.body)
+                    writer.put_item(Item={
+                        'MessageType': 'bundle_notification' if 'notification' in body.keys() else 'other',
+                        'SentTimeMessageId': f"{message.attributes['SentTimestamp']}-{message.message_id}",
```
```suggestion
                        'SentTimeMessageId': message.attributes['SentTimestamp'] + '-' + message.message_id,
```

Looks less dense due to white space.

lambdas/indexer/app.py
```diff
@@ -484,3 +484,39 @@ def nudge(event: chalice.app.CloudWatchEvent):
...
+        while True:
+            received_messages = fail_queue.receive_messages(AttributeNames=['All'],
+                                                            MaxNumberOfMessages=10,
+                                                            WaitTimeSeconds=20,
+                                                            VisibilityTimeout=300)
+            if received_messages:
+                batch_requests = []
+                for message in received_messages:
+                    body = json.loads(message.body)
+                    writer.put_item(Item={
```
hannes-ucsc: Why did you revert back to writing single items. Isn't writing a batch more efficient?

src/azul/health.py
```diff
@@ -100,6 +101,28 @@ def queues(self):
                 }
         return response
 
+    @memoized_property
+    def failures(self):
+        failed_bundle_notifications = []
+        database = boto3.resource('dynamodb')
```
hannes-ucsc:
```suggestion
        dynamodb = boto3.resource('dynamodb')
```

lambdas/indexer/app.py
GPelayo: Leftover code from a previous implementation.

test/health_failures_test_case.py
```diff
@@ -0,0 +1,107 @@
...
+                        "failed_bundle_notifications": bundle_notifications,
+                        "other_failed_messages": other_count
+                    }
+                    with self.subTest(bundle_notification_count=bundle_notification_count,
+                                      document_notification_count=other_count):
+                        indexer_app.retrieve_failure_messages(MagicMock(), MagicMock())
+                        response = requests.get(self.base_url + '/health/failures')
+                        self.assertEqual(200, response.status_code)
+                        actual_response = response.json()
+                        self.assertEqual(expected_response.keys(), actual_response.keys())
```
hannes-ucsc: Why isn't this just `assertEqual(expected_response, actual_response)`?

GPelayo: `assertEqual(expected_response, actual_response)` would fail because the items in `failed_bundle_notifications` list in `actual_response` are not in the same order as the `failed_bundle_notifications` list in `expected_response`. 

The `failed_bundle_notifications` list populated from the SQS fail queue and these messages can't always be retrieved from the SQS in the same order the messages were put into the queue. Instead, the content of the 'failed_bundle_notifications' list is tested in self.assertCountEqual(expected_response['failed_bundle_notifications'], actual_response['failed_bundle_notifications']).

test/health_failures_test_case.py
```diff
@@ -0,0 +1,107 @@
...
+        fail_queue = sqs.get_queue_by_name(QueueName=config.fail_queue_name)
+        with ResponsesHelper() as helper:
+            helper.add_passthru(self.base_url)
+            # The 4th sub-test checks if the indexer lambda can write more than 1 batch of messages to dynamodb.
+            # The max number of messages in a batch is 10 and this sub-test populates the queue with 11 messages.
+            for bundle_notification_count, other_count in ((0, 0), (1, 0), (0, 1), (10, 1)):
+                with self._make_database():
+                    bundle_notifications = [{
+                        'action': 'add',
+                        'notification': IndexerTestCase._make_fake_notification((str(uuid4()),
```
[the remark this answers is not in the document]
GPelayo: Are you referring to `self._make_database()`?

lambdas/service/lambda-policy.json.template.py
GPelayo: Yes, my mistake. I'll remove it.

test/health_failures_test_case.py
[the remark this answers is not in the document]
hannes-ucsc: I'm referring to the only method being invoked in this line: `IndexerTestCase._make_fake_notification`.

test/health_failures_test_case.py
hannes-ucsc: I was afraid you'd give this explanation. Please make this an order-insensitive comparison.

src/azul/health.py
```diff
@@ -108,6 +108,17 @@ def queues(self):
                 }
         return response
 
+    @memoized_property
+    def failures(self):
+        dynamodb = boto3.resource('dynamodb')
+        table = dynamodb.Table(config.dynamo_failure_message_table_name)
+        bundle_key_condition = {'MessageType': {'AttributeValueList': ['bundle'], 'ComparisonOperator': 'EQ'}}
```
hannes-ucsc: These lines are overly too long. Can you wrap all the dict literals?

src/azul/health.py
```diff
@@ -202,3 +215,39 @@ def response(self):
...
+    def archive_fail_messages(self):
+        client = boto3.client('dynamodb')
+        fail_queue = boto3.resource('sqs').get_queue_by_name(QueueName=config.fail_queue_name)
+        while True:
+            received_messages = fail_queue.receive_messages(AttributeNames=['All'],
+                                                            MaxNumberOfMessages=10,
+                                                            WaitTimeSeconds=20,
+                                                            VisibilityTimeout=300)
+            if received_messages:
+                message_batch = [{
```
hannes-ucsc: Once we're in a wrapped literal or comprehension we should wrap it completely and not join brackets `[{`, here and everywhere else you joined brackets.

src/azul/health.py
```diff
@@ -202,3 +215,39 @@ def response(self):
...
+                            }
+                        }
+                    }
+                } for message in received_messages]
+                items = {config.dynamo_failure_message_table_name: message_batch}
+                response = client.batch_write_item(RequestItems=items)
+                unprocessed_items = response['UnprocessedItems']
+                if unprocessed_items and unprocessed_items[config.dynamo_failure_message_table_name]:
+                    raise ConnectionError('Unable to process and write all messages to dynamo db.')
+                fail_queue.delete_messages(Entries=[{
```
hannes-ucsc: else branch as usual

src/azul/health.py
GPelayo: Just wrap all the dict literals in `failures()`?

src/azul/health.py
```diff
@@ -202,3 +226,44 @@ def response(self):
...
+                                    'S': message.body
+                                }
+                            }
+                        }
+                    } for message in received_messages
+                ]
+                items = {config.dynamo_failure_message_table_name: message_batch}
+                response = client.batch_write_item(RequestItems=items)
+                unprocessed_items = response['UnprocessedItems']
+                if unprocessed_items and unprocessed_items[config.dynamo_failure_message_table_name]:
```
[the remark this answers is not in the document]
GPelayo: The condition you describing happens when all the items are processed.

src/azul/queues.py
```diff
@@ -58,6 +58,16 @@ def dump_all(self):
...
+
+    def receive_all_messages(self, queue_name):
```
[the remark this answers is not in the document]
GPelayo: No, `receive_all_messages` is not called anywhere else. I'll remove it.

src/azul/health.py
```diff
@@ -202,3 +226,44 @@ def response(self):
...
+                    raise ConnectionError('Unable to process and write all messages to dynamo db.')
+                else:
+                    fail_queue.delete_messages(Entries=[
```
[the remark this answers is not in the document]
GPelayo: `fail_queue.delete_messages` deletes all messages in the batch from the queue. 

Also, it's not possible to create duplicate items with the same sort_key and hash_key. By default, if an item has the same sort_key and hash_key as an older item in the table and that newer item is being put into that table, the new item will replace the existing item.

.gitlab-ci.yml
```diff
@@ -43,6 +43,7 @@ test:
     - make format
     - make check_clean
     - make pep8
+    - make trufflehog
```
hannes-ucsc: remove

dummy-file
```diff
@@ -0,0 +1 @@
+1
```
hannes-ucsc: Remove

src/azul/health.py
hannes-ucsc:
```suggestion
                if unprocessed_items:
                     assert len(unprocessed_items) == 1 and unprocessed_items[config.dynamo_failure_message_table_name]
```

hannes-ucsc:
Also research if AWS retries the lambda twice when its triggered via a CloudWatch event rule and consider if that could cause trouble.

GPelayo:
Looks like the delay is caused by `WaitTimeSeconds` from `fail_queue.receive_messages()`. 

If `WaitTimeSeconds` set to 20, `receive_messages()` will always wait at least 20 seconds when the queue is empty. Since the while loop in `archive_fail_messages()` only breaks without errors when the error queue is empty, `archive_fail_messages()` will always take at least 20 seconds to finish.

https://github.com/DataBiosphere/azul/blob/53ad7f998c92312d7d395590607cb6b06a0950a2/src/azul/health.py#L234-L237

test/health_failures_test_case.py
```diff
@@ -0,0 +1,120 @@
...
+        indexer_app = load_app_module('indexer')
+        sqs = boto3.resource('sqs', region_name='us-east-1')
+        sqs.create_queue(QueueName=config.fail_queue_name)
+        fail_queue = sqs.get_queue_by_name(QueueName=config.fail_queue_name)
+        original_archive_fail_messages = HealthController.archive_fail_messages
+
+        def mocked_archive_fail_messages(self):
+            original_archive_fail_messages(self, 0)
+
+        with patch.object(HealthController, 'archive_fail_messages', new=mocked_archive_fail_messages):
```
hannes-ucsc: Unless I'm completely delirious, I think I mentioned in our conversation a couple days ago that the wait time should be extracted into a class attribute, not a method parameter. That's much easier to patch.

GPelayo: You said I should parameterize `WaitTimeSeconds`, thats why I extracted it as a parameter.

lambdas/indexer/.chalice/config.json.template.py
```diff
@@ -31,6 +31,9 @@
                     "lambda_memory_size": 128,
                     "lambda_timeout": config.indexer_lambda_timeout,
                 },
+                config.failure_message_lambda_basename: {
+                    'lambda_timeout': config.indexer_lambda_timeout
```
hannes-ucsc: Regarding this timeout, …

src/azul/health.py
```diff
@@ -202,3 +227,45 @@ def response(self):
     def generate_cache(self):
         health_object = dict(time=time.time(), health=self.health.as_json())
         self.storage_service.put(object_key=f'health/{self.health.lambda_name}', data=json.dumps(health_object))
+
+    def archive_fail_messages(self):
+        client = boto3.client('dynamodb')
+        fail_queue = boto3.resource('sqs').get_queue_by_name(QueueName=config.fail_queue_name)
+        while True:
```
hannes-ucsc: … we should make sure that this loop doesn't exceed it. The lambda context exposes the remaining runtime. Check out `LambdaContext` in `types.py`.

src/azul/health.py
```diff
@@ -202,3 +227,45 @@ def response(self):
     def generate_cache(self):
         health_object = dict(time=time.time(), health=self.health.as_json())
         self.storage_service.put(object_key=f'health/{self.health.lambda_name}', data=json.dumps(health_object))
+
+    def archive_fail_messages(self, remaining_context_time):
+        client = boto3.client('dynamodb')
+        fail_queue = boto3.resource('sqs').get_queue_by_name(QueueName=config.fail_queue_name)
+        while remaining_context_time.get() > 0:
```
hannes-ucsc: You need a margin here. What if you have only 1s left but are entering another 5s blocking call?

test/health_failures_test_case.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @mock_sts
+    @mock_sqs
+    @mock_dynamodb2
+    def test_failures_endpoint(self):
+        indexer_app = load_app_module('indexer')
+        sqs = boto3.resource('sqs', region_name='us-east-1')
+        sqs.create_queue(QueueName=config.fail_queue_name)
+        fail_queue = sqs.get_queue_by_name(QueueName=config.fail_queue_name)
+        with patch('azul.time.RemainingLambdaContextTime.get', return_value=1000):
```
hannes-ucsc: How long do you expect the body of this context manager to run for in the worst case?

src/azul/health.py
GPelayo: What do you mean by a blocking call?

GPelayo: By margin, you mean something like this?
```suggestion
        while remaining_context_time.get() > 5:
```

test/health_failures_test_case.py
GPelayo: I forgot to change `return_value` to 5 seconds to per subtest case.

src/azul/health.py
hannes-ucsc: Kind of. Obviously even 5 is cutting it too close since there is work that you do after you get the messages back.

test/health_failures_test_case.py
hannes-ucsc: Can you answer my question, please?

GPelayo: 5 seconds to per subtest case.

hannes-ucsc: That can't be right. You are patching `receive_message_wait_time` to zero. Please describe how you arrived at that 5s estimate.

GPelayo: While `receive_message_wait_time` to zero, the processes after waiting for the messages to be received would a few seconds more than 0. I thought 5 seconds would be the longest worst case while still being strict margin.

hannes-ucsc: I don't understand what you mean by "while still being strict margin". Can you please elaborate.

Can you explain what would take 5s when all the test is doing is using in-memory mocks of DynamodDB and SQS. 

I am also concerned that you are retroactively pretending have thought this fully through. I think you did not but I have no proof. It is just that your estimate of 5s is suspiciously identical to the 5s default for `receive_message_wait_time` and also astronomically high considering the use of mocks.

GPelayo: Yeah, I based my answer from the value given in `receive_message_wait_time`.

lambdas/indexer/app.py
```diff
@@ -310,3 +311,13 @@ def consolidate(self, others: List['DocumentTally']) -> 'DocumentTally':
         assert all(self.entity == other.entity for other in others)
         return replace(self, num_contributions=sum((other.num_contributions for other in others),
                                                    self.num_contributions))
+
+
+@app.schedule('rate(5 minutes)')
```
hannes-ucsc, line 316: We should tie this to the lambda timeout so we don't have more than one concurrent execution of this lambda. I'm thinking 110% of the timeout.

src/azul/__init__.py
```diff
@@ -636,6 +636,10 @@ def cart_export_state_machine_name(self):
 
     access_token_issuer = "https://humancellatlas.auth0.com"
 
+    @property
+    def dynamo_failure_message_table_name(self):
+        return self.qualified_resource_name('failuremessages')
```
hannes-ucsc, line 641:
```suggestion
        return self.qualified_resource_name('failures')
```

src/azul/health.py
```diff
@@ -241,3 +267,54 @@ def update_cache(self) -> None:
     @classproperty
     def all_keys(cls) -> Set[str]:
         return {p.key for p in cls.all_properties}
+
+    # noinspection PyMethodParameters
+    @classproperty
+    def receive_message_wait_time(cls) -> int:
```
hannes-ucsc, line 273: Can't this be a simple attribute?

terraform/dynamo.tf.json.template.py
```diff
@@ -77,6 +77,22 @@
                                 "type": "S"
                             }
                         ]
+                    },
+                    "failed_messages_table": {
```
hannes-ucsc, line 81:
```suggestion
                    "failures_table": {
```

We should be more consistent. There is a mix of "failure", "failed", "failures".

terraform/dynamo.tf.json.template.py
```diff
@@ -77,6 +77,22 @@
...
+                        "name": config.dynamo_failure_message_table_name,
```
hannes-ucsc, line 82:
```suggestion
                        "name": config.dynamo_failures_table_name,
```

hannes-ucsc:
```suggestion
                        "name": config.failures_table_name,
```

Unless there is precedent for the `dynamo` prefix.

src/azul/health.py
```diff
@@ -241,3 +267,54 @@ def update_cache(self) -> None:
...
+        """
+        The number of seconds to wait for a message to arrive in the fail queue
+        before accepting that there are zero messages to be received.
+        """
+        return 5
+
+    def archive_fail_messages(self, remaining_context_time):
+        client = boto3.client('dynamodb')
+        fail_queue = boto3.resource('sqs').get_queue_by_name(QueueName=config.fail_queue_name)
+        while remaining_context_time.get() > 0:
```
hannes-ucsc, line 283: So we still need a margin here so we don't enter another 5s request with only 1s remaining in the context.

test/health_check_test_case.py
```diff
@@ -268,6 +282,10 @@ def _create_mock_queues(self):
         for queue_name in config.all_queue_names:
             sqs.create_queue(QueueName=queue_name)
 
+    def _create_mock_dynamodb(self):
```
hannes-ucsc, line 285:
```suggestion
    def _mock_failures_table(self):
```

test/health_failures_test_case.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class TestHealthFailures(LocalAppTestCase):
+
+    @classmethod
+    def lambda_name(cls) -> str:
+        return 'service'
+
+    @classmethod
+    def dynamo_failure_message_table_settings(cls):
```
hannes-ucsc, line 38: Can this be an attribute?

test/health_failures_test_case.py
```diff
@@ -0,0 +1,124 @@
...
+                'ReadCapacityUnits': 5,
+                'WriteCapacityUnits': 5
+            }
+        }
+
+    @contextmanager
+    def _make_database(self):
+        dynamodb = boto3.resource('dynamodb', region_name='us-east-1')
+        dynamodb.create_table(**self.dynamo_failure_message_table_settings())
+        with mock.patch.dict(os.environ, AWS_DEFAULT_REGION='us-east-1'):
```
hannes-ucsc, line 71: Keep an eye out for #1710.

test/health_failures_test_case.py
```diff
@@ -0,0 +1,124 @@
...
+                }
+            ],
+            'ProvisionedThroughput': {
+                'ReadCapacityUnits': 5,
+                'WriteCapacityUnits': 5
+            }
+        }
+
+    @contextmanager
+    def _make_database(self):
```
hannes-ucsc, line 68:
```suggestion
    def _mock_failures_table(self):
```

And

test/health_failures_test_case.py
```diff
@@ -0,0 +1,124 @@
...
+        dynamodb = boto3.resource('dynamodb', region_name='us-east-1')
+        dynamodb.create_table(**self.dynamo_failure_message_table_settings())
+        with mock.patch.dict(os.environ, AWS_DEFAULT_REGION='us-east-1'):
+            try:
+                yield
+            finally:
+                table = dynamodb.Table(config.dynamo_failure_message_table_name)
```
hannes-ucsc, line 75: Can you reuse the return value from `create_table` above?

test/health_failures_test_case.py
hannes-ucsc: And let's not proliferate that region name all over the place.

terraform/dynamo.tf.json.template.py
danielsotirhos: Changing variable name to `dynamo_failures_table_name` to match other dynamo related configs `dynamo_cart_table_name`, `dynamo_cart_item_table_name`, `dynamo_user_table_name`, and `dynamo_object_version_table_name`.

src/azul/health.py
```diff
@@ -108,6 +110,32 @@ def queues(self):
                 }
         return response
 
+    @health_property
+    def failures(self):
```
hannes-ucsc, line 114: I've arrived at the conclusion that it doesn't make sense to expose the failures as part of a standard /health response and then only return the first 100. Fetching 100 items is both too much for a normal health response and too little for a client that is interested in looking at all errors. This should be exposed as a separate `/health/failures` (not `/health/{keys}`) and a ticket should be created to add paging to that endpoint.

The mocks that were added to the tests of `/health` should be remove again.

lambdas/indexer/app.py
```diff
@@ -310,3 +316,13 @@ def consolidate(self, others: List['DocumentTally']) -> 'DocumentTally':
         assert all(self.entity == other.entity for other in others)
         return replace(self, num_contributions=sum((other.num_contributions for other in others),
                                                    self.num_contributions))
+
+
+@app.schedule(f'rate({(config.indexer_lambda_timeout + 60) // 60} minutes)')
```
danielsotirhos, line 321: Both `rate(5 minutes 30 seconds)` and `rate(330 seconds)` caused failures with `make terraform`:

```
Error: Updating CloudWatch Event Rule failed: ValidationException: Parameter ScheduleExpression is not valid.
  on indexer/chalice.tf.json line 297, in resource.aws_cloudwatch_event_rule.retrieve_fail_messages-event:
```

hannes-ucsc: I don't like the non-linear margin this produces. For example, with indexer_lambda_timeout set to 299, this gives 5 minutes, just 1s of margin. Got anything better? Can you find any documentation that would explain why 330 seconds is considered invalid?

test/health_failures_test_case.py
```diff
@@ -0,0 +1,117 @@
...
+    # Moto's dynamodb backend doesn't support government regions.
+    _aws_test_region = 'ap-south-1'
+
+    @classmethod
+    def lambda_name(cls) -> str:
+        return 'service'
+
+    @contextmanager
+    def _mock_failures_table(self):
+        dynamodb = boto3.resource('dynamodb')  # , region_name=self._aws_test_region)
```
hannes-ucsc, line 67:
```suggestion
        dynamodb = boto3.resource('dynamodb')
```

danielsotirhos:
```
Syntax: rate(Value Unit)
Rate expressions have two required fields. Fields are separated by white space.
Valid values: minute | minutes | hour | hours | day | days
```
https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html#RateExpressions

There is an alternate `cron()` syntax however it only allows granularity to the minute as well.

I believe my latest approach with `math.ceil` gives good results:
```

0 sec -> 0 min
1 sec -> 1 min
...
53 sec -> 1 min
54 sec -> 1 min
55 sec -> 2 min
56 sec -> 2 min
...
108 sec -> 2 min
109 sec -> 2 min
110 sec -> 3 min
111 sec -> 3 min
...
162 sec -> 3 min
163 sec -> 3 min
164 sec -> 4 min
165 sec -> 4 min
...
217 sec -> 4 min
218 sec -> 4 min
219 sec -> 5 min
220 sec -> 5 min
...
271 sec -> 5 min
272 sec -> 5 min
273 sec -> 6 min
274 sec -> 6 min
...
326 sec -> 6 min
327 sec -> 6 min
328 sec -> 7 min
329 sec -> 7 min
...
380 sec -> 7 min
381 sec -> 7 min
382 sec -> 8 min
383 sec -> 8 min
384 sec -> 8 min
```

src/azul/health.py
```diff
@@ -241,3 +270,45 @@ def update_cache(self) -> None:
...
+                                                            MaxNumberOfMessages=10,
+                                                            WaitTimeSeconds=self.receive_message_wait_time,
+                                                            VisibilityTimeout=300)
+            if received_messages:
+                message_batch = [
+                    {
+                        'PutRequest': {
+                            'Item': {
+                                'MessageType': {
+                                    'S': 'bundle' if 'notification' in json.loads(message.body).keys() else 'other'
```
hannes-ucsc, line 288: Since we eliminated tokens this conditional can be removed. All messages in this queue will be notifications.

But in https://github.com/DataBiosphere/azul/issues/1593 we will be adding failed tallies to this queue so it makes sense to hardcode this to `notification` in this PR.

test/health_failures_test_case.py
```diff
@@ -0,0 +1,117 @@
...
+        sqs.create_queue(QueueName=config.fail_queue_name)
+        fail_queue = sqs.get_queue_by_name(QueueName=config.fail_queue_name)
+        with patch('azul.time.RemainingLambdaContextTime.get', return_value=2):
+            with patch.object(HealthController, 'receive_message_wait_time', 0):
+                with ResponsesHelper() as helper:
+                    helper.add_passthru(self.base_url)
+                    # The 4th sub-test checks if the indexer lambda can write more than 1 batch of messages to dynamodb.
+                    # The max number of messages in a batch is 10 and this sub-test populates the queue with 11
+                    # messages.
+                    for num_bundles, num_other in ((0, 0), (1, 0), (0, 1), (10, 1), (10, 0)):
```
danielsotirhos, line 89:
```suggestion
                    for num_bundles, num_other in ((0, 0), (1, 0), (0, 1), (8, 1), (9, 1), (10, 1)):
```

lambdas/indexer/app.py
```diff
@@ -82,6 +84,11 @@ def fast_health():
     return health_controller().fast_health()
 
 
+@app.route('/health/failures', methods=['GET'], cors=True)
```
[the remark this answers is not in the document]
danielsotirhos: Endpoint is documented in `service/app.py` and not here in `indexer/app.py`

melainalegaspi:
This was relevant for DCP1, not DCP2.

Issue #1303, opened 2019-09-19
Improve deployment cleanup process


jessebrennan:
A few steps are missing from the readme. A script should be made that deletes left over IAM roles (see https://github.com/DataBiosphere/azul/issues/1094#issuecomment-533268395).

The readme should include steps to:
 - run this script and
 - run `python scripts/reindex.py --delete`

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-804)
┆Project Name: azul
┆Issue Number: AZUL-804

theathorn:
Spike to investigate possible solutions.

PR #1307, opened 2019-09-23
Improve deployment cleanup process (#1303)

danielsotirhos:


scripts/delete_lambda_iam_role.py
```diff
@@ -0,0 +1,37 @@
+#! /usr/bin/env python3
```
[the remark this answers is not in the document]
hannes-ucsc: Please incorporate into `manage_lambda_iam_role.py`. I mentioned this before: tightly coupled functionality should be in close proximity. Adding a second script to undo the effect of another script is just not how we do things and that should be clear by now.

scripts/delete_lambda_iam_role.py
```diff
@@ -0,0 +1,37 @@
...
+
+lambda_name = sys.argv[1]
+role_name = config.qualified_resource_name(lambda_name)
+
+iam = boto3.client('iam')
+
+try:
+    response = iam.get_role(RoleName=role_name)
+    role = response['Role']
+except iam.exceptions.NoSuchEntityException:
```
hannes-ucsc: Make sure this necessary. I don't think it is.

scripts/delete_lambda_iam_role.py
```diff
@@ -0,0 +1,37 @@
...
+    policies = None
+
+if policies is not None:
+    for policy_name in policies['PolicyNames']:
+        try:
+            print(f"Deleting IAM Policy {policy_name}")
+            iam.delete_role_policy(RoleName=role_name, PolicyName=policy_name)
+        except:
+            print(f"Error deleting Policy {policy_name}")
```
hannes-ucsc: We can't just ignore errors. If we can't remove a policy from a role something is seriously wrong and we need to propagate that problem all the way up so `make` fails.

scripts/delete_lambda_iam_role.py
```diff
@@ -0,0 +1,37 @@
...
+
+try:
+    policies = iam.list_role_policies(RoleName=role_name)
+except iam.exceptions.NoSuchEntityException:
+    policies = None
+
+if policies is not None:
+    for policy_name in policies['PolicyNames']:
+        try:
+            print(f"Deleting IAM Policy {policy_name}")
```
hannes-ucsc: I know the other script uses print() for historic reasons but we should use loggers instead.

lambdas/service/Makefile
```diff
@@ -18,6 +18,7 @@ deploy: state wheels
 	python $(azul_home)/scripts/manage_chalice_deployed.py service upload
 
 delete: state
+    python $(azul_home)/scripts/delete_lambda_iam_role.py service
```
hannes-ucsc: Makefiles must use a tab character for indenting lines in a recipe. This looks like four spaces. Did you actually test this? My `make` fails if I use spaces instead of tabs.

scripts/manage_lambda_iam_role.py
```diff
@@ -1,68 +1,122 @@
...
 from azul import config
 from azul.deployment import aws
+from azul.logging import configure_script_logging
+
+log = logging.getLogger(__name__)
+
+
+def main(argv):
+    configure_script_logging(log)
+    parser = argparse.ArgumentParser(description='Manage lambda IMA roles')
```
hannes-ucsc:
```suggestion
    parser = argparse.ArgumentParser(description='Manage lambda IAM roles')
```

Here and everywhere.

scripts/manage_lambda_iam_role.py
```diff
@@ -1,68 +1,122 @@
...
-                        PolicyDocument=json.dumps(role_policy))
+        policies = iam.list_role_policies(RoleName=role_name)
+    except iam.exceptions.NoSuchEntityException:
+        policies = []
+
+    for policy_name in policies['PolicyNames']:
+        log.info(f"Deleting IAM Policy {policy_name}")
+        iam.delete_role_policy(RoleName=role_name, PolicyName=policy_name)
+
+    if role is not None:
```
hannes-ucsc: I fail to see how it can ever be `None`.

Issue #1194, opened 2019-08-05
Remove keys in the Azul repo using the BFG repo cleaner


GPelayo:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-744)
┆Project Name: azul
┆Issue Number: AZUL-744

GPelayo:
The BFG can redact all the secrets and keys, except the keys in the first commit of the master branch. But this problem can be solved by making a temp commit in the master branch, run the BFG, push the new commits, and delete the new commit. I have a fork of Azul with the secrets and keys removed at https://github.com/GPelayo/azul.

GPelayo:
@shiroyuki fork has been deleted.

hannes-ucsc:
Outstanding work:

- [ ] Delete @shiroyuki fork
- [ ] Delete @hannes fork
- [ ] Ask Github support to delete pull refs for merged and closed branches
- [ ] Commit the BFG report mapping old commits to new commit

hannes-ucsc:
We decided not to delete the PR refs because GH support told us that they are not able to rewrite the refs, they can only delete them. The impact of losing the PR history is not worth it. We already achieved the main goal of removing the secrets from the main branches which allowed us to removed the complicated whitelisting of secrets in Trufflehog and gitsecrets config.

hannes-ucsc:
The repository backups should go into the config bucket, the same bucket that holds the terraform state files. @GPelayo, please reach out to me so we can decide on where in the bucket we place those files.

PR #1308, opened 2019-09-23
Added BFG mapping file and put repos and text files to bucket (#1194)

GPelayo:
A copy of all the repos used to remove the secrets from Azul have been uploaded to the S3 bucket, `org-humancellatlas-azul-dev-config`. In the bucket, there are also a zip of text files that were used to identify the PR refs affected by the removal. 
The following are the keys of each zip file:
```
backup/git/bfg-docs.zip
backup/git/post-bfg-azul-clone.zip
backup/git/post-bfg-azul-mirror.zip
backup/git/pre-bfg-azul-mirror.zip
```


hannes-ucsc:
@GPelayo You're not using fixup correctly. `git help commit` and search for "--fixup" and "--squash". You also need to leave an empty line between the subject line (aka title) of the commit message and the body.

Issue #1276, opened 2019-09-13
Test mode doesn't block deletion notifications


jessebrennan:
I'm not sure if this is a bug, but it's something we should consider. Probably the current behavior is fine the way it is, but I wanted to check.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-792)
┆Project Name: azul
┆Issue Number: AZUL-792

achave11:
By approval of @hannes-ucsc the estimate for this ticket change.

hannes-ucsc:
… because we extended the scope a bit to reduce duplication between the `POST /` and `POST /delete` handlers which requires removing the special `sync` mode for `/`.

PR #1310, opened 2019-09-24
Fix: Test mode doesn't block deletion notifications (#1276)

achave11:


test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +994,38 @@ def test_invalid_auth_for_notification_request(self):
...
+            "match": {
+                'bundle_uuid': 'bb2365b9-5a5b-436f-92e3-4fc6d86a9efd',
+                'bundle_version': '2018-03-28T13:55:26.044Z'
+            },
+            'test_name': f'integration-test_bb2365b9-5a5b-436f-92e3-4fc6d86a9efd_bb2365b9'
+        }
+        for endpoint in ['/', '/delete']:
+            for test_mode in ('0', '1'):
+                for body in (testless_notification, test_name_in_notification):
+                    with self.subTest(test_mode=test_mode):
```
jessebrennan:
```suggestion
                    with self.subTest(test_mode=test_mode, endpoint=endpoint):
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +994,38 @@ def test_invalid_auth_for_notification_request(self):
...
+                        with patch.dict(os.environ, TEST_MODE=test_mode):
+                            with self.subTest(endpoint=endpoint):
```
jessebrennan:
```suggestion
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +994,38 @@ def test_invalid_auth_for_notification_request(self):
...
+                                response = self._test(body, endpoint=endpoint, valid_auth=True)
+                                if 'test_name' not in body and test_mode == '1':
+                                    self.assertEqual(500, response.status_code)
```
jessebrennan: Consider asserting the response message too.

achave11: @jessebrennan Thanks for that!

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+                'bundle_uuid': 'bb2365b9-5a5b-436f-92e3-4fc6d86a9efd',
+                'bundle_version': '2018-03-28T13:55:26.044Z'
+            }
+        }
+        test_name_in_notification = {
+            "match": {
+                'bundle_uuid': 'bb2365b9-5a5b-436f-92e3-4fc6d86a9efd',
+                'bundle_version': '2018-03-28T13:55:26.044Z'
+            },
+            'test_name': f'integration-test_bb2365b9-5a5b-436f-92e3-4fc6d86a9efd_bb2365b9'
```
GPelayo: Just remove this "f".

```suggestion
            'test_name': 'integration-test_bb2365b9-5a5b-436f-92e3-4fc6d86a9efd_bb2365b9'
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+    def test_index_test_mode(self):
+        self._create_mock_notify_queue()
+        testless_notification = {
+            'match': {
+                'bundle_uuid': 'bb2365b9-5a5b-436f-92e3-4fc6d86a9efd',
+                'bundle_version': '2018-03-28T13:55:26.044Z'
+            }
+        }
+        test_name_in_notification = {
+            "match": {
```
GPelayo:
```suggestion
            'match': {
```

hannes-ucsc:
Re commit messsages: 

When you prefix a commit title with `Fix:` what follows should be a description of the problem, not the fix. For example, `Fix: rusty nail sticks through my foot` not `Fix: pull rusty nail out of my foot`.

And `Fix: subtest include both indexing and delete endpoints` is very specific and lacking context. In that case it helps stepping back a little and describing the problem at a more general level. I would suggest one but I don't actually understand what the issue is that this commit addresses.

hannes-ucsc:
> I'm confused... what about the refactoring that we talked about yesterday?

Right, that was the [deal](https://github.com/databiosphere/azul/issues/1276#issuecomment-535326326).

README.md
```diff
@@ -464,12 +464,10 @@ make delete
...
 
    ```
-   python scripts/reindex.py --workers=1 --sync --indexer-url http://127.0.0.1:8000/
+   python scripts/reindex.py --workers=1 --indexer-url http://127.0.0.1:8000/
    ```
 
-   in a separate shell. The `--sync` argument causes the Chalice app to invoke
-   the indexing code directly instead of queuing an SQS message to be consumed
-   by the indexer worker Lambda function in AWS.
+   in a separate shell.
```
hannes-ucsc: We need to remove everything starting with "To hit the indexer" because this does require a complete deployment lambda, queues, ES and all. And worse, it would only do the queuing into the notifications queue locally, the rest of indexing (creating contributions, writing aggregates) would happen in the lambdas. Interestingly, this would also apply to `--sync` where the contributions would be created locally and the aggregate would be created by the lambda. All in all, a mess. So it's a good thing we remove --sync. To test the indexer locally, you just need a write a test. We should also rename the section to `Running service locally`.

lambdas/indexer/app.py
```diff
@@ -74,27 +74,28 @@ def hello():
...
     """
-    Receive a notification event and either queue it for asynchronous indexing or process it synchronously.
+    Receive a notification event and queue it for indexing or deletion.
     """
     hmac.verify(current_request=app.current_request)
     notification = app.current_request.json_body
     log.info("Received notification %r", notification)
     validate_request_syntax(notification)
-    params = app.current_request.query_params
+    if app.current_request.context['path'].endswith('/'):
```
hannes-ucsc: Not sure why this uses `endswith()`. Is it because the handler for `/health` uses it? The situation there is a little more intricate and warrants use of `endswith()`. Here however, I feel that we can just use `==`.

One thing to look out for is how Chalice treats a trailing slash. For example, both, 

https://service.explore.data.humancellatlas.org/repository/summary
https://service.explore.data.humancellatlas.org/repository/summary/

hit the summary endpoint even though it is bound to "/repository/summary" only. It's possible that Chalice strips the trailing slash so `app.current_request.context['path']` would be "/repository/summary" in both cases. It's also possible that it doesn't, and that would confuse your logic here. You'd need to experiment to find out.

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+                "bundle_uuid": "bb2365b9-5a5b-436f-92e3-4fc6d86a9efd",
+                "bundle_version": "2018-03-28T13:55:26.044Z"
+            },
+            "test_name": "integration-test_bb2365b9-5a5b-436f-92e3-4fc6d86a9efd_bb2365b9"
+        }
+        for endpoint in ['/', '/delete']:
+            for test_mode in ('0', '1'):
```
```suggestion
            for test_mode in 0, 1:
```

Tuple parentheses are optional after `in` in `for` statements.

And I would convert to `str` when patching os.environ

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+                for body in (testless_notification, test_name_in_notification):
+                    with self.subTest(test_mode=test_mode, endpoint=endpoint):
+                        with patch.dict(os.environ, TEST_MODE=test_mode):
+                            response = self._test(body, endpoint=endpoint, valid_auth=True)
+                            if 'test_name' not in body and test_mode == '1':
```
```suggestion
                            if 'test_name' not in body and test_mode == 1:
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+                "bundle_uuid": "bb2365b9-5a5b-436f-92e3-4fc6d86a9efd",
+                "bundle_version": "2018-03-28T13:55:26.044Z"
+            },
+            "test_name": "integration-test_bb2365b9-5a5b-436f-92e3-4fc6d86a9efd_bb2365b9"
+        }
+        for endpoint in ['/', '/delete']:
+            for test_mode in ('0', '1'):
+                for body in (testless_notification, test_name_in_notification):
+                    with self.subTest(test_mode=test_mode, endpoint=endpoint):
+                        with patch.dict(os.environ, TEST_MODE=test_mode):
```
```suggestion
                        with patch.dict(os.environ, TEST_MODE=str(test_mode)):
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+            }
+        }
+        test_name_in_notification = {
+            "match": {
+                "bundle_uuid": "bb2365b9-5a5b-436f-92e3-4fc6d86a9efd",
+                "bundle_version": "2018-03-28T13:55:26.044Z"
+            },
+            "test_name": "integration-test_bb2365b9-5a5b-436f-92e3-4fc6d86a9efd_bb2365b9"
+        }
+        for endpoint in ['/', '/delete']:
```
```suggestion
        for endpoint in '/', '/delete':
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+            for test_mode in ('0', '1'):
+                for body in (testless_notification, test_name_in_notification):
```
```suggestion
                for body in testless_notification, test_name_in_notification:
```

test/indexer/test_hca_indexer.py
```diff
@@ -994,9 +995,43 @@ def test_invalid_auth_for_notification_request(self):
...
+                    with self.subTest(test_mode=test_mode, endpoint=endpoint):
+                        with patch.dict(os.environ, TEST_MODE=test_mode):
+                            response = self._test(body, endpoint=endpoint, valid_auth=True)
+                            if 'test_name' not in body and test_mode == '1':
+                                self.assertEqual(500, response.status_code)
+                                self.assertEqual({
+                                    "Code": "ChaliceViewError",
+                                    "Message": "ChaliceViewError: Ignored notification {'match': {'bundle_uuid': "
```
hannes-ucsc: Couldn't you interpolate `testless_notification` here to shorten the literal?

lambdas/indexer/app.py
```diff
@@ -74,27 +74,28 @@ def hello():
...
+    Receive a notification event and queue it for indexing or deletion.
     """
     hmac.verify(current_request=app.current_request)
     notification = app.current_request.json_body
     log.info("Received notification %r", notification)
     validate_request_syntax(notification)
-    params = app.current_request.query_params
+    if app.current_request.context['path'] == '/':
+        return process_notification('add', notification)
+    elif app.current_request.context['path'].startswith('/delete'):
```
hannes-ucsc: This is technically correct, but confusing to a person reading this. I would do, 

```suggestion
    elif app.current_request.context['path'] in ('/delete', '/delete/')
```

And did you confirm that this is indeed needed? Does Chalice not strip the trailing slash before calling the handler?

We should also add `else: assert False` as we always do.

achave11: @hannes-ucsc apologies for not addressing the question in the previous cycle. 
Yes, I confirmed that chalice does not strip the trailing dash. So we must account for both cases.

Issue #1151, opened 2019-07-24
Monitor latency for Azul health checks


jessebrennan:
This is a simple Terraform configuration change.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-721)
┆Project Name: azul
┆Issue Number: AZUL-721

achave11:
@jessebrennan can you elaborate on this issue?

PR #1321, opened 2019-09-27
Measure Azul health checks latency (#1151)

achave11:


jessebrennan:
I guess the only stipulation is make sure you test this in your personal deployment first so that we know that it deploys without problems.

hannes-ucsc:
@jessebrennan, building this in the sandbox takes care of that.

Issue #1318, opened 2019-09-26
Add entity_ids param to GET /integrations endpoint


danielsotirhos:
Add an optional parameter `entity_ids` to the GET /integration endpoint.
Value expected is a comma separated list of UUIDs.
If present in request, response is limited to integrations that are for the matching entity ids.

This ticket builds upon work done for #1243 
This ticket requires #1197 to be updated for the new parameter.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-811)
┆Project Name: azul
┆Issue Number: AZUL-811
┆Epic: Generic Tertiary Portal Handoff

PR #1329, opened 2019-09-28
Add entity_ids param to integrations endpoint (#1318)

danielsotirhos:


test/service/test_response.py
```diff
@@ -1393,24 +1393,51 @@ def lambda_name(cls) -> str:
...
             ('get', 'project', 9),
         ]
         for integration_type, entity_type, num_integrations_expected in test_cases:
-            with self.subTest(integration_type=integration_type,
-                              entity_type=entity_type):
-                url = self.base_url + f'/integrations?integration_type={integration_type}&entity_type={entity_type}'
-                response = requests.get(url)
-                response.raise_for_status()
-                response_json = response.json()
+            with self.subTest(integration_type=integration_type, entity_type=entity_type):
```
hannes-ucsc: swap this and next line, use self.subTest(*params)

test/service/test_response.py
```diff
@@ -1393,24 +1393,51 @@ def lambda_name(cls) -> str:
 
     maxDiff = None
 
+    def _get_integrations(self, params=None):
+        url = self.base_url + '/integrations'
+        response = requests.get(url, params=params)
+        response.raise_for_status()
+        return response.json()
+
+    # TODO: Change test to be more generic, not specific to current contents of portal integrations db
```
hannes-ucsc: Either resolve to do right away or file a ticket and mention here. Having said that, see below.

test/service/test_response.py
```diff
@@ -1393,24 +1393,51 @@ def lambda_name(cls) -> str:
 
     maxDiff = None
 
+    def _get_integrations(self, params=None):
```
hannes-ucsc: `params` should be mandatory. All call sites provide it.

test/service/test_response.py
```diff
@@ -1393,24 +1393,51 @@ def lambda_name(cls) -> str:
...
         for integration_type, entity_type, num_integrations_expected in test_cases:
-            with self.subTest(integration_type=integration_type,
-                              entity_type=entity_type):
-                url = self.base_url + f'/integrations?integration_type={integration_type}&entity_type={entity_type}'
-                response = requests.get(url)
-                response.raise_for_status()
-                response_json = response.json()
+            with self.subTest(integration_type=integration_type, entity_type=entity_type):
+                params = dict(integration_type=integration_type, entity_type=entity_type)
+                response_json = self._get_integrations(params=params)
```
```suggestion
                response_json = self._get_integrations(params)
```

test/service/test_response.py
```diff
@@ -1393,24 +1393,51 @@ def lambda_name(cls) -> str:
...
 
+    def test_integrations_entity_ids(self):
+        def extract_entity_ids(response_json):
+            return [entity_id
+                    for portal in response_json
+                    for integration in portal['integrations']
+                    for entity_id in integration['entity_ids']]
+        # To avoid hardcoding a specific UUID in this test we first make a request without entity_ids specified
+        params = dict(integration_type='get', entity_type='project')
+        response_json = self._get_integrations(params=params)
```
```suggestion
        response_json = self._get_integrations(params)
```

test/service/test_response.py
```diff
@@ -1393,24 +1393,51 @@ def lambda_name(cls) -> str:
...
+        entity_ids = extract_entity_ids(response_json)
+        # If we have an entity_id we can make a request specifying that entity_id to verify only it is returned
+        if len(entity_ids) > 0:
```
hannes-ucsc: If len(entity_ids) is 0 this test does nothing. Whenever you wrap an assertion in a conditional or loop you need to double question your approach. Generic tests are hard to write, complicated and difficult to understand. Mocking is easy. Mocking the portal database will simplify these tests considerably and take care of the TODO above.

test/service/test_response.py
danielsotirhos: Fixing the implementation by using mocked data

test/service/test_response.py
danielsotirhos: Yes, I was trying to handle an incomplete and varying dataset depending on DSS selected, in other words problems that existed because I didn't mock the database. Will correct this implementation with mock.

hannes-ucsc:
> Do we want to handle &entity_ids== differently to entity_ids being absent from the query?

Let's post-pone this. Focus on the tests and the other comments. Then file a ticket about this concern.

test/service/test_response.py
```diff
@@ -1393,24 +1394,164 @@ def lambda_name(cls) -> str:
 
     maxDiff = None
 
-    def test_integrations(self):
+    # Mocked db content for the tests
+    _portal_integrations_db = [
```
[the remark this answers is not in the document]
danielsotirhos: I've trimmed down the mocked db entries and removed extranious properties such as `contact_email` and `title`.
The db now contains 1 `get_manifest, file` integration and 4 `get, project` integrations, 2 of which reference the same hca project uuid. This combination of data allows the testcases that verify requests both return matched data and not unmatched data.

lambdas/service/app.py
```diff
@@ -319,19 +319,28 @@ def validate_params(query_params: Mapping[str, str],
...
         raise BadRequestError('Parameters entity_type and integration_type must be given')
-    body = _fetch_integrations(entity_type, integration_type)
+    body = _fetch_integrations(entity_type, integration_type, entity_ids)
     return Response(status_code=200,
                     headers={"content-type": "application/json"},
                     body=json.dumps(body))
 
 
-def _fetch_integrations(entity_type, integration_type):
+def _fetch_integrations(entity_type: str, integration_type: str, entity_ids: Set) -> List:
```
hannes-ucsc: Set of what? List of what? This is not the first time I point this out. There is a reason why the types in `typing` are generic.

lambdas/service/app.py
```diff
@@ -319,19 +319,29 @@ def validate_params(query_params: Mapping[str, str],
...
+    body = _fetch_integrations(entity_type, integration_type, entity_ids)
     return Response(status_code=200,
                     headers={"content-type": "application/json"},
                     body=json.dumps(body))
 
 
-def _fetch_integrations(entity_type, integration_type):
+def _fetch_integrations(entity_type: str, integration_type: str, entity_ids: Set[str]) -> List[Dict[str, Any]]:
+    """
+    Return matching portal integrations
```
```suggestion
    Return matching portal integrations.
```

lambdas/service/app.py
```diff
@@ -319,19 +319,29 @@ def validate_params(query_params: Mapping[str, str],
...
         raise BadRequestError('Parameters entity_type and integration_type must be given')
-    body = _fetch_integrations(entity_type, integration_type)
+    body = _fetch_integrations(entity_type, integration_type, entity_ids)
     return Response(status_code=200,
                     headers={"content-type": "application/json"},
                     body=json.dumps(body))
 
 
-def _fetch_integrations(entity_type, integration_type):
+def _fetch_integrations(entity_type: str, integration_type: str, entity_ids: Set[str]) -> List[Dict[str, Any]]:
```
hannes-ucsc: https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#16type-hints

hannes-ucsc: IOW

```suggestion
def _fetch_integrations(entity_type: str, integration_type: str, entity_ids: Set[str]) -> Sequence[JSON]:
```

danielsotirhos: Fixing with `Sequence[JSON]` however I noticed in types.py there is a `JSONs = Sequence[JSON]` that doesn't appear to be used anywhere in the codebase yet. Should we start using `JSONs` instead?

hannes-ucsc: Yes. Good catch.

hannes-ucsc:
Could you review my changes, @danielsotirhos?

danielsotirhos:
Rebasing branch on latest develop to fix branch conflicts.

hannes-ucsc:
> Without the # noinspection PyArgumentList, a warning Expected a dictionary, got dict is given for the **params. Was this intentional?

I get no warning if I remove the pragma, which is why I removed it. The warning you quote makes no sense so I guess it's related to the other anomaly in your PyCharm project.

PR #1330, opened 2019-09-29
"integration_type" moved to oneOf objects

VIIgit:
I'm not sure if this is the right and only approach, but this is how  the oneOf example is working


hannes-ucsc:
Given https://github.com/swagger-api/swagger-ui/pull/5530#issuecomment-537686694 we don't need this workaround anymore. Thanks again, @VIIgit !

PR #1331, opened 2019-09-30
Hotfix for staging, DSS bucket swap

achave11-ucsc:


hannes-ucsc:
The definition of critical is: _Issue can cause great damage until resolved, urgency is implied_

Urgent is more appropriate (_Resolution is needed soon_).

src/azul/__init__.py
```diff
@@ -156,7 +156,7 @@ def _dss_bucket(self, qualifier=None):
         stage = self.dss_deployment_stage
         # For domain_part, DSS went from `humancellatlas` to `hca` in 9/2018 and started reverting back to
         # `humancellatlas` in 12/2018. As I write this, only `dev` is back on `humancellatlas`
-        domain_part = 'hca' if stage == 'prod' else 'humancellatlas'
+        domain_part = 'hca' if stage in ('prod', 'staging') else 'humancellatlas'
```
hannes-ucsc: Let's also remove the comment above. I think it is wrong and it would be kinda pointless to try to keep up with the back and forth of bucket names in the DSS.

hannes-ucsc (changes requested):

hannes-ucsc (approved):

hannes-ucsc:
Let's hold this warm until DSS is giving us the go ahead.

[cc: @DailyDreaming]

Issue #1316, opened 2019-09-25
Rename TEST_MODE to AZUL_TEST_MODE


hannes-ucsc:
Since the current name is generic, there is a low chance of a collision with some other process in the Lambda execution environment.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-809)
┆Project Name: azul
┆Issue Number: AZUL-809

Issue #1317, opened 2019-09-26
Enable TEST_MODE last in IT set-up, first in tear-down


hannes-ucsc:
… such that an exception in `setUp`/`tearDown` doesn't leave `TEST_MODE` on.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-810)
┆Project Name: azul
┆Issue Number: AZUL-810

Solve together with #1316

PR #1332, opened 2019-09-30
Prefix TEST_MODE with AZUL_ and reduce chance of leaving it set (#1316, #1317)

GPelayo:


GPelayo:
Also resolves Issue #1316.

Issue #1345, opened 2019-10-02
Add Xena portal integrations


danielsotirhos:
From: Jing Zhu

We just updated our data on Xena with the HCA data portal production site. There are total of 12 projects with matrix gene expression data available on the HCA portal. 

We are ready to support link out from HCA. 

Daniel, please see the attached for the one to one mapping. It has four fields, tab delimitated.

HCA data portal display name
HCA project uuid
xena study name
xena url to link out

Xena urls all have the pattern https://singlecell.xenabrowser.net/datapages/?cohort=$URL_ENCODED_xena_study_name . 

[mapping_HCA_Xena_2019-10-01.txt](https://github.com/DataBiosphere/azul/files/3679565/mapping_HCA_Xena_2019-10-01.txt)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-824)
┆Project Name: azul
┆Issue Number: AZUL-824

PR #1346, opened 2019-10-02
Add Xena portal integrations (#1345)

danielsotirhos:


src/azul/project/hca/__init__.py
```diff
@@ -473,5 +473,171 @@ def portal_integrations_db(self) -> JSON:
...
                 ],
+            },
+            {
+                "portal_id": "2e05f611-16fb-4bf3-b860-aa500f0256de",
+                "portal_name": "Xena",
+                "portal_icon": "https://xenabrowser.net/03340e094d1f3edc51bc3d1a2a589b65.png",
+                "contact_email": "",
+                "organization_name": "UCSC",
+                "portal_description": "",
+                "integrations": [
```
hannes-ucsc: Time to switch to comprehensions, Daniel.

src/azul/project/hca/__init__.py
```diff
@@ -473,5 +474,64 @@ def portal_integrations_db(self) -> JSON:
...
+                        "integration_type": "get",
+                        "entity_type": "project",
+                        "title": title,
+                        "entity_ids": {
+                            "staging": [],
+                            "integration": [prod_uuid],
+                            "prod": [prod_uuid]
+                        },
+                        "portal_url": "https://singlecell.xenabrowser.net/datapages/?cohort=" + quote(title)
+                    } for integration_id, title, prod_uuid in (
```
hannes-ucsc: 1) title should come last
2) rename prod_uuid to project_uuid
3) one tuple per line, and suppress warning

See diff

danielsotirhos:
I apologize and take ownership for not realizing `JSON` was the wrong type in this case due to the return value not being a dict but a list of dict. My mistake was in misunderstanding the `JSON` type to be any structure that would be ready to be serialized as JSON instead of an alias for `Mapping[str, AnyJSON]`.

Regarding PyCharm's warnings, please note I have not ignored your advice and have been aiming for warning free commits. In this particular case I am confused however as in my PyCharm I get no such warning as you describe, instead I see 9 weak warnings: 1 that another method in the file could be static, and 8 for exceeding line length which I fixed have fixed in the `1337-update-SCEA-portal-details` branch.

Screen Shot 2019-10-03 at 9 14 58 AM

I will investigate why my PyCharm doesn't show the same warning you experienced, and increase effort to pay closer attention to type hinting going forward.

hannes-ucsc:
> Regarding PyCharm's warnings, please note I have not ignored your advice and have been aiming for warning free commits. In this particular case I am confused however as in my PyCharm I get no such warning as you describe, instead I see 9 weak warnings: 1 that another method in the file could be static, and 8 for exceeding line length which I fixed have fixed in the 1337-update-SCEA-portal-details branch.

@danielsotirhos I apologize. My accusation was unjustified.

Issue #1299, opened 2019-09-19
Use object tag instead of name object to name full manifest


hannes-ucsc:
Not sure how I missed this but S3 has https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUTtagging.html which can be used to associate tags with an object after it was created.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-802)
┆containerName: azul
┆Issue Number: AZUL-802

PR #1350, opened 2019-10-04
Use object tag instead of name object to name full manifest (#1299)

GPelayo:


src/azul/service/responseobjects/storage_service.py
```diff
@@ -126,6 +137,19 @@ def __complete(self):
 
         try:
             self.mp_upload.complete(MultipartUpload={"Parts": [part.to_dict() for part in self.parts]})
+            if self.filename:
```
achave11: You might want to consider adding this to an `else` branch after the `except`. But I am not sure since this could also raise an error.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -29,10 +29,20 @@ def head(self, object_key: str) -> dict:
     def get(self, object_key: str) -> bytes:
         return self.client.get_object(Bucket=self.bucket_name, Key=object_key)['Body'].read()
 
-    def put(self, object_key: str, data: bytes, content_type: Optional[str] = None, **kwargs) -> str:
+    def put(self, object_key: str, data: bytes, content_type: Optional[str] = None, filename: Optional[str] = None,
```
achave11: Indent all or none.

src/azul/service/responseobjects/storage_service.py
hannes-ucsc: Yes, the code you added should be in a separate `try:`, nested in the `else:` branch of this `try:`.

And you must now handle eventual consistency. If put_object_tagging claims that the object does not exist you should wait 5s and retry for up to a 1 min.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -87,7 +100,7 @@ class MultipartUploadHandler:
     automatically.
     """
 
-    def __init__(self, object_key, content_type: str):
+    def __init__(self, object_key, content_type: str, file_name: str = None):
```
hannes-ucsc: Bad smell: this parameter is never used by any call site.

src/azul/service/responseobjects/storage_service.py
```diff
@@ -135,6 +149,24 @@ def __complete(self):
                          exc_info=exception)
             self.__abort()
             raise MultipartUploadError(self.bucket_name, self.object_key) from exception
+        else:
```
hannes-ucsc: This feels out of place, as if it were tagged on. Setting the name tag on an object is orthogonal to how the object was created. MultipartUploadHandler should only handle the MP upload, not what happens with the objects afterwards.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -352,20 +349,19 @@ def _push_content(self) -> str:
                 elif self.format == 'full':
                     project_short_name = self._write_full(text_buffer)
                     if project_short_name:
-                        self._create_name_object(object_key, project_short_name)
+                        multipart_upload.file_name = self._create_name(project_short_name)
```
hannes-ucsc: Another bad smell: setting this attribute so late also indicates that tagging the object doesn't really have anything to do with the MP upload itself.

src/azul/service/storage_service.py
```diff
@@ -32,10 +35,17 @@ def head(self, object_key: str) -> dict:
     def get(self, object_key: str) -> bytes:
         return self.client.get_object(Bucket=self.bucket_name, Key=object_key)['Body'].read()
 
-    def put(self, object_key: str, data: bytes, content_type: Optional[str] = None, **kwargs) -> str:
+    def put(self,
+            object_key: str,
+            data: bytes,
+            content_type: Optional[str] = None,
+            file_name: Optional[str] = None,
```
hannes-ucsc: Let's generalize this

```suggestion
            tagging: Optional[Mapping[str, str]] = None,
```

src/azul/service/storage_service.py
```diff
@@ -66,6 +76,27 @@ def get_presigned_url(self, key: str, file_name: Optional[str] = None) -> str:
     def create_bucket(self, bucket_name: str = None):
         self.client.create_bucket(Bucket=(bucket_name or self.bucket_name))
 
+    def put_filename_tag(self, object_key: str, file_name: str):
+        deadline = time.time() + 60
+        tagging = {'TagSet': [{'Key': 'azul_file_name', 'Value': file_name}]}
```
hannes-ucsc: The tag name (aka "tag key") should be extracted to a class attribute, like the suffix was before.

src/azul/service/storage_service.py
```diff
@@ -66,6 +76,27 @@ def get_presigned_url(self, key: str, file_name: Optional[str] = None) -> str:
     def create_bucket(self, bucket_name: str = None):
         self.client.create_bucket(Bucket=(bucket_name or self.bucket_name))
 
+    def put_filename_tag(self, object_key: str, file_name: str):
```
hannes-ucsc:
```suggestion
    def put_object_tagging(self, tagging: Optional[Mapping[str, str]]):
```

src/azul/service/storage_service.py
```diff
@@ -97,6 +128,7 @@ def __init__(self, object_key, content_type: str):
         self.parts = []
         self.futures = []
         self.thread_pool = None
+        self.filename = None
```
hannes-ucsc: What's this for?

GPelayo: That should be removed, oops.

src/azul/service/storage_service.py
GPelayo: Why is the object key argument removed?

hannes-ucsc: A mistake.

```suggestion
    def put_object_tagging(self, object_key: str, tagging: Optional[Mapping[str, str]]):
```

src/azul/service/storage_service.py
```diff
@@ -32,10 +38,17 @@ def head(self, object_key: str) -> dict:
...
+    def put(self,
+            object_key: str,
+            data: bytes,
+            content_type: Optional[str] = None,
+            tagging: Optional[Mapping[str, str]] = None,
+            **kwargs) -> str:
         params = {'Bucket': self.bucket_name, 'Key': object_key, 'Body': data, **kwargs}
         if content_type is not None:
             params['ContentType'] = content_type
+        if tagging is not None:
```
hannes-ucsc: put() is never called with `tagging` (or your previous `file_name` argument). You should catch these things.

hannes-ucsc:
Correction: we don't know the value for the tag when we create the MP upload so we'll have to resort to retro-actively applying it. I'll still take this on, though.

src/azul/service/manifest_service.py
```diff
@@ -138,16 +137,13 @@ def _use_cached_manifest(self, generator: 'ManifestGenerator', object_key: str)
...
-            except ClientError as e:
-                if e.response['Error']['Code'] == 'NoSuchKey':
-                    logger.warning("Name object '%s' doesn't exist. "
-                                   "Generating pre-signed URL without Content-Disposition header.", name_object_key)
-                    file_name = None
-                else:
-                    raise
+                file_name = one([tag['Value'] for tag in tag_set if tag['Key'] == 'azul_file_name'])
+            except ValueError:
+                logger.warning("Manifest object '%s' doesn't have the tag, azul_file_name."
```
hannes-ucsc: Forgot to extract the class attribute here.

src/azul/service/manifest_service.py
```diff
@@ -138,16 +137,13 @@ def _use_cached_manifest(self, generator: 'ManifestGenerator', object_key: str)
...
             try:
-                file_name = self.storage_service.get(name_object_key).decode()
-            except ClientError as e:
-                if e.response['Error']['Code'] == 'NoSuchKey':
-                    logger.warning("Name object '%s' doesn't exist. "
-                                   "Generating pre-signed URL without Content-Disposition header.", name_object_key)
-                    file_name = None
-                else:
-                    raise
+                file_name = one([tag['Value'] for tag in tag_set if tag['Key'] == 'azul_file_name'])
```
hannes-ucsc: And here.

Issue #1174, opened 2019-07-31
IT fails at times unless reindex has been run


hannes-ucsc:
https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/2518 was the first run of the  integration test after https://github.com/DataBiosphere/azul/commit/d537d817ed5d7f6678756a7b57dd9b8425c0430e had been deployed to `dev`. That commit changed the contents of index documents by replacing all `null` (`None`) values with other discrete values, like `__null__` for string properties. The test ultimately fails detecting all bundles.

The indexer logs (/aws/lambda/azul-indexer-dev-write) contain failures like this:

```
Traceback (most recent call last):
  File "/var/task/app.py", line 159, in new_handler
    return old_handler(self, event, context)
  File "/var/task/chalice/app.py", line 989, in __call__
    return self.func(event_obj)
  File "/var/task/app.py", line 277, in write
    indexer.aggregate(tallies)
  File "/var/task/azul/indexer.py", line 191, in aggregate
    contributions = self._read_contributions(total_tallies)
  File "/var/task/azul/indexer.py", line 252, in _read_contributions
    contributions = [Contribution.from_index(hit) for hit in hits]
  File "/var/task/azul/indexer.py", line 252, in <listcomp>
    contributions = [Contribution.from_index(hit) for hit in hits]
  File "/var/task/azul/transformer.py", line 181, in from_index
    source = cls.translate_fields(hit['_source'], forward=False)
  File "/var/task/azul/transformer.py", line 128, in translate_fields
    return {key: cls.translate_fields(val, path=path + (key,), forward=forward) for key, val in source.items()}
  File "/var/task/azul/transformer.py", line 128, in <dictcomp>
    return {key: cls.translate_fields(val, path=path + (key,), forward=forward) for key, val in source.items()}
  File "/var/task/azul/transformer.py", line 128, in translate_fields
    return {key: cls.translate_fields(val, path=path + (key,), forward=forward) for key, val in source.items()}
  File "/var/task/azul/transformer.py", line 128, in <dictcomp>
    return {key: cls.translate_fields(val, path=path + (key,), forward=forward) for key, val in source.items()}
  File "/var/task/azul/transformer.py", line 130, in translate_fields
    return [cls.translate_fields(val, path=path, forward=forward) for val in source]
  File "/var/task/azul/transformer.py", line 130, in <listcomp>
    return [cls.translate_fields(val, path=path, forward=forward) for val in source]
[...]

PR #1352, opened 2019-10-05
 Fix: IT fails at times unless reindex has been run (#1174) 

jessebrennan:


src/azul/azulclient.py
```diff
@@ -106,10 +107,13 @@ def test_notifications(self, test_name: str, test_uuid: str) -> Tuple[List[dict]
         effective_bundle_fqids = set()
         for bundle_fqid in real_bundle_fqids:
             new_bundle_uuid = str(uuid.uuid4())
-            _, _, version = bundle_fqid.partition('.')
-            effective_bundle_fqids.add(new_bundle_uuid + '.' + version)
+            # Using a new version helps ensure that any aggregation will choose this entity.
+            # Failing to do this before caused https://github.com/DataBiosphere/azul/issues/1174
+            new_bundle_version = datetime.utcnow().isoformat().replace(':', '') + 'Z'
```
hannes-ucsc: Can you consolidate with this:

https://github.com/DataBiosphere/azul/blob/develop/scripts/copy_bundles.py#L227

by moving that code into the `azul.dss` module and reusing it here. You may have to add that module to the suite in `test_doctest.py`.

Just double check that the code I link to actually generates a version representing the current time in UTC.

src/azul/azulclient.py
```diff
@@ -106,10 +107,13 @@ def test_notifications(self, test_name: str, test_uuid: str) -> Tuple[List[dict]
         effective_bundle_fqids = set()
         for bundle_fqid in real_bundle_fqids:
             new_bundle_uuid = str(uuid.uuid4())
-            _, _, version = bundle_fqid.partition('.')
-            effective_bundle_fqids.add(new_bundle_uuid + '.' + version)
+            # Using a new version helps ensure that any aggregation will choose this entity.
```
hannes-ucsc:
```suggestion
            # Using a new version helps ensure that any aggregation will choose the contribution from the test bundle over that by any other bundle not in the test set.
```

src/azul/azulclient.py
jessebrennan: I missed
 > by moving that code into the azul.dss module and reusing it here. You may have to add that module to the suite in test_doctest.py.

On my first pass. It was simpler and seemed appropriate to put it in a separate PR which I did: #1363.

scripts/copy_bundles.py
```diff
@@ -212,21 +213,15 @@ def _copy_version(self, version: str):
...
                 return mode
 
-    version_format = '%Y-%m-%dT%H%M%S.%fZ'
-
-    @classmethod
-    def _new_version(cls):
-        return datetime.utcfromtimestamp(time.time()).strftime(cls.version_format)
-
     @classmethod
     def _validate_version(cls, version: str):
```
hannes-ucsc: We should move this too.

scripts/copy_bundles.py
```diff
@@ -26,6 +26,8 @@
 from azul.dss import (
     MiniDSS,
     shared_dss_credentials,
```
hannes-ucsc: https://github.com/DataBiosphere/azul/pull/1366/files#diff-61633c57a57b908ccba54e0971f8fd0dR26

Symbols in azul.dss should not duplicate the `dss` string. The way this is implemented here and now, `new_version` is a bit general and lacks context. But if references as `azul.dss.new_version` it becomes clear.

scripts/copy_bundles.py
```diff
@@ -197,7 +197,7 @@ def _copy_bundle(self, bundle_uuid, bundle_version, manifest, attempt=0):
                                         creator_uid=0,
                                         files=manifest['files'])
         except SwaggerAPIException as e:
-            if e.code == 400 and e.reason == 'file_missing' and attempt < 10:
+            if e.code == 401 and e.reason == 'file_missing' and attempt < 10:
```
hannes-ucsc: Remove this change.

src/azul/transformer.py
```diff
@@ -751,10 +751,16 @@ def _select_latest(self, contributions: Sequence[Contribution]) -> MutableMappin
...
                     for entity in entities:
                         entity_id = entity['document_id']  # FIXME: the key 'document_id' is HCA specific
-                        bundle_version, _ = collated_entities.get(entity_id, ('', None))
-                        if bundle_version < contribution.bundle_version:
-                            collated_entities[entity_id] = contribution.bundle_version, entity
+                        cur_bundle_uuid, cur_bundle_version, cur_entity = \
+                            collated_entities.get(entity_id, (None, '', None))
+                        if cur_entity is not None and entity.keys() != cur_entity.keys():
+                            symmetric_difference = set(entity.keys()).symmetric_difference(cur_entity)
+                            logger.warning('Shape of `%s` entities does not match between bundles %s, version %s '
```
jessebrennan:
```suggestion
                            logger.warning('Document shape of `%s` entity does not match between bundles %s, version %s '
```

Issue #1290, opened 2019-09-17
Duplication in health endpoints between service and indexer


achave11:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-798)
┆Project Name: azul
┆Issue Number: AZUL-798

jessebrennan:
Can you explain this a bit more? What is the duplication?

achave11:
`/health` endpoints in Azul (service & indexer) have the same code duplicated when generating the health object:
```
    if keys == 'basic':
        return {
            'up': True,
        }
    else:
        health = Health('{LAMBDA-NAME}')
        if keys is None:
            if app.current_request.context['path'].endswith('/'):
                keys = ()
        else:
            keys = keys.split(',')
        try:
            body = health.as_json(keys)
        except RequirementError:
            raise BadRequestError('Invalid health keys')
        return Response(body=json.dumps(body),
                        status_code=200 if body['up'] else 503)
```

PR #1354, opened 2019-10-05
Remove health endpoint duplication in indexer & service lambdas (#1290)

achave11:


src/azul/health.py
```diff
@@ -151,3 +159,43 @@ def _lambda(self, lambda_name) -> JSON:
             return {
                 'up': up,
             }
+
+
+class HealthHandler(Health):
+    storage_service = StorageService()
```
hannes-ucsc: Should be an instance attribute.

src/azul/health.py
```diff
@@ -151,3 +159,43 @@ def _lambda(self, lambda_name) -> JSON:
             return {
                 'up': up,
             }
+
+
+class HealthHandler(Health):
```
hannes-ucsc:
```suggestion
class HealthController(Health):
```

lambdas/service/app.py
```diff
@@ -191,41 +190,18 @@ def openapi():
...
-
-
-@app.schedule('rate(1 minute)', name=config.service_cache_health_lambda_basename)
-def generate_health_object(event_: chalice.app.CloudWatchEvent):
-    storage_service = StorageService()
-    health = Health('service')
-    health = health.as_json()
-    health_object = dict(time=time.time(), health=health)
-    storage_service.put(object_key='health/service', data=json.dumps(health_object))
+    context_route_path = app.current_request.context['path']
```
hannes-ucsc: Inline context_route_path

hannes-ucsc: Use keyword arguments.

lambdas/service/app.py
```diff
@@ -191,41 +190,18 @@ def openapi():
...
+    health_handler = HealthHandler('service', keys, context_route_path)
+    return health_handler.response()
+
+
+@app.schedule('rate(1 minute)', name=config.cache_health_lambda_basename)
+def generate_health_object(_event: chalice.app.CloudWatchEvent):
+    service_health_handler = HealthHandler('service')
+    service_health_handler.generate_cache()
+
+    indexer_health_handler = HealthHandler('indexer')
```
hannes-ucsc: I'd like there to be a second, separate scheduled lambda for the indexer. We've chosen to implement health symmetrically and until we change our mind we should stick with that decision. Other than being slightly less efficient, do you see other problems with the symmetric approach?

test/health_check_test_case.py
```diff
@@ -115,7 +115,7 @@ def test_cached_health(self):
         # A successful response is obtained when all the systems are functional
         self._create_mock_queues()
         endpoint_states = self._make_endpoint_states(self.endpoints)
-        app = load_app_module(self.lambda_name())
+        app = load_app_module('service')
```
jessebrennan: I don't think this change is related, but for some reason this test fails for me locally with both the service and the indexer. Do you have any idea why?

src/azul/health.py
```diff
@@ -151,3 +159,43 @@ def _lambda(self, lambda_name) -> JSON:
...
+    storage_service = StorageService()
+
+    def __init__(self, lambda_name: str, keys: str = None, request_path: str = '/health'):
+        super().__init__(lambda_name)
+        self.keys = keys
+        self.request_path = request_path
+
+    def response(self):
+        pass
```
hannes-ucsc: Remove

lambdas/service/app.py
achave11: Not at all.

test/health_check_test_case.py
achave11: `AZUL_DEBUG=1`

lambdas/indexer/app.py
```diff
@@ -72,41 +66,16 @@ def version():
...
-                keys = ()
-        else:
-            keys = keys.split(',')
-        try:
-            body = health.as_json(keys)
-        except RequirementError:
-            raise BadRequestError('Invalid health keys')
-    return Response(body=json.dumps(body),
-                    status_code=200 if body['up'] else 503)
+    health_handler = HealthController(lambda_name='indexer',
```
```suggestion
    controller = HealthController(lambda_name='indexer',
```

lambdas/indexer/app.py
```diff
@@ -72,41 +66,16 @@ def version():
...
 
 @app.schedule('rate(1 minute)', name=config.indexer_cache_health_lambda_basename)
-def generate_health_object(event_: chalice.app.CloudWatchEvent):
-    storage_service = StorageService()
-    health = Health('indexer')
-    health = health.as_json()
-    health_object = dict(time=time.time(), health=health)
-    storage_service.put(object_key='health/indexer', data=json.dumps(health_object))
+def generate_health_object(_event: chalice.app.CloudWatchEvent):
+    indexer_health_handler = HealthController(lambda_name='indexer')
```
```suggestion
    controller = HealthController(lambda_name='indexer')
```

lambdas/service/app.py
```diff
@@ -191,41 +190,16 @@ def openapi():
...
-                keys = ()
-        else:
-            keys = keys.split(',')
-        try:
-            body = health.as_json(keys)
-        except RequirementError:
-            raise BadRequestError('Invalid health keys')
-    return Response(body=json.dumps(body),
-                    status_code=200 if body['up'] else 503)
+    health_handler = HealthController(lambda_name='service',
```
```suggestion
    controller = HealthController(lambda_name='service',
```

lambdas/service/app.py
```diff
@@ -191,41 +190,16 @@ def openapi():
...
 
 @app.schedule('rate(1 minute)', name=config.service_cache_health_lambda_basename)
-def generate_health_object(event_: chalice.app.CloudWatchEvent):
-    storage_service = StorageService()
-    health = Health('service')
-    health = health.as_json()
-    health_object = dict(time=time.time(), health=health)
-    storage_service.put(object_key='health/service', data=json.dumps(health_object))
+def generate_health_object(_event: chalice.app.CloudWatchEvent):
+    service_health_handler = HealthController(lambda_name='service')
```
```suggestion
    controller = HealthController(lambda_name='service')
```

src/azul/health.py
```diff
@@ -151,3 +159,46 @@ def _lambda(self, lambda_name) -> JSON:
...
+        if self.keys == 'basic':
+            return {
+                'up': True,
+            }
+        elif self.keys == 'cached':
+            try:
+                health_object = json.loads(self.storage_service.get(f'health/{self.lambda_name}'))
+            except self.storage_service.client.exceptions.NoSuchKey:
+                return Response(status_code=500,
+                                body=json.dumps({'Message': 'Object key does not exist in S3'}))
```
hannes-ucsc:
```suggestion
                                body=json.dumps({'Message': 'Cached health object does not exist'}))
```

Note the linguistic correspondence between the two error messages.

lambdas/indexer/app.py
```diff
@@ -72,41 +66,16 @@ def version():
...
-            keys = keys.split(',')
-        try:
-            body = health.as_json(keys)
-        except RequirementError:
-            raise BadRequestError('Invalid health keys')
-    return Response(body=json.dumps(body),
-                    status_code=200 if body['up'] else 503)
+    controller = HealthController(lambda_name='indexer',
+                                  keys=keys,
+                                  request_path=(app.current_request.context['path']))
```
```suggestion
                                  request_path=app.current_request.context['path'])
```

hannes-ucsc: Does your PyCharm not warn you about this? Mine usually does.

achave11: No warning.

hannes-ucsc:
Re protocol: Amending a commit counts as squashing. If you amend or squash I can't easily see how you addressed my comments since the subsequent force push could also include changes resulting from a rebase which you are encouraged to do on every review cycle.

lambdas/indexer/app.py
hannes-ucsc: Did you see https://humancellatlas.slack.com/archives/CGDQ4DNJY/p1570658294009600?

achave11:
Apologies, I do try to follow protocol. However, I rebase frequently to stay on top of the latest changes. This requires me to force push, so I consolidate all `fixit!` commits before pushing. I'll be more reserved when rebasing and consolidating commits.

src/azul/health.py
```diff
@@ -151,3 +159,46 @@ def _lambda(self, lambda_name) -> JSON:
             return {
                 'up': up,
             }
+
+
+class HealthController(Health):
```
hannes-ucsc: Controllers shouldn't inherit models because the [is-a](https://en.wikipedia.org/wiki/Is-a) relationship does not apply to [controllers and models](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller). The easy counter example is that one controller often need to deal with multiple model instances and then the subtype relation breaks down. Also, in typical web apps the controllers are singletons that outlive the models.

Issue #1360, opened 2019-10-07
Remove SCP495 study entry from portals database


theathorn:
Remove the URL for the unpopulated study: https://portals.broadinstitute.org/single_cell/study/SCP495
for the project titled "Ischaemic sensitivity of human tissue by single cell RNA seq" (https://data.humancellatlas.org/explore/projects/c4077b3c-5c98-4d26-a614-246d12c2e5d7).

From the PM meeting: for the "Ischaemic sensitivity" project the submitter states that it is not complete and they will be adding more data.

From Vicky Horst: I think we should focus on enabling the automated handoff [get_entities] as the next goal. We are looking to the general policy for HCA data and its implications on SCP studies.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-831)
┆Project Name: azul
┆Issue Number: AZUL-831
┆Epic: MVP support for SCP

PR #1362, opened 2019-10-07
Remove SCP495 study entry from portals database (#1360)

danielsotirhos:


hannes-ucsc:
We should stick with protocol, @theathorn. Build must pass before reviewer is assigned.

hannes-ucsc:
I think the conflicts arose from me merging #1329.

Issue #1364, opened 2019-10-08
Add support for `file_core.content description`


hannes-ucsc:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-833)
┆Project Name: azul
┆Issue Number: AZUL-833

PR #1377, opened 2019-10-11
[R] Add support for `file_core.content_description` (#1364)

danielsotirhos:


test/indexer/test_hca_indexer.py
```diff
@@ -814,6 +814,17 @@ def test_cell_line_sample(self):
                 self.assertEqual(one(contents['cell_suspensions'])['organ'], ['blood (child_cell_line)'])
                 self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [config.null_keyword])
 
+    def test_file_core(self):
+        """
+        """
```
hannes-ucsc: Remove or add doc string.

test/indexer/test_hca_indexer.py
```diff
@@ -814,6 +814,17 @@ def test_cell_line_sample(self):
...
+        self._index_canned_bundle(('e0ae8cfa-2b51-4419-9cde-34df44c6458a', '2018-12-05T230917.591044Z'))
```
hannes-ucsc: If we are adding a new test, which requires indexing the bundle, I'd rather can a new bundle from production that has this field filled out so we don't need to modify an existing can.

danielsotirhos:
Changed to coverage 4.5.4, results seem better(?) however still lots of blank lines and comment lines listed as `uncovered` in the results.

test/indexer/test_hca_indexer.py
```diff
@@ -840,6 +840,17 @@ def test_cell_line_sample(self):
                 self.assertEqual(one(contents['cell_suspensions'])['organ'], ['blood (child_cell_line)'])
                 self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [config.null_keyword])
 
+    def test_files_content_description(self):
+        self._index_canned_bundle(('ffac201f-4b1c-4455-bd58-19c1a9e863b4', '2019-10-09T170735.528600Z'))
+        hits = self._get_all_hits()
+        for hit in hits:
+            contents = hit['_source']['contents']
+            entity_type, aggregate = config.parse_es_index_name(hit['_index'])
+            if entity_type in ('files', 'bundles') or not aggregate:
```
hannes-ucsc: Still wrong, I think. The `content_description` should be in documents for all ~entities~ entity types, and for both, contributions and aggregates. Maybe you're missing an aggregator/accumulator?

test/indexer/test_hca_indexer.py
```diff
@@ -840,6 +840,24 @@ def test_cell_line_sample(self):
                 self.assertEqual(one(contents['cell_suspensions'])['organ'], ['blood (child_cell_line)'])
                 self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [config.null_keyword])
 
+    def test_files_content_description(self):
+        self._index_canned_bundle(('ffac201f-4b1c-4455-bd58-19c1a9e863b4', '2019-10-09T170735.528600Z'))
+        hits = self._get_all_hits()
+        # Specify how many files we expect in each index from bundle with 2 sequence files
+        num_files_per_index = {
```
hannes-ucsc: This could be made less brittle. We'd have to modify this after adding a new outer entity. Also don't like the name ("per index" is not accurate) and the fact that it's not obvious that the boolean keys represent the `aggregate` flag. Only files and bundles are special, this mapping hides that to some extent.  

```python
if aggregate:
    # bundle aggregates keep individual files
    num_inner_files = 2 if entity_type == 'bundles' else 1
else:
    # one inner file per file contribution
    num_inner_files = 1 if entity_type == 'files' else 2
```

Fewer and shorter lines, more expressive.

Issue #1378, opened 2019-10-14
Allow sorting on donor count


MillenniumFalconMechanic:
As per @ambrosejcarr on [#858](https://github.com/HumanCellAtlas/data-browser/issues/858):

> I was surprised that I wasn't able to sort based on those columns. At least for the cell count, as an Analysis Roadbuilder, I want to be able to find the datasets with the largest amount of cells because those datasets have the largest information content and promise to be the most effective at helping me train my cell biology-focused machine learning methods.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-841)
┆Project Name: azul
┆Issue Number: AZUL-841

hannes-ucsc:
Azul supports sorting on cell count:

```
$ http 'https://service.explore.data.humancellatlas.org/repository/projects?filters=%7B%7D&size=15&sort=cellCount&order=desc' | jq '.hits[].cellSuspensions[].totalCells'
1480000
1330000
339089
274182
253910
230800
15175
44615
44000
30000
30000
20000
19916
16893
12404
10001
12109
2464
1432
1580
10561
5862
1638
1391
7115
2663
1961
1923
951
1718
4149
1151
5355
981
865
```

hannes-ucsc:
For donor count, it looks like DB is counting the number of items in `hits[].donorOrganisms[0].id`. That's unreliable because I'm pretty sure we cap the number of items in that field. 

Azul should be extended to track the number of donors during aggregation. We should expose that number as a response field as `hits[].donorOrganisms[0].donorCount`. In the index, the corresponding `contents.donors[].donor_count` field should only be present in aggregate documents and it should be computed at aggregation time as the size of the `contents.donors[].biomaterial_id` field set. There may be a few novelties in this approach but I'll discuss those with the assignee.

hannes-ucsc:
Since sorting on cell count is already possible, I changed the title accordingly.

PR #1391, opened 2019-10-22
[R] Allow sorting on donor count (#1378)

noah-aviel-dove:


src/azul/transformer.py
```diff
@@ -648,7 +648,7 @@ def accumulate(self, value) -> bool:
         """
         :return: True, if the given value increased the count of unique values
         """
-        return self.items.accumulate(value)
+        return self.set_accum.accumulate(value)
```
jessebrennan: This change belongs in the previous commit. If you'd rather not move the change, consider squashing these together.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -822,7 +822,14 @@ def make_donor(self, donor):
...
     def make_donors(self, entry):
-        return [self.make_donor(donor) for donor in entry["contents"]["donors"]]
+        donors = entry["contents"]["donors"]
+        if donors:
+            count = donors[0]['donor_count']
+            result = [self.make_donor(donor) for donor in donors]
+            result[0]['donorCount'] = count
+            return result
+        else:
+            return []
```
jessebrennan: I'm confused why this change made here. I think instead you should be changing the `self.make_donor()` function.

jessebrennan: I'll leave this to @hannes-ucsc's discretion.

hannes-ucsc: Unless I'm mistaken, `len(entry["contents"]["donors"]) < 2` invariantly. If, in the future, we started to group donors by some property (like we do for files which are grouped by format), then there may be more than one donor in an aggregate document. In that case we'd want a count in every donor in the response. My instructions where imprecise in that respect. I think we should just modify make_donor as @jessebrennan suggests, and have the count in every donor in the response.

src/azul/transformer.py
hannes-ucsc: By our current, ill-defined standards, the history is a bit fragmented. All of the commits are tightly coupled, so much so that I don't foresee us reverting one of them and leaving the others. That's my goto criterion for deciding what goes into a commit: would I ever want to revert a change separately from another change. If yes, separate commit. Otherwise, same commit.

src/azul/transformer.py
```diff
@@ -635,6 +635,26 @@ def get(self):
         return self.value.get()
 
 
+class SetCountingAccumulator(Accumulator):
```
```suggestion
class CountUniqueValuesAccumulator(Accumulator):
```

src/azul/transformer.py
```diff
@@ -635,6 +635,26 @@ def get(self):
...
+    """
+    Count the number of unique values
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.set_accum = SetAccumulator()
```
```suggestion
        self.value = SetAccumulator()
```

For consistency with other accumulators.

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -814,6 +814,7 @@ def make_cell_lines(self, entry):
     def make_donor(self, donor):
         return {
             "id": donor["biomaterial_id"],
+            "donorCount": donor["donor_count"],
```
hannes-ucsc: The IT fails when hitting the service's health endpoint which in turns hits the repository endpoints. Hitting those manually after enabling AZUL_DEBUG yields.

```
$ http 'https://service.sandbox.dev.explore.data.humancellatlas.org/repository/bundles?size=1'
HTTP/1.1 500 Internal Server Error

  File "/var/task/chalice/app.py", line 726, in _get_view_function_response
    response = view_function(**function_args)
  File "/var/task/app.py", line 580, in get_bundle_data
    return repository_search('bundles', bundle_uuid)
  File "/var/task/app.py", line 435, in repository_search
    return service.get_data(entity_type, pagination, filters, item_id, file_url)
  File "/var/task/azul/service/repository.py", line 86, in get_data
    return self._get_items(entity_type, pagination, filters, file_url_func)
  File "/var/task/azul/service/repository.py", line 60, in _get_items
    response = self._get_data(entity_type, pagination, filters, file_url_func)
  File "/var/task/azul/service/repository.py", line 39, in _get_data
    entity_type=entity_type)
  File "/var/task/azul/service/responseobjects/elastic_request_builder.py", line 526, in transform_request
    final_response = FileSearchResponse(hits, paging, facets, entity_type)
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 990, in __init__
    KeywordSearchResponse.__init__(self, hits, entity_type)
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 898, in __init__
    self.map_entries(x) for x in hits], 'pagination': None}
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 898, in <listcomp>
    self.map_entries(x) for x in hits], 'pagination': None}
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 882, in map_entries
    donorOrganisms=self.make_donors(entry),
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 827, in make_donors
    return [self.make_donor(donor) for donor in entry["contents"]["donors"]]
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 827, in <listcomp>
    return [self.make_donor(donor) for donor in entry["contents"]["donors"]]
  File "/var/task/azul/service/responseobjects/hca_response_v5.py", line 817, in make_donor
    "donorCount": donor["donor_count"],
KeyError: 'donor_count'
```

So while it is a good default approach to use `[]` for keys that you expect to be present, this creates a problem during the integration test which hits the service endpoint *before* a reindexing can be triggered in Gitlab manually. I think we should use `.get` as for the other document fields.

hannes-ucsc:
This requires re-indexing.

hannes-ucsc:
This will require reindexing.

src/azul/service/responseobjects/hca_response_v5.py
noah-aviel-dove: Ah, so in other words, this function needs to process donors that were assembled by older versions of the code that don't have the other changes I made. Got it.

hannes-ucsc: Yes, there is a little bit of a chicken-egg problem. We don't want to reindex until we've vetted the code but in this case the vetting fails because we haven't reindex yet. I'm currently unsure as to how we'll solve this.

hannes-ucsc: I would like the service code to be strict when it comes to consuming the index, but at the moment that's not really possible.

Issue #1353, opened 2019-10-05
Make all `/integrations` params optional or support mandatory params in validation


hannes-ucsc:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-827)
┆Project Name: azul
┆Issue Number: AZUL-827

hannes-ucsc:
The handler for `/integrations` does extra validation to enforce that its parameters are being passed. This should be generalized and moved to the general validator functionality.

PR #1392, opened 2019-10-22
Add check for mandatory integration parameters in validation (#1353)

noah-aviel-dove:


lambdas/service/app.py
```diff
@@ -356,6 +357,7 @@ def validate_facet(value):
 
 def validate_params(query_params: Mapping[str, str],
                     allow_extra_params: bool = False,
+                    mandatory_params: Iterable[str] = None,
```
hannes-ucsc: Instead of adding a new parameter, please consider this approach:

```
class Mandatory:

    def __init__(self, validator) -> None:
        super().__init__()
        self._validator = validator

    def __call__(self, value):
        return self._validator(value)
```

This could then be used as­ `validate_params(…,foo=Mandatory(int))` which I think reads better. You can then compute `mandatory_params` by taking the values in `validators` that are `instanceof(Mandatory)`.

lambdas/service/app.py
```diff
@@ -385,24 +389,53 @@ def validate_params(query_params: Mapping[str, str],
...
+    >>> validate_params({}, mandatory_params=['foo'], foo=str)
+    Traceback (most recent call last):
+        ...
+    chalice.app.BadRequestError: BadRequestError: Missing required query parameter `foo`
+
     """
-    for param_name, param_value in query_params.items():
+
+    def fmt_error(err_description, params):
+        s = 's' if len(params) != 1 else ''
```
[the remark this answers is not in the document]
hannes-ucsc: `pluralize` can then be used directly in the f-string.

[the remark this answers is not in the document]
jessebrennan: What if the word ends with `s`?

lambdas/service/app.py
noah-aviel-dove: Okay. I like this approach.

lambdas/service/app.py
```diff
@@ -385,24 +389,53 @@ def validate_params(query_params: Mapping[str, str],
...
+        joined = ', '.join([f'`{p}`' for p in params])
+        return f'{err_description} query parameter{s} {joined}'
```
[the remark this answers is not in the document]
noah-aviel-dove: I wrote the `fmt_error` function to ensure that each parameter was separately enclosed in backticks. The pluralization was an afterthought. But it's a good idead to make that behavior accessible to other modules.

lambdas/service/app.py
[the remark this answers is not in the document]
noah-aviel-dove: In English, words that end in s, x, z, ch, or sh usually are are pluralized by appending '-es'. I don't think its reasonable to expect our code to do anything more complicated than that ("mouses" is still perfectly understandably even if "mice" is technically more correct).  This is very easy to add to the function Hannes is proposing.

lambdas/service/app.py
```diff
@@ -359,24 +374,52 @@ def validate_params(query_params: Mapping[str, str],
...
+    >>> validate_params({}, foo=Mandatory(str))
+    Traceback (most recent call last):
+        ...
+    chalice.app.BadRequestError: BadRequestError: Missing required query parameter `foo`
+
     """
-    for param_name, param_value in query_params.items():
+
+    def fmt_error(err_description, params):
+        joined = ', '.join([f'`{p}`' for p in params])
```
```suggestion
        joined = ', '.join(f'`{p}`' for p in params)
```

src/azul/strings.py
```diff
@@ -20,3 +20,13 @@ def departition(before, sep, after):
         return before
     else:
         return before + sep + after
+
+
+def pluralize(word: str, count: int) -> str:
```
hannes-ucsc: Some doctests would be nice here.

src/azul/strings.py
```diff
@@ -20,3 +20,29 @@ def departition(before, sep, after):
...
+    'foo'
+
+    >>> pluralize('bar', 2)
+    'bars'
+
+    >>> pluralize('baz', 2)
+    'bazes'
+
+    >>> pluralize('woman', 2)
+    'womans'
```
hannes-ucsc: +1 for adding a test that demonstrates the limitations of the current implementation. In the future we may switch to https://inflection.readthedocs.io/en/latest/ if we think it's worth it.

Issue #1340, opened 2019-10-01
Handle entity_ids param differently if not given vs given empty


danielsotirhos:
Handle `&entity_ids==` differently to `entity_ids` being absent from the query.

`&entity_ids==` could be interpreted as "Give me all integrations that match the other two criteria and that don't have an entity_ids attribute".

https://github.com/DataBiosphere/azul/pull/1329#issuecomment-536842704

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-821)
┆Project Name: azul
┆Issue Number: AZUL-821

hannes-ucsc:
See also #1353

theathorn:
Waiting for #1383 to demo.

PR #1394, opened 2019-10-23
Handle blank entity_ids parameter differently than when parameter is missing (#1340)

noah-aviel-dove:


hannes-ucsc:
I'd like @danielsotirhos to take a look first.

test/service/test_response.py
```diff
@@ -1502,14 +1509,18 @@ def test_integrations_by_entity_ids(self, portal_integrations_db):
...
-                    response_json = self._get_integrations(params)
-                    found_integration_ids = self._extract_integration_ids(response_json)
-                    self.assertEqual(set(integration_ids), set(found_integration_ids))
+                assert_matched_integrations(params, integration_ids)
+            # Test for blank entity_ids param (shouldn't match anything in this case)
+            params['entity_ids'] = ''
+            assert_matched_integrations(params, [])
+            # Test for missing entity_ids param (no filtering; should match all integrations)
+            params['entity_ids'] = None
+            assert_matched_integrations(params, set.union(*[set(iids) for eids, iids in test_cases]))
```
danielsotirhos: Creative approach, however I would suggest keeping the test structure as it was simply by adding the new test cases (entity_id is empty string and entity_id not specified) to the `test_cases` list. (note: this does require a slight change to how `params` is constructed in order to support the `None` case)

Also, setting `entity_ids` to `None` might technically be the same in the current implementation as omitting `entity_ids` from the query, however it is better to have the test case reflect the intended situation as close to as possible.
```suggestion
        test_cases = [
            # One project entity id specified by one integration
            (['cddab57b-6868-4be4-806f-395ed9dd635a'],
             ['f13ddf2d-d913-492b-9ea8-2de4b1881c26']),
            # Two project entity ids specified by two different integrations
            (['cddab57b-6868-4be4-806f-395ed9dd635a', '90bd6933-40c0-48d4-8d76-778c103bf545'],
             ['f13ddf2d-d913-492b-9ea8-2de4b1881c26', 'dbfe9394-a326-4574-9632-fbadb51a7b1a']),
            # One project entity id specified by two different integrations
            (['c4077b3c-5c98-4d26-a614-246d12c2e5d7'],
             ['977854a0-2eea-4fec-9459-d4807fe79f0c', 'e8b3ca4f-bcf5-42eb-b58c-de6d7e0fe138']),
            # No entity ids (empty string) should return no matching integrations
            ([''], []),
            # No entity ids (not specified) should return all integrations
            (None,
             ['f13ddf2d-d913-492b-9ea8-2de4b1881c26', 'dbfe9394-a326-4574-9632-fbadb51a7b1a',
              '977854a0-2eea-4fec-9459-d4807fe79f0c', 'e8b3ca4f-bcf5-42eb-b58c-de6d7e0fe138'])
        ]

        portal_integrations_db.return_value = self._portal_integrations_db
        with mock.patch.object(type(config), 'dss_deployment_stage', 'prod'):
            for entity_ids, integration_ids in test_cases:
                params = dict(integration_type='get', entity_type='project')
                if entity_ids is not None:
                    params['entity_ids'] = ','.join(entity_ids)
                with self.subTest(**params):
                    response_json = self._get_integrations(params)
                    found_integration_ids = self._extract_integration_ids(response_json)
                    self.assertEqual(set(integration_ids), set(found_integration_ids))
```

lambdas/service/app.py
```diff
@@ -406,11 +406,17 @@ def get_integrations():
     try:
         entity_ids = query_params['entity_ids']
     except KeyError:
+        # Case where parameter is absent (any id okay)
```
danielsotirhos: Unsure what is meant here by "any id okay". Could you elaborate or remove if not needed?

noah-aviel-dove: Would these updated comments be sufficiently clear?

409        # Case where parameter is absent (do not filter by id)
413        # Case where parameter is present and has a value (filter for matching id value)
416        # Case where parameter is present but lacks a value (filter for missing entity_id field, i.e., there are no acceptable id values)

danielsotirhos: These comment messages look good, thanks!

test/service/test_response.py
noah-aviel-dove: I agree, yours is the better structured approach.

test/service/test_response.py
```diff
@@ -1500,12 +1501,22 @@ def test_integrations_by_entity_ids(self, portal_integrations_db):
              ['f13ddf2d-d913-492b-9ea8-2de4b1881c26', 'dbfe9394-a326-4574-9632-fbadb51a7b1a']),
             # One project entity id specified by two different integrations
             (['c4077b3c-5c98-4d26-a614-246d12c2e5d7'],
-             ['977854a0-2eea-4fec-9459-d4807fe79f0c', 'e8b3ca4f-bcf5-42eb-b58c-de6d7e0fe138'])
+             ['977854a0-2eea-4fec-9459-d4807fe79f0c', 'e8b3ca4f-bcf5-42eb-b58c-de6d7e0fe138']),
+            # Blank entity id, to match integrations lacking the entity_id field (in this case none of them)
```
hannes-ucsc: Please add such an integration for this case to be more meaningful.

hannes-ucsc: Add it to the mock database that is.

noah-aviel-dove: The current implementation (in my branch and in [develop](https://github.com/DataBiosphere/azul/blob/develop/lambdas/service/app.py#L421)) is to include integrations lacking the `entity_ids` field even when a non-null `entity_ids` query parameter is provided. I noticed this when the unit tests failed due to the new entry (missing the field) being expected in the results of every other subtest. I just wanted to confirm whether this is still the desired behavior, i.e., the new entry should be added to the other tests.

hannes-ucsc:
Don't forget to request review AND assign. It's easier for me to see at a glance what PRs need my attention. I will notice if just a review is requested but usually only when I go through my email notifications from Github.

Issue #1410, opened 2019-10-30
Manifest integration tests increases Gitlab build times


hannes-ucsc:
Since this build https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/pipelines/542 build times went up from ~26 min to 35-45 min. That build was for #1086, PR #1306. 

The build that lands the PR on `develop` is https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/pipelines/543 and it fails for an unrelated reason. The first successful `develop` after that is https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/pipelines/545 and is longer as well. All subsequent builds are > 35min.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-863)
┆Project Name: azul
┆Issue Number: AZUL-863

PR #1419, opened 2019-11-04
Parallelize delete notifications for integration test (#1410)

achave11:


hannes-ucsc:
There is a problem, it appears as if I authored both commits which is not the case. I recently switched to a new Git UI which may be causing this. I will fix by force-pushing develop.

hannes-ucsc:
Fixed

Issue #1412, opened 2019-10-30
Integration and staging use two shards instead of four


hannes-ucsc:
They share the ES domain with dev which has four nodes but their `environment` specifies two nodes. When a deployment shares the ES domain with another deployment, its `AZUL_ES_INSTANCE_COUNT` should match that of the shared deployment.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-864)
┆Project Name: azul
┆Issue Number: AZUL-864

hannes-ucsc:
Hannes still owes demo. Was ill-prepared today.

hannes-ucsc:
The fix is currently deployed to integration but not on staging yet. The screenshot below shows that in `integration` the instance count is configured via `environment` to 2 but the lambda uses 4 since it's sharing the ES domain with `dev` which uses 4.

![image](https://user-images.githubusercontent.com/5143256/69195955-f60a2200-0ae1-11ea-9b02-aebb9222872d.png)

The screenshot below shows that on `staging` both values are 2.

![image](https://user-images.githubusercontent.com/5143256/69196171-906a6580-0ae2-11ea-9688-67e634251793.png)

The number of aggregation shards is derived from the ES instance count. The screenshot below shows that `integration` uses the correct number of shards (4) matching the actual instance count on the ES cluster and overriding the value from `environment`. In contrast, `staging` still uses 2 shards as defined in `environment` even though it is running on the same cluster with 4 nodes.

![image](https://user-images.githubusercontent.com/5143256/69196399-3d44e280-0ae3-11ea-831f-1c449f7910d8.png)

PR #1420, opened 2019-11-04
[R] Ensure that sharing an ES domain uses correct number of shards (#1412)

achave11:


achave11:
@hannes-ucsc I would instead like to suggest the following approach. I find it a bit complicated having to patch boto at set up time. This is required for the docker ES instance to be created in order to begin testing.
```
Index: src/azul/__init__.py
IDEA additional info:
<+>UTF-8
--- src/azul/__init__.py	(date 1573576695000)
+++ src/azul/__init__.py	(date 1573584189582)
@@ -295,10 +295,24 @@
     @property
     def es_instance_count(self) -> int:
         if self.share_es_domain:
-            raise NotImplementedError('Please use `deployment.aws.es_instance_count` instead')
+            _, deployment_stage = self.unqualified_resource_name(self.es_domain)
+            instance_count = self._env_variable_from_shared_deployment('AZUL_ES_INSTANCE_COUNT', deployment_stage)
+            variable, count = instance_count.split()
+            if 'AZUL_ES_INSTANCE_COUNT' == variable:
+                return int(count)
+            else:
+                raise NotImplementedError('Please use `deployment.aws.es_instance_count` instead')
         else:
             return int(os.environ['AZUL_ES_INSTANCE_COUNT'])
 
+    def _env_variable_from_shared_deployment(self, env_variable: str, deployment_stage: str) -> str:
+        with open(f'{self.project_root}/deployments/{deployment_stage}/environment') as f:
+            for line in f:
+                if line.startswith(f'_set {env_variable}'):
+                    return line.replace('_set ', '', 1).strip()
+            f.close()
+        return None
+
     @property
     def es_volume_size(self) -> int:
         return int(os.environ['AZUL_ES_VOLUME_SIZE'])

```

hannes-ucsc:
That's making too many assumptions about the syntax and location of the config files. The Config class should only knows about the environment variables, not how they are set.

Issue #1383, opened 2019-10-15
Move portal integration DB to S3


danielsotirhos:
Move the hard-coded portal integrations DB created in https://github.com/DataBiosphere/azul/issues/1243 to a AWS S3 bucket.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-844)
┆Project Name: azul
┆Issue Number: AZUL-844
┆Epic: Generic Tertiary Portal Handoff

noah-aviel-dove:
I would benefit from some more specific instructions. From reading the commit history of #1243, it would seem that your're referring to [this JSON](https://github.com/DataBiosphere/azul/commit/9b12986bf0915dfa0b12b5f5b4bc3c562f3838a6#diff-48eef37b4334bc720ffa086b444fd6a1). 

Does it go into an existing bucket or a new one? If new, are there naming or organization conventions? 

I searched for 'azul' in the HCA S3 buckets and it's not obvious that there's a correct implementation.

hannes-ucsc:
Good questions. Yes, the JSON structure you link to is the portal database. 

In S3, it should go into the "config" bucket, the same place where the Terraform state is kept (the bucket name is controlled by the `azul.config.terraform_backend_bucket` property. The key of the object should be `f"azul/{azul.config.deployment_stage}/portals/db.json"`.

To ease the roll-out of this feature, the hard-coded DB should remain in place in the source tree and should be used to create the object if it doesn't exist. This roll-out behaviour should be covered by the tests. Note that we use `moto` to mock S3 in unit tests. Once the feature is rolled-out to prod we'll remove the hard-coded version.

The "config" bucket has versioning enabled, so accidental deletions/overwrites of the object are not permanent and can be reverted.

noah-aviel-dove:
Should the moto rollout test be added to `TestPortalIntegrationResponse`, `TestStorageService`, or somewhere else?

hannes-ucsc:
The former, I think.

noah-aviel-dove:
Why was this closed? I never even submitted a pull request.

hannes-ucsc:
> Why was this closed? I never even submitted a pull request.

I must have closed this by accident when I attempted to use Github keyboard shortcuts.

noah-aviel-dove:
Add integration test for concurrency

N threads simultaneously adding entries to portal DB, all entries must be present.
Only runs in sandbox deployment stage

Issue #1414, opened 2019-10-31
Log exceptions in Chalice apps even if AZUL_DEBUG is 0


hannes-ucsc:
https://github.com/hannes-ucsc/chalice/blob/02f0c58766e12f711438b4c7507a9dedd2effe06/chalice/app.py#L738

Chalice either logs the exception traceback and returns it in the body of the 500 response (if `app.debug` is True) or it does neither (if `app.debug` is False). This is, of course, silly, since there is no way to track the cause for a 500 error in the latter case. We want to always log the exception as WARNING and only make the presence of the backtrace in the response conditional on `app.debug` and therefore `AZUL_DEBUG`.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-866)
┆Project Name: azul
┆Issue Number: AZUL-866

hannes-ucsc:
Upstream Chalice fixed this:

https://github.com/aws/chalice/commit/1b21e7a992186cf384c07e3768039cff3e7cdf65

Issue #1399, opened 2019-10-29
Portals Registration CRUD APIs


theathorn:
PUT, POST, DELETE.

PR #1428, opened 2019-11-06
Add code to attempt retrieving portal DB from S3 (#1383)

noah-aviel-dove:


src/azul/__init__.py
```diff
@@ -551,6 +551,14 @@ def github_project(self) -> str:
     def github_access_token(self) -> str:
         return os.environ['azul_github_access_token']
 
+    def portal_integrations_db_params(self):
+        """
+        :return: Bucket name, object key, character encoding
+        """
+        bucket_name = self.terraform_backend_bucket
+        object_key = f'azul/{self.deployment_stage}/portals/db.json'
```
danielsotirhos: `deployment_stage` comes from `AZUL_DEPLOYMENT_STAGE` (i.e. where it was deployed)
`dss_deployment_stage` comes from `AZUL_DSS_ENDPOINT` (i.e. which DSS is being used)
Consider dev's environment:
```
_set AZUL_DEPLOYMENT_STAGE dev
_set AZUL_DSS_ENDPOINT "https://dss.integration.data.humancellatlas.org/v1"
```
Because a deployment can use a diff deployment's DSS, and that the portals db data is DSS specific, I believe we should be using `dss_deployment_stage` here
```suggestion
        object_key = f'azul/{self.dss_deployment_stage}/portals/db.json'
```

hannes-ucsc:
@danielsotirhos brings up good points that I'd like to discuss after scrum.

src/azul/project/hca/__init__.py
```diff
@@ -6,12 +6,17 @@
...
+from azul.portal_service import PortalService
 from azul.project.hca.indexer import Indexer
 from azul.types import JSON
 
 
 class Plugin(azul.plugin.Plugin):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._staged_portal_integrations_db = PortalService.stage_db(Plugin._unstaged_portal_integrations_db())
```
danielsotirhos: It seems a bit excessive to be creating the "staged" portal db every time the Plugin class is loaded, especially since we only expect this hardcoded db to be used once before the db is saved to s3.

Instead of pre-setting up the default db so it can be passed into `get_portal_integrations_db()`, what if the default db was created in `get_portal_integrations_db()` (by calling `PortalService.stage_db()`) when it determined the db on s3 didn't exist?

src/azul/portal_service.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+
+        s3 = boto3.resource('s3')
+        bucket_name, object_key, charset = config.portal_integrations_db_params()
+        obj = s3.Object(bucket_name, object_key)
+        try:
+            retrieval = obj.get()
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'NoSuchKey' and default_db is not None:
+                db = default_db
```
danielsotirhos:
```suggestion
                db = PortalService.stage_db(Plugin._unstaged_portal_integrations_db())
```

noah-aviel-dove: I think there a couple reasons to avoid referencing `azul.project.hca.Plugin` in the `PortalService` class. 
First, it prevents `PortalService` from being used by any other classes. On a current practical note, this makes unit testing more tedious.

src/azul/project/hca/__init__.py
noah-aviel-dove: I would propose that we instead keep the design of `PoralService` the same and change `azul.project.hca.Plugin._staged_portal_integrations_db` to a class attribute that is only initialized once.

danielsotirhos: Ok, yes that sounds better.

src/azul/portal_service.py
```diff
@@ -0,0 +1,64 @@
...
+
+from azul import config
+from azul.types import JSON
+
+log = logging.getLogger(__name__)
+
+
+class PortalService:
+
+    @staticmethod
```
hannes-ucsc: This should be mentioned in `CONTRIBUTING.rst`: we don't use `@staticmethod`, only `@classmethod` even if the first argument (`cls`) of such a method is not used by its body. Whether `cls` is used is often incidental and an implementation detail. We wouldn't want to repeatedly switch from `@staticmethod` to `@classmethod` and back if that implementation detail changes. We'll simply declare all methods that should be invoked through the class (as opposed to through an instance of that class) as `@classmethod` and call it a day. The same consideration goes for instance methods and `self`: some use it, some don't. The ones that don't shouldn't suddenly be considered static methods. The distinction between instance and class methods is driven by higher order concerns than the one about whether a method's body currently references `self` or not.

Could you add a section to `CONTRIBUTING.rst` in a separate commit in this PR? Feel free to use some of the above text for that. It's the best and most in-depth explanation of this rule I've given so far.

hannes-ucsc: Aside from that: these methods should be instance methods and `app.py` should instantiate this class before calling them. This is a pattern that we should follow for all services and controllers.

src/azul/portal_service.py
hannes-ucsc: I think I agree with @noah-aviel-dove here.

src/azul/portal_service.py
```diff
@@ -0,0 +1,64 @@
...
+            json_bytes = retrieval['Body'].read()
+            db = json.loads(json_bytes.decode(charset))
+        return db
+
+    @staticmethod
+    def stage_db(unstaged_db: Sequence[JSON]) -> Sequence[JSON]:
+        """
+        Transform portal integrations database to only contain entity_ids from the current DSS deployment stage.
+        Original is not modified.
+        :param unstaged_db: portal DB where the `entity_ids` fields  are dictionaries whose keys correspond to DSS
```
hannes-ucsc:
```suggestion

        :param unstaged_db: portal DB where the `entity_ids` fields  are dictionaries whose keys correspond to DSS
```

src/azul/__init__.py
```diff
@@ -562,6 +562,14 @@ def github_project(self) -> str:
     def github_access_token(self) -> str:
         return os.environ['azul_github_access_token']
 
+    def portal_integrations_db_params(self):
```
hannes-ucsc: I'd split this up into two methods and make them properties.

I'd remove the encoding setting. I don't see why anyone would use anything but UTF-8. Thankfully, the times when we had to consider other encodings are behind us.

src/azul/portal_service.py
noah-aviel-dove: Sounds good for the most part, but `PortalService` currently is not used anywhere in `app.py`. Should an instance be created there that is accessed for use by other classes, or is it okay to create the instance in the `azul.project.hca.Plugin` class where it's actually used?

noah-aviel-dove:
Comments from sprint demo: integrations without `entity_ids` field don't really belong in the database and really a byproduct of how multiple stages are nested in the hardcoded db. They should be removed when the db is staged by `PortalService` instead of forwarding the empty array. For example an integration with `"entity_ids": {"staging": [], "dev": ..., "prod": ...}` should not be included in staging.

hannes-ucsc:
Given the above review summary, I've bumped the estimate on the associated ticket to 40.

src/azul/portal_service.py
```diff
@@ -0,0 +1,73 @@
...
+
+        s3 = boto3.resource('s3')
+        obj = s3.Object(config.portal_integrations_db_bucket, config.portal_integrations_db_object)
+        try:
+            retrieval = obj.get()
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'NoSuchKey' and default_db is not None:
+                db = default_db
+                log.info('Portal integration DB not found in S3; uploading hardcoded DB.')
+                json_bytes = bytes(json.dumps(default_db), 'UTF-8')
```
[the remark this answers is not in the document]
noah-aviel-dove:
```
>>> bytes('fish')
TypeError: string argument without an encoding
```

src/azul/portal_service.py
```diff
@@ -0,0 +1,73 @@
...
+                if 'entity_ids' not in integration:
+                    yield integration
+                else:
+                    staged_entity_ids = integration['entity_ids'][config.dss_deployment_stage]
+                    if staged_entity_ids:
+                        integration['entity_ids'] = staged_entity_ids
+                        yield integration
+
+        for portal in db:
+            portal['integrations'] = list(transform(portal['integrations']))
```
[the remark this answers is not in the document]
danielsotirhos: To match original behavior portals should only be returned by the `/integrations` endpoint if they have 1 or more integrations matching the query params `entity_type`, `integration_type`, and (optionally) `entity_ids`.

src/azul/portal_service.py
```diff
@@ -0,0 +1,73 @@
...
+        Original is not modified.
+
+        :param unstaged_db: portal DB where the `entity_ids` fields  are dictionaries whose keys correspond to DSS
+        deployment stages.
+        :return: deep copy of that DB where the `entity_ids` fields have been replaced by the entry associated with the
+        current DSS deployment stage. If the `entity_ids` field is present but no entity ids are specified for the
+        current deployment stages, the integration is removed. Portals, however, are not removed even if they have no
+        remaining associated integrations.
+        """
+        db = deepcopy(unstaged_db)
```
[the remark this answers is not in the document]
noah-aviel-dove: I agree it's odd. I'm continuing to think about it.

src/azul/portal_service.py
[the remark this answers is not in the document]
noah-aviel-dove: The endpoint behavior is not an issue. Obviously, a portal with no integrations has no integrations matching the query parameters, so it would never included in the results. That remains unchanged in this PR (see [the updated integrations endpoint](https://github.com/DataBiosphere/azul/pull/1428/files#diff-a3dec4f6b35a08e7cb29086cb8766c3a)).

The question is whether portals with no integrations defined for the current deployment stage should be included in the database at all, following the deployment-specific flattening of the entity ids. Since we've established that they cannot contribute to the integrations endpoint, it would seem they should not?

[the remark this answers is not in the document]
danielsotirhos: Ah, thanks for the clarification. When a new portal registers with the Portal Integrations AP, it first will post to `/portal` to register the portal details, then will post to `/portal/{portal_id}/integrations` to register the integrations, meaning before the portal's registers the first integration the database will only have the portal details with no integrations.

So yes, since portals with no integrations need to be allowed to exist in the deployment-specific database, therefore we might as well include those from the flattened default_db.

src/azul/portal_service.py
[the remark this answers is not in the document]
noah-aviel-dove: I have implemented and tested a version of this method that delays copying the integrations until it has determined that they belong in the result. It's not as hideous as I had worried, but I'm also not convinced the increase in efficiency justifies the increase in complexity. I am including it in a separate (semantically neutral) commit for your review.

src/azul/portal_service.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+
+        client = boto3.client('s3')
+        try:
+            retrieval = client.get_object(Bucket=config.portal_integrations_db_bucket,
+                                          Key=config.portal_integrations_db_object)
+        except client.exceptions.NoSuchKey:
+            plugin = Plugin.load()
+            db = self.stage_db(plugin.portal_integrations_db())
+            log.info('Portal integration DB not found in S3; using and uploading hardcoded DB from plugin.')
```
hannes-ucsc:
```suggestion
            log.info('Portal integration DB not found; staging hard-coded DB.')
```

Log messages should be concise and don't need to convey every implementation detail.

src/azul/portal_service.py
```diff
@@ -0,0 +1,82 @@
...
+            json_bytes = bytes(json.dumps(db), 'UTF-8')
```
hannes-ucsc:
```suggestion
            json_bytes = json.dumps(db).encode()
```

I don't know why the bytes constructor requires the encoding argument.

hannes-ucsc:
```python
with BytesIO() as b:
    with TextIOWrapper(b) as w:
        json.dump({'ü':'ber'}, w)
        w.flush()
        print(repr(b.getvalue()))
    
b'{"\\u00fc": "ber"}'
```

Should use less space since the unencoded string is never materialized. However, I'm not advocating that approach for this PR.

src/azul/portal_service.py
```diff
@@ -0,0 +1,82 @@
...
+            client.put_object(Bucket=config.portal_integrations_db_bucket,
+                              Key=config.portal_integrations_db_object,
+                              Body=json_bytes,
+                              ContentType='application/json')
+        except ClientError as e:
```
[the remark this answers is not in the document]
hannes-ucsc: Also, we prefer using the `exc_info` argument to log methods. You should try it to see what the difference is.

src/azul/portal_service.py
```diff
@@ -0,0 +1,82 @@
...
+            log.error(f'Unexpected error retrieving portal DB from S3: {e}')
+            raise
+        else:
+            log.info("Retrieved portal integration DB from S3")
+            json_bytes = retrieval['Body'].read()
+            db = json.loads(json_bytes.decode('UTF-8'))
+        return db
+
+    def stage_db(self, unstaged_db: Sequence[JSON]) -> Sequence[JSON]:
```
hannes-ucsc:
```suggestion
    def convert_hardcoded_db(self, db: Sequence[JSON]) -> Sequence[JSON]:
```

The name `stage_db` sounds like this method does the staging.

src/azul/portal_service.py
```diff
@@ -0,0 +1,82 @@
...
+
+        :return: deep copy of that DB where the `entity_ids` fields have been
+        replaced by the entry associated with the current DSS deployment stage.
+        If the `entity_ids` field is present but no entity ids are specified for
+        the current deployment stages, the integration is removed. Portals,
+        however, are not removed even if they have no remaining associated
+        integrations.
+        """
+
+        db = deepcopy(unstaged_db)
```
hannes-ucsc:
```suggestion
        db = deepcopy(db)
```

I like this because it signals that you are done with the original argument value from this point forward. It also means that you don't need to invent a discriminating prefix or suffix, like `staged` or `unstaged` in order to distinguish between variable.

src/azul/portal_service.py
```diff
@@ -0,0 +1,82 @@
...
+        Retrieve portal DB from S3.
+
+        If no object is present at the location specified in azul.config, then
+        a default hardcoded DB is obtained from the current plugin, uploaded to
+        S3, and returned.
+        """
+
+        client = boto3.client('s3')
+        try:
+            retrieval = client.get_object(Bucket=config.portal_integrations_db_bucket,
```
hannes-ucsc:
```suggestion
            response = client.get_object(Bucket=config.portal_integrations_db_bucket,
```

is what we typically call this.

src/azul/portal_service.py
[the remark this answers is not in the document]
noah-aviel-dove: I added this to assist in debugging the service response unit tests, which at one point were failing with no diagnostic information besides a 500 internal server error. But I can see how this usefulness doesn't translate into deployment.

src/azul/portal_service.py
noah-aviel-dove: It would be easy to to add that as a utility function in `strings.py` since you've already written it.

src/azul/portal_service.py
noah-aviel-dove: By "unstaged" I mean "including nested entity ids for multiple deployment stages" and by "staged" I mean "including entity ids only for the current deployment stage." Are we on the same page in that regard?

hannes-ucsc: Ohh, I see. That's a rather confusing interpretation of the word _staged_ and _unstage_. 

In fact, I now regret the use of `stage` as a noun in identfiers like `dss_deployment_stage` because it leads to beauties like "the staging stage". But at the time of Azul's inception I borrowed it from the DSS team's terminology. 

I think we should stay away from "stage" alltogether. Open to suggestions. "demultiplex" and "multiplex"?

src/azul/portal_service.py
hannes-ucsc: More like `json.py` but we would first need make reasonably sure that it is indeed an space improvement and has no serious drawbacks.

src/azul/portal_service.py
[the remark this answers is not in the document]
hannes-ucsc: > which at one point were failing with no diagnostic information besides a 500 internal server error

Can you reproduce this? It may indicate that something is wrong with test logging.

src/azul/portal_service.py
noah-aviel-dove: Fair. I'll leave it out for now.

src/azul/portal_service.py
noah-aviel-dove: "[de]multiplex" is satisfactory. I can't think of anything I would prefer.

src/azul/portal_service.py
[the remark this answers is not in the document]
noah-aviel-dove: Steps to reproduce:

1. Checkout `develop` or `issues/noah-aviel-dove/1383-move-portal-integration-db`
2. Add the line `raise ValueError("Can't see me!")` at the top of the `_fetch_integrations` function in app.py.
3. Right-click on test/service/test_response.py and run unit tests
4. Expected output for the first subtest (entity_type='file', integration_type='get_manifest'):

```
SubTest error: Traceback (most recent call last):
  File "/usr/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib/python3.6/unittest/case.py", line 523, in subTest
    yield
  File "/home/nadove/PycharmProjects/azul/test/service/test_response.py", line 1554, in test_integrations
  File "/home/nadove/PycharmProjects/azul/test/service/test_response.py", line 1518, in _get_integrations
    response.raise_for_status()
  File "/home/nadove/PycharmProjects/azul/.venv/lib/python3.6/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://127.0.0.1:43109/integrations?integration_type=get_manifest&entity_type=file
```

5. To programatically demonstrate that the underlying exception does not appear anywhere in the output:
6. `cd azul/test/service`
7. `python -m unittest test_response.py 2>&1 | fgrep ValueError`
8. Expected output: none.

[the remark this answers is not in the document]
hannes-ucsc: I think I know what's going on: https://github.com/DataBiosphere/azul/issues/1414

That actually contradicts what I said about Chalice handling and logging errors: it only does that when self.debug is set and that's currently tied to AZUL_DEBUG > 0.

As a workaround, when I set AZUL_DEBUG to 1.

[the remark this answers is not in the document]
hannes-ucsc: I'll see if I can quickly solve #1414.

src/azul/portal_service.py
```diff
@@ -0,0 +1,80 @@
...
+                else:
+                    yield deepcopy(integration)
+
+        def transform_portal(portal):
+            return {
+                k: deepcopy(v) if k != 'integrations' else list(transform_integrations(v))
+                for k, v in portal.items()
+            }
+
+        return [transform_portal(portal) for portal in db]
```
hannes-ucsc:
```suggestion
        return list(map(transform_portal, db))
```

src/azul/portal_service.py
```diff
@@ -0,0 +1,80 @@
...
+        replaced by the entry associated with the current DSS deployment stage.
+        If the `entity_ids` field is present but no entity ids are specified for
+        the current deployment stages, the integration is removed. Portals,
+        however, are not removed even if they have no remaining associated
+        integrations.
+        """
+
+        def transform_integrations(integrations):
+            for integration in integrations:
+                if 'entity_ids' in integration:
```
hannes-ucsc: This …

src/azul/portal_service.py
```diff
@@ -0,0 +1,80 @@
...
+                    staged_entity_ids = integration['entity_ids'][config.dss_deployment_stage]
```
hannes-ucsc: ... and `integration['entity_ids']` here are two lookups that could be combined with `try: … except KeyError: …`

src/azul/portal_service.py
hannes-ucsc: But overall I like this better.

src/azul/portal_service.py
noah-aviel-dove: Very well. Does implementing the pseudocode in #1399 fall under the scope of this PR? As in, that's what I work on next?

hannes-ucsc: Yes.

src/azul/portal_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _create_db(self) -> None:
+        """
+        Write hardcoded portal integrations DB to S3 if no other DB is present.
+        """
+        plugin = Plugin.load()
+        db = self.demultiplex(plugin.portal_integrations_db())
+        init_version = self._write_db(db)
+        self._update_version(None, init_version)
+        # Ignore failure, just means that someone else created it first
```
hannes-ucsc: This race should not be ignored and a retry should be triggered.

src/azul/version_service.py
```diff
@@ -0,0 +1,71 @@
...
+            item = response['Item']
+        except KeyError:
+            return None
+        else:
+            return item[VersionService.value_name]['S']
+
+    def put(self, object_url: str, version: Optional[str], new_version: str) -> None:
+        """
+        Update object's current version from `version` to `new_version`, or fail
+        if `version` does not match the stored current version.
```
hannes-ucsc: Document what `version=None` means.

src/azul/version_service.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        condition_params = {
+            'ConditionExpression': f'attribute_not_exists({VersionService.key_name})'
+        } if version is None else {
+            'ConditionExpression': f'{VersionService.value_name} = :v',
```
hannes-ucsc: Ditto

src/azul/portal_service.py
```diff
@@ -0,0 +1,167 @@
...
+        missing version.
+        """
+        try:
+            response = self.client.get_object(Bucket=config.portal_integrations_db_bucket,
+                                              Key=config.portal_integrations_db_object,
+                                              VersionId=version)
+        except self.client.exceptions.NoSuckKey:
+            # boto raises the same exception when the key is not found and when
+            # the key IS found but the version is not, so we search the object
+            # versions to disambiguate these cases.
```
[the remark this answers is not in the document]
noah-aviel-dove: Seems I was too literal in interpreting the pseudocode. Good thing too since this was potentially a very inefficient operation.

src/azul/portal_service.py
```diff
@@ -0,0 +1,167 @@
...
+            all_versions = self.client.list_object_versions()['Versions']
+            if any(version == v['VersionId'] for v in all_versions):
+                raise
+            else:
+                raise NoSuchObjectVersion(version) from None
```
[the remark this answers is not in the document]
noah-aviel-dove: It was to prevent the `NoSuchKey` exception from the surrounding `except` block from appearing at the top of the stacktrace of the `NoSuchObjectVersion` exception for clarity. No longer relevant given the previous comment.

src/azul/portal_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+        json_bytes = response['Body'].read()
+        return json.loads(json_bytes.decode('UTF-8'))
```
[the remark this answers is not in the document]
noah-aviel-dove: Not sure how that got in. Must have been copied from somewhere we missed last time.

src/azul/portal_service.py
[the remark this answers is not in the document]
hannes-ucsc: Generally speaking, we'd want the primary exception to be included in the trace. Suppressing information does not add clarity, except when the suppressed information would is distracting (due to volume) or misleading.

src/azul/portal_service.py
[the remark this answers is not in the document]
hannes-ucsc: It could be that the pseudocode is wrong but I don't see where it specifies that we should distinguish the two cases.

src/azul/portal_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+        :param operation: Callable that accepts the latest version of the portal
+        DB and optionally returns an updated version.
+        """
+        while True:
+            version = self.version_service.get(self._db_url())
+            if version is None:
+                log.info('Portal integration DB not found in S3; uploading hard-coded DB.')
+                self._create_db()
+                continue
```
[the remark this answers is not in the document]
noah-aviel-dove: There is no "retry" for the create operation, which currently only happens when there's a request to operate on the DB but no DB exists (i.e. it is never explicitly requested via the API). If no DB exists at the time the operation is requested, but is created by another process before our process creates it, there is no need to re-create it (and trying to do so might actually erase data if a write operations sneaks in before our process manages to catch up).

The continuation is to ensure that we have the correct version before performing the requested operation. The `version` local is `None` before `_create_db` is called and must be updated to the current/initial version after the DB is created before `operation` can be performed. But we cannot correctly obtain the current version ID by returning it from `_create_db` because  while the DB will exist after this call it may have been created by another process. So we `continue` to re-read the current version at the top of the loop.

src/azul/portal_service.py
noah-aviel-dove: https://github.com/DataBiosphere/azul/pull/1428#discussion_r348674257

Let's keep this in one thread

src/azul/portal_service.py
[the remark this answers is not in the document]
noah-aviel-dove: Catching `NoSuchObjectVersion` instead of `NoSuchKey` implies the need for a separate exception to be raised in one case (but not the other).

src/azul/portal_service.py
```diff
@@ -0,0 +1,154 @@
...
+        Retrieve specified version of portal DB from S3.
+        Raises `NoSuchObjectVersion` if the version is not found.
+        """
+        try:
+            response = self.client.get_object(Bucket=config.portal_db_bucket_name,
+                                              Key=config.portal_db_object_key,
+                                              VersionId=version)
+        except self.client.exceptions.NoSuchKey:
+            raise NoSuchObjectVersion(version)
+        json_bytes = response['Body'].read()
```
hannes-ucsc: This could go into `else` branch of the try for clarity.

src/azul/portal_service.py
```diff
@@ -0,0 +1,154 @@
...
+                                          ContentType='application/json')
+        new_version = response['VersionId']
+        try:
+            self.version_service.put(self._db_url, version, new_version)
+        except VersionConflict:
+            # Operation was performed on outdated DB and now erroneously exists
+            # in S3.
+            self._delete_db(new_version)
+            raise
+        return new_version
```
hannes-ucsc: This could go into `else` branch of the try for clarity.

src/azul/portal_service.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def _delete_db(self, version: str) -> None:
+        """
+        Delete the specified version of the portal integrations DB from S3.
+        """
+        self.client.delete_object(Bucket=config.portal_db_bucket_name,
```
hannes-ucsc: I have seen it happen that this raises NoSuchKey, too. We could either retry in a small loop here or catch, log and ignore.

src/azul/version_service.py
```diff
@@ -0,0 +1,74 @@
...
+        except KeyError:
+            return None
+        else:
+            return item[self.value_name]['S']
+
+    def put(self, object_url: str, version: Optional[str], new_version: str) -> None:
+        """
+        Update object's current version from `version` to `new_version`, or fail
+        if `version` does not match the stored current version.
+        Provide `None` for the current version to began tracking a previously
```
hannes-ucsc:
```suggestion
        Provide `None` for the current version to begin tracking a previously
```

test/service/test_portal_service.py
```diff
@@ -0,0 +1,169 @@
...
+)
+
+from azul import config
+from azul.logging import configure_test_logging
+from azul.plugin import Plugin
+from azul.portal_service import PortalService
+from azul.types import JSONs
+from azul.version_service import NoSuchObjectVersion
+from azul_test_case import AzulTestCase
+from service.test_version_service import create_version_table
```
hannes-ucsc: We typically share fixtures or other test functionality via inheritance. I would create a base class that both TestPortalService and TestVersionService inherit and put all the shared functionality there. This would also allow you to share the boto3 DynamoDB client object. I would call that class `VersionTableTestCase`.

src/azul/portal_service.py
noah-aviel-dove: You raise a good point. The only purpose of the `_delete_db` method is to resolve version conflicts in `_write_db`, but it could be that the object to be written won't actually uploaded by the time the version conflict is detected and `_delete_db` is called.

As it stands, the call to `_delete_db` isn't necessary per se, since the copy we delete never has its version ID set as current and doesn't affect further operations. Perhaps we need a stronger strategy in general for removing unneeded versions. For now I think I'll just log and ignore.

hannes-ucsc: OK.

test/service/test_version_service.py
```diff
@@ -0,0 +1,42 @@
...
+
+from azul.logging import configure_test_logging
+from azul.version_service import (
+    VersionService,
+    VersionConflict,
+)
+from version_table_test_case import VersionTableTestCase
+
+
+def setUpModule():
```
hannes-ucsc: This should have a pragma suppressing the warning like so:

https://github.com/DataBiosphere/azul/commit/bccc31d29dfe9da710e0d1c0e4b03082de42886c

test/service/test_portal_service.py
```diff
@@ -0,0 +1,159 @@
...
+from azul import config
+from azul.logging import configure_test_logging
+from azul.plugin import Plugin
+from azul.portal_service import PortalService
+from azul.types import JSONs
+from azul.version_service import NoSuchObjectVersion
+from version_table_test_case import VersionTableTestCase
+
+
+def setUpModule():
```
hannes-ucsc: ditto

lambdas/service/app.py
```diff
@@ -456,21 +457,26 @@ def _fetch_integrations(entity_type: str, integration_type: str, entity_ids: Opt
...
-            if integration['entity_type'] == entity_type and integration['integration_type'] == integration_type
-        ]
-        if entity_ids is not None:
-            integrations = [integration for integration in integrations
-                            if 'entity_ids' not in integration or entity_ids.intersection(integration['entity_ids'])]
-        if len(integrations) > 0:
-            portal = {k: v if k != 'integrations' else integrations for k, v in portal.items()}
-            results.append(portal)
+
+    def list_integrations(portal_db):
```
hannes-ucsc: In the spirit of slimming down `app.py` and moving functionality into controllers and services, this functionality should probably live in the service as well. The handler function here would then essentially look like `return PortalService().list_integrations()` and the `.crud()` method could become protected (`_crud`).

test/integration_test.py
```diff
@@ -404,6 +410,78 @@ def _get_entities_by_project(self, entity_type, project_short_name):
             params['search_after_uid'] = pagination['search_after_uid']
         return entities
 
+    def test_concurrent_portal_db_crud(self):
+        """
+        Use multithreading to simulate multiple users simultaneously modifying
+        the portals database.
+        """
+        if config.deployment_stage not in ('sandbox', 'nadove'):
```
hannes-ucsc: I'd use `@unittest.skipIf` or `.skipUnless`

test/integration_test.py
```diff
@@ -404,6 +410,78 @@ def _get_entities_by_project(self, entity_type, project_short_name):
...
+                    self.portal_service._crud(lambda db: add_entry(db, self.name, i))
+
+        portal_service._crud(get_db)
+        init_db = cast(JSONs, portal_db)
+
+        threads = [TestThread(name=f'T{i}') for i in range(n_threads)]
+        for thread in threads:
+            thread.start()
+
+        while any(thread.is_alive() for thread in threads):
```
hannes-ucsc: Traditionally, one would use `Thread.join()` which is blocking and avoids the busy waiting.

test/integration_test.py
```diff
@@ -404,6 +410,78 @@ def _get_entities_by_project(self, entity_type, project_short_name):
...
+                self.portal_service = PortalService()
+
+            def run(self):
+                for i in range(n_ops):
+                    self.portal_service._crud(lambda db: add_entry(db, self.name, i))
+
+        portal_service._crud(get_db)
+        init_db = cast(JSONs, portal_db)
+
+        threads = [TestThread(name=f'T{i}') for i in range(n_threads)]
```
hannes-ucsc: Let's switch to ThreadPoolExecutor.

test/integration_test.py
```diff
@@ -404,6 +410,78 @@ def _get_entities_by_project(self, entity_type, project_short_name):
...
+        n_ops = 10
+        portal_service = PortalService()
+
+        # Assign to local variable outside of callback so we don't need to put
+        # the entire test inside of _crud call
+        portal_db = None
+
+        def get_db(db):
+            # `make format` is convinced this needs to be on two lines.
+            nonlocal \
```
hannes-ucsc: There must be a less convoluted way. `_crud` could just return what he callback returns.

test/integration_test.py
```diff
@@ -405,6 +410,54 @@ def _get_entities_by_project(self, entity_type, project_short_name):
...
+                    ],
+                    "mock-count": entry_format.format(thread_count, op_count)
+                })
+                portal_service._crud(lambda db: list(db) + [mock_entry])
+
+        old_db = portal_service.read()
+
+        with ThreadPoolExecutor(max_workers=n_threads) as executor:
+            for i in range(n_threads):
+                executor.submit(run, i)
```
hannes-ucsc: To collect exceptions raised in worker threads, one would need to keep track of the return values of submit() and call result() or exception() on them after all tasks have been submitted. If the tasks don't return anything the bare bones is

```
futures = [executor.submit() for i in range(…)]
assert all(f.result() is None for f in futures)
```

This would raise an exception on the first worker exception encountered and ignore any exceptions in the other futures. I think that would suffice here as long as a worker exception propagates to a test failure.

test/integration_test.py
```diff
@@ -405,6 +410,54 @@ def _get_entities_by_project(self, entity_type, project_short_name):
...
+                        }
+                    ],
+                    "mock-count": entry_format.format(thread_count, op_count)
+                })
+                portal_service._crud(lambda db: list(db) + [mock_entry])
+
+        old_db = portal_service.read()
+
+        with ThreadPoolExecutor(max_workers=n_threads) as executor:
+            for i in range(n_threads):
```
hannes-ucsc: I'm not sure how quickly TPE allocates new threads. It might make sense to schedule an order of magnitude more tasks than the pool can run in parallel.

```suggestion
            for i in range(n_threads * 10):
```

src/azul/version_service.py
```diff
@@ -0,0 +1,74 @@
...
+
+from azul import config
+
+
+class VersionService:
+    key_name = 'object_url'
+    value_name = 'current_version'
+
+    def __init__(self):
+        self.client = boto3.client('dynamodb', 'us-east-1')
```
hannes-ucsc:
```suggestion
        self.client = boto3.client('dynamodb')
```

Generally speaking, we cannot hard-code the region. The region is configured like so

https://github.com/DataBiosphere/azul/blob/develop/deployments/.example.local/environment#L38

noah-aviel-dove: Is there a more elegant way to read the value from within the codebase or is it okay to just use `os.environ` here?

hannes-ucsc: I'm pretty sure, you don't have to specify anything. You won't see a region name being passed to a boto3 client instantiation anywhere else in the code base, except for direct DSS access and there is a ticket about eliminating that. I you find yourself needing to pass a region name on your machine/personal deployment, you probably didn't set AWS_DEFAULT_REGION as per the example deployment.

Note this used to be set in [environment.local](https://github.com/DataBiosphere/azul/blob/a49b098a38772d10b06594636e323e76e214f2bc/deployments/.example.local/environment.local#L2) but on develop this now is in [environment](https://github.com/DataBiosphere/azul/blob/hca/develop/deployments/.example.local/environment.local#L2).

lambdas/service/app.py
```diff
@@ -1987,3 +1954,41 @@ def gs_url(file_uuid, version):
 
 class GatewayTimeoutError(ChaliceViewError):
     STATUS_CODE = 504
+
+
+@app.route('/integrations/add_portal', methods=['POST'])
```
hannes-ucsc: I had missed these two endpoints in earlier reviews. We need to disable or remove them. The easiest way to disable would be to embed the method def in `if False`. Instead of `False` an environment variable could be used, with a wrapper property in `azul.config` etc, but that seems overkill.

noah-aviel-dove: Is there any reason to keep them in the source code at all? I only added them for demo purposes, and with the script added in #1511 they aren't even needed for that anymore. I was expecting that the commit that added them would be omitted when rebase/squashing once we were ready to merge.

hannes-ucsc: > Is there any reason to keep them in the source code at all?

I was undecided but it sounds like you're leaning towards removing them. Please do so, ideally by removing the commit that added them.

>  was expecting that the commit that added them would be omitted when rebase/squashing once we were ready to merge.

If you want the reviewer to remove a commit during squashing, you can add the prefix DELETE to the commit title as documented in the contributing guidelines.

noah-aviel-dove: Noted. The commit has been removed.

Issue #1262, opened 2019-09-07
OpenAPI UI renders incorrect server URL


hannes-ucsc:
<img width="468" alt="Screen Shot 2019-09-06 at 7 56 14 PM" src="https://user-images.githubusercontent.com/5143256/64469042-a9b56580-d0e0-11e9-9fa3-729719dfecdc.png">

That should be `service.…` not `drs.…`

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-782)
┆Project Name: azul
┆Issue Number: AZUL-782
┆Epic: User web service API

PR #1431, opened 2019-11-07
Fix: OpenAPI UI renders incorrect server URL (#1262)

jessebrennan:
Even though API Gateway is configured with multiple (namely 2) different
server domains, when requesting the OpenAPI spec from AWS only 1 is
returned. My guess is that because OpenAPI v2 only supported one "host"
entry, AWS doesn't preserve this information for when the spec is
generated, even if v3 is requested. Probably, behind the scenes, they
just convert their v2 spec to v3.

To solve this, I hard coded the URL in the OpenAPI template. and removed
from the spec any servers that are API Gateway lists.

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -92,6 +92,9 @@ openapi_spec = {
             'name': 'Health',
             'description': 'For checking on service health'
         }
+    ],
+    'servers': [
+        {'url': config.service_endpoint()}
     ]
 }
 
--- a/src/azul/openapi.py
+++ b/src/azul/openapi.py
@@ -36,7 +36,7 @@ def annotated_specs(gateway_id, app, toplevel_spec) -> JSON:
     """
     specs = aws.api_gateway_export(gateway_id)
     clean_specs(specs)
-    specs = merge_dicts(specs, toplevel_spec, override=True)
+    specs = merge_dicts(toplevel_spec, specs, override=True)
     docs = get_doc_specs(app)
     for path, doc_spec in docs.items():
         for verb, description in specs['paths'][path].items():
@@ -52,6 +52,8 @@ def clean_specs(specs):
     # Filter out 'options' since it causes linting errors
     for path in specs['paths']:
         specs['paths'][path].pop('options', None)
+    # Remove listed servers since API Gateway give false results
+    specs.pop('servers')
 
 
 def get_doc_specs(app):


jessebrennan:
I can't get the import changes to go away... Something seems to be wrong with my git.

jessebrennan:
@danielsotirhos 

> Ideally duplicated URLs wouldn't be listed in the spec source. Are we ok with relying on Swagger to hide the duplicates?

I think this is the best solution for now. First of all, it works. Secondly, we are considering phasing out the part of the spec that is generated from API Gateway, which eliminate the duplication.

> Do we need / want the 'drs.dev...' URL listed if the 'service.dev..." URL is already included?

I think yes, but let's see what @hannes-ucsc has to say. I will assign him for review and hopefully he can clarify.

src/azul/__init__.py
```diff
@@ -244,7 +244,7 @@ def api_lambda_domain_aliases(self, lambda_name):
         """
         Additional alias domain names for the given API lambda
         """
-        return [self.drs_domain] if lambda_name == 'service' and self.drs_domain else []
+        return [self.drs_endpoint()] if lambda_name == 'service' and self.drs_domain else []
```
hannes-ucsc: An endpoint is a URL. A domain is just a hostname. This method needs to return domains, not URLs. If you look at the callsite, it's in the TF config. Did you test this change by running `make terraform`?

lambdas/service/app.py
```diff
@@ -92,6 +92,10 @@
             'name': 'Health',
             'description': 'For checking on service health'
         }
+    ],
+    'servers': [
+        {'url': url} for url in
```
hannes-ucsc: Wrap before `for`.

src/azul/__init__.py
hannes-ucsc: That would have revealed the mistake.

jessebrennan: I did not test it that way. I had conflated, "endpoint" and "domain". I see my mistake and removed this commit.

hannes-ucsc:
Please stick with protocol and don't squash.

jessebrennan:
@hannes-ucsc I had to remove a commit. I figured that adding fixups on top of it would have made things difficult for you to squash. Should I have removed the commit, but left the fixups?

Issue #1404, opened 2019-10-29
Document manifest endpoints


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-857)
┆containerName: azul
┆Issue Number: AZUL-857
┆Epic: OpenAPI Spec

PR #1435, opened 2019-11-12
Document manifest endpoints (#1404)

danielsotirhos:


lambdas/service/app.py
```diff
@@ -771,36 +806,76 @@ def get_order():
...
+            'description': 'The desired format of the output. Possible values are `compact` (the default) for a '
+                           'tab-separated manifest and `terra.bdbag` for a manifest in the format documented '
+                           '`http://bd2k.ini.usc.edu/tools/bdbag/. The latter is essentially a ZIP file containing two '
+                           'manifests: one for participants (aka Donors) and one for samples (aka specimens). The '
+                           'format of the manifests inside the BDBag is documented here: '
+                           'https://software.broadinstitute.org/firecloud/documentation/article?id=10954',
+            'required': False,
+            'schema': {
+                'type': 'string',
+                'enum': [
```
jessebrennan: Consider consolidating this list with https://github.com/DataBiosphere/azul/blob/225699a3af92a90d6bf1ad3653be2f0b52973b7a/lambdas/service/app.py#L898

jessebrennan: Also, not all of these are listen in the description, and there differences should be described.

lambdas/service/app.py
```diff
@@ -771,36 +806,76 @@ def get_order():
...
+    'description': 'Initiate and check status of a manifest generation job, returning a either a 301 redirecting to a '
+                   'URL to re-check the status of the manifest generation or a 302 redirecting to the location of the '
+                   'completed manifest.',
+    'parameters': [
+        {
+            'in': 'query',
+            'name': 'filters',
+            'description': 'Filters to be applied when generating the manifest',
+            'required': False,
+            'schema': {
```
jessebrennan: I think, instead of "schema", you can specify 

```
"content": {
  "application/json": {
    "schema": {
      "$ref": "#/components/schemas/..."
    }
  }
}
```

And then just start some empty template in components/schemas. See https://swagger.io/docs/specification/describing-parameters/#schema-vs-content

The schema for manifest parameters will be the same as the schema for file any other filter param for the other endpoints (right??). I don't think you need to bother to create the whole schema though, I think it's fair to leave that for someone else father down the line.

lambdas/service/app.py
```diff
@@ -810,69 +885,79 @@ def start_manifest_generation():
                     status_code=301 if wait_time else 302)
 
 
-@app.route('/fetch/manifest/files', methods=['GET'], cors=True)
+@app.route('/fetch/manifest/files', methods=['GET'], cors=True, spec={
+    'description': 'Initiate and check status of a manifest generation job, returning a 200 response with simulated '
+                   'headers in the body.<p>'
```
jessebrennan: Consider putting the `'<p>'` on their own lines to make the separation even more clear.

lambdas/service/app.py
```diff
@@ -810,69 +885,79 @@ def start_manifest_generation():
...
+                   'service and the client should expect a response containing the actual manifest. Currently the '
+                   '`Location` field of the final response is a signed URL to an object in S3 but clients should not '
+                   'depend on that.',
+    'parameters': [
+        {
+            'in': 'query',
+            'name': 'filters',
+            'description': 'Filters to be applied when generating the manifest',
+            'required': False,
+            'schema': {
```
jessebrennan: Same as other 'filters' comment.

lambdas/service/app.py
```diff
@@ -810,69 +885,79 @@ def start_manifest_generation():
...
+                'type': 'string'
+            }
+        },
+        {
+            'in': 'query',
+            'name': 'format',
+            'description': 'The desired format of the output. Possible values are `compact` (the default) for a '
```
jessebrennan: Instead of duplicating the out of date description here, we can just use python's reuse mechanisms and make a variable or something.

lambdas/service/app.py
danielsotirhos: 'tsv' and 'bdbag' are legacy values for 'compact' and 'terra.bdbag' respectively. I suspect we might want them mentioned in the description of the 'format' field, however probably don't want them listed as possible values.

hannes-ucsc: If we want to drive validation from the spec we need to mention the backward-compatible entries in the enum. The description of the `format` parameter should make them as deprecated.

lambdas/service/app.py
```diff
@@ -860,69 +904,68 @@ def start_manifest_generation():
...
+            'description': 'The manifest generation has been started or is still ongoing',
+            'content': {
+                'application/json': {
+                    'schema': schema.object(
+                        Status=int,
+                        Location={'type': 'string', 'format': 'url'},
+                        **{'Retry-After': schema.optional(int)}
+                    )
+                }
+            }
```
jessebrennan: This looks to be the same as https://github.com/DataBiosphere/azul/pull/1435/files#diff-a3dec4f6b35a08e7cb29086cb8766c3aR871-R879

and should be extracted as a variable as well to reduce redundancy.

jessebrennan: Ah.. I see it's not. Maybe there's a way to reduce the redundancy still.

lambdas/service/app.py
```diff
@@ -860,69 +904,68 @@ def start_manifest_generation():
...
+            description=manifest_description,
+        ),
+        params.query(
+            'token',
+            schema.optional({'type': 'string'}),
+            description='Reserved. Do not pass explicitly.',
+        ),
+    ],
+}, method_spec={
+    'description': unwrap('''
```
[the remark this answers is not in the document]
hannes-ucsc: unwrap() was created for this very purpose. Is the problem that it prevents putting paragraph breaks in the description?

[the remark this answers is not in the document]
hannes-ucsc: And by purpose I mean to be used on `description` fields in specs.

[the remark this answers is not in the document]
noah-aviel-dove: I can confirm that the current implementation of the `unwrap` helper does not preserve paragraph breaks. I considered preserving consecutive newlines when writing  but decided not to sink time into adding that feature when my PR had already dragged on for so long. It would not be too difficult to add. The helper is very simple.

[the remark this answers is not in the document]
hannes-ucsc: > but decided not to sink time into adding that feature

It's fine to forgo a change in favour of a simpler implementation that satisfies the current needs. But in this case I would have filed a ticket about this improvement.

> when my PR had already dragged on for so long

That is to be expected. We value quality over quantity because, in the long run, only quality ensures **sustained** productivity.

danielsotirhos:
Unsure if we still want `unwrap_description()` or if all fields needing unwrapping (including description) should be unwrapped by calling unwrap() where the documentation is defined.

danielsotirhos:
I kept the original behaviour of `unwrap()` by adding an optional boolean argument 'keep_paragraphs'. I can make this mode the default or remove the other mode (not preserving paragraphs) if not needed.

lambdas/service/app.py
```diff
@@ -821,36 +825,76 @@ def get_order():
...
-          description: The desired format of the output. Possible values are `compact` (the default) for a tab-separated
-          manifest and `terra.bdbag` for a manifest in the format documented `http://bd2k.ini.usc.edu/tools/bdbag/. The
-          latter is essentially a ZIP file containing two manifests: one for participants (aka Donors) and one for
-          samples (aka specimens). The format of the manifests inside the BDBag is documented here:
-          https://software.broadinstitute.org/firecloud/documentation/article?id=10954
-        - name: token
-          in: query
-          type: string
-          description: Reserved. Do not pass explicitly.
+manifest_description = unwrap('''
```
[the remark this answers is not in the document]
danielsotirhos: Modified to now use a modified `unwrap()` that preserves paragraph breaks (empty lines)

lambdas/service/app.py
```diff
@@ -821,36 +825,76 @@ def get_order():
...
+    <p>The desired format of the output.
```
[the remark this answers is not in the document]
danielsotirhos: Removed html tags and replaced with Markdown and empty lines for paragraph breaks.

lambdas/service/app.py
```diff
@@ -821,36 +825,76 @@ def get_order():
...
+    <ul>
```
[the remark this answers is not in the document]
danielsotirhos: It looks like Markdown and HTML don't mix well. After the HTML was removed the Markdown was able to be rendered correctly.

lambdas/service/app.py
```diff
@@ -821,36 +825,76 @@ def get_order():
...
+    </ul>
+''')
 
-    :return: If the manifest generation has been started or is still ongoing, the response will have a
-    301 status and will redirect to a URL that will get a recheck the status of the manifest.
 
-    If the manifest generation is done and the manifest is ready to be downloaded, the response will
-    have a 302 status and will redirect to the URL of the manifest.
-    """
+@app.route('/manifest/files', methods=['GET'], cors=True, path_spec={
```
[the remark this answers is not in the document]
danielsotirhos: Removed redundancy in the documentation.

lambdas/service/app.py
danielsotirhos: Extracted redundant documentation to a function to allow for the slight difference in the instances where it is used.

lambdas/service/app.py
```diff
@@ -821,36 +825,84 @@ def get_order():
     return {'order': Plugin.load().service_config().order_config}
 
 
-@app.route('/manifest/files', methods=['GET'], cors=True)
-def start_manifest_generation():
-    """
-    Initiate and check status of a manifest generation job, returning a either a 301 or 302 response
-    redirecting to either the location of the manifest or a URL to re-check the status of the manifest job.
+manifest_filters = {
+    facet_name: schema.object(**{'is': schema.array(str)})
```
hannes-ucsc: Why so complicated? 

```suggestion
    facet_name: schema.object(is=schema.array(str))
```

lambdas/service/app.py
```diff
@@ -821,36 +825,84 @@ def get_order():
     return {'order': Plugin.load().service_config().order_config}
 
 
-@app.route('/manifest/files', methods=['GET'], cors=True)
-def start_manifest_generation():
-    """
-    Initiate and check status of a manifest generation job, returning a either a 301 or 302 response
-    redirecting to either the location of the manifest or a URL to re-check the status of the manifest job.
+manifest_filters = {
```
hannes-ucsc: We need to make sure that the meta domain variables such as this one are name in such a way that their "meta-ness" is obvious to avoid confusion and naming conflicts. The suffix `_spec` already been established for this purpose.

```suggestion
manifest_filter_spec = {
```

hannes-ucsc: Secondly, the filter argument is universal so this can be generalized to

```suggestion
filter_spec = {
```

lambdas/service/app.py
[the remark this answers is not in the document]
jessebrennan: `start_manifest_generation` and `start_manifest_generation_fetch` both just call `handle_manifest_generation_request` (but return different forms of responses). This means they use the same code to parse the filters parameter. This implies that either both should have `file=` in the description, or neither should. In any case the descriptions of parameters will be the same.

It looks like all of the parameters (and therefore the entire `path_spec`) in the manifest endpoints should be the same, and therefore extracted out into its own variable. Since the variables you created for `manifest_filters`, `manifest_param_token` and `manifest_param_format` would only be used in once, in this _new_ variable, they should probably just be inlined.

src/azul/strings.py
```diff
@@ -48,12 +48,15 @@ def pluralize(word: str, count: int) -> str:
     return result
 
 
-def unwrap(string: str):
+def unwrap(string: str, keep_paragraphs: bool = False) -> str:
```
hannes-ucsc: Why do we need this modal argument?

lambdas/service/app.py
danielsotirhos: The complicated approach was needed because `is` is a python reserved word.
```
>>> def f(**params) -> bool:
...    print(params)    
    
>>> f(foo='bar')
{'foo': 'bar'}

>>> f(is='test')
  File "<input>", line 1
    f(is='test')
       ^
SyntaxError: invalid syntax

>>> f(**{'is': 'test'})
{'is': 'test'}
```

src/azul/strings.py
jessebrennan: I think unwrap should only ever `keep_paragraphs` and it shouldn't be an option. If you're worried about this breaking something, double check the usages of `unwrap` to be sure (but I don't think it will). 

Also I recall @hannes-ucsc saying that the body of `unwrap` could just be replaced with a call to `dedent`. If this is done, the name "unwrap" doesn't really make sense and perhaps should be replaced with `format_description` or something? IDK. I think @hannes-ucsc might have said something about it during our call on Friday, but I can't remember.

lambdas/service/app.py
hannes-ucsc: Got it, that makes sense.

src/azul/strings.py
hannes-ucsc: Yes, I think `unwrap` should always call `dedent` and not be modal. The name `unwrap` doesn't make sense. I'll leave it to you if you want to rename it to `format_description` as @jessebrennan  suggests or if you'd rather create a ticket for the rename.

danielsotirhos:
- Consolidated `path_spec` values for both manifest endpoints into a function `manifest_path_spec()`. This wouldn't work as a variable as it caused the `manifest_spec` value for `/manifest/files` to be overwritten with the `manifest_spec` from `fetch/manifest/files`.

- Implemented `format_description()` to replace `unwrap()`. This function now uses `textwrap.dedent()` to remove all common leading whitespace from the lines of given text.

- Implemented `format_description_key()` to replace `unwrap_description()`. This function applies `format_description()` to the given dict's `description` key (if it exits)

lambdas/service/app.py
```diff
@@ -825,35 +825,44 @@ def get_order():
...
-          [https://software.broadinstitute.org/firecloud/documentation/article?id=10954]
-    ''',
-)
-
-manifest_param_token = params.query(
-    'token',
-    schema.optional({'type': 'string'}),
-    description='Reserved. Do not pass explicitly.',
-)
+def manifest_path_spec():
```
jessebrennan: This works, or there could be an explicit call to `copy` when the variable is used.

src/azul/strings.py
```diff
@@ -48,38 +55,53 @@ def pluralize(word: str, count: int) -> str:
...
+    Remove common leading whitespace from every line in text.
     Useful for processing triple-quote strings.
 
     :param string: The string to unwrap
-    :param keep_paragraphs: To preserve paragraph breaks as a single empty line
 
-    >>> unwrap(' ce nest pas \\n une chaine \\n de plusieurs lignes. ')
-    'ce nest pas une chaine de plusieurs lignes.'
+    >>> format_description(" c'est \\n une chaine \\n de plusieurs lignes. ")
+    "c'est \\nune chaine \\nde plusieurs lignes. "
```
jessebrennan: I appreciate the corrected string ;)

test/service/test_async_manifest_service.py
```diff
@@ -145,7 +145,7 @@ def test_manifest_endpoint_start_execution(self, mock_uuid, step_function_helper
                     mock_uuid.return_value = execution_name
                     step_function_helper.describe_execution.return_value = {'status': 'RUNNING'}
                     format_ = 'compact'
-                    filters = {'file': {'organ': {'is': ['lymph node']}}}
+                    filters = {'organ': {'is': ['lymph node']}}
```
jessebrennan: Was this a bug? Why was this test passing before?

danielsotirhos: Appears to be a bad test, or at least a test that doesn't check the validity of the filters. This test is only verifying the response after the first request to the manifest service, checking for a 301 status and for the 'Retry-After' and 'Location' fields.
I'm not sure why the filter's value used was chosen, the test passes with `filters = {}` or even `filters = {'foo': 'bar'}`

lambdas/service/app.py
```diff
@@ -821,36 +825,81 @@ def get_order():
     return {'order': Plugin.load().service_config().order_config}
 
 
-@app.route('/manifest/files', methods=['GET'], cors=True)
-def start_manifest_generation():
-    """
-    Initiate and check status of a manifest generation job, returning a either a 301 or 302 response
-    redirecting to either the location of the manifest or a URL to re-check the status of the manifest job.
+def manifest_path_spec():
```
hannes-ucsc: Why is this a function and not a simple variable?

src/azul/strings.py
```diff
@@ -48,20 +55,53 @@ def pluralize(word: str, count: int) -> str:
     return result
 
 
-def unwrap(string: str):
+def format_description(string: str) -> str:
```
hannes-ucsc: With the change to implementation and name, this is not a general string function anymore. It should go where `unwrap_description` used to be.

hannes-ucsc: Ideally, I would want to invoke this function as `api.description(…)`. This implies either an `import azul.openapi as api` in the referring module or that this function is defined in `src/azul/openapi/api.py` in conjunction with `from azul.openapi import api`. I am not sure. I think it is a bad smell if modules import from the `__init__.py` of the package they're in which most certainly will be necessary if this function is defined in `src/azul/openapi/__init__.py`. Let's do this, you move the function back, leave the name and I'll add a commit that gives the function a logical and concise name.

lambdas/service/app.py
```diff
@@ -860,69 +909,39 @@ def start_manifest_generation():
                     status_code=301 if wait_time else 302)
 
 
-@app.route('/fetch/manifest/files', methods=['GET'], cors=True)
+@app.route('/fetch/manifest/files', methods=['GET'], cors=True, path_spec=manifest_path_spec(), method_spec={
+    'description': format_description('''
+        Initiate and check status of a manifest generation job, returning a
```
```suggestion
        Initiate a manifest generation or check the status of an already ongoing generation, returning a
```

lambdas/service/app.py
```diff
@@ -860,69 +909,39 @@ def start_manifest_generation():
...
+        of the final response is a signed URL to an object in S3 but clients
+        should not depend on that.
+    '''),
+    'responses': {
+        '200': {
+            **manifest_response_content(include_status=True),
+            'description': 'The manifest generation has been started or is still ongoing',
+        }
+    },
+    'summary': 'Describe current version of the Azul service',
```
hannes-ucsc: Better to write from hand instead of copying and pasting. Summary should precede description.

lambdas/service/app.py
```diff
@@ -860,69 +909,39 @@ def start_manifest_generation():
                     status_code=301 if wait_time else 302)
 
 
-@app.route('/fetch/manifest/files', methods=['GET'], cors=True)
+@app.route('/fetch/manifest/files', methods=['GET'], cors=True, path_spec=manifest_path_spec(), method_spec={
+    'description': format_description('''
+        Initiate and check status of a manifest generation job, returning a
+        200 response with simulated headers in the body. If the client receives
```
```suggestion
        200 response with simulated HTTP headers in the body. If the client receives
```

src/azul/strings.py
```diff
@@ -48,20 +55,53 @@ def pluralize(word: str, count: int) -> str:
...
     ...        indented,
     ...        triple-quoted string.
     ...        ''')
-    'Multi-lined, indented, triple-quoted string.'
+    '\\nMulti-lined,\\nindented,\\ntriple-quoted string.\\n'
+    """
+    return dedent(string)
+
+
+def format_description_key(kwargs: MutableMapping[str, Any]) -> None:
```
hannes-ucsc: And this one is even more specific to OpenAPI spec, so it's definitely in the wrong place here.

test/service/test_async_manifest_service.py
hannes-ucsc: Since this uses mocks, the filter doesn't matter, right? This used to be a correct filter, for historic reasons. This predates my time, potentially from before the indexer denormalized associated entities into inner ones.

lambdas/service/app.py
```diff
@@ -821,36 +825,81 @@ def get_order():
...
-          samples (aka specimens). The format of the manifests inside the BDBag is documented here:
-          https://software.broadinstitute.org/firecloud/documentation/article?id=10954
-        - name: token
-          in: query
-          type: string
-          description: Reserved. Do not pass explicitly.
 
-    :return: If the manifest generation has been started or is still ongoing, the response will have a
-    301 status and will redirect to a URL that will get a recheck the status of the manifest.
+def manifest_response_content(include_status=False):
```
```suggestion
def manifest_response_body_spec(include_status=False):
```

lambdas/service/app.py
```diff
@@ -821,36 +825,81 @@ def get_order():
...
+@app.route('/manifest/files', methods=['GET'], cors=True, path_spec=manifest_path_spec(), method_spec={
+    'description': format_description('''
+        Initiate and check status of a manifest generation job, returning
+        either a 301 response redirecting to a URL to re-check the status of
+        the manifest generation or a 302 response redirecting to the location
+        of the completed manifest.
+    '''),
+    'responses': {
+        '301': {
+            **manifest_response_content(),
```
hannes-ucsc: This doesn't work. The 301 and 302 responses don't have a body.

lambdas/service/app.py
danielsotirhos: As a variable the functionality of the `method_spec` parameter broke for the two manifest endpoints, causing the value from `/manifest/files` method_spec to be overwritten with the method_spec from `/fetch/manifest/files`.

Two possible solutions for this were to make the `manifest_path_spec` variable a function, or copy/deepcopy the variable when passed in as a `path_spec` parameter. The function approach seemed cleaner however I will change this to the copy approach instead.

lambdas/service/app.py
danielsotirhos: The swagger specification didn't specify a preferred ordering of keys so I thought alphabetical would be a consistent scheme for us to use. It does unfortunately result in `description` key being separated from the `summary` key though. I will try to re-order the keys based on a more contextual ordering.

lambdas/service/app.py
```diff
@@ -94,6 +95,10 @@
         {
             'name': 'Auxiliary',
             'description': 'Describes various aspects of the Azul service'
+        },
+        {
+            'name': 'Manifest',
```
```suggestion
            'name': 'Manifests',
```

lambdas/service/app.py
```diff
@@ -94,6 +95,10 @@
...
+            'description': 'For requesting a manifest of files that match a given search criteria'
```
```suggestion
            'description': 'Complete listing of files matching a given filter in TSV and other formats
```

I usually try to avoid circular definitions.

test/service/test_async_manifest_service.py
danielsotirhos: Yes, normally a request to the manifest endpoint is handled by `handle_manifest_generation_request()` which takes the `filters` param and uses it in a call to `ManifestService.get_cached_manifest`. This test function mock's `get_cached_manifest()` to simply return None regardless of what is passed in.

danielsotirhos:
Added a more complete description for `/fetch/manifest/files` response.
Moved fix of filter param (remove 'file' key) in test function to a separate commit.
Rebased on latest develop

lambdas/service/app.py
```diff
@@ -831,36 +836,79 @@ def get_order():
...
+    'description': format_description('''
+        Initiate and check status of a manifest generation job, returning
+        either a 301 response redirecting to a URL to re-check the status of
+        the manifest generation or a 302 response redirecting to the location
+        of the completed manifest.
+    '''),
+    'responses': {
+        '301': {
+            'description': 'The manifest generation has been started or is still ongoing',
+            **manifest_response_headers_spec
```
hannes-ucsc: I think we should denormalize this i.e., undo the extraction of the variable. The URL in the Location header has a different meaning and its confusing to describe both cases identically (and therefore unnecessarily generic) as URL to be requested next.

lambdas/service/app.py
```diff
@@ -870,69 +918,63 @@ def start_manifest_generation():
                     status_code=301 if wait_time else 302)
 
 
-@app.route('/fetch/manifest/files', methods=['GET'], cors=True)
+@app.route('/fetch/manifest/files', methods=['GET'], cors=True, path_spec=copy.copy(manifest_path_spec), method_spec={
```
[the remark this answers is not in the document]
danielsotirhos: Created https://github.com/DataBiosphere/azul/issues/1646 for bug with `app.route`'s `method_spec` being overwritten by another `method_spec`. Unknown at present why a shallow copy of the `path_spec` param value serves as a work-around fix for the bug however I have verified (through testing) that it does, and included a note about shallow copy in the bug ticket.

lambdas/service/app.py
```diff
@@ -831,36 +836,101 @@ def get_order():
...
+        of the completed manifest.
+    '''),
+    'responses': {
+        '301': {
+            'description': format_description('''
+                The manifest generation has been started or is still ongoing.
+                The response is a redirect back to this endpoint, so the client
+                should expect a subsequent response of the same kind.
+            '''),
+            **{
```
hannes-ucsc: Is the `**` necessary?

lambdas/service/app.py
```diff
@@ -831,36 +836,101 @@ def get_order():
...
+                        'schema': {'type': 'string'},
+                    },
+                }
+            }
+        },
+        '302': {
+            'description': format_description('''
+                The manifest generation is complete and ready for download.
+            '''),
+            **{
```
hannes-ucsc: Is the `**` necessary?

lambdas/service/app.py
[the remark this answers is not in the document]
hannes-ucsc: Thank you.

lambdas/service/app.py
```diff
@@ -870,69 +940,48 @@ def start_manifest_generation():
...
+                here is intended to be processed by client-side Javascript such
+                that the recommended delay in `Retry-After` can be handled in
+                Javascript rather than relying on the native implementation by
+                the web browser.
 
-    The client should request the URL given in the `Location` field. The URL will point to a different service and
-    the client should expect a response containing the actual manifest. Currently the `Location` field of the final
-    response is a signed URL to an object in S3 but clients should not depend on that.
-    """
+                See `/manifest/files` [(link)](#operations-Manifests-get_manifest_files)
```
```suggestion
                See [`/manifest/files`](#operations-Manifests-get_manifest_files) endpoint 
```

It works in GH Markdown:

[`some text`](https://github.com/DataBiosphere/azul/pull/1435)

hannes-ucsc: Also, you're making it a little easy for yourself here. I though we agreed on our last call that it needs to be made clear that the **fields** of a `/fetch/manifest/files` response body directly relate to the response **headers** from the `/manifest/files` endpoint. Let's try to nail this in the next review cycle. I'm getting tired of writing more text in review comments on documentation than there actually is documentation text.

src/azul/openapi/__init__.py
```diff
@@ -288,30 +288,52 @@ def _recursive_merge_dicts(d1: JSON, d2: JSON, path: List[str], override: bool)
...
+    Remove common leading whitespace from every line in text.
+    Useful for processing triple-quote strings.
+
+    :param string: The string to unwrap
+
+    >>> format_description(" c'est \\n une chaine \\n de plusieurs lignes. ")
+    "c'est \\nune chaine \\nde plusieurs lignes. "
+
+    >>> format_description('''
+    ...        Multi-lined,
```
hannes-ucsc: Indent by 4 not 7 spaces.

lambdas/service/app.py
danielsotirhos: Good catch. It was necessary when the `**` was unpacking a dictionary stored in a variable, but certainly not now. My fault for not checking PyCharm's refactor inline results more closely.

lambdas/service/app.py
danielsotirhos: I've reworded the sentence with the link and the preceding paragraph in attempt to better highlight the relationship of the response fields to the response headers they emulate.

Issue #1242, opened 2019-08-29
Invalidate cached manifest when a project is updated


theathorn:
Currently we update the cache version of the project metadata tsv file once a day.

Since projects change infrequently following initial ingestion, it would be better to invalidate the cached version when a project is updated.

When would the new version be generated? On first user demand?

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-772)
┆Project Name: azul
┆Issue Number: AZUL-772

hannes-ucsc:
1) Track version of project as the maximum version of a contributing bundle

2) Incorporate project version in manifest hash

Optimizations:

3) Add a `cron` function to periodically request the full manifest for each project (not exactly part of the solution but this makes it less likely that users will have to wait for the manifest to be generated after it was invalidated)

4) Use a heuristic to prevent multiple concurrent step function executions for the same hash. This is difficult because step functions does not let you retrieve functions by name, only by ID and you don't control the ID.

achave11:
A query that generates a Hash for a specific query.
```
{
  "size": 0, 
  "query": {
    "match_all": {}
  },
  "aggs": {
    "fields": {
      "scripted_metric": {
        "init_script": "params._agg.fields = 0",
        "map_script": "for (bundle in params._source.bundles) { params._agg.fields += (bundle.uuid + bundle.version).hashCode()}",
        "combine_script": "return params._agg.fields.hashCode()",
        "reduce_script": "int result = 0; for (agg in params._aggs) { result += agg } return result"
      }
    }
  }
}
```

PR #1442, opened 2019-11-13
Invalidate cached manifest when index is updated (#1242)

achave11:


src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -599,12 +602,35 @@ def _manifest_params_full(self, filters: JSON) -> Tuple[ManifestConfig, List[str
...
+        es_search.aggs.metric('hash', 'scripted_metric',
+                              init_script='params._agg.fields = 0',
+                              map_script=generate_hash_map_script,
+                              combine_script='return params._agg.fields.hashCode()',
+                              reduce_script=generate_hash_reduce_script)
         es_search = es_search.extra(size=0)
+        # The following is a workaround for ES inability to execute queries immediately after index creation. All shards
+        #  must be active in order to execute hash_code aggregate, not all shards are available on index creation.
+        import time
+        time.sleep(3)
```
hannes-ucsc: Nope

src/azul/service/responseobjects/hca_response_v5.py
```diff
@@ -615,20 +566,88 @@ def return_response(self):
...
         presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
         headers = {'Content-Type': 'application/json', 'Location': presigned_url}
         return Response(body='', headers=headers, status_code=302)
 
 
+class ManifestCacheHelper:
+    """
+    Helper for cached manifest containing validation methods
+    """
+    storage_service = StorageService()
```
hannes-ucsc: No static service instances, please.

src/azul/service/manifest_service.py
```diff
@@ -291,6 +302,14 @@ def manifest_config(self) -> ManifestConfig:
         """
         return self.service.service_config.manifest
 
+    @property
+    def manifest_hash_code(self) -> int:
+        """
+        The manifest hash code this generator uses. A manifest hash code is generated
+        to distinguish the unique set of data in the elasticsearch response.
+        """
+        return self.service.service_config.manifest_hash_code
```
hannes-ucsc: This makes no sense to me. ServiceConfig doesn't have this attribute/property.

src/azul/service/manifest_service.py
```diff
@@ -224,6 +224,17 @@ def _create_name_object(self, manifest_object_key, base_name, extension):
                                  data=file_name.encode())
         return file_name
 
+    def fetch_cached_manifest(self, format_: str, filters: Optional[str] = None):
```
hannes-ucsc: This should come after transform_manifest

src/azul/service/manifest_service.py
```diff
@@ -291,6 +302,14 @@ def manifest_config(self) -> ManifestConfig:
         """
         return self.service.service_config.manifest
 
+    @property
```
hannes-ucsc:
```suggestion
    @cachedproperty
```

Very important.

src/azul/service/manifest_service.py
```diff
@@ -152,7 +152,7 @@ def _use_cached_manifest(self, generator: 'ManifestGenerator', object_key: str)
             file_name = None
         return file_name
 
-    def _derive_manifest_key(self, filters: Filters) -> str:
+    def _derive_manifest_key(self, filters: Filters, hash_code: int, format_: str) -> str:
```
hannes-ucsc:
```suggestion
    def _derive_manifest_key(self, format_: str, filters: Filters, hash_code: int) -> str:
```

src/azul/service/manifest_service.py
hannes-ucsc: And declare return type.

src/azul/service/manifest_service.py
```diff
@@ -224,6 +224,17 @@ def _create_name_object(self, manifest_object_key, base_name, extension):
...
+    def fetch_cached_manifest(self, format_: str, filters: Optional[str] = None):
+        filters = self.parse_filters(filters)
+        generator = ManifestGenerator.for_format(format_, self, filters)
+        hash_code = generator.manifest_hash_code
+        manifest_key = self._derive_manifest_key(filters, hash_code, format_)
+        object_key = f'manifests/{manifest_key}.{generator.file_name_extension}'
+        if generator.is_cacheable and self._can_use_cached_manifest(object_key):
+            file_name = self._use_cached_manifest(generator, object_key)
+            return self.storage_service.get_presigned_url(object_key, file_name=file_name)
+        return False
```
hannes-ucsc: :else branch. Please remember this guideline. It's mentioned in the contributing guide.

src/azul/service/manifest_service.py
```diff
@@ -162,7 +162,7 @@ def _derive_manifest_key(self, filters: Filters) -> str:
         git_commit = config.lambda_git_status['commit']
         manifest_namespace = uuid.UUID('ca1df635-b42c-4671-9322-b0a7209f0235')
         filter_string = repr(sort_frozen(freeze(filters)))
-        return str(uuid.uuid5(manifest_namespace, git_commit + filter_string))
+        return str(uuid.uuid5(manifest_namespace, git_commit + filter_string + str(hash_code) + format_))
```
hannes-ucsc:
```suggestion
        return str(uuid.uuid5(manifest_namespace, git_commit + format_ + filter_string + str(hash_code)))
```

lambdas/service/app.py
```diff
@@ -924,6 +924,11 @@ def handle_manifest_generation_request():
     token = query_params.get('token')
     retry_url = self_url()
     manifest_service = AsyncManifestService()
+    if format_ == 'full':
```
hannes-ucsc: This assumes that the generator for the full manifest is cacheable. In general, you shouldn't make decision based on the format string anymore. The format is used once to determine the generator implementation and then that implementation is interrogated about it's capabilities.

src/azul/indexer.py
```diff
@@ -159,6 +159,7 @@ def _create_indices(self):
                                      ignore=[400],
                                      body=dict(settings=self.settings(index_name),
                                                mappings=dict(doc=self.mapping())))
+            es_client.cluster.put_settings(body=dict(transient={'script.max_compilations_per_minute': 60}))
```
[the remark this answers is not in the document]
achave11: > This method (_create_indices) is called a lot, once per bundle notification. Think of ways to avoid touching the cluster settings every time. If this is a cluster setting, should this go into TF config?
I was unable to find resources that modify the cluster settings via terraform.

achave11:
>Globally s/hash_code/content_hash/

I am unsure as to what this means. 

>The way this is implemented right now it appears that the caching lookup is done twice. Once in the service (app.py) and once in the step function. I could be wrong.

I was under the impression that that was the intended behavior. We discuss in a future flesh out a mechanism where we can inject the object key into the step function to prevent another lookup.

danielsotirhos:
`s/old/new/` is regex speak for substitute the old for the new

src/azul/indexer.py
```diff
@@ -160,6 +160,10 @@ def _create_indices(self):
                                      body=dict(settings=self.settings(index_name),
                                                mappings=dict(doc=self.mapping())))
 
+        # By default `max_compilations_per_minute` is set to 15, we must increase that threshold in order to execute
```
hannes-ucsc: As I wrote earlier, this method is used a lot and I do not want to do this every time a bundle is indexed. You mentioned in an earlier [comment](https://github.com/DataBiosphere/azul/pull/1442#discussion_r356196846) on this section

> I was unable to find resources that modify the cluster settings via terraform.

Please make sure that there really is no way this setting can be applied at cluster creation time. What you don't want is that someone else researches this for five minutes and comes up with a solution.

If there really is no way to do that, we should only limit the application of this setting to tests, and not in production code. Correct me if I am wrong, but I only recall us seeing the rate limit being hit during the tests.

src/azul/service/manifest_service.py
```diff
@@ -175,7 +187,7 @@ def _derive_manifest_key(self, filters: Filters) -> str:
         git_commit = config.lambda_git_status['commit']
         manifest_namespace = uuid.UUID('ca1df635-b42c-4671-9322-b0a7209f0235')
         filter_string = repr(sort_frozen(freeze(filters)))
-        return str(uuid.uuid5(manifest_namespace, git_commit + filter_string))
+        return str(uuid.uuid5(manifest_namespace, git_commit + filter_string + format_ + str(content_hash)))
```
hannes-ucsc: We should use a separator between fields, as in

```suggestion
        content_hash = str(content_hash)
        return str(uuid.uuid5(manifest_namespace, ','.join(git_commit, format_, content_hash, filter_string)))
```
This is a way to avoid collisions as in 

"a" + "ab" == "aa" + "b"

Since commas *can* occur in the filter string, it should go last. We should also assert that commas don't occur in the other fields.

src/azul/service/manifest_service.py
```diff
@@ -175,7 +187,9 @@ def _derive_manifest_key(self, filters: Filters) -> str:
         git_commit = config.lambda_git_status['commit']
         manifest_namespace = uuid.UUID('ca1df635-b42c-4671-9322-b0a7209f0235')
         filter_string = repr(sort_frozen(freeze(filters)))
-        return str(uuid.uuid5(manifest_namespace, git_commit + filter_string))
+        content_hash = str(content_hash)
+        assert all(',' not in param for param in [git_commit, format_, content_hash])
```
hannes-ucsc:
```suggestion
        assert not any(',' in param for param in [git_commit, format_, content_hash])
```

reads better

hannes-ucsc:
```suggestion
        assert all(',' not in param for param in (git_commit, format_, content_hash))
```

since you're passing a tuple to `.join` below.

src/azul/service/manifest_service.py
```diff
@@ -552,6 +566,31 @@ def manifest_config(self) -> ManifestConfig:
             }
         }
 
+    @cachedproperty
+    def manifest_content_hash(self) -> int:
+        es_search = self._create_request()
+        generate_hash_map_script = '''
+                   for (bundle in params._source.bundles) {
```
hannes-ucsc: Looks over-indented.

test/es_test_case.py
```diff
@@ -40,6 +40,12 @@ def setUpClass(cls):
         except BaseException:  # no coverage
             cls._kill_containers()
             raise
+        else:
+            # By default `max_compilations_per_minute` is set to 15, we must increase that threshold in order to execute
+            # dynamic header row and HashCode generation for test_full_hash_validity.
+            # If this also happens in production code, we should move this code to be run as part of `make terraform`,
+            # as a provisioner, for example. Or we can precompile the Painless scripts as part of `make deploy`.
+            cls.es_client.cluster.put_settings(body=dict(persistent={'script.max_compilations_per_minute': 60}))
```
hannes-ucsc: Line is too long. Can we convert the `dict` call to a literal and wrap this a bit?

test/service/test_manifest.py
```diff
@@ -1087,6 +1087,41 @@ def test_manifest_content_disposition_header(self, _can_use_cached_manifest):
                             actual_cd = one(parse_qs(query).get('response-content-disposition'))
                             self.assertEqual(actual_cd, expected_cd)
 
+    @mock_sts
+    @mock_s3
+    def test_full_hash_validity(self):
+        self.maxDiff = None
+        old_bundle = ("aaa96233-bf27-44c7-82df-b4dc15ad4d9d", "2018-11-02T113344.698028Z")
+        self._index_canned_bundle(old_bundle)
+        # When a new bundle is indexed and its full manifest cached,
```
hannes-ucsc: While I like the explanatory comments, the empty lines are in the wrong places it seems. Usually, I'll put an empty line before every comment.

src/azul/service/manifest_service.py
```diff
@@ -79,25 +78,41 @@ def __init__(self, storage_service: StorageService):
 
     def transform_manifest(self, format_: str, filters: Filters):
         generator = ManifestGenerator.for_format(format_, self, filters)
+        object_key, presigned_url = self._fetch_cached_manifest(generator, format_, filters)
+        if not presigned_url:
```
hannes-ucsc:
```suggestion
        if presigned_url is None:
```

src/azul/service/async_manifest_service.py
```diff
@@ -50,7 +53,7 @@ def start_or_inspect_manifest_generation(self,
                                              self_url,
                                              token: Optional[str] = None,
                                              format_: Optional[str] = None,
-                                             filters: Optional[str] = None
+                                             filters: Filters = None
```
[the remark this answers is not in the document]
hannes-ucsc: Some other signatures in your changes are afflicted with this as well.

src/azul/service/manifest_service.py
```diff
@@ -79,13 +79,13 @@ def __init__(self, storage_service: StorageService):
...
         object_key, presigned_url = self._fetch_cached_manifest(generator, format_, filters)
-        if not presigned_url:
+        if presigned_url is None:
             object_key = f'manifests/{uuid.uuid4()}.{generator.file_name_extension}' if not object_key else object_key
             file_name = self._generate_manifest(generator, object_key)
             presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
         return presigned_url
 
-    def fetch_cached_manifest(self, format_: str, filters: Filters = None) -> Optional[str]:
+    def fetch_cached_manifest(self, format_: str, filters: Optional[Filters] = None) -> Optional[str]:
```
hannes-ucsc: I see one call site and it passes a non-None value to `filters`. So there is no need for this to be `Optional` or have a default. `for_format` would fail straight away if `filters` were `None`.

Issue #1436, opened 2019-11-12
Using OpenAPI template from API Gateway adds complications to documenting


jessebrennan:
There are a couple complications that have been pointed out by @danielsotirhos during his development of #1404. They all stem from the problem of overriding the default values sent from API Gateway.

* API Gateway assumes a default 200 response for any of the endpoints. This isn't always correct, specifically the manifest response only returns 301 or 302. Without more extensive changes, it's impossible to override the 200 response with nothing.

* In order to override some properties, one may have to use a less than ideal format. To override the manifest files response, Daniel had to create an entry in components.schemas which is not as straight forward as just inlining the response template.

I suspect that, while useful for initial development, the OpenAPI template downloaded from API Gateway will be completely obsoleted by our coming documentation. The likely solution then, is to generate the entire OpenAPI spec from scratch in the server (or better yet, statically at build time).

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-881)
┆containerName: azul
┆Issue Number: AZUL-881

jessebrennan:
@hannes-ucsc and I discussed this issue and decided that it can be broken down into two parts. One part will be a quick fix that will solve the main issues discussed in bullets above. The scope of the quick fix is much smaller and the work will be covered by this issue. Therefore I'm reducing the estimate to a 1.

The description above also talks about the possibility of creating the API spec at build time (which would also make testing much more straight forward). This work will be tracked in a follow-on ticket: #1450

jessebrennan:
Increased estimate since this now includes the work for #1454 .

PR #1449, opened 2019-11-14
Override API Gateway's values for API documentation (#1436)

jessebrennan:
Before this change, the specs were combined. But now, if any
documentation is provided, we will assume it is correct and complete.
Because of this, it should override the default values provided by API
Gateway.

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -184,19 +184,39 @@ def openapi():
                     body=spec)
 
 
-@app.route('/health', methods=['GET'], cors=True, spec={
+health_spec = {
+    'responses': {
+        '200': {
+            'description': '200 response',
+            'content': {
+                'application/json': {
+                    'schema': {'$ref': '#/components/schemas/Empty'}
+                }
+            }
+        }
+    },
+    'tags': ['Health']
+}
+
+
+@app.route('/health', methods=['GET'], cors=True, method_spec={
+    **health_spec,
+    'summary': 'List health information for webservice'
+})
+@app.route('/health/{keys}', methods=['GET'], cors=True, path_spec={
     'parameters': [
         {
-            'in': 'query',
-            'name': 'foo',
-            'description': 'This is not a real parameter',
-            'required': False,
-            'schema': {'$ref': "#/components/schemas/Empty"}
+            'name': 'keys',
+            'in': 'path',
+            'required': True,
+            'schema': {'type': 'string'}
         }
     ],
-    'tags': ['Health']
+}, method_spec={
+    **health_spec,
+    'summary': 'List health information for a specific key',
+    'description': 'List health information for a specific key, or all keys if `/health/` is used.'
 })
-@app.route('/health/{keys}', methods=['GET'], cors=True)
 def health(keys: Optional[str] = None):
     controller = HealthController(lambda_name='service',
                                   keys=keys,
--- a/src/azul/openapi.py
+++ b/src/azul/openapi.py
@@ -1,24 +1,104 @@
 import copy
-from typing import List
+from typing import (
+    List,
+    cast,
+    Optional,
+)
 
 from azul.deployment import aws
-from azul.types import JSON
+from azul.types import (
+    JSON,
+    MutableJSON,
+)
 
 
-def openapi_spec(description: JSON):
+def openapi_spec(path: str, methods: List[str], path_spec: Optional[JSON] = None, method_spec: Optional[JSON] = None):
     """
-    Add description to decorated function
+    Add description to decorated function at specified route.
 
-    >>> @openapi_spec({'a': 'b'})
+    >>> @openapi_spec('/foo', ['GET'], path_spec={'a': 'b'}, method_spec={'c': 'd'})
     ... def foo():
     ...     pass
+
+    >>> foo.path_specs
+    {'/foo': {'a': 'b'}}
+
+    >>> foo.method_specs
+    {('/foo', 'get'): {'c': 'd'}}
+
+    One of path_spec or method_spec maybe unspecified
+
+    >>> @openapi_spec('/foo', ['GET'], method_spec={'c': 'd'})
+    ... def foo():
+    ...     pass
+
+    >>> @openapi_spec('/foo', ['GET'], path_spec={'a': 'b'})
+    ... def foo():
+    ...     pass
+
+    Although a valid, documented endpoint should not allow both to be none, we
+    have to allow this behavior to prevent breakages from all currently
+    undocumented endpoints.
+
+    >>> @openapi_spec('/foo', ['GET'])
+    ... def foo():
+    ...     pass
+
+    Multiple routes can be made for the same function
+
+    >>> @openapi_spec('/foo', ['GET', 'PUT'], path_spec={'a': 'b'}, method_spec={'c': 'd'})
+    ... @openapi_spec('/foo/too', ['GET'], method_spec={'e': 'f'})
+    ... def foo():
+    ...     pass
+
+    >>> foo.path_specs
+    {'/foo': {'a': 'b'}}
+
+    >>> foo.method_specs
+    {('/foo/too', 'get'): {'e': 'f'}, ('/foo', 'get'): {'c': 'd'}, ('/foo', 'put'): {'c': 'd'}}
+
+    Only one route should define the top level spec
+
+    >>> @openapi_spec('/bar', ['PUT'], path_spec={'bad, duplicate': 'path spec'}, method_spec={'e': 'f'})
+    ... @openapi_spec('/bar', ['GET'], path_spec={'a': 'b'}, method_spec={'c': 'd'})
+    ... def bar():
+    ...     pass
+    Traceback (most recent call last):
     ...
-    >>> foo.api_spec == {'a': 'b'}
-    True
+    AssertionError: Only specify path_spec once per route path
+
+    At this point we can only validate duplicate specs if they occur on the same
+    method. Unfortunately, this succeeds:
+
+    >>> @openapi_spec('/bar', ['PUT'], path_spec={'a': 'b'}, method_spec={'c': 'd'})
+    ... def foo():
+    ...     pass
+    >>> @openapi_spec('/bar', ['GET'], path_spec={'e', 'f'}, method_spec={'g': 'h'})
+    ... def bar():
+    ...     pass
     """
 
     def spec_adder(func):
-        func.api_spec = description
+        try:
+            func.path_specs
+        except AttributeError:
+            func.path_specs = {}
+        try:
+            func.method_specs
+        except AttributeError:
+            func.method_specs = {}
+
+        assert path not in func.path_specs, 'Only specify path_spec once per route path'
+        if path_spec:
+            func.path_specs[path] = path_spec
+
+        if method_spec:
+            for method in methods:
+                # OpenAPI routes must be lower case
+                method = method.lower()
+                # No need to worry about duplicate method_specs since Chalice
+                # will complain in that case
+                func.method_specs[path, method] = method_spec
         return func
 
     return spec_adder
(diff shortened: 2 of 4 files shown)


lambdas/service/app.py
```diff
@@ -184,13 +184,22 @@ def openapi():
...
         {
-            'in': 'query',
-            'name': 'foo',
-            'description': 'This is not a real parameter',
-            'required': False,
-            'schema': {'$ref': "#/components/schemas/Empty"}
+            'name': 'keys',
+            'in': 'path',
+            'required': True,
+            'schema': {'type': 'string'}
```
danielsotirhos: This is specifying that the `/health` endpoint has a required path parameter `keys`. This is incorrect, no?
`/health/{keys}` has a required path parameter but `/health` does not.

jessebrennan:
@hannes-ucsc, I used my discretion to redo the branches history. I did this because the new code is so drastically different than what was there before, the commit history would no longer have made sense (especially as this PR now solved what was #1454).

A few interesting points:
* Although in conversation we had agreed that we should have separate documentation for the `/health/` route, this isn't actually possible in practice. To document this path we would need to add a separate `@route` decorator, but Chalice doesn't allow trailing slashes on routes:
  ```
    File "/home/jesse/gi/azul/.venv/lib/python3.6/site-packages/chalice/deploy/deployer.py", line 371,   in _validate_config
      validate_configuration(config)
    File "/home/jesse/gi/azul/.venv/lib/python3.6/site-packages/chalice/deploy/validate.py", line 41, in validate_configuration
      validate_routes(routes)
    File "/home/jesse/gi/azul/.venv/lib/python3.6/site-packages/chalice/deploy/validate.py", line 107, in validate_routes
      % route_name)
  ValueError: Route cannot end with a trailing slash: /health/
  ```
  Instead I included this information in the `/health/{keys}` path, but then Swagger doesn't allow you to then exercise the `/health/` path through the 'try it out functionality'
required-health
  Unfortunately this is a limitation of having the API this way.
* When reviewing the code you'll probably notice that there is some spec that's shared between health endpoints. This is unavoidable given the limitations of OpenAPI. See the commit message for more details.

src/azul/openapi.py
```diff
@@ -2,23 +2,65 @@
...
 from azul.deployment import aws
-from azul.types import JSON
+from azul.types import (
+    JSON,
+    MutableJSON,
+)
 
 
-def openapi_spec(description: JSON):
+def openapi_spec(path: str, methods: List[str], path_spec: MutableJSON, method_spec: MutableJSON):
```
hannes-ucsc: Shouldn't `path_spec` be optional?

src/azul/openapi.py
```diff
@@ -2,23 +2,65 @@
...
+    >>> foo.method_spec
+    {('/foo/too', 'get'): {'e': 'f'}, ('/foo', 'get'): {'c': 'd'}, ('/foo', 'put'): {'c': 'd'}}
+
+    Only one route should define the top level spec
+
+    >>> @openapi_spec('/bar', ['PUT'], {'bad, duplicate': 'method spec'}, {'e': 'f'})
+    ... @openapi_spec('/bar', ['GET'], {'a': 'b'}, {'c': 'd'})
+    ... def bar():
+    ...     pass
+    Traceback (most recent call last):
```
hannes-ucsc: What about 

```
@openapi_spec('/bar', ['PUT'], {…}, {…})
def foo():
    pass
@openapi_spec('/bar', ['GET'], {…}, {…})
def bar():
    pass
```

Should that not raise an exception, too?

lambdas/service/app.py
```diff
@@ -184,20 +184,40 @@ def openapi():
...
     'parameters': [
         {
-            'in': 'query',
-            'name': 'foo',
-            'description': 'This is not a real parameter',
-            'required': False,
-            'schema': {'$ref': "#/components/schemas/Empty"}
+            'name': 'keys',
+            'in': 'path',
+            'required': True,
```
hannes-ucsc: So despite being required, it's still allowed to be empty? Does `GET /health/` still work?

src/azul/openapi.py
hannes-ucsc: In fact, I think both need to be optional. We should force users to pass these as keyword arguments. We should programmatically enforce that at least one of the two is passed.

lambdas/service/app.py
```diff
@@ -184,20 +184,40 @@ def openapi():
...
     ],
-    'tags': ['Health']
+}, method_spec={
+    **health_spec,
+    'summary': 'List health information for a specific key',
+    'description': 'List health information for a specific key, or all keys if `/health/` is used.'
 })
-@app.route('/health/{keys}', methods=['GET'], cors=True)
-def health(keys: Optional[str] = None):
+def health_keys(keys: Optional[str] = None):
```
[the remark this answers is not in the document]
jessebrennan: Woops! Thanks for catching this.

src/azul/openapi.py
jessebrennan: This is an important error case to check for, but it can't be caught here. It's needs to be tested for in `join_specs`.

src/azul/openapi.py
jessebrennan: I added this and tests.

jessebrennan: But actually I have to remove this behaviour for now. If we prevent both from being None, then any existing endpoints without documentation will cause this assertion to fail. Once all endpoints are documented, we can consider adding this back in.

lambdas/service/app.py
jessebrennan: Path parameters are always required https://swagger.io/docs/specification/describing-parameters/#path-parameters. You can still make the request, just not through the Swagger page.

src/azul/openapi.py
```diff
@@ -37,32 +109,52 @@ def annotated_specs(gateway_id, app, toplevel_spec) -> JSON:
...
+    path_specs = {}
+    method_specs = {}
+    for entries in app.routes.values():
         func = next(iter(entries.values())).view_function
-        if hasattr(func, 'api_spec'):
-            docs[route] = func.api_spec
-    return docs
+        if hasattr(func, 'path_specs'):
+            for path, spec in func.path_specs.items():
+                if path in path_specs:
```
hannes-ucsc: I noticed you do this here and there. It's not explicitly specified in the contribution guide but we follow the commonly accepted [EAFP](https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python) idiom. If you feel it should be mentioned, could you add a section to the guidelines? Either way, though, this should use EAFP.

src/azul/openapi.py
```diff
@@ -37,32 +109,52 @@ def annotated_specs(gateway_id, app, toplevel_spec) -> JSON:
...
+
+    for path, path_spec in path_specs.items():
+        # This may override entries from API Gateway with the same path
+        specs['paths'][path] = path_spec
+
+    for (path, method), method_spec in method_specs.items():
+        # This may override duplicate specs from path_specs or API Gateway
+        specs['paths'][path][method] = method_spec
+
+    return specs
```
hannes-ucsc: Looks like this code isn't covered by any tests: https://codecov.io/gh/DataBiosphere/azul/pull/1449/diff#D3-127.

jessebrennan: It is covered by the integration test. If you want better coverage, I can add more tests.

src/azul/openapi.py
jessebrennan: Sure. I find LBYL to be more legible (and often less code), but I understand the strengths of EAFP, especially with regard to race conditions. I'm happy to switch it over.

src/azul/openapi.py
hannes-ucsc: *Some* of it is covered by IT. The error case isn't. Correct me if I am wrong but during my penultimate review the two-handlers-same-path case was not tested for and NOT handled. In response to that review you added a check for that to the code under test but the code is not present. So yes, I would like *unit* test coverage for at least the error case but ideally all of `get_app_specs`. 

Re `join_specs` not being covered by IT: it looks like it would be easy to do with a few doctests.

For future reference: when I say "coverage" or "covered" without qualification, I'm referring to the unit test coverage as measured by Coveralls and Codecov.

hannes-ucsc: Sorry for the multiple edits. Done now.

jessebrennan: Ok. I agree and I'll add this.

src/azul/openapi.py
hannes-ucsc: Despite EAFP being more verbose, it IS easier to reason about because it doesn't do everything twice.

src/azul/openapi.py
```diff
@@ -125,25 +142,70 @@ def clean_specs(specs):
...
 def get_app_specs(app):
+    """
+    Extract OpenAPI specs from a Chalice app object
+    """
     path_specs = {}
     method_specs = {}
     for entries in app.routes.values():
         func = next(iter(entries.values())).view_function
-        if hasattr(func, 'path_specs'):
+        try:
```
hannes-ucsc: Each `.` in the body of this try could potentially raise an AttributeError. The body of the `try` should be minimal (one `.` or one use of `[]` for dicts) for so as to not mask/conflate other issues.

Also note that it IS ok to use `dict.get()` and `getattr()` with a default since they use EAFP under the hood.

This might seem pedantic, but this is worse now than the previous version. I try to stick with EAFP as habit so I don't need to think about when it does matter (performance-critical or concurrent code) and when it doesn't.

src/azul/openapi.py
```diff
@@ -125,25 +142,70 @@ def clean_specs(specs):
...
             for path, spec in func.path_specs.items():
                 if path in path_specs:
                     assert spec == path_specs[path], f'Path spec for {spec} already exists'
-            path_specs.update(func.path_specs)
-        else:
+        except AttributeError:
             # TODO (jesse): warn if route is undocumented
```
hannes-ucsc: Is this hard to do? When do you plan to do this?

jessebrennan: This can't be done until all of the routes are documented.

hannes-ucsc: You can't *warn* unless all routes are documented?

src/azul/openapi.py
jessebrennan: I agree this is worse. Good point about `getattr()`. I've switched to that since I think it is the most readable, yet still EAFP.

src/azul/openapi.py
jessebrennan: I changed "warn" to "raise", because that was what I intended, and it cannot be done until all of the routes are documented.

jessebrennan: I don't think we want to _actually_ make a warning log here because then it will show up in Airbrake, etc. which I believe we don't want.

lambdas/service/app.py
hannes-ucsc: Old change that crept back in.

scripts/envhook.py
```diff
@@ -1,17 +1,19 @@
 #! /usr/bin/env python3
 
 import errno
+from importlib.abc import (
```
hannes-ucsc: Why is this change in this PR?

jessebrennan: I was experimenting with PyCharm's changelists, and had your envhook fix saved separately in order to run doctests. When I was rebasing I forgot to "shelve" the envhook fix so it got roped in.

jessebrennan:
@hannes-ucsc proposed changing the implementation to store the spec with the Chalice app object. This is a better design than the current solution, but for expediency, and since this current solution is basically ready to merge, I think this re-design should be addressed later, in https://github.com/DataBiosphere/azul/issues/1485.

jessebrennan:
@hannes-ucsc I see your point about the type signatures. I'll pay more attention next time when I compose them. I added (and approved!) your changes into a new fixup commit and rebased.

Issue #1444, opened 2019-11-13
Unit tests, and potentially service, produce too much log output


hannes-ucsc:
Regression from https://github.com/DataBiosphere/azul/issues/1413, I think most of it is due to the logging of ES queries in the service.

Before: 2.6k lines, https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/4717/raw
After: 30k lines, https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/4737/raw

Step one is to evaluate how much eliminating the ES query logging would save in terms of test output volume. We may want to make ES query logging conditional on `AZUL_DEBUG == 2`. I'm worried that the query logging not only affects the test output volume but also the service runtime performance and Kibana.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-886)
┆Project Name: azul
┆Issue Number: AZUL-886

hannes-ucsc:
Also, I am linking to Gitlab builds but this should also be observable on Travis.

achave11:
ES query log for `AZUL_DEBUG=2` [test](https://api.travis-ci.org/v3/job/611707896/log.txt)  is 14893  lines.

achave11:
Setting `AZUL_DEBUG=0` for `sandbox` and Travis [test](https://api.travis-ci.org/v3/job/611721262/log.txt) is 3234 lines.

Note that there is not a significant reduction in run time for the Travis test.

PR #1451, opened 2019-11-14
ES query logging active only when AZUL_DEBUG set to 2 (#1444)

achave11:


deployments/sandbox/environment
```diff
@@ -34,4 +34,4 @@ _set AZUL_ES_DOMAIN "azul-index-dev"
 # Ensure that monitoring and pushing to Grafana is enabled
 _set AZUL_ENABLE_MONITORING 1
 
-_set AZUL_DEBUG 1
+_set AZUL_DEBUG 0
```
hannes-ucsc: We don't want to disable AZUL_DEBUG for the sandbox as a whole. We want to disable it for Gitlab builds, independently of whether they run in the sandbox, dev or any other deployment.

.travis.yml
```diff
@@ -13,6 +13,8 @@ python:
 before_install:
   - (cd deployments && ln -snf dev .active)
   - source environment
+  # This overrides debug mode in order to reduce test output volume
```
hannes-ucsc:
```suggestion
```

.travis.yml
```diff
@@ -13,6 +13,8 @@ python:
...
+  - export AZUL_DEBUG=0
```
hannes-ucsc:
```suggestion
  - export AZUL_DEBUG=0  # reduce test output volume
```

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -439,7 +439,7 @@ def transform_summary(self,
         es_response = es_search.execute(ignore_cache=True)
         self._translate_response_aggs(es_response)
         final_response = SummaryResponse(es_response.to_dict())
-        if logger.isEnabledFor(logging.DEBUG):
+        if logger.isEnabledFor(logging.DEBUG and config.debug == 2):
```
```suggestion
        if logger.isEnabledFor(logging.DEBUG) and config.debug == 2:
```

Do you test you changes?

src/azul/service/responseobjects/elastic_request_builder.py
```diff
@@ -439,7 +439,7 @@ def transform_summary(self,
         es_response = es_search.execute(ignore_cache=True)
         self._translate_response_aggs(es_response)
         final_response = SummaryResponse(es_response.to_dict())
-        if logger.isEnabledFor(logging.DEBUG):
+        if logger.isEnabledFor(logging.DEBUG) and config.debug == 2:
```
```suggestion
        if config.debug == 2 and logger.isEnabledFor(logging.DEBUG):
```

is more natural because the simplest, fastest check should short-circuit other, more complicated checks. Ordering checks this way should be second nature to you, even if the performance difference is insignificant.

I would also like an answer to my previous question. Did you test this in your deployment?

achave11: Yes, refer to log_id: 35096512501426893458947336217053887030665928500080279570 
in Kibana.
It illustrates the elasticsearch query when `AZUL_DEBUG=2`
There is no logs of the elasticsearch query when `AZUL_DEBUG=1` as intended.

hannes-ucsc: Looks good.

Issue #1448, opened 2019-11-14
Queue manager should not use "special" way of enumerating queues


hannes-ucsc:
Instead of [listing all queues and filtering by name pattern](https://github.com/DataBiosphere/azul/blob/911a8f2c7adb8d7c7201b2fa9159b72f8de29c6e/src/azul/queues.py#L139), `queues.py` should determine the candidate queue names from `config.all_queue_names` and gracefully ignore queues that don't exist.

All clients of the queue manger (especially `reindex.py` for example) should be checked if they can benefit from the refactoring.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-890)
┆containerName: azul
┆Issue Number: AZUL-890

danielsotirhos:
Resolved in PR https://github.com/DataBiosphere/azul/pull/1462 by commit https://github.com/DataBiosphere/azul/pull/1462/commits/74fcd4219af822e5c05777d1bd4c2ab769136934

achave11:
@hannes-ucsc it seems like this issue was resolved by @danielsotirhos PR, I confirmed that queues are fetched from SQS by name, compared to the previous implementation of listing all available queues from SQS and parsing out the ones not associated with Azul. I am just wondering if there is alternative work required to close this ticket, perhaps something not specified in the description?

hannes-ucsc:
You are right, @achave11. 

@danielsotirhos to demo.

Issue #1425, opened 2019-11-05
`make reindex` should block until reindexing is done


hannes-ucsc:
… by repeatedly monitoring the queue lengths.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-874)
┆Project Name: azul
┆Issue Number: AZUL-874

PR #1462, opened 2019-11-19
Block 'make reindex' until reindex complete (#1425)

danielsotirhos:


scripts/reindex.py
```diff
@@ -116,6 +122,21 @@ def main(argv: List[str]):
             azul_client.remote_reindex(args.partition_prefix_length)
         else:
             azul_client.reindex()
+        if not args.nowait:
+            logger.info('Waiting for queues to empty ... ')
```
hannes-ucsc: This was copied from the IT. I'll give you one guess as to what I might object to here.

scripts/reindex.py
```diff
@@ -70,6 +71,10 @@ def __init__(self, prog) -> None:
                     default=False,
                     action='store_true',
                     help='Purge the queues before taking any action on the index.')
+parser.add_argument('--nowait', '--no-wait',
```
hannes-ucsc: Use `default=True, dest='wait', action='store_false'` to prevent the negation from leaking.

scripts/reindex.py
```diff
@@ -70,6 +71,10 @@ def __init__(self, prog) -> None:
...
+                    default=False,
+                    action='store_true',
+                    help='Don\'t wait for queues to empty before exiting script.')
```
hannes-ucsc:
```suggestion
                    help="Don't wait for queues to empty before exiting script.")
```

This is a PEP-8 recommendation. I'm getting tired reminding you of this. So even though it should be implied, please add a separate commit in which you amend `CONTRIBUTING.rst` with a section about not using backslash to escape single (double) quotes in single (double) quoted strings. Link that section to specific PEP8 recommendation about that.

scripts/reindex.py
```diff
@@ -93,17 +98,18 @@ def main(argv: List[str]):
...
                              num_workers=args.num_workers,
                              dryrun=args.dryrun)
-    if args.purge:
+    if args.purge or args.index:
         queue_manager = Queues()
         queues = dict(queue_manager.azul_queues())
         # FIXME: eliminate need for filtering (https://github.com/DataBiosphere/azul/issues/1448)
         queues = {k: v for k, v in queues.items() if k in config.work_queue_names}
+    else:
+        queue_manager, queues = None, None
```
hannes-ucsc: Looks like this branch is unnecessary.

scripts/reindex.py
danielsotirhos: Refactored duplicated code into a static method in Queues class.

scripts/reindex.py
danielsotirhos: Added contributing guideline regarding quote escaping after guideline about preference for single quoted strings.

danielsotirhos:
Should there be a builtin timeout exit if the queues don't empty after x minutes? 30min? 60min?

src/azul/queues.py
```diff
@@ -148,6 +148,20 @@ def _is_azul_queue(self, queue_name) -> bool:
         return any(config.unqualified_resource_name_or_none(queue_name, suffix)[1] == config.deployment_stage
                    for suffix in ('', '.fifo'))
 
+    @staticmethod
```
hannes-ucsc:
```suggestion
    @classmethod
```
@noah-aviel-dove is adding a clause explaining this:

https://github.com/DataBiosphere/azul/pull/1428/files#diff-ea1f65a8eaac0cec7ae151dfdbbbc587R352

scripts/reindex.py
```diff
@@ -93,17 +99,16 @@ def main(argv: List[str]):
...
                              dryrun=args.dryrun)
-    if args.purge:
+    if args.purge or args.index:
         queue_manager = Queues()
         queues = dict(queue_manager.azul_queues())
         # FIXME: eliminate need for filtering (https://github.com/DataBiosphere/azul/issues/1448)
         queues = {k: v for k, v in queues.items() if k in config.work_queue_names}
+    if args.purge:
         logger.info('Disabling lambdas ...')
         queue_manager.manage_lambdas(queues, enable=False)
```
hannes-ucsc: PyCharm warns about this line using variables prior to being assigned. Even though we can prove that they will be assigned because `args.purge` implies `args.purge or args.index`, PyCharms static analyzer doesn't catch that. Think about how we can restructure the code so it does not fool PyCharm. Note how the only case in which we don't need the queues is if `args.delete`.

scripts/reindex.py
```diff
@@ -116,6 +121,13 @@ def main(argv: List[str]):
             azul_client.remote_reindex(args.partition_prefix_length)
         else:
             azul_client.reindex()
+        if args.wait:
+            logger.info('Waiting for queues to empty ... ')
+            while True:
+                sleep(5)
+                total_message_count = Queues.count_messages(queues)
```
hannes-ucsc: If you have an instance you should call through the instance because that avoids hard-coding the class-name again. However, see overall comments.

scripts/reindex.py
danielsotirhos: In that case it seems overkill to special case the initiation of `queue_manager` and `queues` based on the script arguments and they might as well always be available in the script much like `azul_client`.

scripts/reindex.py
```diff
@@ -93,29 +98,34 @@ def main(argv: List[str]):
                              prefix=args.prefix,
                              num_workers=args.num_workers,
                              dryrun=args.dryrun)
+    queue_manager = Queues()
+    queues = dict(queue_manager.azul_queues())
+    # FIXME: eliminate need for filtering (https://github.com/DataBiosphere/azul/issues/1448)
+    queues = {k: v for k, v in queues.items() if k in config.work_queue_names}
```
hannes-ucsc: … this section …

scripts/reindex.py
hannes-ucsc: OK

src/azul/queues.py
```diff
@@ -148,6 +149,56 @@ def _is_azul_queue(self, queue_name) -> bool:
...
+    def wait_for_queue_level(cls, queue_names, empty: bool = True, timeout: int = 60):
+        """
+        Wait until the total count of queued messages reaches the desired level
+
+        :param queue_names: The names of the queues to check messages counts in
+        :param empty: If we should wait until the queues are empty. False to wait until not empty.
+        :param timeout: Give up waiting after this many seconds
+        """
+        queues = {queue_name: boto3.resource('sqs').get_queue_by_name(QueueName=queue_name)
+                  for queue_name in queue_names}
```
hannes-ucsc: … and this section could be consolidated. I have the feeling that you will end up with the added Queue methods being instance methods instead of class methods. You *may* have to tackle #1448 as well but I'm not sure.

danielsotirhos: While it wasn't necessary to resolve #1448 here I felt it was related to the work in the previous commit and easy enough so I committed a fix for it in this branch as a separate commit.

Issue #1470, opened 2019-11-21
Running doctests from PyCharm fails with import error


hannes-ucsc:
To reproduce, open `openapi.py`, right-click the `openapi_spec` function and select _Run 'Doctest openapi_spec'_.

```
Testing started at 5:31 PM ...
/home/hannes/workspace/hca/azul/.venv.xps/bin/python /snap/pycharm-professional/171/helpers/pycharm/docrunner.py /home/hannes/workspace/hca/azul/src/azul/openapi.py::::openapi_spec
envhook.py: Setting AWS_DEFAULT_REGION to 'us-east-1'
envhook.py: Setting AWS_PROFILE to 'hca'
envhook.py: Setting AZUL_CART_EXPORT_MAX_BATCH_SIZE to '100'
envhook.py: Setting AZUL_CART_EXPORT_MIN_ACCESS_TOKEN_TTL to 'REDACTED'
envhook.py: Setting AZUL_DEBUG to '1'
envhook.py: Setting AZUL_DEPLOYMENT_STAGE to 'staging'
envhook.py: Setting AZUL_DISABLE_MULTIPART_MANIFESTS to '0'
envhook.py: Setting AZUL_DOMAIN_NAME to 'staging.explore.data.humancellatlas.org'
envhook.py: Setting AZUL_DRS_DOMAIN_NAME to 'drs.staging.data.humancellatlas.org'
envhook.py: Setting AZUL_DSS_ENDPOINT to 'https://dss.staging.data.humancellatlas.org/v1'
envhook.py: Setting AZUL_DSS_WORKERS to '8'
envhook.py: Setting AZUL_ENABLE_MONITORING to '1'
envhook.py: Setting AZUL_ES_DOMAIN to 'azul-index-dev'
envhook.py: Setting AZUL_ES_INSTANCE_COUNT to '2'
envhook.py: Setting AZUL_ES_INSTANCE_TYPE to 'r4.2xlarge.elasticsearch'
envhook.py: Setting AZUL_ES_TIMEOUT to '60'
envhook.py: Setting AZUL_ES_VOLUME_SIZE to '70'
envhook.py: Setting AZUL_FUSILLADE_ENDPOINT to 'https://auth.dev.data.humancellatlas.org'
envhook.py: Setting AZUL_INDEXER_CONCURRENCY to '64'
envhook.py: Setting AZUL_INDEX_PREFIX to 'azul'
[...]

PR #1471, opened 2019-11-21
Sanitize sys.path in envhook.py; defeat PyCharm docrunner.py (#1470)

hannes-ucsc:


hannes-ucsc:
> One reason not to use .index() is that it only give the index of the first occurrence if there are multiple (though I'm not sure that's even possible in this case).

I think it is possible. What do you propose we do?

jessebrennan:
A couple solutions:
```
from collections import OrderedDict
# Remove duplicates, preserving the order of the first occurrence of each element
sys_path = OrderedDict.fromkeys(sys.path)
sys_path.pop(self.bad_path)
sys_path = list(sys_path)
```
This solution is nice because it's O(n) where n is the length of sys.path, regardless of the number of duplicates. It has the consequence of removing all duplicates in sys.path, but that's never going to cause any issues anyway.

```
while self.bad_path in sys_path:
    del sys_path[sys_path.index(self.bad_path)]
```
This solution is O(n*m) where m in the number of duplicates of `bad_path`. On the other hand, I think duplicates are rare, and this solution is simple and perhaps more intuitive than the first.

Issue #1190, opened 2019-08-03
MetadataGenerator is slow


hannes-ucsc:
I profiled the indexer and the amount of time spent in _flatten and _set_value is nuts. Especially _set_value is dumb. Low hanging fruit for optimization. Hint: split—append—set-ify—sort—join is not advisable if the method is called tens of thousands of times for a bundle. Instead maintain a set and convert back to string in a second pass at the end. Also, `if k in d: x = d[k] else: d[k] = …` is dumb, too.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-740)
┆Project Name: azul
┆Issue Number: AZUL-740

PR #1472, opened 2019-11-21
Optimize flattening operation in metadata generator (#1190)

hannes-ucsc:


hannes-ucsc:
On imaging bundles, the metadata_generator is now five times faster.

Profile of `test_imaging_bundle` on base commit 5e3226965f54e0fa7736520c5b9ed69031012bc8:

image

Same test on the tip of this branch: 31b4900ab0c08b7f7eb33a1f0df853566fe1239a

image

PR #1477, opened 2019-11-22
Refactor services

hannes-ucsc:


hannes-ucsc:
Why would it be redundant?

Issue #1474, opened 2019-11-22
Gitlab job for IT is killed before queue waiter times out


hannes-ucsc:
https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/5149

I would like the waiter to timeout well before the job is killed. Killing the job causes AZUL_TEST_MODE to never be reset.

Keep in mind that there are multiple occasions during the IT that it waits for the queues to empty. I think two. Once for the additions and once for the deletions.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-904)
┆Project Name: azul
┆Issue Number: AZUL-904

PR #1478, opened 2019-11-22
Timeout queue waiter before Gitlab times out (#1474)

danielsotirhos:


src/azul/queues.py
```diff
@@ -178,7 +178,10 @@ def wait_for_queue_level(cls, queue_names, empty: bool = True, timeout: int = 60
         wait_start_time = time.time()
         queue_size_history = deque(maxlen=10)
 
-        logger.info('Waiting up to %s seconds for %s queues to empty ...', timeout, len(queues))
+        logger.info('Waiting up to %s seconds for %s queues to %s ...',
+                    timeout,
```
hannes-ucsc:
```suggestion
                    timeout, len(queues), 'empty' if empty else 'not be empty')
```

For log statements, I would actually prefer to exclude the first argument from the wrap all or none rule:

https://github.com/DataBiosphere/azul/pull/1479/files

danielsotirhos:
Adjusted IT timeout for queues to empty to be 4 seconds per bundle (from the logs it takes more like 1-2 seconds per bundle), plus 50 seconds for the zero to propagate through the deque in `wait_for_queue_level()`. For 64 bundles this gives us a timeout just over 5 min.  [(4 sec * 64) + 50 sec = 306 seconds]

hannes-ucsc:
This didn't `rebase --autosquash` without conflict.

Issue #1256, opened 2019-09-05
Add link to .pagination in service response


hannes-ucsc:
For easier programmatic and interactive consumption of the paginated responses the service should return a `pagination.next` link ~instead of~ in addition to the current `pagination.searchAfter` and `pagination.searchAfterUuid` response properties. Ditto with `previous` instead of `searchBefore` and `searchBeforeUuid`. The link property should be a full URL to the next/previous page of items.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-779)
┆containerName: azul
┆Issue Number: AZUL-779
┆Epic: User web service API

PR #1494, opened 2019-12-03
Add next and previous link in pagination for service response (#1256)

achave11:


test/service/test_pagination.py
```diff
@@ -34,23 +35,33 @@ def assert_page1_correct(self, json_response):
...
+        if 'next' in json_response['pagination']:
+            search_after, search_after_uid = self._get_paging_params_from_url(json_response, 'next')
+            self.assertEqual(search_after,
                              json_response['hits'][num_hits - 1]['entryId'],
-                             "search_after not set to last returned document on first page")
-            self.assertIsNotNone(json_response['pagination']['search_after_uid'])
+                             "next not set to last returned document on first page")
+            self.assertIsNotNone(search_after_uid)
         else:
             self.fail("search_after not set on first page of results")
```
danielsotirhos:
```suggestion
            self.fail("previous not set on first page of results")
```

test/service/test_pagination.py
```diff
@@ -34,23 +35,33 @@ def assert_page1_correct(self, json_response):
...
-        if 'search_before' in json_response['pagination']:
-            self.assertIsNone(json_response['pagination']['search_before'],
-                              "search_before != null on first page of results")
-            self.assertIsNone(json_response['pagination']['search_before_uid'],
-                              "search_before != null on first page of results")
+        if 'previous' in json_response['pagination']:
+            self.assertIsNone(json_response['pagination']['previous'],
+                              "previous != null on first page of results")
         else:
             self.fail("search_before not set to null on first page of results")
```
danielsotirhos:
```suggestion
            self.fail("previous not set to null on first page of results")
```

test/service/test_pagination.py
```diff
@@ -34,23 +35,33 @@ def assert_page1_correct(self, json_response):
...
-            self.assertIsNotNone(json_response['pagination']['search_after_uid'])
+                             "next not set to last returned document on first page")
+            self.assertIsNotNone(search_after_uid)
         else:
             self.fail("search_after not set on first page of results")
 
+    def _get_paging_params_from_url(self, json_response, string):
+        query = urlparse(json_response['pagination'][string]).query
+        try:
+            _, search, search_uid = query.split('&')
```
danielsotirhos: Instead of parsing the query string yourself which could break if the ordering of query string params is ever different, have you considered using `parse_qs`?
```
parsed_query = parse_qs(query)
search = parsed_query.get('search_after', [None])[0]
search_uid = parsed_query.get('search_after_uid', [None])[0]
```

danielsotirhos: It would make the duality of the function (working for both search_after and search_before) a little more tricky though

test/service/test_pagination.py
```diff
@@ -150,18 +160,20 @@ def test_search_after_last_page(self):
...
-            self.assertEqual(json_response['pagination']['search_before'], search_after_lrfp,
+        if 'previous' in json_response['pagination']:
+            search_before, search_before_uid = self._get_paging_params_from_url(json_response, 'previous')
+            self.assertEqual(search_before, search_after_lrfp,
                              "search_before on last page is not set correctly")
-            self.assertIsNotNone(json_response['pagination']['search_before_uid'],
+            self.assertIsNotNone(search_before_uid,
                                  "search_before_uid on last page is not set")
         else:
             self.fail("search_before not set on last page of results")
```
danielsotirhos:
```suggestion
            self.fail("previous not set on last page of results")
```

test/service/test_pagination.py
```diff
@@ -150,18 +160,20 @@ def test_search_after_last_page(self):
...
-            self.assertIsNone(json_response['pagination']['search_after'],
+        if 'next' in json_response['pagination']:
+            search_after, search_after_uid = self._get_paging_params_from_url(json_response, 'next')
+            self.assertIsNone(search_after,
                               "search_after is not null on last page")
-            self.assertIsNone(json_response['pagination']['search_after_uid'],
+            self.assertIsNone(search_after_uid,
                               "search_after_uid is not null on last page")
         else:
             self.fail("search_after not set to [] on first page of results")
```
danielsotirhos:
```suggestion
            self.fail("next not set to [] on first page of results")
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,35 +328,45 @@ def _generate_paging_dict(self, es_response, pagination):
...
         es_hits = es_response['hits']['hits']
         count = len(es_hits)
+        self_url = pagination['self_url'] + f"?size={pagination['size']}"
         if 'search_before' in pagination:
             # hits are reverse sorted
             if count > pagination['size']:
                 # There is an extra hit, indicating a previous page.
                 count = count - 1
                 search_before = es_hits[count - 1]['sort']
+                params = f'&search_before={search_before[0]}&search_before_uid={search_before[1].replace("#", "%23")}'
```
hannes-ucsc: Don't manually encode. There are built-ins for that:

https://docs.python.org/3.6/library/urllib.parse.html#urllib.parse.urlencode

src/azul/service/elasticsearch_service.py
```diff
@@ -328,26 +329,49 @@ def _generate_paging_dict(self, es_response, pagination):
...
         if 'search_before' in pagination:
             # hits are reverse sorted
             if count > pagination['size']:
                 # There is an extra hit, indicating a previous page.
                 count = count - 1
                 search_before = es_hits[count - 1]['sort']
+                query = dict(search_before=search_before[0],
+                             search_before_uid=search_before[1])
+                params = f'&{urlencode(query)}'
+                previous = self_url + params
```
danielsotirhos:
```suggestion
                params = dict(search_before=search_before[0],
                              search_before_uid=search_before[1])
                previous = self_url + '&' + urlencode(params)
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,26 +329,49 @@ def _generate_paging_dict(self, es_response, pagination):
...
             else:
                 # No previous page
+                previous = None
                 search_before = [None, None]
             search_after = es_hits[0]['sort']
+            query = dict(search_after=search_after[0],
+                         search_after_uid=search_after[1])
+            params = f'&{urlencode(query)}'
+            next_ = self_url + params
```
danielsotirhos:
```suggestion
            params = dict(search_after=search_after[0],
                          search_after_uid=search_after[1])
            next_ = self_url + '&' + urlencode(params)
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,26 +329,49 @@ def _generate_paging_dict(self, es_response, pagination):
...
         else:
             # hits are normal sorted
             if count > pagination['size']:
                 # There is an extra hit, indicating a next page.
                 count = count - 1
                 search_after = es_hits[count - 1]['sort']
+                query = dict(search_after=search_after[0],
+                             search_after_uid=search_after[1])
+                params = f'&{urlencode(query)}'
+                next_ = self_url + params
```
danielsotirhos:
```suggestion
                params = dict(search_after=search_after[0],
                              search_after_uid=search_after[1])
                next_ = self_url + '&' + urlencode(params)
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,26 +329,49 @@ def _generate_paging_dict(self, es_response, pagination):
...
                 # No next page
+                next_ = None
                 search_after = [None, None]
-            search_before = es_hits[0]['sort'] if 'search_after' in pagination else [None, None]
+            if 'search_after' in pagination:
+                search_before = es_hits[0]['sort']
+                query = dict(search_before=search_before[0],
+                             search_before_uid=search_before[1])
+                params = f'&{urlencode(query)}'
+                previous = self_url + params
```
danielsotirhos:
```suggestion
                params = dict(search_before=search_before[0],
                              search_before_uid=search_before[1])
                previous = self_url + '&' + urlencode(params)
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,26 +329,45 @@ def _generate_paging_dict(self, es_response, pagination):
         # ...else use search_after/search_before pagination
         es_hits = es_response['hits']['hits']
         count = len(es_hits)
+        self_url = pagination['self_url'] + f'?size={pagination["size"]}'
```
hannes-ucsc: Too pedestrian. Yes, size should be a string but you can't bet your life on it. It would be safer and more elegant to refactor this so there is exactly one place where ALL of the parameters are encoded and appended to the URL. Hint: Make `next_` and `previous` a dict compose the actual links as late as possible.

src/azul/service/hca_response_v5.py
```diff
@@ -65,6 +65,8 @@ class PaginationObj(JsonObject):
     search_after_uid = StringProperty()
     search_before = DefaultProperty()
     search_before_uid = StringProperty()
+    next = DefaultProperty()
```
hannes-ucsc: Sure about the type? Why not StringProperty?

test/service/test_pagination.py
```diff
@@ -34,22 +39,33 @@ def assert_page1_correct(self, json_response):
         :param json_response: A JSON response dictionary
         :return:
         """
-        if 'search_before' in json_response['pagination']:
```
hannes-ucsc: I'd feel more confident in the fact that your changes don't break anything if the test changes weren't so invasive. For now we should stick with `search_before` and `search_after` being the pivots, not `previous` and `next`. After all you're adding response elements so the test code should only have additions, no deletions, or minimal deletions.

src/azul/service/elasticsearch_service.py
```diff
@@ -328,27 +329,51 @@ def _generate_paging_dict(self, es_response, pagination):
...
+        next_ = {}
+        previous = {}
         if 'search_before' in pagination:
             # hits are reverse sorted
             if count > pagination['size']:
                 # There is an extra hit, indicating a previous page.
                 count = count - 1
                 search_before = es_hits[count - 1]['sort']
+                previous['search_before'] = search_before[0]
+                previous['search_before_uid'] = search_before[1]
```
GPelayo: Merge the two `previous` assignment expressions to this:
```
previous['search_before'], previous['search_before_uid'] = search_before
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,27 +329,51 @@ def _generate_paging_dict(self, es_response, pagination):
...
             else:
                 # No previous page
                 search_before = [None, None]
             search_after = es_hits[0]['sort']
+            next_['search_after'] = search_after[0]
+            next_['search_after_uid'] = search_after[1]
```
GPelayo: Merge the two `next_` assignment expressions to this:
```
next_['search_after'], next_['search_after_uid'] = search_after
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,27 +329,51 @@ def _generate_paging_dict(self, es_response, pagination):
...
+            pagination.pop('search_before', None)
+            pagination.pop('search_before_uid', None)
         else:
             # hits are normal sorted
             if count > pagination['size']:
                 # There is an extra hit, indicating a next page.
                 count = count - 1
                 search_after = es_hits[count - 1]['sort']
+                next_['search_after'] = search_after[0]
+                next_['search_after_uid'] = search_after[1]
```
GPelayo: Merge the two `next_` assignment expressions to this:
```
next_['search_after'], next_['search_after_uid'] = search_after
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,27 +329,51 @@ def _generate_paging_dict(self, es_response, pagination):
...
             else:
                 # No next page
                 search_after = [None, None]
-            search_before = es_hits[0]['sort'] if 'search_after' in pagination else [None, None]
-
+            if 'search_after' in pagination:
+                search_before = es_hits[0]['sort']
+                previous['search_before'] = search_before[0]
+                previous['search_before_uid'] = search_before[1]
```
GPelayo: Merge the two `previous` assignment expressions to this:
```
previous['search_before'], previous['search_before_uid'] = search_before
```

src/azul/service/hca_response_v5.py
```diff
@@ -65,6 +65,8 @@ class PaginationObj(JsonObject):
     search_after_uid = StringProperty()
     search_before = DefaultProperty()
     search_before_uid = StringProperty()
+    next = StringProperty()
```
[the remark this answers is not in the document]
achave11: @GPelayo this suggestion I will not follow through with because this might affect how the `pagination` response object looks to the user. Which I believe we want to keep it as user friendly as possible, I could be wrong.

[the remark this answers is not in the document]
GPelayo: I see. We can use the `name` argument of `StringProperty`, so the property can be user-friendly without being a reserved word.

```suggestion
    next_ = StringProperty(name='next')
```

src/azul/service/elasticsearch_service.py
```diff
@@ -328,27 +329,47 @@ def _generate_paging_dict(self, es_response, pagination):
...
                 # There is an extra hit, indicating a previous page.
                 count = count - 1
                 search_before = es_hits[count - 1]['sort']
+                previous['search_before'], previous['search_before_uid'] = search_before
             else:
                 # No previous page
                 search_before = [None, None]
             search_after = es_hits[0]['sort']
+            next_['search_after'], next_['search_after_uid'] = search_after
+            pagination.pop('search_before', None)
```
jessebrennan: Why do you have to pop all this stuff at different times from `pagination`? Would it make more sense to build a new dictionary for the `next_` and `previous` URLs? Maybe even making a function that generates the next / previous URLs would separate out some of the logic and make this cleaner.

src/azul/service/elasticsearch_service.py
```diff
@@ -360,16 +358,10 @@ def _generate_paging_dict(self, es_response, pagination, sort_facet):
...
         sort = pagination.pop('sort')
         self_url = pagination.pop('self_url')
-        next_ = self_url + '?' + urlencode(dict(sort=sort_facet,
-                                                **pagination,
-                                                **next_)) if next_ else None
-        previous = self_url + '?' + urlencode(dict(sort=sort_facet,
-                                                   **pagination,
-                                                   **previous)) if previous else None
+        next_ = self.page_link(next_, pagination, self_url, sort_facet)
+        previous = self.page_link(previous, pagination, self_url, sort_facet)
```
jessebrennan: This is definitely an improvement.

I think you could take it a bit further by changing `page_link()`'s signature to look like

```
def page_links(self, search_before, search_after, pagination, sort_facet) -> Tuple[str, str]:
```
and have it return `next_` and `prev` both. If you do this, then there's no need to assign any `next_` nor `prev` variables earlier, but you would need to check if `search_after` or `search_before` were none when building the link.

Just an idea to consider, but the code looks good overall.

src/azul/service/elasticsearch_service.py
```diff
@@ -379,11 +373,19 @@ def _generate_paging_dict(self, es_response, pagination, sort_facet):
 
         return page_field
 
-    def page_link(self, search_params, pagination, self_url, sort_facet):
-        return self_url + '?' + urlencode(dict(sort=sort_facet,
+    def page_link(self, pagination, self_url, sort_facet, search_after, search_before) -> Tuple[str, str]:
```
jessebrennan: I think this is great.

You could probably avoid the duplication by adding a loop within `page_link()`. If you keep the return value as a tuple and remove the duplication with a loop, then the function should probably be called `page_links` instead.

Another option would be to make `search_after` and `search_before` optional arguments, assert exactly one of them is set, then return either next or previous depending on which one it is. I know this is sort of going back on myself, but I like this option better.

src/azul/service/elasticsearch_service.py
```diff
@@ -329,39 +330,32 @@ def _generate_paging_dict(self, es_response, pagination, sort_facet):
...
             else:
                 # No next page
                 search_after = [None, None]
             if 'search_after' in pagination:
                 search_before = es_hits[0]['sort']
-                previous['search_before'], previous['search_before_uid'] = search_before
             else:
                 search_before = [None, None]
         sort = pagination.pop('sort')
         self_url = pagination.pop('self_url')
```
jessebrennan: If you remove the pops here, then you don't have to pass in `self_url` into page_link as a separate arg. You can just extract it from pagination. You don't have to pop sort either I think, you can just extract what you need when you compose the `page_field` dict.

src/azul/service/elasticsearch_service.py
```diff
@@ -357,13 +368,21 @@ def _generate_paging_dict(self, es_response, pagination):
...
+            'next': next_,
+            'previous': previous,
             'pages': pages,
             'sort': pagination['sort'],
             'order': pagination['order']
         }
 
         return page_field
 
+    def page_link(self, pagination, sort_facet, **kwargs) -> str:
```
hannes-ucsc: If you make this a nested function in _generate_paging_dict you can eliminate passing the `pagination` and `sort_facet`.

src/azul/service/elasticsearch_service.py
```diff
@@ -502,8 +521,8 @@ def transform_request(self,
             hits = self.plugin.translate_fields(hits, forward=False)
 
             facets = es_response_dict['aggregations'] if 'aggregations' in es_response_dict else {}
-
-            paging = self._generate_paging_dict(es_response_dict, pagination)
+            sort_facet = inverse_translation[pagination['sort']]
```
hannes-ucsc: I'm a bit worried how the `sort` field flows through this. It seem that it is reverse translated here just so it can be injected into the link. Four lines down it's reverse translated a second time. This feels convoluted but I don't have a clear vision of how it can be cleaned up. I'd be happy to join a paired programming session to figure it out.

hannes-ucsc:
Rather stale base. Remember to rebase daily.

Issue #1490, opened 2019-12-03
Blocking reindex exits prematurely


hannes-ucsc:
The queue waiting functionality added for #1425 in PR #1462 exits prematurely:

https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/5284

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-916)
┆Project Name: azul
┆Issue Number: AZUL-916

PR #1496, opened 2019-12-04
Fix: Blocking reindex exits prematurely (#1490)

danielsotirhos:


danielsotirhos:
Default timeout for `wait_for_queue_level()` has been changed from 60 to 120 seconds to account for the extra 50 seconds now needed when waiting for queues to empty.

PR #1508, opened 2020-01-13
Fix: KeyError: 'dev' on `/integrations` (load-project#57)

achave11:


lambdas/service/app.py
```diff
@@ -481,7 +481,7 @@ def _fetch_integrations(entity_type: str, integration_type: str, entity_ids: Opt
     results = []
     for portal in portals:
         integrations = [
-            {k: v if k != 'entity_ids' else v[config.dss_deployment_stage] for k, v in integration.items()}
+            {k: v if k != 'entity_ids' else [] for k, v in integration.items()}
```
```suggestion
            {k: v if k != 'entity_ids' else v.get(config.dss_deployment_stage, []) for k, v in integration.items()}
```

achave11: @danielsotirhos Nice, that is much more elegant!

Issue #1509, opened 2020-01-28
Update check_branch.py to handle `hca/*` branches


hannes-ucsc:
```
$ make tag
python /home/hannes/workspace/hca/azul.stable/scripts/check_branch.py
Traceback (most recent call last):
  File "/home/hannes/workspace/hca/azul.stable/scripts/check_branch.py", line 34, in check_branch
    expected_stage = stage_by_branch[branch]
KeyError: 'hca/integration'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/hannes/workspace/hca/azul.stable/scripts/check_branch.py", line 88, in <module>
    main(sys.argv[1:])
  File "/home/hannes/workspace/hca/azul.stable/scripts/check_branch.py", line 81, in main
    check_branch(branch, stage)
  File "/home/hannes/workspace/hca/azul.stable/scripts/check_branch.py", line 37, in check_branch
    raise RuntimeError(f"Non-protected branch '{branch}' can't be deployed to main deployment '{stage}'")
RuntimeError: Non-protected branch 'hca/integration' can't be deployed to main deployment 'integration'
make: *** [common.mk:37: check_branch] Error 1
```

That should not happen. Also, when an `hca/{develop,integration,staging,prod}` branch is checked out, we should ensure that the HCA account is configured (not sure how exactly that would work). Likewise, when `{develop,integration,staging,prod}` is checked out, we should ensure that credentials to the UCSC account is configured.

PR #1516, opened 2020-01-29
Update check_branch.py to handle `hca/*` branches (#1509)

jessebrennan:
This also will ensure that a branch is not deployed unless the expected corresponding AWS account is selected / active.

This commit will also need to be merged into `hca/develop`. @hannes-ucsc, should I make a separate PR for this purpose?

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -8,3 +8,6 @@
 # This overrides one of the default patterns used by the AWS provider for
 # `git secrets. We don't mind exposing account IDs.
 AWS_ACCOUNT_ID=861229788715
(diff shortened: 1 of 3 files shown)


hannes-ucsc:
> should I make a separate PR for this purpose?

Yes, please. Once it shown to work on HCA `dev`, promote normally with merges though the `hca/*` branches, up to and including the `hca/prod` branch. The only steps we'll post-pone for `hca/prod` is  is the creation of the deployment tag, and pushing to the prod instance of Gitlab.

Issue #1446, opened 2019-11-13
Always sort multi-valued fields


hannes-ucsc:
```
$ http 'https://service.explore.data.humancellatlas.org/repository/files?filters=%7B%22projectTitle%22%3A%7B%22is%22%3A%5B%22Single%20cell%20profiling%20of%20human%20induced%20dendritic%20cells%20generated%20by%20direct%20reprogramming%20of%20embryonic%20fibroblasts%22%5D%7D%2C%22fileFormat%22%3A%7B%22is%22%3A%5B%22csv%22%5D%7D%7D&size=15' | jq .hits[].samples[].id
[
  "Specimen1"
]
[
  "Cell_line_2",
  "Specimen1"
]
[
  "Specimen1",
  "Cell_line_1"
]
```

That should be 

```
$ http 'https://service.explore.data.humancellatlas.org/repository/files?filters=%7B%22projectTitle%22%3A%7B%22is%22%3A%5B%22Single%20cell%20profiling%20of%20human%20induced%20dendritic%20cells%20generated%20by%20direct%20reprogramming%20of%20embryonic%20fibroblasts%22%5D%7D%2C%22fileFormat%22%3A%7B%22is%22%3A%5B%22csv%22%5D%7D%7D&size=15' | jq .hits[].samples[].id
[
  "Specimen1"
]
[
  "Cell_line_2",
  "Specimen1"
]
[
  "Cell_line_1",
  "Specimen1"
]
```

And I think this should be applied uniformly for all fields that can potentially contain multiple values.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-888)
┆containerName: azul
┆Issue Number: AZUL-888

PR #1524, opened 2020-01-31
Always sort multi-valued fields (#1446)

danielsotirhos:


src/azul/service/elasticsearch_service.py
```diff
@@ -295,19 +295,20 @@ def _apply_paging(self, es_search, pagination):
 
         _sort = pagination['sort'] + ".keyword"
         _order = pagination['order']
+        _mode = 'min' if _order == 'asc' else 'max'
```
hannes-ucsc: I don't think so. We want to sort by the smallest value in the list in all cases. 

Two documents: `x={"foo": "x", "bar": [1, 4]}` and `y={"foo": "x", "bar": [2,3]}` sorted by field `bar` ascending would give `[x, y]` in that order. With your implementation, the descending sort would give `[x, y]` as well which is broken. The descending sort order should be the inverse of the ascending one.

src/azul/transformer.py
```diff
@@ -454,8 +454,11 @@ def accumulate(self, value) -> bool:
         else:
             return False
 
-    def get(self) -> List[Any]:
-        return list(self.value)
+    def get(self, sort: bool = True) -> List[Any]:
```
hannes-ucsc: Can you explain why this needs to be modal? Why can't it always sort?

src/azul/service/hca_response_v5.py
```diff
@@ -254,26 +241,12 @@ class KeywordSearchResponse(AbstractResponse, EntryFetcher):
...
-                    # Make each dict hashable so we can deduplicate the list
-                    merged_dict[key] = list(map(thaw, set(map(freeze, cleaned_list))))
-                else:
-                    merged_dict[key] = list(set(cleaned_list))
-            elif value is None:
-                merged_dict[key] = []
-        merged_dict[identifier] = dict_id
-        return dict(merged_dict)
+    @classmethod
+    def sorted(cls, value):
```
hannes-ucsc: What kinds of types do you expect? AFAICT, only None or list or set are possible. This method should fail if anything else is passed to it as that constitutes an error. Sounds familiar?

src/azul/service/hca_response_v5.py
```diff
@@ -254,26 +241,12 @@ class KeywordSearchResponse(AbstractResponse, EntryFetcher):
...
-                if len(cleaned_list) > 0 and isinstance(cleaned_list[0], dict):
-                    # Make each dict hashable so we can deduplicate the list
-                    merged_dict[key] = list(map(thaw, set(map(freeze, cleaned_list))))
-                else:
-                    merged_dict[key] = list(set(cleaned_list))
-            elif value is None:
-                merged_dict[key] = []
-        merged_dict[identifier] = dict_id
-        return dict(merged_dict)
+    @classmethod
```
hannes-ucsc: AFAICT, this is only ever called through `self` and doesn't need to be a class method. In fact, since this is called so frequently, I'd consider making this a function in this module and calling it `_sort`.

src/azul/service/hca_response_v5.py
```diff
@@ -285,11 +258,11 @@ def make_protocols(self, entry):
         protocols = []
         for protocol in entry["contents"]["protocols"]:
             translated_process = {
-                "libraryConstructionApproach": protocol.get("library_construction_approach", []),
-                "instrumentManufacturerModel": protocol.get("instrument_manufacturer_model", []),
-                "pairedEnd": protocol.get("paired_end", []),
-                "workflow": protocol.get("workflow", []),
-                "assayType": protocol.get("assay_type", []),
+                "libraryConstructionApproach": self.sorted(protocol.get("library_construction_approach", [])),
```
hannes-ucsc: Why are we sorting the response again? Shouldn't the document fields already be sorted?

src/azul/transformer.py
danielsotirhos: The `get` method of `SetAccumulator` requires a `sort` param because the `get` method of the child class `SetOfDictAccumulator` calls `super().get()` and fails when attempting to produce a sorted list from a set of frozendict.

src/azul/service/hca_response_v5.py
danielsotirhos: Many various value types will pass through this function including `int` (donorCount, totalCells) and `str` (projectTitle, modelOrgan). While the `int` values might always be `int`, the `str` values such as projectTitle and projectShortname will appear as either `str` or a list of `str`

src/azul/transformer.py
hannes-ucsc: This isn't at all obvious and would have deserved an explanation in the form of a comment.

However, I think there are better ways of dealing with this. I am starting to think that we should move the sorting out of the accumulators and sort at the call sites to get(). We can then either have a protocol by which the caller interrogates the accumulator if the accumulated value is sortable or catch an exception that unmistakably indicates the inability to sort it. Or we just make sure that frozendict is sortable by implementing __lt__. If you do this, there maybe some interaction/overlap with `sort_frozen`.

src/azul/service/hca_response_v5.py
hannes-ucsc: Can you demonstrate this to me?

danielsotirhos: Yes, a request to `/repository/projects` results in `make_projects` being sent hits that contain `hit['contents']['projects']['project_title']` values that are `str`,

test/indexer/test_hca_indexer.py
```diff
@@ -768,6 +768,56 @@ def test_organoid_priority(self):
         self.assertEqual(inner_cell_suspensions_in_contributions + inner_cell_suspensions_in_aggregates,
                          inner_cell_suspensions)
 
+    def test_field_values_are_sorted(self):
```
hannes-ucsc: I think we should write a generic sort validator for all documents, both aggregates and contributions, and plug it into every test method such that every index document from every test bundle is checked. Because the setup for these tests is so time-consuming we simply don't have the luxury to accumulate test cases like this.

You already have the beginnings of such a validator in `_verify_sorted_values`. It just needs to be adapted to handle documents rather than service responses but they are not that dissimilar.

test/indexer/__init__.py
```diff
@@ -136,3 +136,31 @@ def mocked_get_bundle(bundle_uuid, bundle_version):
             with patch.object(indexer, '_get_bundle', new=mocked_get_bundle):
                 contributions = indexer.transform(notification, delete=False)
                 return indexer.contribute(contributions)
+
+    def _verify_sorted_lists(self, value, path):
```
hannes-ucsc: Please add type hints

test/indexer/__init__.py
```diff
@@ -136,3 +136,31 @@ def mocked_get_bundle(bundle_uuid, bundle_version):
...
+        """
+        Traverse through 'value' to verify all lists of primitives are sorted
```
hannes-ucsc: Explain what sort of JSON this function accepts (index documents and service responses).

test/indexer/__init__.py
hannes-ucsc: Path should be a tuple and have a default of `()`

test/indexer/test_hca_indexer.py
```diff
@@ -60,11 +60,15 @@ def setUpModule():
 
 class TestHCAIndexer(IndexerTestCase):
 
-    def _get_all_hits(self):
+    def _get_all_hits(self, verify_lists_sorted=True):
```
hannes-ucsc: If `verify_lists_sorted` is never passed as `False` we should eliminate it. I was expecting it to get in the way of some of the tests but I guess I was wrong.

test/indexer/__init__.py
hannes-ucsc:
```suggestion
    def _assert_sorted_leaf_lists(self, value, path):
```

hannes-ucsc: Scratch the default argument for path. I would like the bulk of the body of this method to move into a nested function so the path argument to the method disappears completely and the return value becomes None. The assertion of `> 0` becomes part of the method and can be eliminated from the two call sites. That simplifies the signature and you have to explain less in the docstring.

test/indexer/test_hca_indexer.py
```diff
@@ -60,11 +60,15 @@ def setUpModule():
...
         hits = scan(client=self.es_client,
                     index=','.join(self.get_hca_indexer().index_names()),
                     doc_type="doc")
-        return list(hits)
+        hits = list(hits)
+        if verify_lists_sorted:
+            num_lists_checked = self._verify_sorted_lists(hits, [])
```
hannes-ucsc: The verifier needs to be called separately for every hit. That makes the assertion a little stronger.

Issue #1481, opened 2019-11-25
Gitlab stages `reindex` and `reindex_early` time out


hannes-ucsc:
Regression from #1462.

https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/5189

Solution might be to use job-specific timeouts.

https://docs.gitlab.com/ee/ci/yaml/#timeout

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-908)
┆Project Name: azul
┆Issue Number: AZUL-908

PR #1529, opened 2020-02-01
Gitlab reindex timeout (#1481)

danielsotirhos:


.gitlab-ci.yml
```diff
@@ -85,10 +85,12 @@ early_reindex:
   when: manual
   script:
     - make reindex
+  timeout: 1h 10m
```
hannes-ucsc: This might need an explanatory comment. This timeout is coupled with another timeout in another location. That's a brittle coupling which should be documented at least. In both places.

Issue #1530, opened 2020-02-01
Mention backend bucket in deployment README


jessebrennan:
And other things that need to be provisioned manually.

PR #1537, opened 2020-02-04
Document golden cloud infrastructure eggs (#1530)

hannes-ucsc:
… and other fixes.


hannes-ucsc:
Deployment passes on Gitlab, which implies that the sandbox has been created. IT fail for same reason as on develop branch: direct access is still broken.

Issue #1400, opened 2019-10-29
Document /health endpoints


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-853)
┆containerName: azul
┆Issue Number: AZUL-853
┆Epic: OpenAPI Spec

PR #1538, opened 2020-02-04
Document /health endpoints (#1400)

achave11:


lambdas/service/app.py
```diff
@@ -185,37 +239,56 @@ def openapi():
...
+                    ]
+                }
+            }
         }
     ],
 }, method_spec={
     **health_spec,
-    'summary': 'List health information for a specific key',
-    'description': 'List health information for a specific key, or all keys if `/health/` is used.'
+    'description': 'List health information for a specific key, or all keys if `/health/` is used.',
```
```suggestion
    'description': 'List health information for a specific key
```
There is another endpoint available for `/health` with no components specified, so documenting that here seems redundant.

achave11: @danielsotirhos I addressed your request but I left the last section. The path that ends with `/` is not registered and can not be used in the example, so there is a not that addresses its existence. Let me know if you disagree.

danielsotirhos: Then perhaps `keys` should not be a required parameters?
Screen Shot 2020-02-13 at 2 57 53 PM
vs
Screen Shot 2020-02-13 at 3 01 25 PM

lambdas/service/app.py
```diff
@@ -185,37 +239,56 @@ def openapi():
...
+            '$ref': '#/components/responses/UnavailableHealth'
+        }
+    }
 })
 @app.route('/health/{keys}', methods=['GET'], cors=True, path_spec={
     'parameters': [
         {
             'name': 'keys',
             'in': 'path',
             'required': True,
```
```suggestion
            'required': False,
```

lambdas/service/app.py
```diff
@@ -179,37 +233,55 @@ def openapi():
 
 
 health_spec = {
```
jessebrennan: As long as the responses are the same for all of the `method_spec`s, then it makes sense to use this variable to reduce duplication.

lambdas/service/app.py
```diff
@@ -179,37 +233,55 @@ def openapi():
...
+    'description': 'List health information for Azul service.',
+    'summary': 'Check Azul service health',
+    'responses': {
+        '200': {
+            '$ref': '#/components/responses/OkHealth'
+        },
+        '503': {
+            '$ref': '#/components/responses/UnavailableHealth'
+        }
+    }
```
jessebrennan: This should be moved into the `health_spec` variable.

lambdas/service/app.py
```diff
@@ -84,12 +84,66 @@
...
             'name': 'Health',
-            'description': 'For checking on service health'
+            'description': 'Describes the health status of Azul service.'
         }
     ],
     'servers': [
         {'url': config.service_endpoint()}
-    ]
+    ],
+    'components': {
```
jessebrennan: Instead of defining the responses in the global spec object, you could define it in the `health_spec` variable like
```
health_spec = {
    'tags': ['Health'],
    'responses': {
        '200': {
            'description': 'Specified service component is reachable.',
            'content': {
...
```
I think, if we can we should avoid defining responses in `components.responses` and instead rely on Python's reuse functionality (i.e. the variable in this case).

lambdas/service/app.py
achave11: @danielsotirhos it seems like the specifications require this param to be se tot true, https://swagger.io/specification/#parameterObject.

lambdas/service/app.py
```diff
@@ -179,37 +179,95 @@ def openapi():
...
+                                'type': 'boolean'
+                            }
+                        }
+                    }
+                }
+            }
+        },
+        '503': {
+            'description': 'Specified service component is unreachable.',
+            'content': {
```
hannes-ucsc: This seems to be almost identical to the 200 case. Try to deduplicate.

lambdas/service/app.py
```diff
@@ -179,37 +179,95 @@ def openapi():
...
+                                    'type': 'boolean',
+                                    'default': False
+                                }
+                            },
+                            'additionalProperties': True
+                        },
+                        'properties': {
+                            'up': {
+                                'type': 'boolean',
+                                'default': False
```
hannes-ucsc: See if you can use the `enum` workaround described in https://github.com/OAI/OpenAPI-Specification/issues/1313

lambdas/service/app.py
```diff
@@ -183,36 +183,54 @@ def openapi():
...
 @app.route('/health', methods=['GET'], cors=True, method_spec={
     **health_spec,
-    'summary': 'List health information for webservice'
+    'description': 'List health information for Azul service',
+    'summary': 'Check Azul health'
 })
 @app.route('/health/{keys}', methods=['GET'], cors=True, path_spec={
     'parameters': [
         {
             'name': 'keys',
```
hannes-ucsc: This parameter needs a description. The fact that `keys` is a comma separated list of these enum values isn't captured anywhere. Maybe an enum isn't the right way to spec this. Consider restricting by regex pattern like `^(api_endpoints|basic|cached|…)(,(api_endpoints|basic|cached|…))*$`

src/azul/openapi/schema.py
```diff
@@ -220,3 +225,27 @@ def make_type(t: TYPE) -> JSON:
...
+        "enum": [
+            "foo",
+            "bar"
+        ]
+    }
+    """
+    if type_:
+        assert all(isinstance(i, type_) for i in items)
+    else:
+        type_ = one(set(map(type, items)))
```
hannes-ucsc: Looks like this line isn't covered by a doc test.

src/azul/openapi/schema.py
```diff
@@ -220,3 +225,27 @@ def make_type(t: TYPE) -> JSON:
...
+    >>> assert_json(enum('foo', 'bar', type_=str))
+    {
+        "type": "string",
+        "enum": [
+            "foo",
+            "bar"
+        ]
+    }
+    """
+    if type_:
```
hannes-ucsc: `is None` or `is not None`. This keeps popping up in your contributions, please try to remember the distinction between `bool(x)` and `x is None`. It may not matter here because no instance of `type` should ever be False-ish but it's just not good practice to rely on the complicated assumptions. If you want to test whether something is `None` use `is None`.

src/azul/openapi/schema.py
```diff
@@ -220,3 +225,27 @@ def make_type(t: TYPE) -> JSON:
...
+        assert all(isinstance(i, type_) for i in items)
```
hannes-ucsc: It would be easy to add a test for the case where this assertion fails.

src/azul/openapi/schema.py
hannes-ucsc: According to the function signature, `type_` can be a string or a dictionary. This assertion wouldn't work correctly in those cases. The assertion should only be made if `type_` is an instance of `type`, otherwise we should trust the caller.

achave11:
The previous implementation of the health controller and class did not allowed access to the internal health check keys such as `basic`, `cached`, `fast` for the comma separated keys health endpoints. Notice the changes were added to the last commit so that those changes can easily be discarded if need be.

lambdas/service/app.py
```diff
@@ -183,36 +183,48 @@ def openapi():
...
             'in': 'path',
             'required': True,
-            'schema': {'type': 'string'}
+            'description': 'Available component(s) to query health check status: '
```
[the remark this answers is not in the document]
hannes-ucsc: Use `all_keys` and `internal_keys` attributes to derive this line.

lambdas/service/app.py
```diff
@@ -183,36 +183,48 @@ def openapi():
...
+                           '[api_endpoints, basic, cached, elasticsearch, fast, progress, other_lambdas, queues]',
+            'schema': schema.array(str,
+                                   pattern='^(api_endpoints|basic|cached|elasticsearch|fast|progress|other_lambdas|'
```
hannes-ucsc: Derive the regex from the aforementioned property as well.

lambdas/service/app.py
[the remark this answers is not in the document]
hannes-ucsc: I also think we shouldn't introduce "component" as a synonym for the _keys_ concept. If we do, we need to strongly associate the two in the description so users understand that they are the same thing. But again, I'd rather not introduce another alias, especially since it does not completely fit. For example, `fast` isn't a component. Try to find a description that ties the _keys_ parameter to the fact that it selects certain entries for the response body. Writing concise and accurate documentation is difficult, so you'll need to think about this very hard.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,51 @@ def array(item: TYPE, *items: TYPE, **kwargs):
     return array_type(make_type(item), *map(make_type, items), **kwargs)
 
 
+def enum(*items: PrimitiveJSON, type_: TYPE = None) -> JSON:
```
hannes-ucsc: Can you chain your PR to @noah-aviel-dove's?

lambdas/service/app.py
```diff
@@ -209,42 +209,84 @@ def openapi():
...
+    'description': 'Basic health check response, confirms Azul service is reachable.',
+    'summary': 'Check Azul service is reachable'
+})
+def basic_health():
+    return health_controller().basic_response()
+
+
+@app.route('/health/cached', methods=['GET'], cors=True, method_spec={
+    **health_spec,
+    'description': 'Azul service health check status, updates every minute.',
```
```suggestion
    'description': 'Azul service health check status, updates periodically.',
```

lambdas/service/app.py
```diff
@@ -209,42 +209,84 @@ def openapi():
...
+    'summary': 'Minute updates of Azul service health'
 })
+def cached_health():
+    return health_controller().cached_response()
+
+
+@app.route('/health/fast', methods=['GET'], cors=True, method_spec={
+    **health_spec,
+    'description': 'Realtime Azul service health check status.',
```
hannes-ucsc: "Real-time" doesn't capture what this does.

lambdas/service/app.py
```diff
@@ -219,42 +219,110 @@ def openapi_raw():
...
+
+def health_controller():
+    return HealthController(lambda_name='service')
+
 
 @app.route('/health', methods=['GET'], cors=True, method_spec={
     **health_spec,
-    'summary': 'List health information for webservice'
+    'description': 'This endpoint is a complete status check of service '
+                   'and alternate resources that make service operational. '
```
```suggestion
    'description': 'Return a complete health status report for the service '
                   'and related resources. '
```

lambdas/service/app.py
```diff
@@ -219,42 +219,110 @@ def openapi_raw():
...
+    'summary': 'Check overall Azul health'
+})
+def health():
+    return health_controller().full_response()
+
+
+@app.route('/health/basic', methods=['GET'], cors=True, method_spec={
+    **health_spec,
+    'description': 'This endpoint returns a basic response that indicates to '
+                   'the client that service is reachable via http request.',
```
```suggestion
    'description': 'Return a basic health status report that indicates to '
                   'the client if the service is reachable via http request.',
```

lambdas/service/app.py
```diff
@@ -219,42 +219,110 @@ def openapi_raw():
...
+def basic_health():
+    return health_controller().basic_response()
+
+
+@app.route('/health/cached', methods=['GET'], cors=True, method_spec={
+    **health_spec,
+    'description': 'This endpoint returns the same response as `health/fast` endpoint. '
+                   'The main difference is that this endpoints updates the health object '
+                   'every minute and is stored in S3. The client is guaranteed a health '
+                   'object no older than two minutes. '
```
```suggestion
    'description': 'Return a cached copy of the `health/fast` endpoint response. The '
                   'cached copy is periodically updated to ensure that the client '
                   'receives a health status report no older than two minutes. '
```

lambdas/service/app.py
```diff
@@ -219,42 +219,110 @@ def openapi_raw():
...
+            'schema': schema.array(str,
+                                   pattern=f'^({"|".join(health_controller().all_keys)})'
+                                           f'(,({"|".join(health_controller().all_keys)}))*$')
         }
     ],
 }, method_spec={
     **health_spec,
-    'summary': 'List health information for a specific key',
-    'description': 'List health information for a specific key, or all keys if `/health/` is used.'
+    'description': 'A health check status is generated for one or more keys specified by the client. '
```
```suggestion
    'description': 'Return a health status report generated for one or more keys '
                   'specified by the client. '
```

lambdas/service/app.py
```diff
@@ -219,42 +219,110 @@ def openapi_raw():
...
+    'summary': 'Minute updates of Azul service health'
 })
+def cached_health():
+    return health_controller().cached_response()
+
+
+@app.route('/health/fast', methods=['GET'], cors=True, method_spec={
+    **health_spec,
+    'description': 'This endpoint generates a health status derived from two core '
+                   'service components. '
```
```suggestion
    'description': 'Return a health status report derived from two core service '
                   'components. '
```

lambdas/service/app.py
```diff
@@ -219,42 +219,110 @@ def openapi_raw():
...
     },
     'tags': ['Auxiliary']
 }
 
+health_service_keys = '`elasticsearch` indicates whether service can establish a connection with Elasticsearch, ' \
+                      'ensuring that indexed documents can be access through Elasticsearch. ' \
+                      '`api_endpoints` shows a list of service API endpoints being requested by service, ' \
+                      'confirming that a client making a similar request will obtain a OK response.'
+
+health_all_keys = '`queues` lists queues utilized by indexer to process transactions and index documents into ' \
```
hannes-ucsc: Use triple quoted strings and `format_description` from https://github.com/DataBiosphere/azul/pull/1435/files. Base your PR branch on that PR's branch to get that function. Use a bulleted list with markdown.

Then again, maybe this should be more formally spec-ed out in the form of a JSONSchema for the response body. The schema doesn't have to be 100% complete. It merely needs to list all the top-level keys along with a description. You may have had that already and I asked you to remove it. Please reach out to me if that is the case so we can discuss.

achave11:
@danielsotirhos it is worth noting that Hannes and I discussed an approach to make a bulleted list of the keys returned by the response with the meaning of each key. We also concluded that the `Schema` object for the response should remain generic i.e. `{'additionalProperties': True}` and finally, to provide the `Example` object as a replication of the response object but only displaying the high level keys i.e. queues, up, progress, etc. 
I included your suggestions because it makes sense to have them closer to the response. I also manage to remove the duplication on the `503` response.

lambdas/service/app.py
```diff
@@ -217,44 +217,133 @@ def openapi_raw():
...
-            }
-        }
-    },
-    'tags': ['Auxiliary']
+health_service_keys = {
+    'elasticsearch': 'indicates whether service can establish a connection with Elasticsearch, '
+                     'ensuring that indexed documents can be access through Elasticsearch.\n',
+    'api_endpoints': 'shows a list of service API endpoints being requested by service, '
+                     'confirming that a client making a similar request will obtain a OK response.\n',
+    'up': 'indicates with a boolean value if the service is reachable.\n',
```
```suggestion
health_up_key = {
    'up': 'indicates with a boolean value if the service is reachable.\n',
}

health_service_keys = {
    'elasticsearch': 'indicates whether service can establish a connection with Elasticsearch, '
                     'ensuring that indexed documents can be access through Elasticsearch.\n',
    'api_endpoints': 'shows a list of service API endpoints being requested by service, '
                     'confirming that a client making a similar request will obtain a OK response.\n',
    **health_up_key,
```

lambdas/service/app.py
```diff
@@ -217,44 +217,133 @@ def openapi_raw():
...
+                   'and related resources. ',
+
+    'summary': 'Check overall Azul health'
+})
+def health():
+    return health_controller().full_response()
+
+
+@app.route('/health/basic', methods=['GET'], cors=True, method_spec={
+    **health_spec({}),
```
```suggestion
    **health_spec(health_up_key),
```

achave11:
It seems like `continuous-integration/travis-ci` is not updating here. But here is a link to the [Travis build.](https://travis-ci.org/github/DataBiosphere/azul/builds/666531441?utm_medium=notification&utm_source=github_status)

lambdas/service/app.py
```diff
@@ -222,44 +222,154 @@ def openapi_raw():
...
-            'content': {
-                'application/json': {
-                    'schema': {'$ref': '#/components/schemas/Empty'}
-                }
-            }
-        }
-    },
-    'tags': ['Auxiliary']
+health_up_key = {
+    'up': 'indicates with a boolean value if the service is reachable.\n',
```
hannes-ucsc: If it has a period at the end, it must start with an upper case letter. Even if it doesn't—list or enumerations or enumeration items often don't end in punctuation—typically these things need to begin with underscore.

lambdas/service/app.py
```diff
@@ -222,44 +222,154 @@ def openapi_raw():
...
+    'api_endpoints': 'shows a list of service API endpoints being requested by service, '
+                     'confirming that a client making a similar request will obtain a OK response.\n',
+    **health_up_key
+}
+
+health_all_keys = {
+    'queues': 'lists queues utilized by indexer to process transactions and index documents into '
+              'Elasticsearch. A queue is denoted as healthy if queue stats can be collected, stats are then '
+              'displayed for each available queue.\n',
+    'other_lambdas': 'represent other server-less applications used by Azul in order to make '
```
hannes-ucsc: The varying indention of these text blocks annoys me. Using format_description() and starting the paragraph on a new line will fix this.

lambdas/service/app.py
```diff
@@ -222,44 +222,154 @@ def openapi_raw():
...
+    **health_service_keys
 }
 
 
+def health_docs_spec(keys: dict) -> str:
+    return ''.join(f'* __{k}__ {v} ' for k, v in keys.items())
+
+
+def health_example_spec(keys: dict, up: Optional[bool] = None):
+    example = {key: {} for key in keys.keys()}
```
hannes-ucsc: This function can be made a one-liner using `**`

lambdas/service/app.py
```diff
@@ -222,44 +222,154 @@ def openapi_raw():
...
+    example['up'] = up
+    return example
+
+
+def health_spec(health_keys):
+    return {
+        'responses': {
+            f'{200 if up else 503}': {
+                'description': f'Service component is {"" if up else "un"}reachable.'
```
hannes-ucsc: format_description

lambdas/service/app.py
```diff
@@ -222,44 +222,154 @@ def openapi_raw():
...
+    '''),
+    **health_spec(health_all_keys)
 })
+def health():
+    return health_controller().full_response()
+
+
+@app.route('/health/basic', methods=['GET'], cors=True, method_spec={
+    'summary': 'Basic health check',
+    'description': format_description(f'''
```
```suggestion
    'description': format_description('''
```

Check for this problem elsewhere, too.

lambdas/service/app.py
```diff
@@ -222,44 +222,154 @@ def openapi_raw():
...
+    '''),
+    **health_spec(health_up_key)
+})
+def basic_health():
+    return health_controller().basic_response()
+
+
+@app.route('/health/cached', methods=['GET'], cors=True, method_spec={
+    'summary': 'Cached health check for continuous monitoring',
+    'description': format_description('''
```
hannes-ucsc: Wrap all your descriptions at column 80. Use a tool for that. PyCharm can wrap text blocks for you.

lambdas/service/app.py
achave11: @hannes-ucsc it is worth noting that I was unable to integrate unwrapping and documenting of the `health_keys` inside the `f` string. Doing so deformed the whole `description` object. Which is why I concatenated it.

lambdas/service/app.py
```diff
@@ -222,44 +222,163 @@ def openapi_raw():
...
+
+
 @app.route('/health/{keys}', methods=['GET'], cors=True, path_spec={
     'parameters': [
         {
             'name': 'keys',
             'in': 'path',
             'required': True,
-            'schema': {'type': 'string'}
+            'description': f'Available keys to query health status: {[*health_controller().all_keys]}',
```
hannes-ucsc: This should be a static reference to all_keys.

lambdas/service/app.py
```diff
@@ -222,44 +222,163 @@ def openapi_raw():
...
+    '''),
+    **health_spec(health_service_keys)
+})
+def fast_health():
+    return health_controller().fast_response()
+
+
 @app.route('/health/{keys}', methods=['GET'], cors=True, path_spec={
     'parameters': [
         {
```
hannes-ucsc: This should use the param.path helper.

lambdas/service/app.py
```diff
@@ -222,44 +222,163 @@ def openapi_raw():
...
             'name': 'keys',
             'in': 'path',
             'required': True,
-            'schema': {'type': 'string'}
+            'description': f'Available keys to query health status: {[*health_controller().all_keys]}',
+            'schema': schema.array(str,
+                                   pattern=f'^({"|".join(health_controller().all_keys)})'
```
hannes-ucsc: I saw no evidence of the pattern anywhere in the Swagger UI so it should be ripped out.

lambdas/service/app.py
```diff
@@ -222,44 +222,163 @@ def openapi_raw():
...
+                                           f'(,({"|".join(health_controller().all_keys)}))*$')
         }
     ],
 }, method_spec={
-    **health_spec,
-    'summary': 'List health information for a specific key',
-    'description': 'List health information for a specific key, or all keys if `/health/` is used.'
+    'summary': 'Custom health check',
```
hannes-ucsc: path_spec and method_spec should be swapped such that `summary` comes first.

hannes-ucsc:
Could you review my changes for technical correctness, spelling grammar, etc?

achave11:
@hannes-ucsc Looks good, I approve!

hannes-ucsc:
I've had a few ideas of how this could be further improved.

Issue #1188, opened 2019-08-02
labCount in summary aggregation off by +1 due to None value


danielsotirhos:
The `/summary` response contains a metric aggregation on the `contents.projects.laboratory` field which now (after #989) is including a value of `None` as a value it adds in to the count, causing the unique count of laboratory values to be off by +1 if any of the optional laboratory fields were `None`

This is being caused by the changes in #989 whereby `None` values are converted to a special keyword to be stored in Elasticsearch, whereas before a field with a `None` value simply wasn't stored in Elasticsearch.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-739)
┆containerName: azul
┆Issue Number: AZUL-739

danielsotirhos:
This was first resolved by #1188 (PR #1543) where it was found that labCount was in fact not being incremented for None values due to an explicit skip of these values when collecting them in the project entity.

Later it was decided that all unique laboratory values including None should be represented in the labCount value, and this change was included as part of #1575 (PR #1587) verified in the test function `test_summary_response()`

danielsotirhos:
Ticket re-demoed to show the new desired behavior where a `laboratory` value of None (eg not having a laboratory value) counts as a distinct value towards the `labCount` value in the summary endpoint.

PR #1543, opened 2020-02-05
Verify labCount in summary aggregation (#1188)

danielsotirhos:


danielsotirhos:
Not an issue due to how `laboratory` value is collected into the `project` entity

PR #1548, opened 2020-02-06
Only 'index' and 'write' lambdas should reserve concurrency (#1546)

achave11:


achave11:
#### TODO: 
Modify `example` deployment to reduce the number of concurrent lambda executions for personal deployments.

PR #1549, opened 2020-02-06
Stop replacing . with _ in participant_id

noah-aviel-dove:


hannes-ucsc:
@noah-aviel-dove be sure to include ticket number in commit title

Issue #1445, opened 2019-11-13
Sorting by multi-valued fields


hannes-ucsc:
It is very common, especially in aggregate documents, that a particular document field is the result of the set union of the field values from the contributions. How do we sort by these fields?

Ideally one would want a tuple-based comparison, i.e. `("a", "b") < ("a", "c")` but I don't think ES can do that. ES needs to be told to pick one value out of the many: https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-sort.html#_sort_mode_option

Another option would be to used [script-based sorting](https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-sort.html#_script_based_sorting) to sort by the concatenation of all values.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-887)
┆containerName: azul
┆Issue Number: AZUL-887

theathorn:
Solved by #1446?

PR #1554, opened 2020-02-08
Sorting by multi-valued fields (#1445)

danielsotirhos:


danielsotirhos:
Note that this branch was based off of `issues/danielsotirhos/sort-multi-value-fields`

test/service/test_response.py
```diff
@@ -1420,6 +1420,27 @@ def test_ordering(self):
             self.assertEqual(hit_sort_values['asc'], sorted(hit_sort_values['asc']))
             self.assertEqual(hit_sort_values['desc'], sorted(hit_sort_values['desc'], reverse=True))
 
+    def test_multivalued_field_sorting(self):
+        """
+        Test that sorting by a multi-valued field responds with hits that are
+        correctly sorted based on the first value from each multi-valued field, and
+        that each multi-valued field itself is sorted low to high regardless of the search sort
```
hannes-ucsc: Is the statement in the last line true?

```suggestion
        that each multi-valued field itself is sorted low to high regardless of the search sort.
```

danielsotirhos: Yes. This test uses the `laboratory` field to verify 2 aspects of sorting:
- A request made using `sort` and `order` parameters against a multi-valued field like `laboratory` will order the hits based on the first (smallest) element from each hit's `laboratory` field. 
```
1442                self.assertEqual(laboratories, sorted(laboratories, reverse=reverse))
```
- Multi-valued fields (only `laboratory` verified here) is itself sorted low to high
```
1439                    self.assertEqual(laboratory, sorted(laboratory))
```

hannes-ucsc: I see, thank you!

hannes-ucsc:
Not reviewing this right now since I requested changes on the base PR.

PR #1578, opened 2020-02-14
Add GET/PUT script for portal database access

noah-aviel-dove:


noah-aviel-dove:
@danielsotirhos Did you run `make terraform`? The object version table is deployment specific, and the error message looks to me like the entire table is missing rather than a specific entry.

danielsotirhos:
> Did you run `make terraform`?
That is what I was missing. Thanks.

scripts/manage_portals.py
```diff
@@ -0,0 +1,35 @@
...
+import sys
+
+from azul.portal_service import PortalService
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(
+        description='Upload/download portal integrations database with safe concurrency.'
+    )
+    parser.add_argument('action', choices=('GET', 'PUT'), help='Whether to download (GET) or upload (PUT).')
```
hannes-ucsc: Apologies, I meant to ask for this in my previous review: Can we make the usage `manage_portals.py --get | --put [FILE | -]`? I know mandatory options are a contradiction but it just feels more natural to me in this case.

https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group might come in handy for this.

hannes-ucsc:
Try to remember to add the ticket references to the commit title.

Also, there is no good reason to split this up into two commits. I squashed everything down to one commit.

Issue #1401, opened 2019-10-29
Document /version endpoint


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-854)
┆containerName: azul
┆Issue Number: AZUL-854
┆Epic: OpenAPI Spec

PR #1580, opened 2020-02-15
Document /version endpoint (#1401)

noah-aviel-dove:
Potential issues with this commit:

 - Only 200 response is documented (I saw that @achave11 wrote an entire schema for the 503 for [his PR](https://github.com/DataBiosphere/azul/pull/1538/files) but I wasn't sure how that would apply to this endpoint).

- The schema for .changes[] isn't 100% consistent as some items lack the "notes" property. My current approach is to document the property, but to omit it from the list of _required properties_. The swagger UI renders this rather poorly though. An alternative is to omit explicitly naming the notes property and use `additionalProperties` instead.


lambdas/service/app.py
```diff
@@ -85,11 +85,71 @@
         {
             'name': 'Health',
             'description': 'For checking on service health'
+        },
+        {
+            'name': 'Version',
```
hannes-ucsc: Let's merge the Health and Version tags into a single `Auxiliary` tag.

lambdas/service/app.py
```diff
@@ -85,11 +85,71 @@
...
+            'description': 'Describes the current version of Azul service'
         }
     ],
     'servers': [
         {'url': config.service_endpoint()}
-    ]
+    ],
+    'components': {
+        'responses': {
+            'OkVersion': {
```
hannes-ucsc: Why is this extracted into a component? AFAICT, it doesn't need to be since there is only one reference. 

Related things should be in close proximity unless the value of reusing something or abstracting it outweighs the proximity concern. There is no abstraction or reuse in this case it seems.

lambdas/service/app.py
```diff
@@ -224,7 +224,67 @@ def generate_health_object(_event: chalice.app.CloudWatchEvent):
...
+                                            'items': {
+                                                'type': 'string'
+                                            }
+                                        },
+                                        'notes': {
+                                            'type': 'string'
+                                        }
+                                    },
+                                    'required': [
+                                        'title', 'issues', 'upgrade'
```
```suggestion
                                        ['title', 'issues', 'upgrade']
```

Wrap all or none.

lambdas/service/app.py
```diff
@@ -224,7 +224,67 @@ def generate_health_object(_event: chalice.app.CloudWatchEvent):
     controller.generate_cache()
 
 
-@app.route('/version', methods=['GET'], cors=True)
+version_spec = {
```
hannes-ucsc: This should be inlined so the spec textually follows the `@app.route` line. Extraction of specs into variables should only be done for reuse.

hannes-ucsc:
> Only 200 response is documented

That's fine in this case. The body of 503 responses returned by `/health` is well-defined and meaningful.

Issue #1402, opened 2019-10-29
Document /dss/files endpoint


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-856)
┆containerName: azul
┆Issue Number: AZUL-856
┆Epic: OpenAPI Spec

PR #1586, opened 2020-02-19
Document /dss/files and /fetch/dss/files endpoints (#1402)

noah-aviel-dove:
The /fetch endpoint didn't have its own ticket so I figured I'd cover it here.

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -39,6 +39,8 @@ from azul.logging import configure_app_logging
 from azul.openapi import (
     annotated_specs,
     schema,
+    params,
+    responses,
 )
 from azul.plugin import Plugin
 from azul.portal_service import PortalService
@@ -978,57 +980,84 @@ def generate_manifest(event, context):
     return {'Location': presigned_url}
 
 
-@app.route('/dss/files/{uuid}', methods=['GET'], cors=True)
+dss_files_spec = {
+    'tags': ['DSS'],
+    'parameters': [
+        params.path('uuid', str, description='UUID of the file to be checked out'),
+        params.query('fileName',
+                     schema.optional(str),
+                     description='The desired name of the file. If absent, the UUID of the file will be used.'),
+        params.query('requestIndex',
+                     schema.optional(int),
+                     description='Number of attempts made through the endpoint to fetch the desired file.'),
+        params.query('wait',
+                     schema.optional(int),
+                     description='If this parameter is 1 and the checkout is still in process, the server will wait '
+                                 'before returning a response. This parameter should only be set to 1 by clients who '
+                                 'don\'t honor the Retry-After header, preventing them from quickly exhausting the '
+                                 'maximum number of redirects. If the server cannot wait the full amount, any amount '
+                                 'of wait time left will still be returned in the Retry-After header of the response.'),
+        params.query('replica',
+                     schema.enum('aws'),
+                     description='All query parameters not mentioned above are forwarded to DSS in order to initiate '
+                                 'checkout process for the correct file. For more information refer to '
+                                 'https://dss.data.humancellatlas.org under GET `/files/{uuid}`. The only mandatory '
+                                 'forwarded header parameter is `replica`, for which Azul only supports AWS.')
+    ]
+}
+
+
+@app.route('/dss/files/{uuid}', methods=['GET'], cors=True, method_spec={
+    **dss_files_spec,
+    'summary': 'Request a download link to a Data Store file and redirect',
+    'responses': {
+        '301': {
+            'description': 'DSS checkout has been started or is ongoing. The '
+                           'response is a redirect to this very endpoint. The '
+                           'response MAY carry a `Retry-After` header, even if '
+                           'server-side waiting was requested via `wait=1`. '
+                           '`Retry-After` is the recommended number of seconds '
+                           'to wait before requesting the URL specified in the '
+                           '`Location` header, which will point back to this '
+                           'endpoint (so the client should expect a subsequent '
+                           'response of the same kind).',
+
+            'headers': {
+                'Location': responses.header(str, description='This endpoint.'),
+                'Retry-After': responses.header(int, description='Recommended number of seconds to wait before '
+                                                                 'requesting the URL specified in the `Location` '
+                                                                 'header. The response MAY carry this header even if '
+                                                                 'server-side waiting was requested via `wait=1`.')
+            }
+        },
+        '302': {
+            'description': 'DSS checkout is complete. The response is a '
+                           'redirect to an entirely different service. '
+                           'The client should requesting the URL in the '
+                           '`Location` header and expect a response containing '
+                           'the actual file. Currently the `Location` header '
+                           'is a signed URL to an object in S3, but clients '
+                           'should not depend on that. The response will also '
+                           'include a `Content-Disposition` header set to '
+                           '`attachment; filename=` followed by the value of '
+                           'the `fileName` parameter specified in the initial '
+                           'request (or the UUID of the file if that parameter '
+                           'was omitted).',
+            'headers': {
+                'Location': responses.header(str, description='Signed URL that will yield the actual content of the '
+                                                              'file.'),
+                'Content-Disposition': responses.header(str, description='Set to `attachment; filename=` followed by '
+                                                                         'the name of the file as determined by the '
+                                                                         '`uuid`/`fileName` parameters.')
+            }
+        },
+    }
+})
 def dss_files(uuid):
     """
     Initiate checking out a file for download from the HCA data store (DSS)
 
-    parameters:
-        - name: uuid
-          in: path
-          type: string
-          description: UUID of the file to be checked out
-        - name: fileName
-          in: query
-          type: string
-          description: The desired name of the file. If absent, the UUID of the file will be used.
-        - name: requestIndex
-          in: query
-          type: int
-          description: Number of attempts made through the endpoint to fetch  the desired file.
-        - name: wait
-          in: query
-          type: int
-          description: If this parameter is 1 and the checkout is still in process, the server will wait before
-          returning a response. This parameter should only be set to 1 by clients who don't honor the Retry-After
-          header, preventing them from quickly exhausting the maximum number of redirects. If the server cannot wait
-          the full amount, any amount of wait time left will still be returned in the Retry-After header of the
-          response.
-
     :return: A 301 or 302 response describing the status of the checkout performed by DSS.
-
-    All query parameters not mentioned above are forwarded to DSS in order to initiate checkout process for the
-    correct file. For more information refer to https://dss.data.humancellatlas.org under GET `/files/{uuid}`.
-
-    If the file checkout has been started or is still ongoing, the response will be a 301 redirect to this very
-    endpoint. The response MAY carry a Retry-After header, even if server-side waiting was requested via `wait=1`.
-
-    `Retry-After` is the recommended number of seconds to wait before requesting the URL specified in the `Location`
-    header.
-
-    If the client receives a 301 response, the client should wait the number of seconds specified in `Retry-After`
-    and then request the URL given in the `Location` header. The URL in the `Location` header will point back at this
-    endpoint so the client should expect a response of the same kind.
-
-    If the file checkout is done and the file is ready to be downloaded, the response will be a 302 redirect with the
-    `Location` header set to a signed URL. Requesting that URL will yield the actual content of the file.
-
-    The client should request the URL given in the `Location` field. The URL will point to an entirely different
-    service and when requesting the URL, the client should expect a response containing the actual file. Currently
-    the `Location` header of the 302 response is a signed URL to an object in S3 but clients should not depend on
-    that. The response will also include a `Content-Disposition` header set to `attachment; filename=` followed by
-    the value of the `fileName` parameter specified in the initial request or the UUID of the file if that parameter
-    was omitted.
     """
     body = _dss_files(uuid, fetch=False)
     status_code = body.pop('Status')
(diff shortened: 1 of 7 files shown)


lambdas/service/app.py
```diff
@@ -1062,7 +1143,7 @@ def fetch_dss_files(uuid):
 
 
 def _dss_files(uuid, fetch=True):
-    query_params = app.current_request.query_params
+    query_params = app.current_request.query_params or {}
```
achave11: This change isn't specific to the documentation, should it be in a different commit? 
Also this seems like a regression from the present implementation, I could be wrong.

lambdas/service/app.py
```diff
@@ -85,6 +85,10 @@
         {
             'name': 'Health',
             'description': 'For checking on service health'
+        },
+        {
+            'name': 'DSS',
```
achave11: We might want to be explicit about this endpoints being a `get` file proxy from the DSS aws replica.

lambdas/service/app.py
```diff
@@ -925,7 +929,80 @@ def generate_manifest(event, context):
     return {'Location': presigned_url}
 
 
-@app.route('/dss/files/{uuid}', methods=['GET'], cors=True)
+dss_files_spec = {
+    'description': 'Request a signed download link to a file from the Data Store',
```
achave11: We should make the distinction of the `/fetch/dss/files/{uuid}` and `/dss/files/{uuid} ` endpoint descriptions since it is a significant one.

lambdas/service/app.py
noah-aviel-dove: It IS in its own commit. Did you mean a different pull request?

achave11: Yea that was a mistake.

achave11: Yea that was a mistake on my behalf.

lambdas/service/app.py
```diff
@@ -925,7 +929,80 @@ def generate_manifest(event, context):
     return {'Location': presigned_url}
 
 
-@app.route('/dss/files/{uuid}', methods=['GET'], cors=True)
+dss_files_spec = {
```
achave11: You can add the 'tags' field in this dict to reduce duplication.

achave11:
For the next review cycle, ensure the `Try it out` example is functional. Also please fill free to talk to me about any comments or questions you might have.

noah-aviel-dove:
There appears to be an issue with the "Try it out" functionality with the `/dss/files/` endpoint. Swagger follows redirects, meaning that instead of seeing the 30X responses, it spins for a seconds before redirecting all the way to the requested file and displaying its contents with a 200 response. 

While this is an accurate demonstration of how the endpoint behaves, it appears to conflict with the documented responses, since we never see the 30X. I'm not sure whether this should be added to the documentation or not.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,90 @@ def array(item: TYPE, *items: TYPE, **kwargs):
     return array_type(make_type(item), *map(make_type, items), **kwargs)
 
 
+def enum_type(items: Iterable) -> JSON:
```
hannes-ucsc: This should use varargs like other helpers. The name should be just `enum`.

@achave11 is working on this too. You guys should coordinate.

I like the implicit type determination but there should be a way to explicitly pass the type.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,90 @@ def array(item: TYPE, *items: TYPE, **kwargs):
...
+    return make_param(name, 'query', type_, required=not is_optional, description=description)
+
+
+def header_response(type_: TYPE, description: Optional[str] = None) -> JSON:
+    """
+    Returns the schema and description for a response header.
+    """
+    return {
+        'schema': make_type(type_),
+        **({} if description is None else {'description': description})
```
hannes-ucsc: This conditional can be eliminated simply by allowing and interpolating `**kwargs` like some other helpers do. 

This is not supposed to be an abstraction over OpenAPI spec, but merely a way to reduce its verbosity when specified as a JSON literal.

src/azul/openapi/schema.py
```diff
@@ -220,3 +309,27 @@ def make_type(t: TYPE) -> JSON:
         return t
     else:
         assert False, type(t)
+
+
+def make_param(name: str, in_: str, type_: TYPE, required: bool, description: Optional[str]) -> JSON:
```
hannes-ucsc:
```suggestion
def make_param(name: str, in_: str, type_: TYPE, optional: bool, description: Optional[str]) -> JSON:
```

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,90 @@ def array(item: TYPE, *items: TYPE, **kwargs):
...
+            "foo"
+        ]
+    }
+    """
+    items = sorted(items)
+    type_ = one(set(map(type, items)))
+    return dict(make_type(type_), enum=items)
+
+
+def path_param(name: str, type_: TYPE, description: Optional[str] = None) -> JSON:
```
hannes-ucsc: The param stuff should be in a separate `params` module. `schema` is for the JSONSchema subset of OpenAPI.

Instead of `schema.path_param(…)` an invocation would be `params.path` which is a bit shorter.

src/azul/openapi/schema.py
```diff
@@ -220,3 +309,27 @@ def make_type(t: TYPE) -> JSON:
...
+        },
+        "description": "razzle that dazzle"
+    }
+    """
+    return {
+        'name': name,
+        'in': in_,
+        'required': required,
+        'schema': make_type(type_),
+        **({} if description is None else {'description': description})
```
hannes-ucsc: **kwargs

src/azul/openapi/schema.py
```diff
@@ -220,3 +309,27 @@ def make_type(t: TYPE) -> JSON:
         return t
     else:
         assert False, type(t)
+
+
+def make_param(name: str, in_: str, type_: TYPE, required: bool, description: Optional[str]) -> JSON:
+    """
+    >>> from azul.doctests import assert_json
+    >>> assert_json(make_param('foo', 'query', int, False, 'razzle that dazzle'))
```
hannes-ucsc: Because of the large number of heterogeneous positional args, this invocation is hard to read if it were to occur in an actual spec. To make it easier to read, one would need to pass most of the positionals as keywords anyway which wouldn't be much shorter than the raw JSON literal. OTOH, having this helper enables the IDE's type checker to warn about missing arguments and type mismatches. I think we should leave this helper in, but make it protected and convert all calls to use keyword argument passing even for positional arguments.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,90 @@ def array(item: TYPE, *items: TYPE, **kwargs):
...
+        "required": false,
+        "schema": {
+            "type": "integer",
+            "format": "int64"
+        }
+    }
+    """
+    is_optional = isinstance(type_, optional)
+    if is_optional:
+        type_ = type_.property
```
hannes-ucsc: I think we should rename `optional.property` to `optional.type_`

achave11:
@noah-aviel-dove As described [here](https://github.com/swagger-api/swagger-ui/issues/1569#issuecomment-390877095), swagger-ui follows redirects.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,29 @@ def array(item: TYPE, *items: TYPE, **kwargs):
...
+    >>> assert_json(enum('foo', 'bar', type_=str))
+    {
+        "type": "string",
+        "enum": [
+            "foo",
+            "bar"
+        ]
+    }
+    """
+    if type_:
```
achave11: I will need to address this [requested changes](https://github.com/DataBiosphere/azul/pull/1538/files#r385474794). I am at fault for this.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,51 @@ def array(item: TYPE, *items: TYPE, **kwargs):
     return array_type(make_type(item), *map(make_type, items), **kwargs)
 
 
+def enum(*items: PrimitiveJSON, type_: TYPE = None) -> JSON:
+    """
+    Returns a JSON schema `enum` items of a given type.
+    >>> from azul.doctests import assert_json
```
hannes-ucsc: Empty line after description and each assertion.

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,51 @@ def array(item: TYPE, *items: TYPE, **kwargs):
...
+    {
+        "type": "integer",
+        "format": "int64",
+        "enum": [
+            2,
+            5,
+            7
+        ]
+    }
+    >>> assert_json(enum('bar', 'foo', 0))
```
hannes-ucsc: assert_json not necessary here

src/azul/openapi/schema.py
```diff
@@ -117,6 +122,51 @@ def array(item: TYPE, *items: TYPE, **kwargs):
...
+    ...
+    AssertionError
+    """
+    if type_ is None:
+        type_ = one(set(map(type, items)))
+    elif isinstance(type_, type):
+        assert all(isinstance(item, type_) for item in items)
+    else:
+        # Can't verify type when string or JSON
+        pass
```
hannes-ucsc: not covered.

src/azul/openapi/params.py
```diff
@@ -0,0 +1,80 @@
...
+            "format": "int64"
+        }
+    }
+    """
+    return _make_param(name, in_='query', type_=type_, **kwargs)
+
+
+def _make_param(name: str, in_: str, type_: Union[TYPE, schema.optional], **kwargs: PrimitiveJSON) -> JSON:
+    """
+    >>> from azul.doctests import assert_json
```
hannes-ucsc: These doctests add no coverage. Please remove.

lambdas/service/app.py
```diff
@@ -957,7 +959,59 @@ def generate_manifest(event, context):
...
+        params.path('uuid', str, description='UUID of the file to be checked out'),
+        params.query('fileName',
+                     schema.optional(str),
+                     description='The desired name of the file. If absent, the UUID of the file will be used.'),
+        params.query('requestIndex',
+                     schema.optional(int),
+                     description='Number of attempts made through the endpoint to fetch the desired file.'),
+        params.query('wait',
+                     schema.optional(int),
+                     description='If this parameter is 1 and the checkout is still in process, the server will wait '
```
hannes-ucsc: Wouldn't a triple quoted string be easier here? If indention and newlines are a problem, add a helper to post-process the string.

lambdas/service/app.py
```diff
@@ -957,7 +959,59 @@ def generate_manifest(event, context):
...
 
 
-@app.route('/dss/files/{uuid}', methods=['GET'], cors=True)
+dss_files_spec = {
+    'tags': ['DSS'],
+    'parameters': [
+        params.path('uuid', str, description='UUID of the file to be checked out'),
+        params.query('fileName',
+                     schema.optional(str),
+                     description='The desired name of the file. If absent, the UUID of the file will be used.'),
```
hannes-ucsc: Don't duplicate code or documentation. Remove the docstring sections after porting them to the spec. Ideally, the entire docstring should be ported and removed alltogether.

src/azul/openapi/params.py
```diff
@@ -0,0 +1,80 @@
...
+from azul.types import (
+    JSON,
+    PrimitiveJSON,
+)
+
+
+# Path params cannot be optional
+def path(name: str, type_: TYPE, **kwargs: PrimitiveJSON) -> JSON:
+    """
+    Returns a JSON `parameters` entry for an endpoint parameter specified via
```
hannes-ucsc: Slightly imprecise and a bit awkward.

```suggestion
    Returns an OpenAPI `parameters` specification of a URL path parameter.
```

See if you can improve the other docstrings accordingly. I haven't look at them in detail.

src/azul/openapi/params.py
```diff
@@ -0,0 +1,80 @@
...
+from azul.openapi.schema import (
+    TYPE
+)
+from azul.types import (
+    JSON,
+    PrimitiveJSON,
+)
+
+
+# Path params cannot be optional
```
hannes-ucsc: I'd move this into the docstring.

src/azul/strings.py
```diff
@@ -46,3 +46,22 @@ def pluralize(word: str, count: int) -> str:
         else:
             result += 's'
     return result
+
+
+def collapse_lines(string: str):
```
hannes-ucsc:
```suggestion
def unwrap(string: str):
```

This will be called a lot so a shorter name makes sense.

lambdas/service/app.py
```diff
@@ -957,57 +962,101 @@ def generate_manifest(event, context):
...
+            'headers': {
+                'Location': responses.header(str, description='This endpoint.'),
+                'Retry-After': responses.header(int, description=collapse_lines(
+                    '''
+                    Recommended number of seconds to wait before requesting the
+                    URL specified in the `Location` header. The response MAY
+                    carry this header even if server-side waiting was requested
+                    via `wait=1`.
+                    '''
+                ))
```
```suggestion
                'Retry-After': responses.header(int, description=unwrap('''
                    Recommended number of seconds to wait before requesting the
                    URL specified in the `Location` header. The response MAY
                    carry this header even if server-side waiting was requested
                    via `wait=1`.
                '''))
```

unless `make format` mangles this

hannes-ucsc: Or even

```suggestion
                'Retry-After': responses.header(int, description=unwrap('''
                    Recommended number of seconds to wait before requesting the
                    URL specified in the `Location` header. The response MAY
                    carry this header even if server-side waiting was requested
                    via `wait=1`.'''))
```

hannes-ucsc: I'm also wondering if we should build the call to unwrap into the helpers. If a keyword argument named `description` is passed, apply `unwrap()` to it automatically.

noah-aviel-dove: I think it's a good idea to implement this.

PR #1587, opened 2020-02-19
Fix: Paging broken when sorting fields with Null values (#1575)

danielsotirhos:
- Fixed bug where next/previous link was empty if sorted field had a value of null.

### DIFF
--- a/src/azul/collections.py
+++ b/src/azul/collections.py
@@ -3,6 +3,8 @@ from itertools import (
     product,
 )
 from typing import (
+    Any,
+    Callable,
     Dict,
     Iterable,
     List,
@@ -62,35 +64,49 @@ def explode_dict(d: Mapping[K, Union[V, List[V], Set[V], Tuple[V]]]) -> Iterable
         yield dict(zip(d.keys(), t))
 
 
-def none_safe_key(v):
+def none_safe_key(none_last: bool = False) -> Callable[[Any], Any]:
     """
-    A sort key that handles None values.
+    Returns a sort key that handles None values.
 
     >>> sorted([1, None, 2])
     Traceback (most recent call last):
     ...
     TypeError: '<' not supported between instances of 'NoneType' and 'int'
 
-    >>> sorted([1, None, 2], key=none_safe_key)
+    >>> sorted([1, None, 2], key=none_safe_key())
     [None, 1, 2]
+
+    >>> sorted([1, None, 2], key=none_safe_key(none_last=True))
+    [1, 2, None]
     """
-    return v is not None, v
+
+    def inner_func(v):
+        return (v is None, v) if none_last else (v is not None, v)
+
+    return inner_func
 
 
-def none_safe_tuple_key(t):
+def none_safe_tuple_key(none_last: bool = False) -> Callable[[Tuple[Any]], Any]:
     """
-    A sort key that handles tuples containing None value.
+    Returns a sort key that handles tuples containing None values.
 
     >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')])
     Traceback (most recent call last):
     ...
     TypeError: '<' not supported between instances of 'NoneType' and 'int'
 
-    >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')], key=none_safe_tuple_key)
+    >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')], key=none_safe_tuple_key())
     [(None, 'a'), (1, 'b'), (2, None), (2, 'c')]
+
+    >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')], key=none_safe_tuple_key(none_last=True))
+    [(1, 'b'), (2, 'c'), (2, None), (None, 'a')]
     """
-    assert isinstance(t, tuple)
-    return tuple(map(none_safe_key, t))
+
+    def inner_func(t):
+        assert isinstance(t, tuple)
+        return tuple(map(none_safe_key(none_last=none_last), t))
+
+    return inner_func
 
 
 def compose_keys(f, g):
@@ -102,7 +118,7 @@ def compose_keys(f, g):
     >>> v = dict(b=1, a=None)
     >>> key(v)
     (None, 1)
-    >>> composed_key = compose_keys(none_safe_tuple_key, key)
+    >>> composed_key = compose_keys(none_safe_tuple_key(), key)
     >>> composed_key(v)
     ((False, None), (True, 1))
 
--- a/src/azul/transformer.py
+++ b/src/azul/transformer.py
@@ -47,8 +47,6 @@ from azul.types import (
     JSON,
 )
 
-MIN_INT = -sys.maxsize - 1
-
 logger = logging.getLogger(__name__)
 
 Entities = List[JSON]
@@ -120,25 +118,35 @@ class Document:
         :param forward: If we should translate forward or backward (aka un-translate)
         :return: The translated value
         """
+        # Note that the replacement values for `None` used for each data type
+        # ensure that `None` values are placed at the end of a sorted list.
+        null_string = '~null'
+        # Maximum int that can be represented as a 64-bit int and double IEEE
+        # floating point number. This prevents loss when converting between the two.
+        null_int = sys.maxsize - 1023
+        assert null_int == int(float(null_int))
+        bool_translation_forward = {False: 0, True: 1, None: null_int}
+        bool_translation_backward = {0: False, 1: True, null_int: None}
+
         if field_type is bool:
             if forward:
-                return {None: -1, False: 0, True: 1}[value]
+                return bool_translation_forward[value]
             else:
-                return {-1: None, 0: False, 1: True}[value]
+                return bool_translation_backward[value]
         elif field_type is int or field_type is float:
             if forward:
                 if value is None:
-                    return MIN_INT
+                    return null_int
             else:
-                if value == MIN_INT:
+                if value == null_int:
                     return None
             return value
         elif field_type is str:
             if forward:
                 if value is None:
-                    return config.null_keyword
+                    return null_string
             else:
-                if value == config.null_keyword:
+                if value == null_string:
                     return None
             return value
         elif field_type is dict:
@@ -158,6 +166,7 @@ class Document:
         """
         Traverse a document to translate field values for insert into Elasticsearch, or to translate back
         response data. This is done to support None/null values since Elasticsearch does not index these values.
+        Values that are empty lists ([]) and lists of None ([None]) are both forward converted to [null_string]
 
         :param doc: A document dict of values
         :param field_types: A mapping of field paths to field type
@@ -185,7 +194,15 @@ class Document:
                         new_dict[key + '_'] = val
             return new_dict
         elif isinstance(doc, list):
-            return [cls.translate_fields(val, field_types, forward=forward) for val in doc]
+            # Translate an empty list to a list containing a single None value
+            # (and then further translate that None value according to the field
+            # type), but do so only at the field level (to avoid case of
+            # contents['organoids'] == []).
+            if doc or isinstance(field_types, dict):
+                return [cls.translate_fields(val, field_types, forward=forward) for val in doc]
+            else:
+                assert forward
+                return cls.translate_fields([None], field_types)
         else:
             return cls.translate_field(doc, field_types, forward=forward)
 
@@ -485,7 +502,7 @@ class SetAccumulator(Accumulator):
         super().__init__()
         self.value = set()
         self.max_size = max_size
-        self.key = none_safe_key if key is None else key
+        self.key = none_safe_key(none_last=True) if key is None else key
 
     def accumulate(self, value) -> bool:
         """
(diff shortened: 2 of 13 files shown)


src/azul/transformer.py
```diff
@@ -101,25 +102,33 @@ def translate_field(cls, value: AnyJSON, field_type: FieldType, forward: bool =
         """
         # Note that the replacement values for `None` used for each data type
         # ensure that `None` values are placed at the end of a sorted list.
+        null_keyword = '~null'
```
```suggestion
        null_string = '~null'
```

???

test/service/test_pagination.py
```diff
@@ -45,7 +45,7 @@ def assert_page1_correct(self, json_response):
         num_hits = len(json_response['hits'])
         if 'search_after' in json_response['pagination']:
             self.assertEqual(json_response['pagination']['search_after'],
-                             json_response['hits'][num_hits - 1]['entryId'],
+                             '"{}"'.format(json_response['hits'][num_hits - 1]['entryId']),
```
hannes-ucsc: We don't use format(), only f-strings or concatenation. The latter would look the the least weird in this case I think.

hannes-ucsc: Also, shouldn't this involve json.dumps/loads instead?

test/indexer/test_hca_indexer.py
```diff
@@ -844,7 +861,7 @@ def test_cell_line_sample(self):
                     cell_lines_model_organ = {cl['model_organ'] for cl in contents['cell_lines']}
                 self.assertEqual(cell_lines_model_organ, {'blood (parent_cell_line)', 'blood (child_cell_line)'})
                 self.assertEqual(one(contents['cell_suspensions'])['organ'], ['blood (child_cell_line)'])
-                self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [config.null_keyword])
+                self.assertEqual(one(contents['cell_suspensions'])['organ_part'], [Document.translate_field(None, str)])
```
hannes-ucsc: Might make sense to extract this into a variable like you do above even though it may only be used once.

test/indexer/test_hca_indexer.py
```diff
@@ -613,9 +614,13 @@ def test_plate_bundle(self):
                 if aggregate:
                     bundles = hit['_source']['bundles']
                     self.assertEqual(1, len(bundles))
-                    self.assertEqual(one(contents['protocols'])['paired_end'], [True])
+                    self.assertEqual(one(contents['protocols'])['paired_end'], [
+                        Document.translate_field(True, bool),
```
hannes-ucsc: Try to extract these into a variable or a class attribute.

test/service/test_pagination.py
danielsotirhos: I went with `format()` to match its use throughout that file, but yes `json.dumps()` is a better way to produce the verification values in this case.

danielsotirhos:
> The null-safe key stuff isn't available on `hca/develop` so we might need to port that as well. Start looking into what it would take to port this to `hca/develop` by doing a trial rebase. Once you've addressed my other comments, you should open a second PR, that one one based on `hca/develop` instead of `hca/develop`.

Looking now at back-porting to `hca/develop`. First attempt was a broken mess, trying this time with a bit more care in the cherry-picking.

src/azul/collections.py
```diff
@@ -62,35 +64,44 @@ def explode_dict(d: Mapping[K, Union[V, List[V], Set[V], Tuple[V]]]) -> Iterable
         yield dict(zip(d.keys(), t))
 
 
-def none_safe_key(v):
+def none_safe_key(none_last: bool = True) -> Callable[[Any], Any]:
     """
-    A sort key that handles None values.
+    A sort key factory that handles None values.
```
hannes-ucsc:
```suggestion
    Returns a sort key that handles None values.
```

src/azul/collections.py
```diff
@@ -62,35 +64,44 @@ def explode_dict(d: Mapping[K, Union[V, List[V], Set[V], Tuple[V]]]) -> Iterable
...
+        return lambda v: (v is None, v)
+    else:
+        return lambda v: (v is not None, v)
 
 
-def none_safe_tuple_key(t):
+def none_safe_tuple_key(none_last: bool = True) -> Callable[[Any], Any]:
     """
-    A sort key that handles tuples containing None value.
+    A sort key factory that handles tuples containing None value.
```
hannes-ucsc: See above

src/azul/collections.py
```diff
@@ -62,35 +64,44 @@ def explode_dict(d: Mapping[K, Union[V, List[V], Set[V], Tuple[V]]]) -> Iterable
...
     """
-    return v is not None, v
+    if none_last:
+        return lambda v: (v is None, v)
+    else:
+        return lambda v: (v is not None, v)
 
 
-def none_safe_tuple_key(t):
+def none_safe_tuple_key(none_last: bool = True) -> Callable[[Any], Any]:
```
hannes-ucsc: The signature should reflect that the returned callable accepts a tuple.

src/azul/collections.py
```diff
@@ -62,35 +64,44 @@ def explode_dict(d: Mapping[K, Union[V, List[V], Set[V], Tuple[V]]]) -> Iterable
...
-    >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')], key=none_safe_tuple_key)
+    >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')], key=none_safe_tuple_key())
+    [(1, 'b'), (2, 'c'), (2, None), (None, 'a')]
+
+    >>> sorted([(2, 'c'), (None, 'a'), (2, None), (1, 'b')], key=none_safe_tuple_key(none_last=False))
     [(None, 'a'), (1, 'b'), (2, None), (2, 'c')]
     """
-    assert isinstance(t, tuple)
-    return tuple(map(none_safe_key, t))
+    # assert isinstance(t, tuple)
```
hannes-ucsc: The assert should stay. It just needs to move into the body of the returned callable.

src/azul/collections.py
hannes-ucsc: Wouldn't `False` be a better default value for `none_last`? In older Pythons with total ordering on all values, the None value would come first naturally.

src/azul/transformer.py
```diff
@@ -167,7 +176,12 @@ def translate_fields(cls,
                         new_dict[key + '_'] = val
             return new_dict
         elif isinstance(doc, list):
-            return [cls.translate_fields(val, field_types, forward=forward) for val in doc]
+            # Translate [] to [null_string] but only at the field level (to avoid case of contents['organoids'] == [])
```
```suggestion
            # Translate an empty list to a list containing a single None value (and then further translate that None value according to the field type), but do so only at the field level (to avoid case of contents['organoids'] == []).
```

And wrap at 80

src/azul/transformer.py
```diff
@@ -167,7 +176,12 @@ def translate_fields(cls,
...
+            if len(doc) == 0 and not isinstance(field_types, dict):
```
```suggestion
            if doc or isinstance(field_types, dict):
```

and swap the if-else branches.

danielsotirhos:
> Can you do [this](https://github.com/DataBiosphere/azul/pull/1587#pullrequestreview-367495515) now? It's OK to force push for that purpose.

Not sure why that commit link still works, I thought I did squash the two commits and force push.
Yes, 8bd595fe5bab96101a9992b77639cc51d4b92c5e is the new combined commit where elasticsearch_service.py no longer adds and removes the line swapping None for empty strings, and also has the change related to the `-=` operator.

hannes-ucsc:
Just squash all the fixups down and force push.

danielsotirhos:
Kept the first commit separate, however squashed the following commits into one as it was problematic with the test cases when trying to split up the commits into stand alone changes.

danielsotirhos:
Rebased on latest develop at 5e4f62442c560631022d44d5ac37997fb688048f

danielsotirhos:
> For the next review cycle, please make sure the IT passes on this branch in your personal deployment.

The following AssertionError was found in the sandbox lambda logs (CloudWatch). A similar error was experienced running the IT against my personal deployment before a reindex.

```
: AssertionError
  File "/var/task/app.py", line 156, in new_handler
    return old_handler(self, event, context)
  File "/var/task/chalice/app.py", line 1422, in __call__
    return self.func(event_obj)
  File "/var/task/app.py", line 274, in write
    indexer.aggregate(tallies)
  File "/var/task/azul/indexer.py", line 244, in aggregate
    contributions = self._read_contributions(total_tallies)
  File "/var/task/azul/indexer.py", line 305, in _read_contributions
    contributions = [Contribution.from_index(self.field_types(), hit) for hit in hits]
  File "/var/task/azul/indexer.py", line 305, in <listcomp>
    contributions = [Contribution.from_index(self.field_types(), hit) for hit in hits]
  File "/var/task/azul/transformer.py", line 254, in from_index
    source = cls.translate_fields(hit['_source'], field_types, forward=False)
  File "/var/task/azul/transformer.py", line 191, in translate_fields
    new_dict[key] = cls.translate_fields(val, field_type, forward=forward)
  File "/var/task/azul/transformer.py", line 191, in translate_fields
    new_dict[key] = cls.translate_fields(val, field_type, forward=forward)
  File "/var/task/azul/transformer.py", line 202, in translate_fields
    return [cls.translate_fields(val, field_types, forward=forward) for val in doc]
  File "/var/task/azul/transformer.py", line 202, in <listcomp>
    return [cls.translate_fields(val, field_types, forward=forward) for val in doc]
  File "/var/task/azul/transformer.py", line 191, in translate_fields
    new_dict[key] = cls.translate_fields(val, field_type, forward=forward)
  File "/var/task/azul/transformer.py", line 204, in translate_fields
    assert forward
```

After a reindex on my personal deployment with `staging` DSS selected, I have confirmed this branch is able to complete the IT successfully.
[IT_results1.txt](https://github.com/DataBiosphere/azul/files/4341219/IT_results1.txt)
[IT_results2.txt](https://github.com/DataBiosphere/azul/files/4341222/IT_results2.txt)

The "DELETEME" commit adds a few extra but unnecessary assert statements used during debugging. It has not yet been deleted in case we want to include it during sandbox testing.

hannes-ucsc:
I ran `make delete` against sandbox before pushing to GL. Not exactly the same as reindexing but it eliminates the need to wait for the first build to fail in order to manually trigger `early_reindex`.

The DELETEME commit is too complex for me to just remove before merging so this branch needs to pass the sandbox again after that commit has been removed.

src/azul/transformer.py
```diff
@@ -162,7 +162,8 @@ def translate_field(cls, value: AnyJSON, field_type: FieldType, forward: bool =
     def translate_fields(cls,
                          doc: AnyJSON,
                          field_types: Union[FieldType, FieldTypes],
-                         forward: bool = True) -> AnyMutableJSON:
+                         forward: bool = True,
+                         path: list = None) -> AnyMutableJSON:
```
hannes-ucsc: Always use the generic types from `typing` for type annotation if possible.

hannes-ucsc: Note that tuples are usually cheaper than lists, especially when they are passed by value (a new copy of the value is made for each recursion level).

src/azul/transformer.py
```diff
@@ -188,7 +192,7 @@ def translate_fields(cls,
                         raise KeyError(f'Key {key} not defined in field_types')
                     except TypeError:
                         raise TypeError(f'Key {key} not defined in field_types')
-                    new_dict[key] = cls.translate_fields(val, field_type, forward=forward)
+                    new_dict[key] = cls.translate_fields(val, field_type, forward=forward, path=path + [key])
```
hannes-ucsc: This is the copy.

```suggestion
                    new_dict[key] = cls.translate_fields(val, field_type, forward=forward, path=(**path, key))
```

should be more efficient.

src/azul/transformer.py
hannes-ucsc: Also, tuples are immutable so you can use the empty tuple as the default value and …

(I feel like I've already had this conversation.)

src/azul/transformer.py
```diff
@@ -171,13 +172,16 @@ def translate_fields(cls,
...
         :param forward: If we should translate forward or backward (aka un-translate)
+        :param path:
         :return: A copy of the original document with values translated according to their type
         """
         if field_types is None:
             return doc
         elif isinstance(doc, dict):
             new_dict = {}
             for key, val in doc.items():
+                if path is None:
```
hannes-ucsc: … eliminate this.

danielsotirhos:
> I commented on the last commit, in case we decide to leave the additional assertions in. You say they are unnecessary. Could you elaborate?

The DELETEME additions to translate_fields() were useful in debugging however it adds no assertions on the document data that wasn't already present. It does add help in tracking down what part of a document tripped the assertion, but it is at the cost of maintaining the `path` variable with a list of keys as the function descends into the document.

Similarly the addition to `from_index()` was useful in debugging although the assertion it adds simply catches illegal values (unexpected empty list) before it would be caught by the assertion in `translate_fields()`

danielsotirhos:
I locally made the changes to the DELETEME commit and ran a profile of unittests in /test/indexer twice without and twice with the DELETEME changes. The results appear to be indistinguishable from standard fluctuation of testing times.  
profile results

danielsotirhos:
@hannes-ucsc Based on the profile results with and without the commit, would you like me to delete the DELETEME commit or push the changes do you want me to delete the DELETEME commit, or rename and update it with the `Tuple` changes to the `path` parameter?

PR #1597, opened 2020-02-22
Simplify indexing retry that allows tallies of 0 (#1570)

jessebrennan:
This simplifies the retry mechanism for duplicate notifications. It also clarifies why a tally might be 0.

### DIFF
--- a/src/azul/transformer.py
+++ b/src/azul/transformer.py
@@ -7,6 +7,10 @@ from collections import (
     Counter,
     defaultdict,
 )
+from enum import (
+    Enum,
+    auto,
+)
 import logging
 import sys
 from typing import (
@@ -26,6 +30,7 @@ from typing import (
 from dataclasses import (
     dataclass,
     fields,
+    field,
 )
 from humancellatlas.data.metadata import api
 from more_itertools import one
@@ -76,12 +81,25 @@ FieldTypes1 = Union[Mapping[str, FieldTypes2], FieldType]
 FieldTypes = Mapping[str, FieldTypes1]
 
 
+class VersionType(Enum):
+    # No versioning; document is created or overwritten as needed
+    none = auto()
+
+    # Writing a document fails with 409 conflict if one with the same ID already
+    # exists in the index
+    create_only = auto()
+
+    # Use the Elasticsearch "internal" versioning type
+    # https://www.elastic.co/guide/en/elasticsearch/reference/5.6/docs-index_.html#_version_types
+    internal = auto()
+
+
 @dataclass
 class Document:
     entity: EntityReference
     type: ClassVar[str] = 'doc'
+    version_type: VersionType = field(init=False)
     version: Optional[int]
-    version_type: ClassVar[Optional[str]] = None
     contents: Optional[JSON]
 
     @classmethod
@@ -226,14 +244,12 @@ class Document:
                    **cls._from_source(source))
         return self
 
-    def to_index(self, field_types, bulk=False, update=False) -> dict:
+    def to_index(self, field_types, bulk=False) -> dict:
         """
         Build request parameters from the document for indexing
 
         :param field_types: A mapping of field paths to field type
         :param bulk: If bulk indexing
-        :param update: If creation failed and we need to try update. This will only affect docs with version_type None
-                       (AKA contributions)
         :return: Request parameters for indexing
         """
         delete = self.delete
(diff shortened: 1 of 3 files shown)


src/azul/transformer.py
```diff
@@ -269,6 +295,12 @@ class Contribution(Document):
     bundle_version: BundleVersion
     bundle_deleted: bool
 
+    def __post_init__(self):
```
noah-aviel-dove: Nice, I didn't know about this feature.

src/azul/transformer.py
```diff
@@ -76,12 +78,23 @@ class DocumentCoordinates(NamedTuple):
 FieldTypes = Mapping[str, FieldTypes1]
 
 
+class VersionType(Enum):
+    # No versioning; document is created or overwritten as needed
+    none = 1
```
noah-aviel-dove: Since the values of these constants are arbitrary, I would avoid writing them out. See https://docs.python.org/3.6/library/enum.html#omitting-values

src/azul/transformer.py
```diff
@@ -269,6 +295,12 @@ class Contribution(Document):
     bundle_version: BundleVersion
     bundle_deleted: bool
 
+    def __post_init__(self):
+        # Cannot provide a default for version_type at the class level due to
+        # limitations with @dataclass. Its value may change to VersionType.none
```
noah-aviel-dove: I was initially confused by this comment because I thought the two sentences were related. 

```suggestion
        # limitations with @dataclass.
        # May change to VersionType.none
```

src/azul/transformer.py
jessebrennan: While I agree, @hannes-ucsc requested writing them explicitly.

src/azul/transformer.py
jessebrennan: I see your point. I'll break it up.

src/azul/transformer.py
```diff
@@ -243,19 +254,34 @@ def to_index(self, field_types, bulk=False, update=False) -> dict:
...
             '_id' if bulk else 'id': self.document_id
         }
-        if self.version_type is None:
-            # TODO: FIXME: (jesse) why only for bulk???
-            if False and bulk:
+        if self.version_type is VersionType.none:
+            assert self.version is None
+            if bulk:
                 result['_op_type'] = 'delete' if delete else 'index'
-            op_key = '_op_type' if bulk else 'op_type'
```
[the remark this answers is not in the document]
noah-aviel-dove: In fact, I would probably remove all of the `_key if bulk else key` redundancies in this method using a local function like this:

```
def result_key(keyname):
  bulk_keys = {
     'index': '_index',
     'doc_type': '_type',
     'body': '_source',
     'id': '_id',
     'op_type': '_op_type',
     'version_type': '_version_type', 
     'version': '_version'
  }
  return bulk_keys[keyname] if bulk else keyname
```

And then:

```
result = { 
  result_key('index'): self.document_index,
  result_key('doc_type'): self.type
  ...
}
op_key = result_key('op_type')
if self.version_type is VersionType.none:
  assert self.version is None
  if bulk:
    result[op_key] = 'delete' if delete else 'index'
  ...
    result[result_key('version_type')] = 'internal'
    ...
```
This isn't as important for the keys that are only used in one place though.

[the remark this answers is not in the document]
noah-aviel-dove: Actually, if it were 100% up to me, I'd create separate methods for the bulk and non-bulk case, or create classes to streamline handling each case (replace these nested `if` with polymorphic method calls). But I'm not convinced that's consistent with design decision precedent for Azul so I won't request it. If you think it's a good idea though you should do it since this method is the one part of the code where these changes _increased_ complexity.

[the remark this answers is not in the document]
jessebrennan: You inspired me to try a completely different approach. I tried first splitting on `if bulk:` I think the makes the `result_key()` method redundant (though I liked the idea). I think this is more clear than before, but not as extreme as making separate classes.

src/azul/transformer.py
```diff
@@ -243,19 +254,34 @@ def to_index(self, field_types, bulk=False, update=False) -> dict:
...
-            elif self.version is None and not delete:
+                result['_op_type'] = 'delete' if delete else 'create'
+            else:
                 result['op_type'] = 'create'
-            result['_version_type' if bulk else 'version_type'] = self.version_type
+        elif self.version_type is VersionType.internal:
+            if bulk:
+                result['_op_type'] = 'delete' if delete else ('create' if self.version is None else 'index')
+            elif not delete:
+                if self.version is None:
```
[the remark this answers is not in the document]
jessebrennan: The answer to your original question was so that the comment bellow was more accurate. With my refactor though, this point is more or less void.

jessebrennan:
@noah-aviel-dove, actually I just had a chat with @hannes-ucsc about this and he said he preferred `to_index()` the way it was. Your suggestions were a matter of personal preference which I appreciate and value. Since Hannes had a hand in how this code was written, he prefers to leave it the way it was.

hannes-ucsc:
When doing peer reviews, focus on correctness, test coverage and adherence to the contributing guides. Leave the nitpicking and matters of personal style to me. Frequently in software engineering, there is a trade-off to be made or a balance to be found. What that trade-off is or where the balance is struck depends on experience and often is based on intuition. If a single person makes those choices we can at least hope to make them consistently across the code base and over time.

src/azul/transformer.py
hannes-ucsc: Maybe we should do what @noah-aviel-dove suggests. I thought using auto() for a three-item enum is overkill but this sets a bad precedent. We'd have to define in CONTRIBUTING when to use auto() and when not to. Explicitly assigning a value suggests that the value is significant, which I don't think it is in this case.

src/azul/transformer.py
```diff
@@ -76,12 +78,23 @@ class DocumentCoordinates(NamedTuple):
 FieldTypes = Mapping[str, FieldTypes1]
 
 
+class VersionType(Enum):
+    # No versioning; document is created or overwritten as needed
+    none = 1
+    # Writing a document fails with 409 conflict if one with the same ID already
+    # exists in the index
+    create_only = 2
+    # Use the Elasticsearch "internal" versioning type
```
hannes-ucsc: Newline after each item

```suggestion
    
    # Use the Elasticsearch "internal" versioning type
```

jessebrennan: Added these changes.

src/azul/transformer.py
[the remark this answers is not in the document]
hannes-ucsc: The `and` was broken up so we get an `else` branch to put a `pass` and a comment in.

Issue #1527, opened 2020-02-01
Change storage backend bucket environment variable name  


jessebrennan:
Currently the code calls this the `terraform_backend_bucket` but this isn't reflective of all of its purposes.

PR #1598, opened 2020-02-22
Change storage backend bucket environment variable name (#1527)

jessebrennan:
~Notice the FIXME that still needs to be addressed.~

### DIFF
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -337,9 +337,13 @@ class Config:
 
     @property
     def terraform_backend_bucket(self) -> str:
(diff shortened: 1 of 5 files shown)


src/azul/__init__.py
```diff
@@ -649,7 +653,8 @@ def github_access_token(self) -> str:
 
     @property
     def portal_db_bucket_name(self) -> str:
-        return self.terraform_backend_bucket
+        # FIXME: (Jesse) I'm confused why this used the backend bucket
```
GPelayo: I am also curious why there are three properties returning  `AZUL_BACKEND_BUCKET`.

PR #1599, opened 2020-02-24
Disable cart endpoints (#1513)

GPelayo:
- [x] Remove route decorators for endpoints
- [x] Prevent cart related resources from being built in Terraform
- [x] Disable Cart Related Tests
- [x] Ignore cart related code for coverage


achave11:
Another option would be to move the code to the attic, specify the Azul home path to the file where the removal is happening in the file name for what is going in the attic.

GPelayo:
Hey @achave11. As per the advice of @hannes-ucsc, we are not going to disable the deployment of any cart related resources in this ticket. The commenting out code will just be uncommented.

hannes-ucsc:
Superseded by #1614

PR #1603, opened 2020-02-25
Use AWS CLI cache for ESClientFactory.get() (#1602)

jessebrennan:


src/azul/es.py
```diff
@@ -1,7 +1,10 @@
 import logging
 from functools import lru_cache
 
-from aws_requests_auth.boto_utils import BotoAWSRequestsAuth
+from aws_requests_auth.boto_utils import (
```
achave11: Check [Contributing](https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#12imports).

Issue #1604, opened 2020-02-25
Ensure that = in access key for DRS does not cause problems


jessebrennan:
Also the = can probably just be removed and inferred from the length of the key.

PR #1610, opened 2020-03-02
Strip trailing = from DRS access key (#1604)

jessebrennan:


src/azul/drs.py
```diff
@@ -120,20 +120,25 @@ def encode_access_id(token_str: str) -> str:
     """
     Encode the token as an access id in a URL safe way.
 
-    >>> encode_access_id('tedeschi trucks')
-    'dGVkZXNjaGkgdHJ1Y2tz'
+    Notice that base64 padding is stripped off
```
danielsotirhos: This comment sits atop an example using `urlsafe_b64encode` so either it needs more details to explain that the following example does not strip padding but that `encode_access_id` does,  or be moved so it sits above the `encode_access_id` example.
```suggestion
    Notice that with `encode_access_id` base64 padding is stripped off
```

jessebrennan: Good point.

hannes-ucsc:
Made a mistake and merged without going through sandbox. Will revert if the develop build fails on Gitlab.

Issue #1557, opened 2020-02-09
Be less demanding when waiting for queues to be non-empty


hannes-ucsc:
```
2020-02-09 22:48:16,356 INFO    MainThread: Waiting up to 120 seconds for 2 queues to not be empty ...
2020-02-09 22:48:16,367 DEBUG   MainThread: Queue azul-notify-dev has 96 message(s) (89 available, 7 in flight and 0 delayed).
2020-02-09 22:48:16,392 DEBUG   MainThread: Queue azul-documents-dev.fifo has 0 message(s) (0 available, 0 in flight and 0 delayed).
2020-02-09 22:48:16,392 INFO    MainThread: Counting 96 message(s) in 2 queue(s).
2020-02-09 22:48:16,392 INFO    MainThread: The most recently sampled queue sizes are deque([96], maxlen=10).
2020-02-09 22:48:21,406 DEBUG   MainThread: Queue azul-notify-dev has 94 message(s) (81 available, 13 in flight and 0 delayed).
2020-02-09 22:48:21,429 DEBUG   MainThread: Queue azul-documents-dev.fifo has 12 message(s) (0 available, 0 in flight and 12 delayed).
2020-02-09 22:48:21,430 INFO    MainThread: Counting 106 message(s) in 2 queue(s).
2020-02-09 22:48:21,430 INFO    MainThread: The most recently sampled queue sizes are deque([96, 106], maxlen=10).
2020-02-09 22:48:26,446 DEBUG   MainThread: Queue azul-notify-dev has 84 message(s) (78 available, 6 in flight and 0 delayed).
2020-02-09 22:48:26,471 DEBUG   MainThread: Queue azul-documents-dev.fifo has 108 message(s) (0 available, 0 in flight and 108 delayed).
2020-02-09 22:48:26,471 INFO    MainThread: Counting 192 message(s) in 2 queue(s).
2020-02-09 22:48:26,472 INFO    MainThread: The most recently sampled queue sizes are deque([96, 106, 192], maxlen=10).
2020-02-09 22:48:31,486 DEBUG   MainThread: Queue azul-notify-dev has 71 message(s) (62 available, 9 in flight and 0 delayed).
2020-02-09 22:48:31,508 DEBUG   MainThread: Queue azul-documents-dev.fifo has 187 message(s) (1 available, 11 in flight and 175 delayed).
2020-02-09 22:48:31,508 INFO    MainThread: Counting 258 message(s) in 2 queue(s).
[...]

PR #1616, opened 2020-03-05
Make queue waiting less demanding (#1557)

danielsotirhos:


danielsotirhos:
Ok. Went with that implementation to avoid the `deque_size` value `10` needing to be specified in both `_queue_empty_timeout()` and `wait_for_queue_level()`, but the value is always either `10` or `1` based on the `empty` parameter so it's a bit redundant I suppose.

The `deque_size` parameter has been removed.

test/integration_test.py
```diff
@@ -350,11 +352,10 @@ def _prune_test_bundles(self, bundle_fqids, max_bundles):
 
     def _queue_empty_timeout(self, num_bundles: int):
         time_per_bundle = 5
-        time_for_deque_to_empty = 10 * 5  # time for a 0 count to propagate through the deque in wait_for_queue_level()
-        min_timeout = 60
-        max_timeout = 10 * 60
-        timeout = max((num_bundles * time_per_bundle) + time_for_deque_to_empty, min_timeout)
-        return min(timeout, max_timeout)
+        deque_size = 10
```
hannes-ucsc: This reveals an ugly brittle coupling. While this is not Please eliminate this method and the `timeout` argument to `wait_for_queue_level`. Replace it with a `num_bundles` argument that defaults to `None` (meaning "assume many bundles"). In `wait_for_queue_level`, derive the timeout from that `num_bundles` argument.

hannes-ucsc: It's only fair that this is fixed in this PR because the coupling appears to have been introduced in https://github.com/DataBiosphere/azul/commit/87aff4ddb02e989d603dda7ca87c0b774caff638.

Issue #1626, opened 2020-03-10
Get API Gateway ID from request header


jessebrennan:
Currently this is obtained from an environment variable.

jessebrennan:
Also proxy OpenAPI spec from API Gateway with service at `/openapi/raw` for `apidev.py` script.

PR #1627, opened 2020-03-11
Get Gateway ID from app.current_request, not environment (#1626)

jessebrennan:


scripts/apidev.py
```diff
@@ -1,24 +1,25 @@
...
 from azul.openapi import annotated_specs
 
 spec_file = 'openapi.json'
 parent_dir = os.path.realpath(os.path.dirname(__file__))
 
 
-def write_specs(gateway_id, app, openapi_spec):
-    specs = annotated_specs(gateway_id, app, openapi_spec)
+def write_specs(raw_specs, app, openapi_spec):
+    specs = annotated_specs(copy.copy(raw_specs), app, openapi_spec)
```
hannes-ucsc: This seems like a unrelated change or fix to an unrelated issue. Please elaborate here or in a code comment as to why A) a copy is being made and B) the copy is only shallow?

jessebrennan: I made this change because I was concerned that it would prevent the `apidev.py` script from displaying the correct specifications in some cases. After looking into this more deeply, I don't believe this would be the case.

If it was necessary however, it should have been a deep copy.

jessebrennan: There is a separate issue here however, which I will address in a separate commit

jessebrennan: On further reflection, I don't believe there's an issue here.

Issue #556, opened 2018-11-15
Use Terraform for deploying lambdas


hannes-ucsc:
Chalice repeatedly borked deployments with renamed lambdas or modified triggers, requiring manual intervention post deployment. Terraform is much better at determining the delta between what is currently deployed and what should be deployed, and the resulting action plan i.e., what needs to be deleted, renamed and created. We can still use Chalice for creating the distribution (`chalice package`) but the actual cloud resource deployment could be performed by Terraform.

http://www.squidopus.net/technology/1643/

This would also eliminate the [chicken-egginess](https://github.com/DataBiosphere/azul#23-provisioning-stable-api-domain-names) with respect to SSL certificates/DNS records (deployed by Terraform) and API gateways (deployed by Chalice).

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-355)
┆Project Name: azul
┆Issue Number: AZUL-355

hannes-ucsc:
Chalice now supports this natively: https://chalice.readthedocs.io/en/latest/topics/tf.html?highlight=terraform

THX to @jessebrennan for the tip

PR #1629, opened 2020-03-12
Use Terraform for deploying lambdas (#556)

jessebrennan:
TODO:

- [ ] Edit deployment documentation to reflect new process
- [ ] Remove old deployment scripts after all existing deployments have been moved over. Hey if we're redoing all of the deployments, should be take this opportunity to change `qualified_resource_name` to put deployment stage first?
- [ ] Update `.gitlab-ci.yml`


jessebrennan:
@achave11, yes deployed when there is already an existing deployment is more complicated. You have to tear down all of the lambda resources and then rebuild it with Terraform.

Issue #1193, opened 2019-08-05
Replace all uses of str.format() with f-strings


hannes-ucsc:
Also replace idiotic use of format with string concatenation. For example,replace `'{}y'.format(x)` with `x + 'y'` if x is a `str`.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-743)
┆containerName: azul
┆Issue Number: AZUL-743

PR #1631, opened 2020-03-12
Replace str.format with f-strings (#1193)

noah-aviel-dove:
A few usages of `str.format` remain in cases where it seemed preferable. In these cases, the format string is not a literal and cannot reasonably be replaced with one:

https://github.com/DataBiosphere/azul/blob/21ef13a8091a6a524682e782a4afa47faa3821e3/lambdas/service/app.py#L1185

https://github.com/DataBiosphere/azul/blob/21ef13a8091a6a524682e782a4afa47faa3821e3/src/azul/__init__.py#L163

https://github.com/DataBiosphere/azul/blob/21ef13a8091a6a524682e782a4afa47faa3821e3/src/azul/__init__.py#L289

In this case, the code was deliberately designed to avoid literals to reduce redundancy: 

https://github.com/DataBiosphere/azul/blob/21ef13a8091a6a524682e782a4afa47faa3821e3/test/integration_test.py#L441

### DIFF
--- a/src/azul/queues.py
+++ b/src/azul/queues.py
@@ -36,16 +36,16 @@ class Queues:
 
     def list(self):
         logger.info('Listing queues')
-        print('\n{:<35s}{:^20s}{:^20s}{:^18s}\n'.format('Queue Name',
-                                                        'Messages Available',
-                                                        'Messages in Flight',
-                                                        'Messages Delayed'))
+        print(f'\n{"Queue Name":<35s}'
+              f'{"Messages Available":^20s}'
+              f'{"Messages In Flight":^20s}'
+              f'{"Messages Delayed":^18s}\n')
         queues = self.azul_queues()
         for queue_name, queue in queues.items():
-            print('{:<35s}{:^20s}{:^20s}{:^18s}'.format(queue_name,
-                                                        queue.attributes['ApproximateNumberOfMessages'],
-                                                        queue.attributes['ApproximateNumberOfMessagesNotVisible'],
-                                                        queue.attributes['ApproximateNumberOfMessagesDelayed']))
+            print(f'{queue_name:<35s}'
+                  f'{queue.attributes["ApproximateNumberOfMessages"]:^20s}'
+                  f'{queue.attributes["ApproximateNumberOfMessagesNotVisible"]:^20s}'
+                  f'{queue.attributes["ApproximateNumberOfMessagesDelayed"]:^18s}')
 
     def dump(self, queue_name, path):
         sqs = boto3.resource('sqs')
--- a/src/azul/service/elasticsearch_service.py
+++ b/src/azul/service/elasticsearch_service.py
@@ -281,7 +281,7 @@ class ElasticsearchService(AbstractService):
         search_field = field_mapping[search_field] if search_field in field_mapping else search_field
         es_filter_query = self._create_query(filters)
         es_search = es_search.post_filter(es_filter_query)
-        es_search = es_search.query(Q('prefix', **{'{}'.format(search_field): _query}))
+        es_search = es_search.query(Q('prefix', **{str(search_field): _query}))
         return es_search
 
     def _apply_paging(self, es_search, pagination):
@@ -438,7 +438,7 @@ class ElasticsearchService(AbstractService):
         ):
             es_search.aggs.metric(
                 agg_name, 'cardinality',
-                field='{}.keyword'.format(cardinality),
+                field=cardinality + '.keyword',
                 precision_threshold="40000")
 
         self._annotate_aggs_for_translation(es_search)
--- a/test/service/test_pagination.py
+++ b/test/service/test_pagination.py
@@ -91,7 +91,7 @@ class PaginationTestCase(WebServiceTestCase):
         Tests that search_after pagination works for the first returned page.
         :return:
         """
-        content = requests.get("{}?sort=entryId&order=desc".format(self.get_base_url())).content
+        content = requests.get(self.get_base_url() + '?sort=entryId&order=desc').content
         json_response = json.loads(content)
         self.assert_page1_correct(json_response)
 
@@ -101,7 +101,7 @@ class PaginationTestCase(WebServiceTestCase):
         passes from and size variables.
         :return:
         """
-        content = requests.get("{}?sort=entryId&size=10&order=desc".format(self.get_base_url())).content
+        content = requests.get(self.get_base_url() + '?sort=entryId&size=10&order=desc').content
         json_response = json.loads(content)
         self.assert_page1_correct(json_response)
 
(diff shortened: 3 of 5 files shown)


lambdas/indexer/app.py
```diff
@@ -169,7 +169,7 @@ def index(event: chalice.app.SQSEvent):
     for record in event:
         message = json.loads(record.body)
         attempts = record.to_dict()['attributes']['ApproximateReceiveCount']
-        log.info(f'Worker handling message {message}, attempt #{attempts} (approx).')
+        log.info('Worker handling message %s, attempt #%s (approx).', message, attempts)
```
danielsotirhos: I thought log messages were the exception to using f-strings. I don't see it mentioned in the ticket description but if the policy regarding log messages is being changed with this ticket then the contributing guide will need to be updated also

CONTRIBUTING.rst
> We don't use f'' strings or string concatenation when interpolating dynamic values into log messages:

scripts/count-bundles.py
```diff
@@ -157,23 +157,23 @@ def get_project_bundle_count(project_name):
...
         except JSONDecodeError:
-            logger.info('Bundle Not Found')
+            log.info('Bundle Not Found')
             missing_projects.append(f'#{accountable_id} {beta_project_id}')
             continue
 
-        total_bundle_count, analysis_bundle_count = get_project_bundle_count(beta_project_name) \
-            if beta_project_id else None
+        total_bundle_count, analysis_bundle_count = (get_project_bundle_count(beta_project_name)
+                                                     if beta_project_id else None)
```
danielsotirhos:
```suggestion
        total_bundle_count, analysis_bundle_count = (get_project_bundle_count(beta_project_name)
                                                     if beta_project_id else (None, None))
```

lambdas/indexer/app.py
noah-aviel-dove: My changes extensively _removed_ the use of f-strings from log messages, except for a few cases where it seemed to seriously harm readability. I certainly didn't _add_ any f-strings in log messages. I don't understand what feedback I'm supposed to take away from this comment.

danielsotirhos: Yes sorry. I don't know how I got confused by this but I somehow thought the diff was telling me the reverse of what it actually is. Please ignore original comment.

test/service/test_pagination.py
```diff
@@ -111,16 +111,17 @@ def test_search_after_page2(self):
...
         # Fetch the second page using search_after
         search_after = json_response['pagination']['search_after']
         search_after_uid = json_response['pagination']['search_after_uid']
-        url = "{}?sort=entryId&order=desc&search_after={}&search_after_uid={}".format(self.get_base_url(),
-                                                                                      search_after, search_after_uid)
-        content = requests.get(url).content
+        content = requests.get(f'{self.get_base_url()}'
+                               f'?sort=entryId&order=desc'
+                               f'&search_after={search_after}'
+                               f'&search_after_uid={search_after_uid}').content
```
danielsotirhos: A bit cleaner, but not as uniform with all f-strings. Your thoughts? I could agree if you think all f-strings is the better approach here.
```suggestion
        content = requests.get(self.get_base_url() +
                               f'?sort=entryId&order=desc'
                               f'&search_after={search_after}'
                               f'&search_after_uid={search_after_uid}').content
```

noah-aviel-dove: I considered this and thought it was distractingly irregular. But now I'm having second thoughts. I think I'll commit it separately and leave it up to Hannes.

test/service/test_pagination.py
```diff
@@ -129,26 +130,29 @@ def test_search_after_last_page(self):
...
-        content = requests.get("{}?sort=entryId&order=asc".format(self.get_base_url())).content
+        content = requests.get(self.get_base_url() + '?sort=entryId&order=asc').content
         json_response = json.loads(content)
         self.assert_page1_correct(json_response)
         # Store the search_after for the last result of the first page.
         search_after_lrfp = json_response['pagination']['search_after']
         search_after_lrfp_uid = json_response['pagination']['search_after_uid']
-        content = requests.get("{}?sort=entryId&order=asc&search_after={}&search_after_uid={}"
-                               .format(self.get_base_url(), search_after_lrfp, search_after_lrfp_uid)).content
+        content = requests.get(f'{self.get_base_url()}'
```
danielsotirhos:
```suggestion
        content = requests.get(self.get_base_url() +
```

test/service/test_pagination.py
```diff
@@ -129,26 +130,29 @@ def test_search_after_last_page(self):
...
         search_after = json_response_second['pagination']['search_before']
         search_after_uid = json_response_second['pagination']['search_before_uid']
 
-        content = requests.get(f"{self.get_base_url()}"
-                               f"?sort=entryId"
-                               f"&order=desc"
-                               f"&search_after={search_after}"
-                               f"&search_after_uid={search_after_uid}"
-                               f"&order=desc").content
+        content = requests.get(f'{self.get_base_url()}'
```
danielsotirhos:
```suggestion
        content = requests.get(self.get_base_url() +
```

noah-aviel-dove:
@hannes-ucsc You explicitly instructed for f-strings to be replaced with concatenation in [the ticket description](https://github.com/databiosphere/azul/issues/1193#issue-476661453). I don't object to partitioning my changes into multiple PR's but if that's what you want to do the ticket should be updated.

hannes-ucsc:
I apologize. You are right, the ticket asked for string concatenation changes. I should have read the ticket more carefully. Or even better, I should not have conflated those two types of changes in a single ticket. Could you segregate the `str.format()` => f-string changes and the f-string => concatenation changes in two separate commits? As a reviewer, that would enable me to review them separately and keep my eye trained on a particular pattern.

test/service/test_pagination.py
```diff
@@ -129,14 +130,16 @@ def test_search_after_last_page(self):
...
-        content = requests.get("{}?sort=entryId&order=asc".format(self.get_base_url())).content
+        content = requests.get(self.get_base_url() + '?sort=entryId&order=asc').content
         json_response = json.loads(content)
         self.assert_page1_correct(json_response)
         # Store the search_after for the last result of the first page.
         search_after_lrfp = json_response['pagination']['search_after']
         search_after_lrfp_uid = json_response['pagination']['search_after_uid']
-        content = requests.get("{}?sort=entryId&order=asc&search_after={}&search_after_uid={}"
-                               .format(self.get_base_url(), search_after_lrfp, search_after_lrfp_uid)).content
+        content = requests.get(
```
hannes-ucsc: Let's not mix f strings and concatenation. Since you're within balanced constructs, you can split the string literals. There already is a code section like that a few lines down which sets the precedent.

src/azul/service/elasticsearch_service.py
```diff
@@ -430,7 +430,7 @@ def transform_summary(self,
         ):
             es_search.aggs.metric(
                 agg_name, 'cardinality',
-                field='{}.keyword'.format(cardinality),
+                field=f'{cardinality}.keyword',
```
noah-aviel-dove: I think this should use concatenation instead. Is it okay if I make this change in the fixup commit?

src/azul/queues.py
```diff
@@ -36,16 +36,13 @@ def __init__(self, delete=False, json_body=True):
 
     def list(self):
         logger.info('Listing queues')
-        print('\n{:<35s}{:^20s}{:^20s}{:^18s}\n'.format('Queue Name',
-                                                        'Messages Available',
-                                                        'Messages in Flight',
-                                                        'Messages Delayed'))
+        print(f'\n{"Queue Name":<35s}{"Messages Available":^20s}{"Messages In Flight":^20s}{"Messages Delayed":^18s}\n')
```
hannes-ucsc: Could we break this line up into four so it correlates more to the `print` statement in the loop?

src/azul/service/elasticsearch_service.py
hannes-ucsc: Yes.

test/service/test_pagination.py
hannes-ucsc: I like Daniel's suggestion. The first line looks different to the subsequent ones anyway.

test/service/test_pagination.py
hannes-ucsc: Same here and the other places like this.

hannes-ucsc:
There's a typo in your commit title, @noah-aviel-dove. Try to avoid those.

Issue #1443, opened 2019-11-13
Duplication of test code for ElasticSearch health check


jessebrennan:
https://github.com/DataBiosphere/azul/commit/b1ac38f576cebf9d1fd6a9e443b33d1a153f9293#diff-c217391d5a5a682082cd00492ac664a0R63

and 

https://github.com/DataBiosphere/azul/commit/b1ac38f576cebf9d1fd6a9e443b33d1a153f9293#diff-6dd868b395ebb06b586989b8639e8b73R56

As pointed out by @hannes-ucsc

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-885)
┆Project Name: azul
┆Issue Number: AZUL-885

PR #1633, opened 2020-03-13
Refactor redundant tests (#1443)

noah-aviel-dove:


test/indexer/test_health_check.py
```diff
@@ -26,21 +23,12 @@ def lambda_name(cls) -> str:
     @mock_sts
     @mock_sqs
     def test_elasticsearch_down(self):
-        self._create_mock_queues()
-        mock_endpoint = ('nonexisting-index.com', 80)
         endpoint_states = self._make_endpoint_states(self.endpoints)
```
hannes-ucsc: Why can't this line be part of the extracted method?

noah-aviel-dove: Because the service health check uses it to generate the expected API endpoints which are passed to it as the first argument: https://github.com/DataBiosphere/azul/pull/1633/files/3749c6fbfb1534e36a3a02475524cecdf0241917#diff-c217391d5a5a682082cd00492ac664a0R55

hannes-ucsc: I think we misunderstand each other. The line is identical in both places and is not used outside of the extracted method. I see no need to exclude it from the extracted method. Maybe I'm overlooking something. I'll give this refactoring a shot, that's the only way to resolve this misunderstanding, I think.

hannes-ucsc: > and is not used outside of the extracted method

I was wrong. It's return value is used outside.

Issue #1405, opened 2019-10-29
Document /repository endpoints


theathorn:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-858)
┆Project Name: azul
┆Issue Number: AZUL-858
┆Epic: OpenAPI Spec

PR #1647, opened 2020-03-19
Document /index endpoints (#1405)

noah-aviel-dove:


noah-aviel-dove:
The usage of `service_config()` is an immaculate suggestion... wish I'd found that by myself, I will definitely be updating my approach. However I'm having trouble understanding your claim that current is not accurate. When I hit the `repository/file` endpoint I get exactly what the schema describes:
Screenshot from 2020-03-19 15-09-58

I want to make sure there's not something I'm misunderstanding about the endpoint.

achave11:
Screen Shot 2020-03-20 at 11 05 56 AM

@noah-aviel-dove this is what I got for the 3rd and 4th results for the same endpoint you're showing. If you use `service_config()` this type of problem should go away. It will require a bit of ingenuity to distinguish each entity from the `translation` dictionary but I am sure you can see the pattern. Please feel free to reach out to me via slack or by using @ handle for a more immediate response.

noah-aviel-dove:
@achave11 Ah, I see how we got different results now. You're looking at the HCA service while I looked at the SC one, where those hits are far from the first page. 

I spent several hours last night trying to find a way to programmatically build the schema and couldn't see any reasonable way to do it. `service_config` defines the field names but doesn't contain any information on how they're nested, what datatype they are, or which fields are present for which endpoints. AFAICT the actual structure of the API response is determined by the classes in src/azul/service/hca_response_v5.py, which construct the objects imperatively with very little declarative structure to unpack. I did tried to write a method that could convert the `JsonObject` subclasses in that file into openapi schemas by interrogating their class attributes, but ran into a wall with `ListProperty` not providing a reliable way to determine the type of the items it contains.

So far I've been able to replace the hardcoded list of term facets with `service_config().facets`, and my `JsonObject` unpacker could be used to build the `fileTypeSummaries` and `pagination` schemas. But I remain unconvinced that it's feasible to programmatically build a schema for the hit entries. I remain open to being proven wrong if you can suggest a novel approach.

lambdas/service/app.py
```diff
@@ -546,204 +546,229 @@ def repository_search(entity_type: str, item_id: str):
...
+                    'application/json': {
+                        'schema': page_schema
+                    }
+                }
+            }
+        }
+    }
+
+
+def repository_id_spec(entity_type: str):
```
achave11: This method takes in a parameter in order to be used by other entities, but I notice that only `files` uses it. Is that intended? I don't believe it is, but if so maybe there is no need to pass in an entity.

lambdas/service/app.py
```diff
@@ -546,204 +546,229 @@ def repository_search(entity_type: str, item_id: str):
...
-    the facets and/or table data
+    A dictionary with the entries to be used when generating the facets and/or
+    table data and can be used by the browser to display the data and facets.
     """
     return repository_search('files', file_id)
 
 
-@app.route('/repository/samples', methods=['HEAD', 'GET'], cors=True)
+@app.route('/repository/samples', methods=['HEAD', 'GET'], method_spec=repository_search_spec('sample'), cors=True)
 @app.route('/repository/samples/{sample_id}', methods=['GET'], cors=True)
```
achave11: `repository_id_spec('sample')` ?

lambdas/service/app.py
```diff
@@ -546,204 +546,229 @@ def repository_search(entity_type: str, item_id: str):
...
-    the facets and/or table data
+    A dictionary with the entries to be used when generating the facets and/or
+    table data and can be used by the browser to display the data and facets.
     """
     return repository_search('samples', sample_id)
 
 
-@app.route('/repository/bundles', methods=['HEAD', 'GET'], cors=True)
+@app.route('/repository/bundles', methods=['HEAD', 'GET'], method_spec=repository_search_spec('bundle'), cors=True)
 @app.route('/repository/bundles/{bundle_uuid}', methods=['GET'], cors=True)
```
achave11: `repository_id_spec('bundle')` ?

lambdas/service/app.py
```diff
@@ -546,204 +546,229 @@ def repository_search(entity_type: str, item_id: str):
...
+    A dictionary with the entries to be used when generating the facets and/or
+    table data and can be used by the browser to display the data and facets.
     """
-    return repository_search('bundles', bundle_uuid)
+    return repository_search('bundles', bundle_id)
 
 
-@app.route('/repository/projects', methods=['HEAD', 'GET'], cors=True)
+@app.route('/repository/projects', methods=['HEAD', 'GET'], method_spec=repository_search_spec('project'), cors=True)
 @app.route('/repository/projects/{project_id}', methods=['GET'], cors=True)
```
achave11: `repository_id_spec('project')` ?

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+                    'application/json': {
+                        'schema': hit_schema
+                    }
+                }
+            }
+        }
+    }
+
+
+repository_head_spec = {
```
achave11: Good call!

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
         raise NotFoundError(msg=e)
 
 
-@app.route('/repository/files', methods=['HEAD', 'GET'], cors=True)
-@app.route('/repository/files/{file_id}', methods=['GET'], cors=True)
+generic_dict = schema.object(additional_properties=True)
+array_of_dict = schema.array(generic_dict)
```
hannes-ucsc: Array of object would be JSON terminology. List of dict would be Python.

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+hit_schema = schema.object(
```
```suggestion
hit_spec = schema.object(
```
and so on.

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+    return {
+        'tags': ['Repository'],
+        'summary': f'Detailed information on a particular {entity_type} entity.',
+        'parameters': [
+            params.path(f'{entity_type}_id', str, description=f'The UUID of the desired {entity_type}')
+        ],
+        'responses': {
+            '200': {
+                'description': format_description(f'''
+                    This response describes a single {entity_type}, possibly at
```
hannes-ucsc: Weasel words. Does it or does it not have additional detail? I don't think it does.

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+)
+
+
+def repository_search_spec(entity_type):
+    return {
+        'tags': ['Repository'],
+        'parameters': [
+            params.query('filters',
+                         schema.optional(str),
+                         description='Filters to be applied when calling ElasticSearch.'),
```
hannes-ucsc: This is disappointingly brief (if you were a reader, unfamiliar with the API, would you know what to pass for `filters` given this documentation?) and incorrect. The `filter` parameter is a JSON structure and that needs to be expressed somehow. Either formally via JSONSchema or OpenAPI or, if that's technically not possible, verbally.

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+            params.query(
+                'order',
+                schema.optional(schema.enum('asc', 'desc')),
+                description='Whether results should be returned in ascending or descending order.'
+            ),
+            params.query(
+                'search_after',
+                schema.optional(str),
+                description='''
+                        The value of the \'sort\' field for the hit after which all
```
hannes-ucsc: Consistently use back quotes, please. Description is not very clear.

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+
+
+def repository_search_spec(entity_type):
+    return {
+        'tags': ['Repository'],
+        'parameters': [
+            params.query('filters',
+                         schema.optional(str),
+                         description='Filters to be applied when calling ElasticSearch.'),
+            params.query('size', schema.optional(int), description='Number of hits displayed per page.'),
```
hannes-ucsc: Displayed? The service does not display anything.

lambdas/service/app.py
```diff
@@ -546,204 +546,244 @@ def repository_search(entity_type: str, item_id: str):
...
+            params.query('sort', schema.optional(str), description='Which field to sort by.'),
```
hannes-ucsc: Full sentences, please, for description. What set of values is this recruited from? Can you make that more specific? The same applies to `filters` by the way.

lambdas/service/app.py
noah-aviel-dove: I found some instances in which it does. Compare 

` http 'https://service.dev.explore.data.humancellatlas.org/repository/projects?filters={"entryId":{"is":["efb89bf7-5191-4dd2-8fe6-c6cb15f38dce"]}}' | jq .hits[].projects`

With

`http 'https://service.dev.explore.data.humancellatlas.org/repository/projects/efb89bf7-5191-4dd2-8fe6-c6cb15f38dce' | jq .projects`

hannes-ucsc: I forgot about that hack. This quirk should be documented for the /projects endpoint specifically.

hannes-ucsc: Or even better, the hack could be eliminated. I don't think it would incur a big performance penalty if those additional project properties were present on each hit of the regular `/projects` response.

hannes-ucsc: Could you quickly gauge the performance impact? If it is negligible, file a ticket to add that behaviour and leave the quirk undocumented. Otherwise, document the quirk.

src/azul/openapi/params.py
```diff
@@ -15,7 +15,16 @@
 )
 
 
-def path(name: str, type_: TYPE, **kwargs: PrimitiveJSON) -> JSON:
+class JsonParam:
```
hannes-ucsc, line 18: I'm dubious whether this is needed. Can't this just make use of the `JSON` alternative of the [`TYPE` union](https://github.com/DataBiosphere/azul/pull/1647/files#diff-80fb4e5dfa369b5d82f4d6a888f0d83bR22)?

hannes-ucsc: If the answer is complicated, don't respond here but rather jump on a call with me. I may also just want to play with this myself once the remaining issues are addressed.

noah-aviel-dove: It's a little more complicated than that, but I was able to come up with a better, simpler solution anyway. The complication stems from the fact that OpenAPI uses the `content` property to describe JSON-formatted parameters instead of the `schema` property, so this case needs to handled separately from all non-JSON parameters.

hannes-ucsc: It appears that you were able to avoid adding another alternative to the type union exactly as I suspected. That's progress but I don't think you've found the most elegant solution yet, though. I'll comment on the latest changes.

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
+        content=schema.object(**{
+            facet: schema.optional(
+                schema.object(**{
+                    relation: schema.optional(schema.array({}))
+                    for relation in ('is', 'contains', 'within', 'intersects')
+                })
+            ) for facet in facets
+        }),
+        description=format_description('''
+            Filters to be applied when calling ElasticSearch. Omit
```
hannes-ucsc, line 698: ES is an implementation detail that clients should not care about.

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
+            parameter or use `{}` to skip filtering. The relationship
+            parameters are always nested within a singleton array, so
```
hannes-ucsc, line 700: Not true. At least the "is" arrays can have more than one element. If the range-based operators only accept one that would constitute a bug.

"The relationship parameters are always nested within" is awkwardly expressed. Not really sure what you mean there.

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
+                    criteria ("hits"). The structure of these hits is documented
+                    under the [corresponding endpoint for a specific entity]({id_spec_link}).
+
+                    The `pagination` section describes the total number of hits
+                    and total number of pages, as well as user-supplied search
+                    parameters for page size and sorting behavior. It also
+                    provides links for navigating forwards and backwards between
+                    pages of results.
+
+                    The `termFacets` section tabulates the occurrence of unique
```
hannes-ucsc, line 755: Somehow it needs to be documented that termFacets doesn't just cover the current page, but all entities matching the filter.

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
+
+                    The properties that are common to all entity types are
+                    listed in the schema below; however, additional properties
+                    may be present for certain entity types. With the exception
+                    of the {entity_type}'s unique identifier, all properties are
+                    arrays, even in cases where only one value is present.
+
+                    The inner structures of the objects within these arrays are
+                    not consistent between entity types, e.g. projects and
+                    files, and in a very small number of cases, not consistent
```
hannes-ucsc, line 789: What cases? We need to fix these inconsistencies.This is exactly the type of language that is sure to instil confusion among clients. If we can't fix them, we need to be more specific in documenting them.

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
+)
+
+page_spec = schema.object(
+    hits=schema.array(hit_spec),
+    pagination=generic_object_spec,
+    termFacets=generic_object_spec
+)
+
+
+def repository_filters_spec(facets):
```
hannes-ucsc, line 685:
```suggestion
def filter_param_spec(facets):
```

Filters are used not only by the repository endpoints.

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
+                    different files.
+                '''),
+                **schema.wrap_json_schema_content(hit_spec)
+            }
+        }
+    }
+
+
+repository_head_spec = {
+    'summary': 'Used by the /health endpoint for monitoring application status.',
```
hannes-ucsc, line 800: That's an implementation detail.

```suggestion
    'summary': 'Perform a query without returning its result.',
```

The description should say sth like "Can be used to check the validity of query parameters or to test whether the {entity_type} index is operational."

lambdas/service/app.py
```diff
@@ -655,204 +661,237 @@ def repository_search(entity_type: str, item_id: str):
...
-          type: string
-          description: The value of the 'sort' field for the hit before which all results should be returned.  Not valid
-          to set both this and search_after.
-        - name: search_before_uid
-          in: query
-          type: string
-          description: The value of the elasticsearch UID corresponding to the hit above, if search_before is set.
-    :return: Returns a dictionary with the entries to be used when generating
-    the facets and/or table data
+    A dictionary with the entries to be used when generating the facets and/or
```
hannes-ucsc, line 815: These gutted docstrings don't add anything and are duplicative.

src/azul/openapi/schema.py
```diff
@@ -22,6 +22,10 @@
 TYPE = Union[None, Type, str, JSON]
 
 
+def wrap_json_schema_content(json):
```
hannes-ucsc, line 25: I have my doubts about this. For one, it is not used for composing a schema so it's not in the right module.

I also see only three call sites so I am not sure its even worth it. Especially considering that the name of the method is half as long as its body.

But leave this as is for now. I reserve the right to refactor it.

src/azul/openapi/params.py
```diff
@@ -54,15 +54,18 @@ def query(name: str, type_: Union[TYPE, schema.optional], **kwargs: PrimitiveJSO
     return _make_param(name, in_='query', type_=type_, **kwargs)
 
 
-def _make_param(name: str, in_: str, type_: Union[TYPE, schema.optional], **kwargs: PrimitiveJSON) -> JSON:
+def _make_param(name: str, in_: str, type_: Union[TYPE, schema.optional], **kwargs: AnyJSON) -> JSON:
     is_optional = isinstance(type_, schema.optional)
     if is_optional:
         type_ = type_.type_
+    type_property = (schema.wrap_json_schema_content(kwargs.pop('content'))
```
hannes-ucsc, line 61: This is too much hidden and brittle magic for my taste. Popping the `content` key only to then add it again seems awkward. The brittleness arises from the fact that the condition looks at the type but `if` branch then acts on the kwargs.

src/azul/openapi/schema.py
noah-aviel-dove: There are many other endpoints where it could be used, and it saves one level of indentation in cases where the JSON spec is inlined instead of using a variable. A function is also less error prone than repeatedly typing nested string/dict literals.

lambdas/service/app.py
noah-aviel-dove: @hannes-ucsc Is this inconsistency worth documenting or should I remove my allusion to it?

hannes-ucsc: Samples are polymorphic. Any biomaterial that yields a cell suspension which yields a sequence file will be considered a *sample*. In the current HCA metadata schema, a sample can be either a specimen, an organoid or a cell line. Its content, when it occurs as an inner entity, depends on the concrete type of entity it represents. And that explains the variation you are observing. The discriminating property is `sampleEntityType`. Only on `/repository/samples` can we be sure that each`hits[*].samples` is a singleton and that its sole element represents an individual sample. For other repository endpoints, that document path—while still pointing at a a singleton list—represents an aggregate of potentially many samples that are all associated with the given outer entity represented by the hit. That's why `hits[*].samples[0].sampleEntityType` could be a list with more than one entry on those endpoints.

src/azul/openapi/schema.py
hannes-ucsc: Makes sense. When you introduce such an improvement you should either refactor all potential usages or create a ticket to remind us to do so. I will be doing the former in a separate commit to your branch.

lambdas/service/app.py
```diff
@@ -655,204 +660,236 @@ def repository_search(entity_type: str, item_id: str):
...
+    termFacets=generic_object_spec
+)
+
+
+def filters_param_spec(facets):
+    return params.query(
+        'filters',
+        schema.optional(
+            params.json_type(
+                'filters',
```
hannes-ucsc, line 689: The fact that the parameter name occurs twice may be a hint that the current implementation of `json_type` is not the end of the rope as far as simplicity is concerned. I think I have an idea on how to improve it and will be adding it in a separate commit on your branch.

lambdas/service/app.py
hannes-ucsc: Now that I put some effort into explaining this, please incorporate that explanation into the documentation so that our users benefit from it as well.

src/azul/openapi/params.py
```diff
@@ -35,7 +47,7 @@ def path(name: str, type_: TYPE, **kwargs: PrimitiveJSON) -> JSON:
     return _make_param(name, in_='path', type_=type_, **kwargs)
 
 
-def query(name: str, type_: Union[TYPE, schema.optional], **kwargs: PrimitiveJSON) -> JSON:
+def query(name: str, type_: Union[TYPE, schema.optional], **kwargs: AnyJSON) -> JSON:
```
hannes-ucsc, line 50: Can you point me at one call site where a keyword argument is passed that is not a primitive JSON value? If there is none, this should be reverted.

src/azul/openapi/params.py
```diff
@@ -54,15 +66,18 @@ def query(name: str, type_: Union[TYPE, schema.optional], **kwargs: PrimitiveJSO
     return _make_param(name, in_='query', type_=type_, **kwargs)
 
 
-def _make_param(name: str, in_: str, type_: Union[TYPE, schema.optional], **kwargs: PrimitiveJSON) -> JSON:
+def _make_param(name: str, in_: str, type_: Union[TYPE, schema.optional], **kwargs: AnyJSON) -> JSON:
```
hannes-ucsc, line 69: Same here.

lambdas/service/app.py
```diff
@@ -655,204 +655,233 @@ def repository_search(entity_type: str, item_id: str):
...
+def filters_param_spec(facets):
+    return params.query(
+        'filters',
+        schema.optional(
+            application_json(
+                schema.object(**{
+                    facet: schema.optional(
+                        schema.object(**{
+                            relation: schema.optional(schema.array({}))
+                            for relation in ('is', 'contains', 'within', 'intersects')
```
hannes-ucsc, line 688: I don't think this is correct. Only [one of](https://json-schema.org/understanding-json-schema/reference/combining.html#oneof) the relations is allowed.

lambdas/service/app.py
```diff
@@ -655,204 +655,233 @@ def repository_search(entity_type: str, item_id: str):
...
+                        })
+                    ) for facet in facets
```
hannes-ucsc, line 690: This results in a very verbose example. I think you should hard-code a more typical one.

hannes-ucsc:
@noah-aviel-dove, for reasons to complicated to explain here, I did three separate reviews just now and would expect all three of them to be addressed.

src/azul/openapi/params.py
noah-aviel-dove: It will be reverted. This change was made when I passed `content` as a keyword arg for the `filters` parameter, and I forgot to revert it when I dropped that approach.

lambdas/service/app.py
```diff
@@ -655,204 +655,256 @@ def repository_search(entity_type: str, item_id: str):
...
+                    # schema.object(**{
+                    #   facet: schema.optional({
+                    #       'oneOf': [
+                    #           schema.object(**{rel: schema.array({})})
+                    #           for rel in ['is', 'contains', 'within', 'intersects']
+                    #       ]
+                    #   })
+                    #   for facet in facets
+                    # })
+                    # But Swagger does not informatively render the `oneOf`
```
hannes-ucsc, line 696: But the Swagger doc explicitly mentions that you can use these qualifiers: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/ 

Is it possible that when you say "render", you're referring to the generated example instead of the how the schema is rendered? The unmerged PR I was referring to is about the generated examples: https://github.com/swagger-api/swagger-ui/pull/5530#issuecomment-531365696

If so, we should definitely use them. And I think you may have missed

https://github.com/DataBiosphere/azul/pull/1647#discussion_r406567769

addressing which potentially side-steps all of this by providing a hand-made example.

noah-aviel-dove: I've implemented the manual examples. Oddly, Apidev reports structural errors, claiming that the manual example feature is incompatible with the `content` parameter. The rendered UI seems to work fine though.

lambdas/service/app.py
```diff
@@ -655,204 +655,270 @@ def repository_search(entity_type: str, item_id: str):
...
+            `{"fileFormat": {"is": ["fastq", "fastq.gz"]}}` will select entities
+            where the file format is either "fastq" or "fastq.gz". For the other
+            relationships, the parameters are pairs specifying upper and lower
+            bounds, and multiple parameter pairs are combined using "and" logic.
+            For example, `{"donorCount": {"within": [[1,5], [5,10]]}}` will
+            select entities whose donor organism count falls within both ranges,
+            i.e., is exactly 5.
+
+        '''),
+        # apidev.py claims that this parameter should not be provided when using
```
hannes-ucsc, line 720: This is a bit hand-waving. If this is indeed a claim made by `apidev.py`, could you file an issue and turn this into a FIXME comment? Otherwise, could you please point to the component that actually makes this claim?

lambdas/service/app.py
```diff
@@ -655,204 +655,270 @@ def repository_search(entity_type: str, item_id: str):
...
+                        ]
+                    })
+                    for facet in facets
+                })
+            )
+        ),
+        description=format_description('''
+            Criteria to filter entities from the search results.
+
+            Each filter consists of a facet, a relationship, and an array of
```
hannes-ucsc, line 704:
```suggestion
            Each filter consists of a facet, a relational operator, and an array of
```

here and then just `operator` henceforth.

hannes-ucsc:
```suggestion
            Each filter consists of a facet name, a relationship, and an array of
```

lambdas/service/app.py
```diff
@@ -655,204 +655,270 @@ def repository_search(entity_type: str, item_id: str):
...
+            relationship parameters. The available relationships are "is",
```
hannes-ucsc, line 705:
```suggestion
            facet values. The available relationships are "is",
```

src/azul/openapi/params.py
hannes-ucsc: You said this would be reverted?

src/azul/openapi/responses.py
```diff
@@ -1,16 +1,27 @@
 from azul.openapi import (
     format_description_key,
     schema,
+    application_json,
 )
 from azul.openapi.schema import (
-    TYPE,
```
hannes-ucsc, line 6: Could you revert this and fix your PyCharm settings? The recommended PyCharm formatter settings are in the repo at the project root so you can import them. I think they specify the trailing comma. Let me know if they don't.

src/azul/openapi/params.py
hannes-ucsc: > It will be reverted.

In that case, why is that change still here?

lambdas/service/app.py
noah-aviel-dove: From the localhost Swagger UI:

```
Structural error at paths./repository/files.get.parameters.0
should not have `style`, `explode`, `allowReserved`, `example`, or `examples` when `content` is present
Jump to line 483
```
Repeated for every route that uses the filters param.

noah-aviel-dove: Perhaps I should have been more clear that the error comes from within Swagger, not the code in our script.

src/azul/openapi/responses.py
noah-aviel-dove: My formatter settings are up to date. The issue is that the pertinent formatter setting is only triggered for _multiline_ `from` imports:

```
        <option name="FROM_IMPORT_TRAILING_COMMA_IF_MULTILINE" value="true"/>
```

I ran a test:
```
$ head -8 src/azul/openapi/params.py
from typing import (
    Union
)

from azul.openapi import (
    format_description_key,
    schema
)
$ make format
...
$ head -8 src/azul/openapi/params.py
from typing import (
    Union
)

from azul.openapi import (
    format_description_key,
    schema,  # <- comma only added here
)

File: src/azul/openapi/params.py (hunk 24 above)
Comment (noah-aviel-dove) [reply src/azul/openapi/params.py:38]: It was reverted [here](https://github.com/DataBiosphere/azul/pull/1647/commits/523b5d36f1202ea097641faf1a6a1759421f4c74) following your refactoring how the `content` parameter is handled, and then un-reverted [here](https://github.com/DataBiosphere/azul/pull/1647/commits/681e178dbcb2ff21be92fbb825fedcb28af5717f) to accommodate the `examples` parameter.

File: src/azul/openapi/responses.py (hunk 32 above)
Comment (hannes-ucsc) [reply src/azul/openapi/responses.py:6]: I think some team members may not use these settings, which could explain the inconsistencies. I don't catch them all. Until we have a process to enforce this, here's what *should* happen: multiline imports (including multiline imports with just a single imported symbol) should have a trailing comma after each symbol.

File: src/azul/openapi/params.py (hunk 24 above)
Comment (hannes-ucsc) [reply src/azul/openapi/params.py:38]: Sounds good.

File: lambdas/service/app.py (hunk 29 above)
Comment (hannes-ucsc) [reply lambdas/service/app.py:717]: Yes. If a library used by a scripts rejects an input, the rejection should be attributed to the library, not the script, or at least not only the script.

Comment (hannes-ucsc) [reply lambdas/service/app.py:717]: To fix the validation error, the example must be nested underneath `application/json` as [documented](https://swagger.io/docs/specification/adding-examples/).

Comment (hannes-ucsc) [reply lambdas/service/app.py:717]: Oh and this might lead to a reversion of the reversion of the reversion of that type hint.

File: lambdas/service/app.py [hunk 33]
```diff
@@ -745,204 +745,249 @@ def repository_search(entity_type: str, item_id: str):
...
+def repository_search_spec(entity_type):
+    id_spec_link = f'#operations-Repository-get_repository_{entity_type}s__{entity_type}_id_'
+    facets = sorted(valid_facets())
+    return {
+        'summary': f'Search the {entity_type}s index for entities of interest.',
+        'tags': ['Repository'],
+        'parameters': [
+            filters_param_spec(facets),
+            params.query(
+                'size',
```
hannes-ucsc, line 817: We should document the default value.

lambdas/service/app.py
```diff
@@ -745,204 +745,249 @@ def repository_search(entity_type: str, item_id: str):
...
+                schema.optional(int),
+                description='The number of hits included per page.'),
+            params.query(
+                'sort',
```
hannes-ucsc, line 821: If there is a default, we should document it. If there isn't, we should enforce one. I have the feeling that the default order is what we get back from ES which may be by document ID but I am not sure. If you want to punt this to another ticket, feel free to do so. In fact, I would prefer to punt if it's anything but a trivial fix to this PR.

lambdas/service/app.py
```diff
@@ -745,204 +745,249 @@ def repository_search(entity_type: str, item_id: str):
...
+                schema.object(**{'is': schema.array({})}),
+                *[
+                    schema.object(**{op: array_pair_spec})
+                    for op in ['contains', 'within', 'intersects']
+                ]
+            ]
+        })
+        for facet in facets
+    })
+    filter_schema['example'] = {'cellCount': {'within': [[10000, 1.e10]]}}
```
hannes-ucsc, line 784: `1.e10` is a float literal but somehow the example is rendered in the editor as an integer. I'd rather not rely on that conversion (because I don't understand where and why it happens) and instead use an int literal here.

lambdas/service/app.py
noah-aviel-dove: When `default` is provided as a dictionary (`{}`), it is incorrectly rendered as `default value : OrderedMap {}` for some reason. Passing it as a string (`"{}"`) renders as expected (`default value : {}`). How do you evaluate the brittleness of this workaround?

lambdas/service/app.py
noah-aviel-dove: At first glance there doesn't appear to be one (I can verify that results are NOT sorted by `entryId` by default). Ticket opened: [#1782 ](https://github.com/DataBiosphere/azul/issues/1782)

noah-aviel-dove: I've realized my current documentation of the `order` parameter is incorrect and that the solution to #1782 (which will fix this mistake) is indeed a low hanging-fruit. As such I'll be resolving that ticket in this PR instead of opening a separate one.

hannes-ucsc:
```
(.venv) hannes@hannes-xps:~/workspace/hca/azul$ git prsq
Auto-merging src/azul/openapi/responses.py
Auto-merging src/azul/openapi/params.py
CONFLICT (content): Merge conflict in src/azul/openapi/params.py
Auto-merging lambdas/service/app.py
CONFLICT (content): Merge conflict in lambdas/service/app.py
error: could not apply a90a48a4... fixup! Document repositoy endpoints (#1405)
Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".
Could not apply a90a48a4... fixup! Document repositoy endpoints (#1405)
(.venv) hannes@hannes-xps:~/workspace/hca/azul$ cat ~/.gitconfig | grep prsq
        prsq = "!f() { GIT_SEQUENCE_EDITOR=: git rebase -i --autosquash ${1:-develop}; }; f"
```

Before asking for another review, please ensure the squashability with the `prsq` alias.

noah-aviel-dove:
```
$ git prsq
git: 'prsq' is not a git command. See 'git --help'.

The most similar command is
        push
```

lambdas/service/app.py
```diff
@@ -737,204 +737,252 @@ def repository_search(entity_type: str, item_id: str):
...
+    facets = sorted(valid_facets())
+    sort_default, order_default = sort_defaults[entity_type + 's']
+    return {
+        'summary': f'Search the {entity_type}s index for entities of interest.',
+        'tags': ['Repository'],
+        'parameters': [
+            filters_param_spec(facets),
+            params.query(
+                'size',
+                schema.optional(schema.with_default(int, default=10)),
```
hannes-ucsc, line 813:
```suggestion
                schema.optional(schema.with_default(int, 10)),
```

hannes-ucsc: Also consider that `int, 10` is quite redundant. The type can often be derived from the default.

noah-aviel-dove:
Comments addressed and fixup commits properly squashed.

lambdas/service/app.py
```diff
@@ -758,57 +758,69 @@ def repository_search(entity_type: str, item_id: str):
...
-            `{"is":["foo", "bar"]}` will check for equality with either "foo" or
-            "bar". For the other relationships, the parameters are pairs
-            specifying upper and lower bounds, and multiple parameters are
-            combined using "and" logic. For example,
-            `{"within":[[1, 20], [10, 50]]}` will check for falling within both of
-            the specified ranges, i.e. between 10 and 20.
-        '''))
+            Criteria to filter entities from the search results.
+
+            Each filter consists of a facet, an operator, and an array of facet
```
hannes-ucsc, line 784: https://github.com/DataBiosphere/azul/pull/1647#discussion_r410340722

Please make sure my change requests are honoured either by applying the requested changes or by replying with a comment. If I see this occurring more frequently, my reviews are going to get very, very focused and there are going to be many more cycles.

lambdas/service/app.py
```diff
@@ -758,57 +758,69 @@ def repository_search(entity_type: str, item_id: str):
 
 
 def filters_param_spec(facets):
+    array_pair_spec = schema.array({}, minItems=2, maxItems=2)
+
+    filter_schema = schema.object(**{
```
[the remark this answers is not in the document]
noah-aviel-dove: Unfortunately there are still places in app.py where we have to use `**`, for example to handle the hyphens in `Retry-After`. But obviously it's futile to try adding special behavior to handle every case in which a JSON property  isn't a valid Python identifier.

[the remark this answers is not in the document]
noah-aviel-dove: With `is` brought into alignment with the other operators, I'm now realizing that this snippet is much easier to read when written without the list comprehension, which eliminates another usage of `**`.

noah-aviel-dove:
Since #1782 now has its own commit, I am moving it to another branch as per the contributing guide, and restoring the original estimate.

lambdas/service/app.py
```diff
@@ -758,57 +758,69 @@ def repository_search(entity_type: str, item_id: str):
...
+                schema.object(**{'is': schema.array({})}),
+                *[
+                    schema.object(**{op: array_pair_spec})
+                    for op in ['contains', 'within', 'intersects']
+                ]
+            ]
+        })
+        for facet in facets
+    })
+    filter_schema['example'] = {'cellCount': {'within': [[10000, 1000000000]]}}
```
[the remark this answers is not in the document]
noah-aviel-dove: I don't think you looked at the most recent commit. https://github.com/DataBiosphere/azul/pull/1647/files#diff-a3dec4f6b35a08e7cb29086cb8766c3aR764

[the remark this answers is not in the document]
hannes-ucsc: I'm not looking at a specific commit, Noah. I'm looking at all commits combined. Specifically, this URL

https://github.com/DataBiosphere/azul/pull/1647/files

THE VERY LINE I was commenting in this thread on modifies `filter_schema` retroactively and the diff I sent eliminates the need to do that:

image

When I send a patch you should always copy it and apply it to your working copy. Ask on the channel how to do that with PyCharm if you don't know.

lambdas/service/app.py
hannes-ucsc: Please address ALL my previous comments on this `description` section.

hannes-ucsc:
Also, there is a typo in a commit title.

hannes-ucsc:
I deleted the branch and closed this PR by accident. Restored and reopened.

This is labelled as chain but there are no blocked PRs.

noah-aviel-dove:
> I deleted the branch and closed this PR by accident. Restored and reopened.
> 
> This is labelled as chain but there are no blocked PRs.

It was blocking https://github.com/DataBiosphere/azul/pull/1798 and https://github.com/DataBiosphere/azul/pull/1799 but they were closed.

lambdas/service/app.py
noah-aviel-dove: Since it ended up having it's own commit, which is bad code smell for a freebie according to the contributing guide, #1782 is now a chained PR.

lambdas/service/app.py
```diff
@@ -762,204 +757,252 @@ def repository_search(entity_type: str, item_id: str):
...
+    return params.query(
+        'filters',
+        schema.optional(application_json(filter_schema)),
+        description=format_description('''
+            Criteria to filter entities from the search results.
+
+            Each filter consists of a facet name, a relational operator, and an
+            array of facet values. The available operators are "is", "within",
+            "contains", and "intersects". See the `sort` parameter for a list of
+            filter-able facets.
```
hannes-ucsc, line 809:
```suggestion
            filter-able facets. Multiple filters are combined using "and" logic. An entity must match all filters to be included in the response. How multiple facet values within a single filter are combined depends on the operator.
```

lambdas/service/app.py
```diff
@@ -762,204 +757,252 @@ def repository_search(entity_type: str, item_id: str):
...
+
+            For the "is" operator, multiple facet values are combined using "or"
```
hannes-ucsc, line 811:
```suggestion
            For the "is" operator, multiple values are combined using "or"
```

lambdas/service/app.py
```diff
@@ -762,204 +757,252 @@ def repository_search(entity_type: str, item_id: str):
...
+            logic. For example, `{"fileFormat": {"is": ["fastq", "fastq.gz"]}}`
+            selects entities where the file format is either "fastq" or
+            "fastq.gz". For the other operators, the facet values must come
```
hannes-ucsc, line 814:
```suggestion
            "fastq.gz". For the "within", "intersects" and "contains" operators, the facet values must come
```

We may add other operators.

lambdas/service/app.py
```diff
@@ -762,204 +757,252 @@ def repository_search(entity_type: str, item_id: str):
...
+            in nested pairs specifying upper and lower bounds, and multiple
+            pairs are combined using "and" logic. For example,
```
hannes-ucsc, line 816: Please confirm that the conjunctive combination applies to "contains" and "intersects" as well. 

And I think this is a bug. For "within" this really should be "or" logic. For "intersects" and "contains" I'm not sure. Either way, please create a ticket that reminds us to reconsider the "and" logic at least for "within".

noah-aviel-dove: Confirmed for `contains` and `intersects`. https://github.com/DataBiosphere/azul/issues/1833

hannes-ucsc:
Changed the target branch of #1782 and #1793 to develop as per blocking relationship:

image

Deleting branch now.

PR #1506, opened 2019-12-10
Cache `bdbag` and `compact` manifest using object_key (#1476)

achave11:

PR #1648, opened 2020-03-19
Cache `terra.bdbag` and `compact` manifest formats

achave11:
Continued from #1506

### DIFF
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -76,13 +76,27 @@ class ManifestService(ElasticsearchService):
         super().__init__()
         self.storage_service = storage_service
 
-    def get_manifest(self, format_: str, filters: Filters):
+    def get_manifest(self, format_: str, filters: Filters) -> Tuple[str, bool]:
+        """
+        Returns a tuple in which the first element is pre-signed URL to a
+        manifest in the given format and of the file entities matching the given
+        filter. If a suitable manifest already exists, its location will be
+        returned. Otherwise, a new manifest will be generated and its location
+        returned. Subsequent invocations of this method with the same arguments
+        are likely to reuse that manifest, skipping the time-consuming manifest
+        generation.
+
+        The second element of the returned tuple is True if an existing
+        manifest was reused and False if a new manifest was generated.
+        """
         generator = ManifestGenerator.for_format(format_, self, filters)
         object_key, presigned_url = self._get_cached_manifest(generator, format_, filters)
         if presigned_url is None:
             file_name = self._generate_manifest(generator, object_key)
             presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
-        return presigned_url
+            return presigned_url, False
+        else:
+            return presigned_url, True
 
     def get_cached_manifest(self, format_: str, filters: Filters) -> Optional[str]:
         generator = ManifestGenerator.for_format(format_, self, filters)
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -79,8 +79,9 @@ class ManifestService(ElasticsearchService):
     def get_manifest(self, format_: str, filters: Filters):
         generator = ManifestGenerator.for_format(format_, self, filters)
         object_key, presigned_url = self._get_cached_manifest(generator, format_, filters)
-        file_name = self._generate_manifest(generator, object_key)
-        presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
+        if presigned_url is None:
+            file_name = self._generate_manifest(generator, object_key)
+            presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
         return presigned_url
 
     def get_cached_manifest(self, format_: str, filters: Filters) -> Optional[str]:
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -79,10 +79,8 @@ class ManifestService(ElasticsearchService):
     def get_manifest(self, format_: str, filters: Filters):
         generator = ManifestGenerator.for_format(format_, self, filters)
         object_key, presigned_url = self._get_cached_manifest(generator, format_, filters)
-        if presigned_url is None:
-            object_key = f'manifests/{uuid.uuid4()}.{generator.file_name_extension}' if not object_key else object_key
-            file_name = self._generate_manifest(generator, object_key)
-            presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
+        file_name = self._generate_manifest(generator, object_key)
+        presigned_url = self.storage_service.get_presigned_url(object_key, file_name=file_name)
         return presigned_url
 
     def get_cached_manifest(self, format_: str, filters: Filters) -> Optional[str]:
@@ -94,19 +92,14 @@ class ManifestService(ElasticsearchService):
                              generator: 'ManifestGenerator',
                              format_: str,
                              filters: Filters,
-                             ) -> Tuple[Optional[str], Optional[str]]:
-        if generator.is_cacheable:
-            # Assertion to be removed in https://github.com/DataBiosphere/azul/issues/1476
-            assert isinstance(generator, FullManifestGenerator)
-            manifest_key = self._derive_manifest_key(format_, filters, generator.manifest_content_hash)
-            object_key = f'manifests/{manifest_key}.{generator.file_name_extension}'
-            if self._can_use_cached_manifest(object_key):
-                file_name = self._use_cached_manifest(generator, object_key)
-                return object_key, self.storage_service.get_presigned_url(object_key, file_name=file_name)
-            else:
-                return object_key, None
+                             ) -> Tuple[str, Optional[str]]:
+        manifest_key = self._derive_manifest_key(format_, filters, generator.manifest_content_hash)
+        object_key = f'manifests/{manifest_key}.{generator.file_name_extension}'
+        if self._can_use_cached_manifest(object_key):
+            file_name = self._use_cached_manifest(generator, object_key)
+            return object_key, self.storage_service.get_presigned_url(object_key, file_name=file_name)
         else:
-            return None, None
+            return object_key, None
 
     file_name_tag = 'azul_file_name'
 
@@ -259,13 +252,6 @@ class ManifestGenerator(metaclass=ABCMeta):
     # Note to implementors: all property getters in this class and its
     # descendants must be inexpensive. If a property getter performs and
     # expensive computation or I/O, it should cache its return value.
-
-    @property
-    def is_cacheable(self) -> bool:
-        """
-        True if output of this generator can be cached.
-        """
-        return False
 
     @property
     @abstractmethod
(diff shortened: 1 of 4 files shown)


test/service/test_manifest.py
```diff
@@ -198,7 +200,9 @@ def _assert_tsv(self, expected, actual):
 
     @mock_sts
     @mock_s3
-    def test_manifest_zarr(self):
+    @mock.patch('azul.service.manifest_service.ManifestService._can_use_cached_manifest')
+    def test_manifest_zarr(self, can_use_cached_manifest):
+        can_use_cached_manifest.return_value = False
```
hannes-ucsc: displaces docstring

test/service/test_manifest.py
```diff
@@ -1089,41 +1099,50 @@ def test_manifest_content_disposition_header(self, _can_use_cached_manifest):
...
-        # ... invalidating the cached object previously used for the same filter.
-        self.assertNotEqual(old_bundle_object_key, new_bundle_object_key)
-
-        # Updates or additions, unrelated to that project do not affect object key generation
-        self._index_canned_bundle(("f79257a7-dfc6-46d6-ae00-ba4b25313c10", "2018-09-14T133314.453337Z"))
-        generator = manifest_service.FullManifestGenerator(service, filters)
-        latest_bundle_object_key = service._derive_manifest_key(format_, filters, generator.manifest_content_hash)
-
-        self.assertEqual(latest_bundle_object_key, new_bundle_object_key)
+        for format_ in ['compact', 'full', 'terra.bdbag']:
```
hannes-ucsc: Why is this done in three smaller loops rather than just one big loop?

test/service/test_manifest.py
```diff
@@ -503,7 +509,9 @@ def test_terra_bdbag_manifest(self):
 
     @mock_sts
     @mock_s3
-    def test_full_metadata(self):
+    @mock.patch('azul.service.manifest_service.ManifestService._can_use_cached_manifest')
```
hannes-ucsc: I'm a little doubtful about the blanket patch of `_can_use_cached_manifest`. Some tests, like this one, pass without this patch. You should be more deliberate in your choice in which tests need patching. If multiple tests need the same patch, you should write a custom patch decorator and explain the reason for the patch in a comment or docstring. If I remove this patch from all methods, the ones that fail all fail in the same way: a KeyError due to a missing `Expiration` key in moto's mock `HEAD` response.  A more surgical patch might therefore be one that targets `StorageService.head` to return include the `Expiration` key. The advantage of a more surgical patch is that more code is covered by unit tests.

test/service/test_manifest.py
achave11: It has to do with indexing. In the iteration for the first format, the indexer would go through all object key update scenarios (old bundle, new bundle, non related bundle). Leaving the consecutive formats with a non updatable index. Unless the index is cleared at the end of each iteration.

test/service/test_manifest.py
```diff
@@ -82,6 +86,24 @@ def _get_manifest_url(self, format_, filters):
...
+class TestManifestEndpoints(ManifestTestCase):
+
+    def run(self, result: Optional[unittest.result.TestResult] = ...) -> Optional[unittest.result.TestResult]:
+        """
+        This patch introduces a context manager at the unittest level `run`. Allowing for
+        test in this suite to 'run' in the patched context.
+        The context manager patches 'ManifestService._can_use_cached_manifest' in order to
+        prevent a cached manifest from being used. It allows for testing of manifest
+        generation while ignoring cache functionality.
+        """
```
hannes-ucsc:
```suggestion
        # Suppress caching of manifests to prevent false assertion positives
```

Generally speaking, this is not a comment for a docstring. It's also verbose and mostly obvious.

test/service/test_manifest.py
```diff
@@ -82,6 +86,24 @@ def _get_manifest_url(self, format_, filters):
...
+        with mock.patch(
+            'azul.service.manifest_service.ManifestService._can_use_cached_manifest'
+        ) as mock_can_use_cached_manifest:
```
hannes-ucsc:
```suggestion
        ) as m:
```

Remember, the smaller the scope, the shorter your variable names can be.

test/service/test_manifest.py
```diff
@@ -82,6 +86,24 @@ def _get_manifest_url(self, format_, filters):
...
+            mock_can_use_cached_manifest.return_value = False
+            run_ = super().run(result)
```
hannes-ucsc:
```suggestion
            return super().run(result)
```

Be sure to understand why the variable is unnecessary.

test/service/test_manifest.py
```diff
@@ -975,6 +997,55 @@ def test_full_metadata_missing_fields(self):
...
+                    (
+                        {'project': {'is': ['Single of human pancreas']}},
+                        'Single of human pancreas 1985-10-25 01.21'
+                    ),
+                    # In all other cases, the standard content disposition file name
+                    # should be "hca-manifest-" followed by the manifest key, a
+                    # v5 UUID deterministically derived from the filter and
+                    (
+                        {'project': {'is': ['Single of human pancreas', 'Mouse Melanoma']}},
+                        f'hca-manifest-{filtered_uuids[single_part]}',
```
hannes-ucsc: replace the dict lookup with a conditional. To prevent and overly long line it may help to convert the f string into concatenation

src/azul/service/manifest_service.py
```diff
@@ -419,6 +407,30 @@ def _dss_url(self, file):
         dss_url = config.dss_endpoint + '/' + path
         return dss_url
 
+    @cachedproperty
+    def manifest_content_hash(self) -> int:
+        es_search = self._create_request()
+        es_search.aggs.metric('hash',
```
hannes-ucsc: [My suggestion](https://github.com/DataBiosphere/azul/pull/1648#discussion_r398185311) indented differently and I still prefer it that way.

hannes-ucsc: And then make all scripts triple-single-quoted strings for symmetry.

test/service/test_manifest.py
```diff
@@ -82,6 +86,15 @@ def _get_manifest_url(self, format_, filters):
...
         return service.get_manifest(format_, filters)
 
+
+class TestManifestEndpoints(ManifestTestCase):
+
+    def run(self, result: Optional[unittest.result.TestResult] = ...) -> Optional[unittest.result.TestResult]:
+        # Suppress caching of manifests to prevent false assertion positives
+        with mock.patch('azul.service.manifest_service.ManifestService._can_use_cached_manifest') as m:
+            m.return_value = False
+        return super().run(result)
```
hannes-ucsc: Still wrong. I think you're confused about the evaluation order of return statements in context managers but that's essential Python knowledge that you need to understand. Of course, this could just be an oversight, I don't know.

Anyways, now I have my doubts if this is even effective. I would like you leave this broken as is and add an assertion to one of the tests that this should—but currently doesn't—patch. The assertion should fail if a cached manifest is reused in the test. I need to see the Travis build fail on that assertion. Then we can fix problem.

achave11: @hannes-ucsc yes this is completely a mishap on my behave. 
For your doubts, if I may remind you, we added `Cache single and multipart manifests separately` meaning we use single part upload boolean as a parameter to determine the object key. Being the reason why this test passed with out having to use the deactivated use cached manifest. 
I'll still add the assertion.

test/service/test_manifest.py
```diff
@@ -184,6 +197,11 @@ def test_manifest(self):
                         self.assertEqual(200, response.status_code, 'Unable to download manifest')
                         self._assert_tsv(expected, response)
 
+            # The following assertion fails due to ManifestService attempt to use a cached manifest rather a
```
hannes-ucsc, line 200: This comment is not accurate. As you can see in https://travis-ci.org/github/DataBiosphere/azul/builds/669322139#L6920, the build does not fail at the assertion, it fails in the line above the assertion.

test/service/test_manifest.py
```diff
@@ -184,6 +197,11 @@ def test_manifest(self):
...
+            # freshly generated one.
+            response = self._get_manifest('compact', {})
+            self.assertEqual(200, response.status_code, 'Unable to download manifest')
```
hannes-ucsc, line 203: This does not assert that the response was freshly generated.

 I think your proof works but only because I understand the details here and know how to interpret the KeyError. No one else is going to be able to follow your thought process in the future. That's why this change needs to be dropped which you anticipated by labelling the commit as such.

But can you assert the freshness of the manifest in a way that's accurately documented and that can remain in the codebase? I would compare the two signed URLs that come back from the service. Maybe there is a difference between them that indicates freshness?

I have seen evidence for the effectiveness of your patch in the `run()` override. If the above can't be achieved, I'd be happy to move on.

achave11: @hannes-ucsc  
No information regarding cache utilization can be extracted from manifest generation response. The generator will use different pre-signed urls regardless of whether a cached manifest is used or not. Also, manifest object_key/file_name will always be the same provided that the `filters`, `format` and `enable_multipart_upload` boolean are consistent. 
However, this could change if you decide it is better to separate the object_key generation from the pre-signed url generation, currently in the same method. With a separation in place, we could distinguish methods from taking a 'utilize cached manifest' path to a 'generate manifest' one, removing the need for the patch.

src/azul/service/manifest_service.py
```diff
@@ -79,10 +79,8 @@ def __init__(self, storage_service: StorageService):
     def get_manifest(self, format_: str, filters: Filters):
         generator = ManifestGenerator.for_format(format_, self, filters)
         object_key, presigned_url = self._get_cached_manifest(generator, format_, filters)
-        if presigned_url is None:
```
hannes-ucsc, line 82: The removal of this conditional is a bug. It makes no sense to look for a cached manifest only to then ignore it and generate a new one.

hannes-ucsc:
@achave11, 

Last time I looked at this PR around 11:11am today I was assigned and went ahead with my review and improvements. It appears that you then unassigned and reassigned me with at least one force push in between. 

I will force push my version of this branch now. If you just rebased but made no other changes in that force push, that should be fine since I did the same rebase. If you added changes, the onus to restore them should be on you since deassigning me was against protocol.

hannes-ucsc:
Build fails as expected and in the expected assertion: https://travis-ci.org/github/DataBiosphere/azul/builds/669895849#L6916

Other cache-related tests fail, too, but that's also expected due to the hacky way the temporary commit disarms the cache objects expiration check which would otherwise fail due to moto's inability to mock the object expiration header.

https://github.com/DataBiosphere/azul/pull/1648/commits/5c22aad92c77b83852fc5bd8ce7144149b4f4566#diff-c2aad48ac91ce2386409a633dfa0034cR217

test/service/test_manifest.py
hannes-ucsc: I think 639c075 addresses this nicely.

test/service/test_manifest.py
hannes-ucsc: @achave11 You may be right but, you see, only if you didn't make any additional mistakes. Since you made one mistake it was justified for me to assume that there were others. And that's why I had doubts and that's why I asked for the added tests.

PR #1650, opened 2020-03-19
Avoid generating manifest content hash more than once

achave11:


test/service/test_async_manifest_service.py
```diff
@@ -131,7 +131,7 @@ def run(self, result: Optional[unittest.result.TestResult] = ...) -> Optional[un
         with mock.patch('azul.service.manifest_service.ManifestService.__init__') as init_:
             init_.return_value = None
             with mock.patch('azul.service.manifest_service.ManifestService.get_cached_manifest') as get_cache_manifest:
-                get_cache_manifest.return_value = None
+                get_cache_manifest.return_value = None, None  # This returns Tuple[object_key, presign_url] of type str
```
jessebrennan: I think this comment is incorrect because the return type is a tuple of optional strings. Having said that, I think the comment is unnecessary because the return type for `get_cached_manifest()` is already documented in its definition.

test/service/test_async_manifest_service.py
```diff
@@ -131,7 +131,7 @@ def run(self, result: Optional[unittest.result.TestResult] = ...) -> Optional[un
         with mock.patch('azul.service.manifest_service.ManifestService.__init__') as init_:
             init_.return_value = None
             with mock.patch('azul.service.manifest_service.ManifestService.get_cached_manifest') as get_cache_manifest:
```
jessebrennan:
```suggestion
            with mock.patch('azul.service.manifest_service.ManifestService.get_cached_manifest') as get_cached_manifest:
```

test/service/test_async_manifest_service.py
```diff
@@ -118,25 +120,36 @@ def test_manifest_status_failed(self, step_function_helper):
 
 class TestAsyncManifestServiceEndpoints(LocalAppTestCase):
 
+    def run(self, result: Optional[unittest.result.TestResult] = ...) -> Optional[unittest.result.TestResult]:
```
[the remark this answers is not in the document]
achave11: Thanks, it was a suggestion by Hannes in a different PR. I thought this test could also benefit from that approach.

test/service/test_async_manifest_service.py
achave11: Done! :)

test/service/test_async_manifest_service.py
achave11: I shortened the variable, so it's partially addressed. :)

achave11:
> I'm confused by the commit message "Fix: Error tests passing for unintended reasons" because I'm not sure what "Error tests" are. Also I don't understand how the tests were broken before nor how your changes fix it (although perhaps I don't need to 🤷‍♂)

I attempted to address this by being more descriptive in the commit message. 

Note:  The tests being 'fixed', contain a docstring describing the type/form of the error expected. When test postfixed with `..._error` run without the patch, the status code asserted is correct but the stack trace provides a message different to what is described in the docstrigs.

test/service/test_async_manifest_service.py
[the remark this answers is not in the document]
jessebrennan: I just noticed the ellipses here. What does that mean in the context of a function definition?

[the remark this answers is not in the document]
achave11: It seems to be a place holder. [Ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis)
I use the override method option from PyCharm, that automatically wrote the method and signature, I just had to modify the content.

hannes-ucsc:
Let's wait with reviews until base PR is merged.

hannes-ucsc:
Note the `hold_warm` label. In this case, the author should hold this branch warm since I unassigned myself.

test/service/test_async_manifest_service.py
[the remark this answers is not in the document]
hannes-ucsc: This is bug in PyCharm, I think. `...` only makes sense in stubs (see https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values).

For overrides it doesn't work:

```python
class C:
...     def f(self, x=123):
...         return x + 1
... 
class D(C):
...     def f(self, x=...):
...         return x + super().f(x)
... 
D().f()
  File "<input>", line 3, in f
  File "<input>", line 3, in f
TypeError: unsupported operand type(s) for +: 'ellipsis' and 'int'
```

src/azul/service/manifest_service.py
```diff
@@ -90,7 +90,12 @@ def get_manifest(self, format_: str, filters: Filters) -> Tuple[str, bool]:
         manifest was reused and False if a new manifest was generated.
         """
         generator = ManifestGenerator.for_format(format_, self, filters)
-        object_key, presigned_url = self._get_cached_manifest(generator, format_, filters)
+
+        # Passing 'object_key' as a parameter to `get_manifest()` indicates a
```
hannes-ucsc, line 94: Rethink the wording. It is imprecise. Also, this needs to go into the docstring.

src/azul/service/manifest_service.py
```diff
@@ -90,7 +90,12 @@ def get_manifest(self, format_: str, filters: Filters) -> Tuple[str, bool]:
...
+        # cached manifest was not available, it must be generated.
+        object_key, presigned_url = self._get_cached_manifest(generator,
+                                                              format_,
+                                                              filters) if object_key is None else [object_key, None]
```
hannes-ucsc, line 98: This is a bit insidious. I had to look at it three time before I realized that the _get_cached_manifest call is conditional.

A if statement would be better.

src/azul/service/manifest_service.py
```diff
@@ -425,27 +429,22 @@ def _dss_url(self, file):
...
-                return params._agg.fields.hashCode()
-            ''',
-            reduce_script='''
-                int result = 0;
-                for (agg in params._aggs) {
-                    result += agg
-                }
-                return result
-          ''')
+        es_search.aggs.metric('hash',
```
hannes-ucsc, line 432: Why does this undo a change I had requested in a previous PR?

achave11: It was a bad conflict resolution. I had fixed it and failed to catch it again on the rebase of the latest changes.

achave11:
> [#1650 (comment)](https://github.com/DataBiosphere/azul/pull/1650#discussion_r404496939)?

https://github.com/DataBiosphere/azul/pull/1650#discussion_r409156922
>It was a bad conflict resolution. I had fixed it and failed to catch it again on the rebase of the latest changes.

src/azul/service/async_manifest_service.py
```diff
@@ -52,8 +52,8 @@ def decode_token(cls, token: str) -> MutableJSON:
     def start_or_inspect_manifest_generation(self,
                                              self_url,
                                              token: Optional[str] = None,
-                                             format_: Optional[str] = None,
-                                             filters: Optional[Filters] = None,
+                                             format_: str = None,
```
hannes-ucsc, line 55: If `None` is the default, the hint needs to reflect that.

src/azul/service/async_manifest_service.py
```diff
@@ -52,8 +52,8 @@ def decode_token(cls, token: str) -> MutableJSON:
...
+                                             filters: Filters = None,
```
hannes-ucsc, line 56: Same here.

PR #1652, opened 2020-03-21
Link to /dss/files for response description (#1642)

noah-aviel-dove:
Note that anchor links do not work within the swagger editor. To observe the new functionality, use [my personal deployment](https://service.nadove.dev.singlecell.gi.ucsc.edu/).

Also note that the link opens in a new tab. This is obviously gross but there doesn't seem to be any way around it, although I'd be happy to be proven wrong.

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -1134,17 +1134,19 @@ def dss_files(uuid):
     'responses': {
         '200': {
             'description': format_description('''
-                DSS checkout with status report, emulating the response code and
-                headers of the `/dss/files/` endpoint. Note that the actual HTTP
-                response will have status 200 while the `Status` field of the
-                body will be 301 or 302. The intent is to emulate HTTP while
-                bypassing the default client behavior, which (in most web
-                browsers) is to ignore `Retry-After`. The response described
-                here is intended to be processed by client-side Javascript such
-                that the recommended delay in `Retry-After` can be handled in
-                Javascript rather that relying on the native implementation by
-                the web browser. See `/dss/files` for the meaning of the
-                response elements.
+                Emulates the response code and headers of the
+                [`/dss/files/`](#operations-DSS-get_dss_files__uuid_) endpoint
+                using JSON (response elements are documented there). Note that
+                the actual HTTP response will have status 200 while the `Status`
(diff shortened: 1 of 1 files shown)


lambdas/service/app.py
```diff
@@ -1097,8 +1097,8 @@ def dss_files(uuid):
                 here is intended to be processed by client-side Javascript such
                 that the recommended delay in `Retry-After` can be handled in
                 Javascript rather that relying on the native implementation by
-                the web browser. See `/dss/files` for the meaning of the
-                response elements.
+                the web browser. See [/dss/files](#/DSS/get_dss_files__uuid_)
+                for the meaning of the response elements.
```
```suggestion
                the web browser.

                For a detailed description of the fields in the response see
                the documentation for the headers they emulate in the
                [`/dss/files`](#operations-DSS-get_dss_files__uuid_)
                endpoint response.                
```
This anchor link will jump to the endpoint without a new tab, tested in Chrome & FireFox. The link target is in reference to the `id` field specified in the outer div of the endpoint block:
```
<div class="opblock opblock-get is-open" id="operations-DSS-get_dss_files__uuid_">
```


Also, the first paragraph is a bit much for the reader to take in, so I recommend putting the link in a second paragraph with some verbiage to reiterate the relationship between the fields of this response to the headers of the other endpoint's response.

noah-aviel-dove: Awesome, thanks. I did try looking at the source but somehow missed that ID field.

lambdas/service/app.py
```diff
@@ -1134,17 +1134,18 @@ def dss_files(uuid):
...
-                here is intended to be processed by client-side Javascript such
-                that the recommended delay in `Retry-After` can be handled in
-                Javascript rather that relying on the native implementation by
-                the web browser. See `/dss/files` for the meaning of the
-                response elements.
+                Emulates the response code and headers of the
+                [`/dss/files/`](#operations-DSS-get_dss_files__uuid_) endpoint
+                using JSON (response elements are documented there). Note that
+                the actual HTTP response will have status 200 while the `Status`
+                field of the body will be 301 or 302.\n
```
```suggestion
                field of the body will be 301 or 302.

```
I saw the pep8 warning on the last push but this should work with an empty line, just be sure to remove all whitespace from the line.

PR #1635, opened 2020-03-16
Use Terraform for deploying lambdas (#556)

jessebrennan:
This is a continuation of #1629. It's unclear to me both why @hannes-ucsc closed the previous PR, and why I was unable to reopen it.

TODO:

- [x] Edit deployment documentation to reflect new process
- [x] Remove old deployment scripts after all existing deployments have been moved over. Hey if we're redoing all of the deployments, should be take this opportunity to change `qualified_resource_name` to put deployment stage first?
- [x] Update `.gitlab-ci.yml`

Issue #1272, opened 2019-09-12
Use Lambda layers to speed up `make deploy`


hannes-ucsc:
… as demoed by DSS team.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-788)
┆Project Name: azul
┆Issue Number: AZUL-788

Issue #1654, opened 2020-03-24
Rename `deploy` target since it doesn't actually deploy anything


jessebrennan:
This isn't actually an issue at time of creating, but it will be once #556 is resolved.

jessebrennan:
@amarjandu suggested `build` which I think is fitting.

hannes-ucsc:
`build` is ambiguous. We say "Travis build", "Gitlab build", "the build failed" and so on. In a way running make with any target could qualify as a "build".

I like `package`.

PR #1655, opened 2020-03-24
Use Lambda layers to speed up `make deploy` (#1272)

jessebrennan:
~This branch is currently chained off of #1635~


scripts/provision_layers.py
```diff
@@ -0,0 +1,73 @@
...
+            if e.response['Error']['Code'] == '404':
+                return True
+        else:
+            return False
+
+    def update_layer(self, bucket, key):
+        input_zip = self.out_dir / 'deployment.zip'
+        layer_zip = self.out_dir / 'layer.zip'
+        self.build_layer_zip(layer_zip, input_zip)
+        self.client.upload_file(str(layer_zip), bucket, key)
```
amarjandu: does terraform handle the layer registration? there might need to be some layer publishing. https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html#configuration-layers-manage

lambdas/layer/Makefile
```diff
@@ -0,0 +1,13 @@
+all: deploy
+
+include ../../common.mk
+
+config: .chalice/config.json
+
+deploy: config
```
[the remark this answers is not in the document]
jessebrennan: I appreciate the nit. I think it's a better name. I left it at deploy for consistency with what was in the other lambdas, but we have #1654 to cover renaming. I'll note your suggestion there.

.travis.yml
```diff
@@ -23,6 +23,7 @@ branches:
     - integration
     - staging
     - prod
+    - issues/jessebrennan/556-chalice-terraform
```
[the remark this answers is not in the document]
jessebrennan: Yes. If you look at the commit history it should be clear that this is a temporary change that will be removed before this PR is merged. It's a result of this PR being "chained" onto #556. It will need to be rebased once #556 is merged, at which point I will remove the commit containing this change.

scripts/provision_layers.py
jessebrennan: Yes, Terraform handles registering the layer. The Terraform configuration for that is here https://github.com/DataBiosphere/azul/pull/1655/files#diff-6d607bdc1bf3a3126d91f66c2c027df1R1

.travis.yml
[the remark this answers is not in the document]
amarjandu: thx!

jessebrennan:
@amarjandu I responded to your comments and assigned you back for another review (with the eventual goal of approval). Please let me know if you have more in-depth questions about how this works (and Azul's deployment code in general). I'd be happy to have a call if you like. I can imagine it would be pretty difficult to do a exhaustive review on an unfamiliar code base.

jessebrennan:
I'm considering making it so that `make deploy` happens every time a `make` command the deals with Terraform is invoked.

amarjandu:
> I'm considering making it so that `make deploy` happens every time a `make` command the deals with Terraform is invoked.

meaning that everytime terraform is called to deplo (or plan); there is a check for updating the layer?
yes that would be a good idea if so, less chances for the layer to go out of sync with the deployment (and weird bugs that might pop up)

hannes-ucsc:
The base PR has landed and I switched this PRs target to `develop`. Please resolve conflicts and decide in what order you would like to address #1672 and #1673. I have the feeling that #1673 might be affected or even fixed by this PR.

scripts/provision_layers.py
```diff
@@ -0,0 +1,73 @@
...
+    def build_layer_zip(self, destination, input_zip):
+        command = f'chalice package {self.out_dir}'
+        log.info('Running: %s', command)
+        subprocess.check_call(command.split(), cwd=self.layer_dir)
+        log.info('Packaging %s', destination)
+        with ZipFile(input_zip, 'r') as deployment_zip:
+            with ZipFile(destination, 'w') as layer_zip:
+                for file in deployment_zip.namelist():
+                    if file != 'app.py':
+                        data = deployment_zip.read(file)
```
hannes-ucsc, line 58: Can't we stream files between the two archives? ZipFile.open() can return a writeable handle? If this is significantly more code, don't bother, the complexity isn't worth it.

scripts/provision_layers.py
```diff
@@ -0,0 +1,73 @@
...
+
+import boto3
+
+from azul import config
+from azul.logging import configure_script_logging
+
+log = logging.getLogger(__name__)
+
+
+class LayersProvisioner:
```
hannes-ucsc, line 16: This merely stages the layer package and TerraForm actually provisions the lambda layer, correct? If so, could the name of the class and the script reflect that?

scripts/provision_layers.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def update_required(self, bucket, key) -> bool:
+        log.info('Checking if layer.zip needs to be updated')
+        try:
+            # Since the object key is content-addressable, just checking for
+            # the file's presence is sufficient
+            self.client.head_object(Bucket=bucket, Key=key)
+        except self.client.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == '404':
+                return True
```
hannes-ucsc, line 39: else branch?

hannes-ucsc: This is where the rule to create symmetric control flow branches helps because mistakes like this attract the eye as visual violation of that rule making those things easier to catch.

src/azul/files.py
```diff
@@ -15,3 +16,8 @@ def write_file_atomically(path, mode=0o644):
     except BaseException:
         os.unlink(temp_path)
         raise
+
+
+def file_sha(path):
```
hannes-ucsc, line 21:
```suggestion
def file_sha1(path):
```

hannes-ucsc: The way this is written, its not general enough to deserve a place in this module. Either rewrite it to support streaming and therefore large files which should be relatively easy (use a chunk size 1M) or inline it. AFAICT, there is only one call site. If you go the former route, consider adding doctests, one for `"/dev/null"` and one for `__file__`. I'd also accept a docstring that warns about the caveat of memory consumption for large files but consider it lame since streaming should be easy to add.

terraform/lambda.tf.json.template.py
```diff
@@ -0,0 +1,16 @@
+from azul import config
```
hannes-ucsc, line 1: Rename to `lambda_layer…` ?

terraform/lambda.tf.json.template.py
```diff
@@ -0,0 +1,16 @@
...
+from azul.deployment import emit_tf
+
+emit_tf({
+    "resource": [
+        {
+            "aws_lambda_layer_version": {
+                "dependency_layer": {
+                    "layer_name": config.qualified_resource_name("dependency_layer"),
```
hannes-ucsc, line 9:
```suggestion
                    "layer_name": config.qualified_resource_name("dependencies"),
```

It's generally not necessary to encode the resource type in the resource name since each type defines its own isolated space for resource names.

scripts/provision_layers.py
jessebrennan: Correct. This class name is a holdover from my original implementation.

scripts/provision_layers.py
jessebrennan: Thanks for catching this. I see your point.

src/azul/files.py
jessebrennan: I considered adding a doctest for `__file__` but decided that finding a collision for sha1 was a bit out of scope for this PR.

jessebrennan:
@hannes-ucsc 
>I have the feeling this breaks certain build dependency relationships. Are there any? How could someone shoot themselves in the foot but running submakes, for example?

I don't think so. If you want to update dependencies then you run `make -C lambdas layer`. If you only want to update service then run `make -C lambdas layer`. It's true that those two steps are not interlocked, but I don't think that's a problem.

>And could you summarize how you tested this?

I set up a personal deployment using this code and made sure that some endpoints are working as expected. I am certain that code from the dependency layer is running in the lambdas.

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+            else:
+                raise
+        else:
+            return False
+
+    def update_layer(self, bucket, key):
+        input_zip = self.out_dir / 'deployment.zip'
+        layer_zip = self.out_dir / 'layer.zip'
+        log.info('Tainting old lambda layer')
+        command = ['terraform', 'taint', 'aws_lambda_layer_version.dependencies']
```
hannes-ucsc, line 53: Tainting should only be necessary when update is forced.

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+                        dst_zip_info.date_time = src_zip_info.date_time
+                        dst_zip_info.compress_type = ZIP_DEFLATED
+                        with deployment_zip.open(src_zip_info, 'r') as rf:
+                            with layer_zip.open(dst_zip_info, 'w') as wf:
+                                wf.write(rf.read())
+
+
+def main(argv):
+    parser = ArgumentParser(description='Package the requirements layer')
+    parser.add_argument('--bucket', default=config.layer_bucket, help='bucket to hold layer file')
```
hannes-ucsc, line 81: While I see the point for `--force`, I don't think `--bucket` and `--key` are necessary.

src/azul/__init__.py
```diff
@@ -655,6 +658,15 @@ def portal_db_bucket(self) -> str:
...
         return f'azul/{self.deployment_stage}/portals/{self.dss_deployment_stage(self.dss_endpoint)}-db.json'
 
+    @property
+    def layer_bucket(self) -> str:
+        return self.versioned_bucket
+
+    @property
+    def layer_object_key(self) -> str:
+        requirements_hash = file_sha1(Path(self.project_root) / 'requirements.txt')
+        return f'azul/{self.deployment_stage}/layers/{requirements_hash}.zip'
```
hannes-ucsc, line 668:
```suggestion
        return f'azul/lambda_layers/{requirements_hash}.zip'
```

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def update_layer(self, bucket, key):
+        input_zip = self.out_dir / 'deployment.zip'
+        layer_zip = self.out_dir / 'layer.zip'
+        log.info('Tainting old lambda layer')
+        command = ['terraform', 'taint', 'aws_lambda_layer_version.dependencies']
+        subprocess.run(command, cwd=Path(config.project_root) / 'terraform')
+        self.build_layer_zip(layer_zip, input_zip)
+        log.info('Uploading layer zip to S3')
+        self.client.upload_file(str(layer_zip), bucket, key)
```
hannes-ucsc, line 57: Since we removed the stage name from the key, it might make sense to [tag](https://docs.aws.amazon.com/AmazonS3/latest/dev/object-tagging.html) the object with the deployment, just so we can easily find the layer used by a deployment, when looking at the objects in the bucket directly. @natanlao is adding generic tagging so the tag name should be consistent with what he chose.

jessebrennan: I'd rather not do this for a couple reasons:
- it's already easy to find which layer is connected to which lambda from the lambda console
- since layers are now shared between deployments this creates a race on updating the tag. If the tag key contained the deployment name there would be no race, but I think this would be ugly.

hannes-ucsc: Sounds good.

src/azul/__init__.py
jessebrennan: I was considering making this a `cachedproperty` but decided not to because there is no precedence of it in this file.

hannes-ucsc: In fact, it shouldn't be a property at all. It should be a function (method) of the hash and the has should not be computed here. It's a bad smell that `config` looks at a file that's only present at build time. If I were to call this method at runtime it would fail.

UPGRADING.rst
```diff
@@ -10,6 +10,18 @@ branch that does not have the listed changes, the steps would need to be
 reverted. This is all fairly informal and loosely defined. Hopefully we won't
 have too many entries in this file.
 
+#1272 Use Lambda layers to speed up \`make deploy\`
```
hannes-ucsc, line 13:
```suggestion
#1272 Use Lambda layers to speed up ``make deploy``
```

No?

lambdas/layer/.chalice/config.json.template.py
```diff
@@ -0,0 +1,12 @@
+from azul import config
+from azul.template import emit
+
+emit({
+    "version": "2.0",
+    "app_name": config.qualified_resource_name("dummyapp"),
```
hannes-ucsc, line 6:
```suggestion
    "app_name": config.qualified_resource_name("azul-dependencies"),
```

lambdas/layer/app.py
```diff
@@ -0,0 +1,15 @@
+from azul import (
+    config,
+)
+from azul.chalice import AzulChaliceApp
+
+# This whole file only exists so that we can use Chalice to create the layer
+# package and is removed from the final result.
+
+app = AzulChaliceApp(app_name=config.indexer_name,
```
hannes-ucsc, line 9:
```suggestion
app = AzulChaliceApp(app_name=config.qualified_resource_name('dependencies'),
```

lambdas/layer/app.py
```diff
@@ -0,0 +1,15 @@
...
+                     unit_test=globals().get('unit_test', False))
```
hannes-ucsc, line 10:
```suggestion
                     unit_test=False)
```

lambdas/layer/app.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@app.route('foo')
```
hannes-ucsc, line 13:
```suggestion
@app.route('/')
```

lambdas/layer/app.py
```diff
@@ -0,0 +1,15 @@
...
+def foo():
+    return {'bar': 'foo'}
```
hannes-ucsc, line 15:
```suggestion
    return {}
```

Does this even have to return anything?

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+            else:
+                raise
+        else:
+            return False
+
+    def update_layer(self, force: bool = False):
+        input_zip = self.out_dir / 'deployment.zip'
+        layer_zip = self.out_dir / 'layer.zip'
+        if force:
+            log.info('Tainting old lambda layer')
```
hannes-ucsc, line 53:
```suggestion
            log.info('Tainting current lambda layer resource to force update.')
```

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+            command = ['terraform', 'taint', 'aws_lambda_layer_version.dependencies']
+            subprocess.run(command, cwd=Path(config.project_root) / 'terraform')
+        self.build_layer_zip(layer_zip, input_zip)
+        log.info('Uploading layer zip to S3')
```
hannes-ucsc, line 57:
```suggestion
        log.info('Uploading layer ZIP to S3')
```

https://en.wikipedia.org/wiki/Zip_(file_format)

src/azul/files.py
```diff
@@ -15,3 +16,18 @@ def write_file_atomically(path, mode=0o644):
...
+def file_sha1(path):
+    """
+    >>> file_sha1('/dev/null')
+    'da39a3ee5e6b4b0d3255bfef95601890afd80709'
+    """
+    with open(path, 'rb') as f:
+        sha1 = hashlib.sha1()
+        while True:
+            data = f.read(1024 * 1024)
+            if data == b'':
```
hannes-ucsc, line 30:
```suggestion
            if data:
```

I don't want to live in a world in which .read returned a False-ish value other than empty bytes. In that world, your version would pass that False-ish value to .update and then what?

src/azul/files.py
```diff
@@ -15,3 +16,18 @@ def write_file_atomically(path, mode=0o644):
     except BaseException:
         os.unlink(temp_path)
         raise
+
+
+def file_sha1(path):
+    """
+    >>> file_sha1('/dev/null')
```
hannes-ucsc, line 23: I could see why testing on `__file__` is not a good idea but collisions weren't my motivation for my comment so I don't really understand your [response](https://github.com/DataBiosphere/azul/pull/1655#discussion_r406519238) to it.

Can you think of an easy way to test this with a file size of 1M -1, 1M and 1M+1? How about os.urandom and BytesIO? That would still be a one-liner AFAICT.

terraform/lambda_layer.tf.json.template.py
```diff
@@ -0,0 +1,16 @@
+from azul import config
+from azul.deployment import emit_tf
+
+emit_tf({
+    "resource": [
+        {
+            "aws_lambda_layer_version": {
+                "dependencies": {
+                    "layer_name": config.qualified_resource_name("dependency_layer"),
```
hannes-ucsc, line 9:
```suggestion
                    "layer_name": config.qualified_resource_name("dependencies"),
```

hannes-ucsc: AFAICT, I requested this [already](https://github.com/DataBiosphere/azul/pull/1655#discussion_r406360545). Assuming that you simply forgot to address that comment I would recommend that you first make the change and then thumb-up the comment.

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class LayerBuilder:
+    layer_dir = Path(config.project_root) / 'lambdas' / 'layer'
+    out_dir = layer_dir / '.chalice' / 'terraform'
+
+    def __init__(self):
+        self.client = boto3.client('s3')
```
hannes-ucsc, line 25:
```suggestion
        self.s3 = boto3.client('s3')
```

Is (or should be) the convention. `client` doesn't scale to more than one boto3 client instance.

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.update_required():
+            log.info('Updating layer.zip')
+            self.update_layer()
+        else:
+            log.info('No update necessary')
+
+    def update_required(self) -> bool:
+        log.info('Checking if layer.zip needs to be updated')
+        try:
+            # Since the object key is content-addressable, just checking for
```
hannes-ucsc, line 38:
```suggestion
            # Since the object is content-addressed, just checking for
```

scripts/stage_layer.py
```diff
@@ -0,0 +1,94 @@
...
+            # the file's presence is sufficient
```
hannes-ucsc, line 39:
```suggestion
            # the object's presence is sufficient
```

src/azul/files.py
hannes-ucsc: `files.py` should now be included in the doctests suite.

UPGRADING.rst
jessebrennan: Yes. Funny to witness that strange brain malfunction of mine.

src/azul/files.py
jessebrennan: I changed this to `if not data:`

src/azul/files.py
jessebrennan: > I don't really understand your response to it.

I think you could exploit collisions in a hashing function in order to get it to contain it's own hash. I'm not sure of this anyway, but that was my thinking.

terraform/lambda_layer.tf.json.template.py
jessebrennan: I try to be careful about this. I'll be sure I've actually made the change before adding the thumb.

lambdas/layer/app.py
jessebrennan: No. I changed it to just `pass`.

terraform/lambda_layer.tf.json.template.py
hannes-ucsc: Still wrong (singular instead of plural) in latest push.

scripts/stage_layer.py
```diff
@@ -0,0 +1,102 @@
...
+        else:
+            return False
+
+    def update_layer(self, force: bool = False):
+        input_zip = self.out_dir / 'deployment.zip'
+        layer_zip = self.out_dir / 'layer.zip'
+        if force:
+            log.info('Tainting current lambda layer resource to force update')
+            command = ['terraform', 'taint', 'aws_lambda_layer_version.dependencies']
+            subprocess.run(command, cwd=Path(config.project_root) / 'terraform')
```
hannes-ucsc, line 59: Check exit code or document why its not checked, say, because the error is expected. If the latter, we may want to log a info message saying that the error is benign.

scripts/stage_layer.py
```diff
@@ -0,0 +1,102 @@
...
+        return config.layer_object_key(file_sha1(Path(config.project_root) / 'requirements.txt'))
+
+
+def main(argv):
+    parser = ArgumentParser(description='Package the requirements layer')
+    parser.add_argument('-f', '--force', action='store_true', help='force build and re-upload of layer. Also taint '
+                                                                   'Terraform resource to ensure it is re-provisioned')
+    options = parser.parse_args(argv)
+    packager = LayerBuilder()
+    if options.force:
```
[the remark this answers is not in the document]
hannes-ucsc: I think my refactoring simplifies this a bit.

src/azul/__init__.py
```diff
@@ -680,6 +680,13 @@ def portal_db_bucket(self) -> str:
     def portal_db_object_key(self) -> str:
         return f'azul/{self.deployment_stage}/portals/{self.dss_deployment_stage(self.dss_endpoint)}-db.json'
 
+    @property
+    def layer_bucket(self) -> str:
+        return self.versioned_bucket
+
+    def layer_object_key(self, requirements_hash) -> str:
```
hannes-ucsc, line 687: It's funny how a small method like this can still fail to cleanly separate concerns.

lambdas/layer/.chalice/config.json.template.py
```diff
@@ -0,0 +1,12 @@
+from azul import config
+from azul.template import emit
+
+emit({
+    "version": "2.0",
+    "app_name": config.qualified_resource_name("azul_dependencies"),
```
hannes-ucsc, line 6: Doesn't this name the app `azul_azul_dependencies`?

lambdas/layer/.chalice/config.json.template.py
```diff
@@ -0,0 +1,12 @@
...
+    "api_gateway_stage": config.deployment_stage,
+    "manage_iam_role": False,
+    "iam_role_arn": "${var.role_arn}",
```
hannes-ucsc, line 9: This may not be necessary.

lambdas/layer/.chalice/config.json.template.py
```diff
@@ -0,0 +1,12 @@
...
+    'lambda_timeout': config.api_gateway_timeout + config.api_gateway_timeout_padding,
```
hannes-ucsc, line 10: Same here. Default is fine for the dummy.

scripts/stage_layer.py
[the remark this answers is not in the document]
jessebrennan: I agree, this is neater.

scripts/stage_layer.py
```diff
@@ -0,0 +1,102 @@
...
+    layer_dir = Path(config.project_root) / 'lambdas' / 'layer'
+    out_dir = layer_dir / '.chalice' / 'terraform'
+
+    def __init__(self):
+        self.s3 = boto3.client('s3')
+
+    def update_layer_if_necessary(self):
+        log.info('Updating layer in bucket: %s with key %s', config.layer_bucket, self.object_key)
+        if self.update_required():
+            log.info('Updating layer.zip')
```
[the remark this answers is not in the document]
jessebrennan: I think this is a holdover from an old implementation.

scripts/stage_layer.py
```diff
@@ -0,0 +1,102 @@
...
+                        dst_zip_info.external_attr = src_zip_info.external_attr
+                        dst_zip_info.date_time = src_zip_info.date_time
+                        dst_zip_info.compress_type = src_zip_info.compress_type
+                        with deployment_zip.open(src_zip_info, 'r') as rf:
+                            with layer_zip.open(dst_zip_info, 'w') as wf:
+                                shutil.copyfileobj(rf, wf, length=1024 * 1024)
+
+    @cachedproperty
+    def object_key(self):
+        return config.layer_object_key(file_sha1(Path(config.project_root) / 'requirements.txt'))
```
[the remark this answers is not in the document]
jessebrennan: I was stuck on how to solve this. Moving the file I think is the right move.

scripts/stage_layer.py
jessebrennan: An error will happen if you run `--force` but nothing is deployed yet, which is a conceivable (but unlikely) use case. I'll add a log message.

lambdas/layer/.chalice/config.json.template.py
jessebrennan: Something like that yes. It also doesn't matter since the app file is deleted, but I see value in changing it.

scripts/stage_layer.py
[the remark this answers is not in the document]
jessebrennan: I take this back. This an accurate log message, but I see how it's seems to imply what you're saying.

lambdas/layer/.chalice/config.json.template.py
jessebrennan: It's not!

lambdas/layer/.chalice/config.json.template.py
jessebrennan: It's not!

src/azul/lambda_layer.py
```diff
@@ -41,18 +41,20 @@ def _update_required(self) -> bool:
...
     def update_layer(self, force: bool = False):
-        log.info('Using dependency layer package at s3://%s/%s.', config.layer_bucket, self.object_key)
+        log.info('Using dependency layer package at s3://%s/%s.', config.lambda_layer_bucket, self.object_key)
         if force or self._update_required():
             log.info('Staging layer package ...')
             input_zip = self.out_dir / 'deployment.zip'
             layer_zip = self.out_dir / 'layer.zip'
             if force:
                 log.info('Tainting current lambda layer resource to force update')
+                subprocess.run(['make', 'init'], cwd=Path(config.project_root) / 'terraform').check_returncode()
```
hannes-ucsc, line 51: We should add a `taint_dependency_layer` to the makefile and have it depend on the `init` target. This here is a little brittle.

src/azul/lambda_layer.py
```diff
@@ -41,18 +41,20 @@ def _update_required(self) -> bool:
...
                 command = ['terraform', 'taint', 'aws_lambda_layer_version.dependencies']
-                subprocess.run(command, cwd=Path(config.project_root) / 'terraform')
+                if subprocess.run(command, cwd=Path(config.project_root) / 'terraform').returncode != 0:
```
hannes-ucsc, line 53: See if `-allow-missing` can be used instead of ignoring all errors.

src/azul/lambda_layer.py
jessebrennan: Good call.

PR #1657, opened 2020-03-25
Azul Readme: notes about branch merging

amarjandu:
adds note about the branch merging section being out of date in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,6 +597,8 @@ deploy`) in the affected deployment.
 
 # 6. Branch flow & development process
 
+**This section should be considered a draft. It describes a future extension to the current branching flow.**
+
 The section below describes the flow we want to get to eventually, not the one
 we are currently using while this repository recovers from the aftermath of its
 inception.


README.md
```diff
@@ -597,6 +597,8 @@ deploy`) in the affected deployment.
 
 # 6. Branch flow & development process
 
+## Note: The section below regarding branch management is out-of-date.
```
hannes-ucsc: I wouldn't use a heading style for the note. IIRC, GH markdown does not support note boxes. If that's the case, I would use just bold.

```suggestion
## Note: This section should be considered a draft. It describes a future extension to the current branching flow.
```

amarjandu: went with bold && used suggested message change.

hannes-ucsc:
@amarjandu For !fixup commits to work as intended, the string that follows "!fixup" needs to exactly match the title of a previous commit. Otherwise, git --auto-squash won't know what original commit to squash the fixup commit with. The easiest way to achieve that is via `git commit --fixup` or right clicking on a commit in the change log view in PyCharm.

amarjandu:
> @amarjandu For !fixup commits to work as intended, the string that follows "!fixup" needs to exactly match the title of a previous commit. Otherwise, git --auto-squash won't know what original commit to squash the fixup commit with. The easiest way to achieve that is via `git commit --fixup` or right clicking on a commit in the change log view in PyCharm.

👍  thanks for letting me know i'll use `git commit --fixup` next time.

Issue #1617, opened 2020-03-06
Queue waiter should log token queue length


jessebrennan:
Only log the length; don't include it in the total.

danielsotirhos:
@jessebrennan Some questions regarding this ticket:

- `Queues.wait_for_queue_level()` is used by `reindex.py` and by `integration_test.py`. Does this ticket only relate to one of those or both? Currently `reindex.py` waits for the notify, token, and document queues; `integration_test.py` waits for the notify and document queues.

- An example of the logs produced by the queue waiter in the integration test is shown below. If another "Queue x has y message(s).." line is added, is there any recommended styling to identify which queues totals are being added into the overall total, and which queues are not?

Example queue waiter output:
```
 2020-03-21 03:25:52,887 INFO    MainThread: Waiting up to 370 seconds for 2 queues to empty ...
 2020-03-21 03:25:52,898 DEBUG   MainThread: Queue azul-notify-dev has 27 message(s) (27 available, 0 in flight and 0 delayed).
 2020-03-21 03:25:52,909 DEBUG   MainThread: Queue azul-documents-dev.fifo has 30 message(s) (0 available, 0 in flight and 30 delayed).
 2020-03-21 03:25:52,909 INFO    MainThread: Counting 57 message(s) in 2 queue(s).
 2020-03-21 03:25:52,910 INFO    MainThread: The most recently sampled queue sizes are deque([57], maxlen=10).
 2020-03-21 03:25:57,926 DEBUG   MainThread: Queue azul-notify-dev has 24 message(s) (10 available, 14 in flight and 0 delayed).
 2020-03-21 03:25:57,954 DEBUG   MainThread: Queue azul-documents-dev.fifo has 47 message(s) (0 available, 0 in flight and 47 delayed).
 2020-03-21 03:25:57,954 INFO    MainThread: Counting 71 message(s) in 2 queue(s).
 2020-03-21 03:25:57,954 INFO    MainThread: The most recently sampled queue sizes are deque([57, 71], maxlen=10).
```

jessebrennan:
> Queues.wait_for_queue_level() is used by reindex.py and by integration_test.py. Does this ticket only relate to one of those or both? Currently reindex.py waits for the notify, token, and document queues; integration_test.py waits for the notify and document queues

Both instances makes sense.

[...]

PR #1658, opened 2020-03-25
Log token queue length in queue waiter (#1617)

danielsotirhos:


danielsotirhos:
See comment in ticket about current approach of solution https://github.com/DataBiosphere/azul/issues/1617#issuecomment-604143275

danielsotirhos:
I agree `wait_for_queue_level()` could be more useful if it had the queues passed in as a parameter instead of being set to a static list of queues, but it would only makes sense if the feature was going to be used. Otherwise it's a bit of overhead for the caller of the function that might as well be avoided.

I'll keep the current approach with no queue parameters for now and let @hannes-ucsc offer his thoughts.

src/azul/queues.py
```diff
@@ -146,11 +146,12 @@ def get_queues(cls, queue_names: List[str]) -> Mapping[str, Any]:
         return {queue_name: sqs.get_queue_by_name(QueueName=queue_name) for queue_name in queue_names}
 
     @classmethod
-    def count_messages(cls, queues: Mapping[str, Any]) -> int:
+    def count_messages(cls, queues: Mapping[str, Any], log_total: bool = True) -> int:
```
hannes-ucsc: You and your modal args. A function with a complex signature is a bad smell because it indicates that the method doesn't just do one thing or doesn't do it well or is a leaky abstraction.  Logging is an auxiliary concern that shouldn't pollute method signatures unless the purpose of those methods is exclusively logging. 

In this particular case, I don't see the point. Why can't that one call site that's passing False for this argument log the total? If there's a redundant log line so be it or maybe that can be solved more elegantly.

danielsotirhos: Agreed. I got caught up in trying to have `count_messages()` support multiple use cases when I should have been considering if it needed to.

Issue #1582, opened 2020-02-18
Validate filters in manifest endpoint


noah-aviel-dove:
Following passed validation:

```
filters={"fileFormat":["pdf"]}
```

PR #1660, opened 2020-03-30
Validate filters for manifest requests (#1582)

achave11:


achave11:
@noah-aviel-dove remember we don't assign the author of the PR to the PR nor do we assign ourselves. An unassigned PR implies the author is assigned. 
Thanks for the review BTW.

lambdas/service/app.py
```diff
@@ -995,12 +995,14 @@ def handle_manifest_generation_request():
     a retry URL for the view function to handle.
     """
     query_params = app.current_request.query_params or {}
+    filters = query_params.get('filters', '{}')
```
hannes-ucsc: Swap with next line. First validate, then use.

hannes-ucsc:
I'm trying to push the commits separately but the current GH outage makes that unreliable. The [Travis build](https://travis-ci.org/github/DataBiosphere/azul/builds/670311272) for bf92eb0 has already finished but the build status still shows pending:

image

Also, note that the above screenshot shows that the force push that pushed the commit is listed after the commit, which is illogical,

achave11:
@hannes-ucsc it also looks like Travis is acting up. 
In more than one occasion now I've seen builds pass e.g. `bf92eb0` the first commit in this PR which wasn't supposed to.. As well as failures on ones that should have build.

hannes-ucsc:
I was able to produce the desired Github timeline:

image

Well, sort of. The force-push is still listed after the commits that it pushed but what matters is that the two commits are listed together, without a force-push in between. Not sure why this didn't work when you did it. I used `git push origin sha1:branch` whereas you recommitted and pushed. That's the only difference between your approach and mine. I recommend that use my approach in the future so that we can establish if that difference was causative.

Issue #1618, opened 2020-03-06
Indexing sometimes stalls


jessebrennan:
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1032

jessebrennan:
Top theory: one execution crashes or times out and this blocks all affected fifo groups for 5 min (the message visibility timeout)

hannes-ucsc:
I can correlate a stall in [this build](https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1057) with a ScanError in the logs:


```
| 122796619775:/aws/lambda/azul-indexer-dev-write | 2020-03-06 19:10:34.795 | START RequestId: d48b834f-c309-5e30-a462-d593616305ab Version: $LATEST                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
| 122796619775:/aws/lambda/azul-indexer-dev-write | 2020-03-06 19:10:38.703 | [INFO] 2020-03-06T19:10:38.703Z d48b834f-c309-5e30-a462-d593616305ab Received 7 messages for 2 token(s) with a total value of 8                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
| 122796619775:/aws/lambda/azul-indexer-dev-write | 2020-03-06 19:10:38.703 | [INFO] 2020-03-06T19:10:38.703Z d48b834f-c309-5e30-a462-d593616305ab Document grouping for messages received: {'98b65104-4450-4883-b0e4-be1cb7011d46': 1, 'a920d0c4-02fb-41cf-ba1e-28b9be8cf143': 1, '6abb9e58-282e-482c-b5c4-b851d6853a09': 1, '22a0d8a6-0bdc-4af8-8e95-cf02b252d497': 1, '19a52e02-1cda-4c75-8bef-4d2ee884cd46': 1, 'd90e6465-1649-47aa-acbb-bf17644875ed': 1, 'bde5496c-60b0-4cab-bdef-657b5b350fb5': 1}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
[...]

PR #1662, opened 2020-03-30
Retry `scan()` during indexing (#1618)

jessebrennan:
~I may have gone overboard, coping the retry mechanism from Toil, but I thought it was generally useful code, more versatile than what I would have written for this specific case.~


hannes-ucsc:
Here's another stab at a retry decorator:

https://github.com/HumanCellAtlas/data-store/blob/6b8c2d6767d55d3e6d45bd75e107d882427854e1/dss/util/retry.py

I think it's better to have the implicit lambda retry take care of this.

Issue #1661, opened 2020-03-30
Unable to run lambda locally due to AWS Credential conflict


amarjandu:
Branch: develop
AWS Account: None set during error
```
$ aws sts get-caller-identity
Unable to locate credentials. You can configure credentials by running "aws configure".


$ make local
python .chalice/config.json.template.py .chalice/config.json
Traceback (most recent call last):
  File ".chalice/config.json.template.py", line 13, in <module>
    "iam_role_arn": f"arn:aws:iam::{aws.account}:role/{config.indexer_name}",
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/boltons/cacheutils.py", line 610, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/Users/amar/dev/hca/azul/src/azul/deployment.py", line 59, in account
    return self.sts.get_caller_identity()['Account']
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/client.py", line 648, in _make_api_call
    operation_model, request_dict, request_context)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/client.py", line 667, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/endpoint.py", line 102, in make_request
    return self._send_request(request_dict, operation_model)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/endpoint.py", line 132, in _send_request
    request = self.create_request(request_dict, operation_model)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/endpoint.py", line 116, in create_request
    operation_name=operation_model.name)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/hooks.py", line 356, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
  File "/Users/amar/dev/venv/venv-azul/lib/python3.6/site-packages/botocore/hooks.py", line 228, in emit
    return self._emit(event_name, kwargs)
[...]

PR #1664, opened 2020-03-30
Check AWS Account ID (#1661)

amarjandu:
Verifies that an operators AWS Account ID matches the the branch/deployment Account ID.

Connected To: #1661

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,6 +548,10 @@ make delete
 
 # 4. Running indexer or service locally
 
+While this method *does* run the service or indexer locally on your machine, it
+still requires that the cloud resources used by them are already deployed.
+See sections [2](#2-getting-started) and [3](#3-deployment) on how to do that.
+ 
 1. As usual, activate the virtual environment and `source environment` if you
    haven't done so already
 
--- a/common.mk
+++ b/common.mk
@@ -33,6 +33,16 @@ ifneq ($(shell python -c "import wheel as w; \
 $(error Looks like the `wheel` package is outdated or missing. See README for instructions on how to fix this.)
 endif
 
+# This check should not occur within CI environments, where AWS Credentials might not be supplied
+ifneq ($(shell python -c "import os, sys; \
+                          import boto3 as b; \
+                          print(os.environ.get('TRAVIS') == 'true' or \
+                          b.client('sts').get_caller_identity()['Account'] == os.environ['AZUL_AWS_ACCOUNT_ID'])\
+                          "),True)
+$(error Looks like there is a mismatch between AZUL_AWS_ACCOUNT_ID and the currently active AWS credentials. \
+        Check the output from 'aws sts get-caller-identity' against the value of that environment variable.))
+endif
+
 check_branch:
 	python $(azul_home)/scripts/check_branch.py
 
--- a/test/azul_test_case.py
+++ b/test/azul_test_case.py
@@ -1,7 +1,6 @@
 import os
 from unittest import TestCase
 
-from moto.organizations.utils import MASTER_ACCOUNT_ID as moto_account_id
 import boto3.session
 from botocore.credentials import Credentials
 import botocore.session
@@ -53,16 +52,21 @@ class AzulTestCase(TestCase):
     get_credentials_botocore = None
     get_credentials_boto3 = None
     _saved_boto3_default_session = None
+    aws_account_id = None
 
     @classmethod
     def setUpClass(cls) -> None:
         super().setUpClass()
 
-        # Set the AZUL_AWS_ACCOUNT_ID to what the moto package is using, this ensures
-        # that there will not be assertion errors when using the azul.deployment.aws.account
-        # property.
+        # Set AZUL_AWS_ACCOUNT_ID to what the Moto is using. This circumvents
+        # assertion errors in azul.deployment.aws.account.
         cls.aws_account_id = os.environ['AZUL_AWS_ACCOUNT_ID']
-        os.environ['AZUL_AWS_ACCOUNT_ID'] = moto_account_id
+        # The fake Moto account ID is defined as a constant in a Moto module
+        # but we cannot import any Moto modules since doing so has a bad side
+        # effect: https://github.com/spulec/moto/issues/2058.
+        # FIXME: Switch to overriding MOTO_ACCOUNT_ID as part of
+        #        https://github.com/DataBiosphere/azul/issues/1718
+        os.environ['AZUL_AWS_ACCOUNT_ID'] = '123456789012'
 
         # Save and then reset the default boto3session. This overrides any
         # session customizations such as those performed by envhook.py which
--- a/test/azul_test_case.py
+++ b/test/azul_test_case.py
@@ -1,5 +1,7 @@
+import os
 from unittest import TestCase
 
+from moto.organizations.utils import MASTER_ACCOUNT_ID as moto_account_id
 import boto3.session
 from botocore.credentials import Credentials
 import botocore.session
@@ -56,6 +58,12 @@ class AzulTestCase(TestCase):
     def setUpClass(cls) -> None:
         super().setUpClass()
 
+        # Set the AZUL_AWS_ACCOUNT_ID to what the moto package is using, this ensures
+        # that there will not be assertion errors when using the azul.deployment.aws.account
+        # property.
+        cls.aws_account_id = os.environ['AZUL_AWS_ACCOUNT_ID']
+        os.environ['AZUL_AWS_ACCOUNT_ID'] = moto_account_id
+
         # Save and then reset the default boto3session. This overrides any
         # session customizations such as those performed by envhook.py which
         # interfere with moto patchers, rendering them ineffective.
@@ -87,6 +95,7 @@ class AzulTestCase(TestCase):
         boto3.session.Session.get_credentials = cls.get_credentials_boto3
         botocore.session.Session.get_credentials = cls.get_credentials_botocore
         boto3.DEFAULT_SESSION = cls._saved_boto3_default_session
+        os.environ['AZUL_AWS_ACCOUNT_ID'] = cls.aws_account_id
         super().tearDownClass()
 
 
--- a/deployments/dev/environment.py
+++ b/deployments/dev/environment.py
@@ -49,7 +49,8 @@ def env() -> Mapping[str, Optional[str]]:
         'AZUL_DEBUG': '1',
         
         'AZUL_OWNER': 'hannes@ucsc.edu',
-        
+
+        'AZUL_AWS_ACCOUNT_ID': "122796619775",
         'AWS_DEFAULT_REGION': 'us-east-1',
         
         'GOOGLE_PROJECT': 'human-cell-atlas-travis-test',
--- a/deployments/sandbox/environment.py
+++ b/deployments/sandbox/environment.py
@@ -58,7 +58,8 @@ def env() -> Mapping[str, Optional[str]]:
         'AZUL_DEBUG': '1',
         
         'AZUL_OWNER': 'hannes@ucsc.edu',
-        
+
+        'AZUL_AWS_ACCOUNT_ID': "122796619775",
         'AWS_DEFAULT_REGION': 'us-east-1',
         
         'GOOGLE_PROJECT': 'human-cell-atlas-travis-test',
--- a/environment.py
+++ b/environment.py
@@ -26,6 +26,9 @@ def env() -> Mapping[str, Optional[str]]:
         # FIXME: can't use '{project_root}' due to https://github.com/DataBiosphere/azul/issues/1645
         'azul_home': os.environ['project_root'],
 
+        # The Account ID number for AWS
+        'AZUL_AWS_ACCOUNT_ID': None,
+
         # The region of the Azul deployment. This variable is primarily used by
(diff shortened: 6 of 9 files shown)


hannes-ucsc:
By our own convention, a PR author is implicitly assigned. So no one assigned means author is assigned.

README.md
```diff
@@ -545,6 +545,8 @@ make delete
 
 # 4. Running indexer or service locally
 
+**Ensure infrastructure for lambda(s) are deployed before trying to run locally. See Sections: [2](#2-getting-started), [3](#3-deployment)**
```
hannes-ucsc, line 548:
```suggestion
While this method *does* run the service or indexer locally on your machine, it still requires that the cloud resources used by them are already deployed. See sections [2](#2-getting-started) and [3](#3-deployment) on how to do that.
```

and check if I botched the links. And wrap at 80.

common.mk
```diff
@@ -33,6 +33,16 @@ ifneq ($(shell python -c "import wheel as w; \
 $(error Looks like the `wheel` package is outdated or missing. See README for instructions on how to fix this.)
 endif
 
+# This check should not occur within CI environments, where AWS Credentials might not be supplied
+ifneq ($(shell python -c "import os, sys; \
+                          import boto3 as b; \
+                          print(True) if os.getenv('CI') else \
+                          print(b.client('sts').get_caller_identity()['Account'] == \
+                          os.environ['AZUL_AWS_ACCOUNT_ID'])"),True)
```
hannes-ucsc, line 41:
```suggestion
ifneq ($(shell python -c "import os, sys; \
                          import boto3 as b; \
                          print(os.getenv('CI') or \
                          b.client('sts').get_caller_identity()['Account'] == os.environ['AZUL_AWS_ACCOUNT_ID'])"),True)
```

and don't mix os.getenv with os.environ

common.mk
```diff
@@ -33,6 +33,16 @@ ifneq ($(shell python -c "import wheel as w; \
 $(error Looks like the `wheel` package is outdated or missing. See README for instructions on how to fix this.)
 endif
 
+# This check should not occur within CI environments, where AWS Credentials might not be supplied
+ifneq ($(shell python -c "import os, sys; \
+                          import boto3 as b; \
+                          print(True) if os.getenv('CI') else \
```
hannes-ucsc, line 39: Another thing: check if Gitlab defines that variable. If not, this might fail on our Gitlab instance.

amarjandu: It's used within both:
Travis: https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
Gitlab: https://docs.gitlab.com/ee/ci/variables/predefined_variables.html

Was there any other CI that Azul is tested with?

hannes-ucsc: Just those two.

common.mk
amarjandu: @hannes-ucsc i'll switch to getenv for both.

hannes-ucsc: I should have been more clear. There is precedence in common.mk and the rest of Azul for `os.environ` and you should use that. The reason we prefer `os.environ` is that we can more clearly express whether we do or do not expect a variable to be present(`os.environ['…']` vs `os.environ.get('…')`).

amarjandu: in this case i dont expect an operators machine to have the env var `CI` set.
In that case if we were to use `os.environ['CI']` there would be KeyError raised when an operator would run any Make command. 

I could perform additional checks within the env files that could set CI to False or True,and move to explicitly check if CI is set to True. However there would be additional complexity added for both Travis and Gitlab. 

Unless there is an easier way. 🤔

README.md
amarjandu: Done.

common.mk
hannes-ucsc: My post on Slack should resolve this. Let me know if it does not.

amarjandu: @hannes-ucsc  There was a new issue with using the `or` statement within the print, Travis and Gitlab both set the CI value to `true` while our comparison is for a boolean `True`; it ends up always = failin due to `true != True`. 
Was considering using https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool to get around this.

amarjandu: `strtobool` fails to handle None case, considering; 
```
('true' if b.client('sts').get_caller_identity()['Account'] == os.environ['AZUL_AWS_ACCOUNT_ID']))\
```
and setting the comparison to `true` within the make statement. 

```
ifneq ($(shell python -c "import os, sys; \
                          import boto3 as b; \
                          print(os.environ.get('CI') or \
                          ('true' if b.client('sts').get_caller_identity()['Account'] == os.environ['AZUL_AWS_ACCOUNT_ID']))\
                          "),true)
```

hannes-ucsc: I'll reply on the most recent changes.

common.mk
```diff
@@ -33,6 +33,16 @@ ifneq ($(shell python -c "import wheel as w; \
 $(error Looks like the `wheel` package is outdated or missing. See README for instructions on how to fix this.)
 endif
 
+# This check should not occur within CI environments, where AWS Credentials might not be supplied
+ifneq ($(shell python -c "import os, sys; \
+                          import boto3 as b; \
+                          print(os.environ.get('CI') or \
```
hannes-ucsc, line 39:
```suggestion
                          print(os.environ.get('CI') == 'true' or \
```

amarjandu: 👍

amarjandu:
@hannes-ucsc regarding the addition of `azul.config.aws_account` I added the functionality in the commit that resolves the issue, but added a fixup to remove it  after Jesse's [request](https://github.com/DataBiosphere/azul/pull/1664#pullrequestreview-385890685).

Although I added the functionality to the Config class it seemed a little out-of-place when using it. 
for example with terraform templates
```
f"arn:aws:sns:{aws.region_name}:{config.aws_account}:cloudwatch-alarms",
f"arn:aws:sns:{aws.region_name}:{config.aws_account}:dcp-events"
```

while for something like `scripts/check_branch.py` the modification "fits" easier.
```
    stage_by_branch = config.main_deployments_by_branch
    account = azul.config.aws_account
```

The idea that they are "out-of-place" stems from there being 2 classes where the information can come. Within the TF template, the use of the `config` class to get the `aws_account` seems strange, especially because the `aws` class can give that information without additional imports. 

I can re-add the functionality, but thought it should be brought up, if its a non-issue, please let me know. :)

hannes-ucsc:
I wouldn't interpret @jessebrennan post as a request to remove all the changes to the `azul.config` and `azul.deployment.aws` classes. Let's go with what you and I agreed to and we'll take it from there in my next review.

common.mk
```diff
@@ -33,6 +33,16 @@ ifneq ($(shell python -c "import wheel as w; \
 $(error Looks like the `wheel` package is outdated or missing. See README for instructions on how to fix this.)
 endif
 
+# This check should not occur within CI environments, where AWS Credentials might not be supplied
+ifneq ($(shell python -c "import os, sys; \
+                          import boto3 as b; \
+                          print(os.environ.get('CI') == 'true' or \
+                          b.client('sts').get_caller_identity()['Account'] == os.environ['AZUL_AWS_ACCOUNT_ID'])\
+                          "),True)
+$(error Looks like there is a mismatch between the AWS account you have configured, and what the deployment expects. \
```
hannes-ucsc, line 42:
```suggestion
$(error Looks like there is a mismatch between AZUL_AWS_ACCOUNT_ID and the currently active AWS credentials. \
```

common.mk
```diff
@@ -33,6 +33,16 @@ ifneq ($(shell python -c "import wheel as w; \
...
+        Compare output from 'aws sts get-caller-identity' and the deployment environment file)
```
hannes-ucsc, line 43:
```suggestion
        Check the output from 'aws sts get-caller-identity' against the value of that environment variable.)
```

amarjandu:
@hannes-ucsc our proposed solution to asserting that the dynamically evaluated AccountID is equivalent to the static AccountID from the Environment Variable does not seem to past out CI testing environment, I'm not sure that we provide Travis with AWS Credentials, as i dont see any within the ENV configuration, however during run time on Travis, the dynamic account value is [123456789012](https://travis-ci.org/github/DataBiosphere/azul/builds/670764112#L697).

hannes-ucsc:
> I'm not sure that we provide Travis with AWS Credentials

We do not and that's intentional.

> This causes a break on the assertion check that we have placed.

I am surprised that that code (`azul.deployment.aws.account`) is even executed on Travis. Since Travis does not have AWS creds, that property would have to raise an exception on `develop`. Why does it not?

Another thing: Gitlab does have cloud access and we don't need to circumvent the common.mk check there. AFAICT, the current code circumvents the check on both GL and Travis. It only needs to be circumvented on Travis.

amarjandu:
> Another thing: Gitlab does have cloud access and we don't need to circumvent the common.mk check there. AFAICT, the current code circumvents the check on both GL and Travis. It only needs to be circumvented on Travis.

I've updated the `common.mk` to ensure that the `make` checks are only bypassed within the `TRAVIS` CI system.

I've also adjusted the `azul_test_case` class to modify the environment value allowing `unit_test` to pass like we discussed after standup.

attic/deployments/dev/environment
```diff
@@ -30,6 +30,8 @@ _set AZUL_DEBUG 1
 
 _set AZUL_OWNER hannes@ucsc.edu
 
+_set AZUL_AWS_ACCOUNT_ID 122796619775
```
hannes-ucsc, line 33: The attic is for "deleted" files. We don't need to, and should not, modify them.

attic/deployments/sandbox/environment
```diff
@@ -39,6 +39,8 @@ _set AZUL_DEBUG 1
 
 _set AZUL_OWNER hannes@ucsc.edu
 
+_set AZUL_AWS_ACCOUNT_ID 122796619775
```
hannes-ucsc, line 42: Same here

test/azul_test_case.py
```diff
@@ -56,6 +59,9 @@ class AzulTestCase(TestCase):
     def setUpClass(cls) -> None:
         super().setUpClass()
 
+        cls.aws_account_id = config.aws_account_id
```
hannes-ucsc, line 62:
```suggestion
        cls.aws_account_id = os.environ['AZUL_AWS_ACCOUNT_ID']
```

For symmetry. There is no point in using the abstraction if the line below does not. The long term solution would be to extend the abstraction to handle patches like this.

test/azul_test_case.py
```diff
@@ -56,6 +59,9 @@ class AzulTestCase(TestCase):
...
+        os.environ['AZUL_AWS_ACCOUNT_ID'] = '123456789012'
```
hannes-ucsc, line 63: You should mention in a comment where the value comes from or, even better, use the corresponding moto constant here.

test/azul_test_case.py
```diff
@@ -1,5 +1,7 @@
+import os
 from unittest import TestCase
 
+from moto.organizations import utils
```
hannes-ucsc, line 4:
```suggestion
from moto.organizations.utils import MASTER_ACCOUNT_ID as moto_account_id
```

test/azul_test_case.py
```diff
@@ -56,6 +58,12 @@ class AzulTestCase(TestCase):
     def setUpClass(cls) -> None:
         super().setUpClass()
 
+        # Set the AZUL_AWS_ACCOUNT_ID to what the moto package is using, this ensures
+        # that there will not be assertion errors when using the azul.deployment.aws.account
+        # property.
+        cls.aws_account_id = os.environ['AZUL_AWS_ACCOUNT_ID']
+        os.environ['AZUL_AWS_ACCOUNT_ID'] = utils.MASTER_ACCOUNT_ID
```
hannes-ucsc, line 65:
```suggestion
        os.environ['AZUL_AWS_ACCOUNT_ID'] = moto_account_id
```

is more clear IMO

hannes-ucsc:
Despite the scope creep, I think the estimate of 3 is still accurate.

amarjandu:
> https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1647

ah, I should have checked for usage of the object; I've repaired the error, awaiting testing .

hannes-ucsc:
Your changes in response to my last review should have been made as a separate !fixup commit.

amarjandu:
> Your changes in response to my last review should have been made as a separate !fixup commit.

@hannes-ucsc 
Strange IIRC that was what i had done. Ill keep my eyes on that, thanks for informing me. 

Screen Shot 2020-04-16 at 10 59 55 AM

hannes-ucsc:
> Strange IIRC that was what i had done. Ill keep my eyes on that, thanks for informing me.

I think you are right. My force push actually shows that I obliterated the fixup commit in your screenshot.

image

And https://github.com/DataBiosphere/azul/commit/4cb2db59c4958e7d9014db5e376368c4ec50b08b only contains the changes that address my review comments.

My apologies.

hannes-ucsc:
The IT fail seems to be limited to this branch. A subsequent sandbox build on a different branch succeeded: https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1760

hannes-ucsc:
This was merged.

Issue #1653, opened 2020-03-23
Switch to SQS FIFO queue trigger for AWS lambda 


melainalegaspi:
https://aws.amazon.com/about-aws/whats-new/2019/11/aws-lambda-supports-amazon-sqs-fifo-event-source/

https://aws.amazon.com/blogs/compute/new-for-aws-lambda-sqs-fifo-as-an-event-source/

It appears that we can now trigger the `azul-indexer-$AZUL_DEPLOYMENT_STAGE-write` lambda directly from the FIFO queue it is actually consuming. Until now, we've been emulating that with token messages in a standard queue.  That emulation is complex and imprecise logic. It affects performance negatively and increases cost by causing a significant number of idle receive message calls.

Remove the emulation, mainly the token concept, the token queue (azul-documents-$AZUL_DEPLOYMENT_STAGE) and the nudge lambda.

amarjandu:
@hannes-ucsc i was looking into what changes we would have to make to the chalice wrappers, and i found this note within the chalice documentation; https://chalice.readthedocs.io/en/latest/topics/events.html#sqs-events it states that FIFO queues are currently not supported by chalice. Should we place this ticket into the backlog? I do not see any issues created to adding FIFO support to Chalice within their github page.

hannes-ucsc:
This is so important to me that I would even consider adding FIFO queue support to Chalice, in a fork. I think we already use a fork anyways. Could you investigate what it would take to do that? I can't imaging it would be that difficult. You may also raise this with the Chalice team to see why the support is not there yet. I'm gonna spike this to one day to cover the research.

amarjandu:
yeah, I'll raise the question with the chalice team and investigate what a solution would look like.

amarjandu:
Issue Raised: https://github.com/aws/chalice/issues/1384

PR #1667, opened 2020-04-01
Remove SQS Token Queue (#1653)

amarjandu:
Removes the usages of tokens to alert the index writer about new tallies.
Removes nudge lambda. 

Connected To: #1653

### DIFF
--- a/lambdas/indexer/app.py
+++ b/lambdas/indexer/app.py
@@ -2,15 +2,11 @@
 Chalice application module to receive and process DSS event notifications.
 """
 from collections import (
-    Counter,
-    defaultdict,
+    defaultdict
 )
-from datetime import datetime
 import http
-from itertools import chain
 import json
 import logging
-import random
 import time
 from typing import (
     List,
@@ -22,14 +18,13 @@ import uuid
 import boto3
 # noinspection PyPackageRequirements
 import chalice
+from chalice.app import SQSRecord
 from dataclasses import (
-    asdict,
     dataclass,
     replace,
 )
 from more_itertools import (
-    chunked,
-    partition,
+    chunked
 )
 
 from azul import (
@@ -41,7 +36,6 @@ from azul.chalice import AzulChaliceApp
 from azul.health import HealthController
 from azul.logging import configure_app_logging
 from azul.plugin import Plugin
-from azul.time import RemainingLambdaContextTime
 from azul.transformer import EntityReference
 from azul.types import JSON
 
@@ -212,10 +206,8 @@ def index(event: chalice.app.SQSEvent):
 
                 log.info("Queueing %i entities for aggregating a total of %i contributions.",
                          len(tallies), sum(tally.num_contributions for tally in tallies))
-                token_queue = queue(config.token_queue_name)
                 document_queue = queue(config.document_queue_name)
                 for batch in chunked(tallies, document_batch_size):
-                    token_queue.send_message(MessageBody=json.dumps(Token.mint(len(batch)).to_json()))
                     document_queue.send_messages(Entries=[dict(tally.to_message(), Id=str(i))
                                                           for i, tally in enumerate(batch)])
         except BaseException:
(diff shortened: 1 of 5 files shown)


lambdas/indexer/app.py
```diff
@@ -231,95 +225,52 @@ def index(event: chalice.app.SQSEvent):
...
+    # Consolidate multiple tallies for the same entity and process entities with only one message. Because SQS FIFO
+    # queues try to put as many messages from the same message group in a reception batch, a single message per
+    # group may indicate that that message is the last one in the group. Inversely, multiple messages per group
+    # in a batch are a likely indicator for the presence of even more queued messages in that group. The more
+    # bundle contributions we defer, the higher the amortized savings on aggregation become. Aggregating bundle
+    # contributions is a costly operation for any entity with many contributions e.g., a large project.
+    #
+    tallies_by_id: MutableMapping[EntityReference, List[DocumentTally]] = defaultdict(list)
+    for record in event['Records']:
+        tally = DocumentTally.from_message(record)
```
jessebrennan, line 242: I think these names should agree, so change `from_message()` to `from_record` or `record` to `message`. I think message makes the most sense, assuming they are the same.

lambdas/indexer/app.py
```diff
@@ -231,95 +225,52 @@ def index(event: chalice.app.SQSEvent):
...
+        log.info('Attempt %i of handling %i contribution(s) for entity %s/%s',
+                 tally.attempts, tally.num_contributions, tally.entity.entity_type, tally.entity.entity_id)
+        tallies_by_id[tally.entity].append(tally)
+        # NOTE here we are grouping the tallies based on their entity
```
jessebrennan, line 246: I think this comment is unnecessary since this information can be quickly deduced by looking at the type annotation for `tallies_by_id`.

lambdas/indexer/app.py
```diff
@@ -231,95 +225,52 @@ def index(event: chalice.app.SQSEvent):
...
+
+    if deferrals:
+        for tally in deferrals:
+            log.info('Deferring aggregation of %i contribution(s) to entity %s/%s',
+                     tally.num_contributions, tally.entity.entity_type, tally.entity.entity_id)
+        document_queue.send_messages(Entries=[dict(tally.to_message(), Id=str(i))
+                                              for i, tally in enumerate(deferrals)])
+
+    document_queue.delete_messages(Entries=[dict(Id=str(i), ReceiptHandle=message.receipt_handle)
+                                            for i, message in enumerate(chain(record))])
```
jessebrennan, line 273: I imagine you meant `event['Records']`, not `record`. Having said that, it's unnecessary to delete the messages from the queue that triggers the lambda. From the [Chalice docs](https://chalice.readthedocs.io/en/latest/topics/events.html#sqs-events):

>If your lambda functions completes without raising an exception, then Lambda will automatically delete all the messages associated with the SQSEvent. You don’t need to manually call sqs.delete_message() in your lambda function.

We had to delete them before since this lambda was triggered by the token queue, It's safe to just delete these two lines.

amarjandu: When i was reading through the documentation i opted to add the manual deletion back to the method the reasoning being at the bottom of the documented page you sent...

``` 
Note that if you are using a batch size of more than one, the entire batch succeeds or fails. This means that it is possible for your lambda function to see a message multiple times, even if it’s successfully processed the message previously....
```

I'm not sure if it would be wrong to requeue the whole batch if there was a failure of a single message, or the implications of how that would effect the system overall. If you think I should omit the manual deletion, I'll remove it.

lambdas/indexer/app.py
amarjandu: whoops, that was a personal note.

lambdas/indexer/app.py
amarjandu: I changed the outer loop to `for message in event['Records']` the format of the `message` and objects within the list `event['Records']` do have the same format.

amarjandu:
> I'm excited to see this refactor.
> 
> Some general comments:
> 
>     * Aren't there supposed to be some changes to Chalice for this to work?
> 
>     * How did you test this? At some point this should be deployed to your personal deployment and integration tests should be run there.
> 
>     * The diff for `indexer/app.py` is unnecessarily huge when viewed on Github. When viewing on PyCharm, the diff is much smaller. I'm not sure why this is, but if you can figure out how, we should fix this.

Actually, In the the past chalice did not "support" lambda invocation by SQS-FIFO, but there's not anything in the Chalice codebase that stops you from attempting to register them; i believe they were relying on the AWS API to decline the registration of the FIFO queue as the lambda invocation source at deployment time, rather than have something in the Chalice codebase to perform this check.

I've been testing this in a local deployment, and have been testing reindexing operation on that deployment at well. I am concerned about potential performance implications that this has; but due to the stalling issues seen in #1618 I'm unable to really say if there is an measurable improvement, I've observed the stalls occur multiple times within one `reindex` operation. 

as for why the `git diff` is ugly, 🤷‍♂️  but yes, i am curious if there is something that can be done to make it easier on the eyes.

lambdas/indexer/app.py
jessebrennan: I hear your concern about duplicate messages. Deleting a message explicitly will only prevent a successful message from being re-queued if it's done before the failure occurs. Here, we delete all of the messages all at once, explicitly and after _all_ of the aggregation happens. If a failure happens, then all of the messages will be re-queued. (Notice that this was the behaviour before your changes too.)

Fortunately, it doesn't matter. Azul can handle re-processing of a notification. If we wanted to delete messages as they were processed it would be a much more extensive change.

The precedence in the indexer is to not do the explicit delete (look at the `index` lambda for example) so I think you should omit the manual deletion.

lambdas/indexer/app.py
amarjandu: I ended up changing the method to `from_event_record` .

src/azul/queues.py
```diff
@@ -176,7 +176,6 @@ def wait_for_queue_level(cls, empty: bool = True, num_bundles: int = None):
         sleep_time = 5
         deque_size = 10 if empty else 1
         queues = cls.get_queues((config.notify_queue_name, config.document_queue_name))
```
danielsotirhos, line 178:
```suggestion
        queues = cls.get_queues(config.work_queue_names)
```
Now that `token_queue_name` has been removed from `work_queue_names` we can use it here.

lambdas/indexer/app.py
```diff
@@ -1,16 +1,10 @@
 """
 Chalice application module to receive and process DSS event notifications.
 """
-from collections import (
-    Counter,
-    defaultdict,
-)
-from datetime import datetime
+from collections import defaultdict
```
hannes-ucsc, line 4: Once an import statement is multi line we don't go back to single line, even if just one imported symbol remains. You couldn't have known that but the PyCharm formatter settings at the project root define that. Be sure to import it. If you don't use PyCharm you'll have to enforce those settings manually.

lambdas/indexer/app.py
```diff
@@ -334,12 +274,12 @@ class DocumentTally:
     attempts: int
 
     @classmethod
-    def from_message(cls, message) -> 'DocumentTally':
-        body = json.loads(message.body)
+    def from_event_record(cls, record: dict) -> 'DocumentTally':
```
hannes-ucsc, line 277:
```suggestion
    def from_sqs_event(cls, record: dict) -> 'DocumentTally':
```

and read https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#16type-hints

amarjandu: The method name has been changed.

 Ended up changing the typing hint to `Dict[str,str]` do you think that it is too generic, or have have another idea for a hint?

amarjandu:
> A few nits. This passes IT I assume?

@hannes-ucsc Yes it is able to deploy successfully, `make reindex`, and integration tests also pass successfully. 

I added another fixup, and had a followup question regarding typing hints.

lambdas/indexer/app.py
hannes-ucsc: The event records holds JSON, no? See fourth bullet in the contrib guid section I linked to above.

amarjandu: the record would could have JSON present in the `body`, there is also use the 'record ["attributes"]' which contain strings of information.

```
{'attributes': {'ApproximateFirstReceiveTimestamp': '1587423132081',
                'ApproximateReceiveCount': '1',
                'MessageDeduplicationId': 'a4f0df47-3c56-4cae-b908-857be7fd7840',
                'MessageGroupId': '06923068-8202-40c4-a42e-d51070fb574c',
                'SenderId': 'AROARZFZ7W77UV6O5LBHR:azul-indexer-amar-index',
                'SentTimestamp': '1587423122076',
                'SequenceNumber': '18853124392961007616'},
 'awsRegion': 'us-east-1',
 'body': '{"entity_type": "cell_suspensions", "entity_id": '
         '"06923068-8202-40c4-a42e-d51070fb574c", "num_contributions": 1}',
 'eventSource': 'aws:sqs',
 'eventSourceARN': 'arn:aws:sqs:us-east-1:122796619775:azul-documents-amar.fifo',
 'md5OfBody': '40709903053d7dfea1e3fb04b29bb12f',
 'messageAttributes': {},
 'messageId': '22d70fac-7ee4-471c-9377-211fd341c7b6'}
```

For that reason i ~think~ this should work.
   def from_sqs_record(cls, record: Mapping[str, Union[str, JSON]]) -> 'DocumentTally':

However because we know which values contain JSON,  a typing.TypedDict might even be more suitable for this... (certainly is not a "generic" approach)

Thanks for your patience with my questions, I've not used Typing much in the past and i would like learn correct usage + habits going forward.

amarjandu: Changed to `Mapping[str, JSON]`

lambdas/indexer/app.py
```diff
@@ -334,12 +280,12 @@ class DocumentTally:
     attempts: int
 
     @classmethod
-    def from_message(cls, message) -> 'DocumentTally':
-        body = json.loads(message.body)
+    def from_sqs_record(cls, record: Mapping[str, JSON]) -> 'DocumentTally':
```
hannes-ucsc, line 283: The type hint is incorrect and I would excuse a mistake like that if it weren't for the fact that PyCharm's type checker warns about it. I would check my PyCharm settings to ensure you are made aware of type checker warnings in PyCharm.

The reason I am such a stickler for type hints is that type hints are extremely useful in declaring intent AND and having that declaration be partially checked by tooling.

`JSON` is `Mapping`. So `Mapping[str, JSON]` implies that the function expects a dictionary where the values are dictionaries. But the very first line passes such a value to a function that expects a string (`json.loads`). This is obviously wrong.

Either way, I want to simplify this further. See my diff.

amarjandu:
@hannes-ucsc yeah that is a little strange, i ran the IT on my personal deployment again prior to marking this for review. I'll investigate.

hannes-ucsc:
It is possible that the failure is intermittent. Note that the IT picks a prefix randomly. For the best reproduction, you should temporarily pin the bundle prefix and the seed for bundle subset selection.

https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1951#L496
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1951#L498

hannes-ucsc:
IT failure seemed to have been caused by overload due to an ongoing `dev` reindex.

hannes-ucsc:
When choosing commit titles, a safe bet is to use the issue title.

Issue #1485, opened 2019-11-27
Save OpenAPI spec in Chalice app object


jessebrennan:
Currently, OpenAPI spec for each route is stored as an attribute to the decorated function. A more elegant and efficient solution is to store this spec as an attribute of the Chalice app object.

This way, the spec only needs to be complied together once and testing and error checking is easier.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-911)
┆Project Name: azul
┆Issue Number: AZUL-911

Issue #1450, opened 2019-11-14
Build OpenAPI spec statically


jessebrennan:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-891)
┆Project Name: azul
┆Issue Number: AZUL-891

jessebrennan:
After #556 is merged we can integrate the spec manually into the Terraform generated by Chalice. This would allow more fine grained control of what API Gateway makes available. Also see what TF says about template_file https://www.terraform.io/docs/providers/template/d/file.html

PR #1668, opened 2020-04-01
Store openapi spec in chalice object (#1485)

noah-aviel-dove:


src/azul/chalice.py
```diff
@@ -64,6 +79,25 @@ def test_route(self, *args, **kwargs):
...
+                       path_spec: Optional[JSON],
+                       method_spec: Optional[JSON],
+                       methods: Iterable[str]):
+        assert path not in self.path_specs, 'Only specify path_spec once per route path'
+
+        if path_spec is not None:
+            self.path_specs[path] = path_spec
+
+        if method_spec is not None:
+            # OpenAPI routes must be lower case
```
jessebrennan, line 93: Consider moving this comment directly preceding the relevant line.

src/azul/chalice.py
```diff
@@ -44,11 +59,11 @@ def route(self, path, enabled=True, path_spec=None, method_spec=None, **kwargs):
...
-            methods = kwargs.get('methods')
+            methods = kwargs.get('methods', ())
             decorator = super().route(path, **kwargs)
 
             def _decorator(view_func):
-                view_func = openapi_spec(path, methods, path_spec=path_spec, method_spec=method_spec)(view_func)
+                self._register_spec(path, path_spec, method_spec, methods)
                 # Stash the URL path a view function is bound to as an attribute of
                 # the function itself.
                 view_func.path = path
```
hannes-ucsc, line 69: Do we still need to do this?

src/azul/chalice.py
```diff
@@ -64,6 +79,25 @@ def test_route(self, *args, **kwargs):
         """
         return self.route(*args, enabled=self.unit_test, **kwargs)
 
+    def _register_spec(self,
+                       path: str,
+                       path_spec: Optional[JSON],
+                       method_spec: Optional[JSON],
+                       methods: Iterable[str]):
+        assert path not in self.path_specs, 'Only specify path_spec once per route path'
```
hannes-ucsc, line 87: `_register_spec('/foo', path_spec={}, …)` followed by `_register_spec('/foo', path_spec=None, …)` would raise, unjustifiably so, I think.

src/azul/chalice.py
```diff
@@ -64,6 +79,25 @@ def test_route(self, *args, **kwargs):
...
+
+        if path_spec is not None:
+            self.path_specs[path] = path_spec
+
+        if method_spec is not None:
+            # No need to worry about duplicate method_specs since Chalice will
+            # complain in that case
```
hannes-ucsc, line 94: That's an assumption. There should be an explicit assertion preventing dupes, without relying on this assumption that Chalice does that for us.

src/azul/chalice.py
```diff
@@ -64,6 +79,25 @@ def test_route(self, *args, **kwargs):
...
+            self.method_specs.update({
+                # OpenAPI routes must be lower case
```
hannes-ucsc, line 96:
```suggestion
                # Method names in OpenAPI routes must be lower case
```

src/azul/openapi/__init__.py
```diff
@@ -14,98 +13,6 @@
 )
 
 
-def openapi_spec(path: str, methods: List[str], path_spec: Optional[JSON] = None, method_spec: Optional[JSON] = None):
-    """
-    Add description to decorated function at specified route.
-
-    >>> @openapi_spec('/foo', ['GET'], path_spec={'a': 'b'}, method_spec={'c': 'd'})
```
hannes-ucsc, line 21: Don't we want to retain these tests?

src/azul/openapi/__init__.py
```diff
@@ -150,24 +57,12 @@ def get_app_specs(app):
...
-            path_specs.update(func.path_specs)
-        else:
-            # TODO (jesse): raise if route is undocumented
-            pass
-        func_method_specs = getattr(func, 'method_specs', None)
-        if func_method_specs:
-            method_specs.update(func_method_specs)
+    undocumented = app.routes.keys() - app.path_specs.keys()
+    if undocumented:
+        # TODO (jesse): raise if route is undocumented
```
hannes-ucsc, line 62: @jessebrennan is there a ticket for this? If so, post that here so @noah-aviel-dove can add a reference.

hannes-ucsc: … and if not create one and post ticket number here.

src/azul/openapi/__init__.py
noah-aviel-dove: I believe they are now redundant given that the `openapi_spec` decorator has been absorbed (in a dramatically simplified form) by `AuzlChaliceApp.route`, which is covered by `test_openapi.TestGetAppSpecs.` Do you have any specific tests for the route method that you'd like to see added?

src/azul/chalice.py
noah-aviel-dove: My changes up to this point have not removed the need for it, if that's what you're asking. I can very easily make additional changes to implement an approach that avoids storing this attribute (which AFAICT is only ever used for two specific routes in a single function).

src/azul/openapi/__init__.py
jessebrennan: In my head this would be taken care of in #1450.

src/azul/chalice.py
hannes-ucsc: The ticket says

> Currently, OpenAPI spec for each route is stored as an attribute to the decorated function. A more elegant and efficient solution is to store this spec as an attribute of the Chalice app object.

which is misleading. It's not the spec that is stored in an attribute, it's the path. But the intend is to remove the attribute, whatever it stores. So please do make those changes. For future reference, and in case you were already aware of that error in the ticket description, feel free to proactively bring things like that to my attention.

src/azul/openapi/__init__.py
hannes-ucsc: No. Just make sure that all cases currently covered by the doctest are covered by the test you mention.

src/azul/chalice.py
noah-aviel-dove: I will certainly make the changes to remove the path attribute from the functions, but I do not believe the ticket description was inaccurate. `path_spec` and `method_spec` attributes are indeed stored in the decorated functions on the develop branch. See https://github.com/DataBiosphere/azul/pull/1668/files#diff-faf2c5ae5050f2419fdbdbd6a8af190aL95

jessebrennan: FWIW, I didn't store the path as an attribute. It looks like this was done by @hannes-ucsc in b277a14af879a1db5c7f72862494064feab42922.

src/azul/openapi/__init__.py
noah-aviel-dove: It looks like there was one test I missed. It's been added.

src/azul/chalice.py
hannes-ucsc: You both are right and I was wrong. The ticket description IS accurate and the I did at the `.path` attribute. After @jessebrennan pointed out the commit, I remembered why I added it. It was for a good reason so we'll need to revert the removal. I owe you a favor, @noah-aviel-dove.

lambdas/service/app.py
```diff
@@ -1348,8 +1348,7 @@ def _dss_files(uuid, fetch=True):
 
 def file_url(uuid: str, fetch: bool = True, **params: str):
     uuid = urllib.parse.quote(uuid, safe="")
-    view_function = fetch_dss_files if fetch else dss_files
-    url = self_url(endpoint_path=view_function.path.format(uuid=uuid))
+    url = self_url(endpoint_path=f'{"/fetch" if fetch else ""}/dss/files/{uuid}')
```
hannes-ucsc, line 1351: I remember now that I added `.path` so we don't have to duplicate the URL here. We need revert this change.

src/azul/openapi/__init__.py
hannes-ucsc: The ticket reference still needs to be added here.

noah-aviel-dove:
@hannes-ucsc I've moved the methods into the App class, and also wrote a 2nd commit that removes the `get_app_specs` method entirely since it felt superfluous in that setup. Take it or leave it.

test/test_openapi.py
```diff
@@ -11,28 +11,30 @@
...
 
     def test_unannotated(self):
-        # TODO: Once all endpoints are documented, this test should assert failure
+        # TODO: (jesse) Once all endpoints are documented, this test should assert failure
+        # (https://github.com/DataBiosphere/azul/issues/1450)
         app = self._mock_app_object()
         self._mock_app_route(app, '/foo')
-        app.get_all_specs()
+        undocumented = app.routes.keys() - app.path_specs.keys()
+        self.assertFalse(0 and undocumented)
```
hannes-ucsc, line 19: `@skip("https://github.com/DataBiosphere/azul/issues/1450")` before the test is what we typically use to disable tests until resolution of a ticket lands.

```suggestion
        self.assertEqual({}, undocumented)
```

test/test_openapi.py
```diff
@@ -1,91 +1,112 @@
...
 
+    @skip('https://github.com/DataBiosphere/azul/issues/1450')
     def test_unannotated(self):
-        # TODO: Once all endpoints are documented, this test should assert failure
-        app = self._mock_app_object([lambda x: x])
-        get_app_specs(app)
+        app = self._mock_app_object()
+        self._mock_app_route(app, '/foo')
+        undocumented = app.routes.keys() - app.path_specs.keys()
+        self.assertFalse(undocumented)
```
hannes-ucsc, line 21: I prefer

```suggestion
        self.assertEqual(set(), undocumented)
```

My [earlier comment](https://github.com/DataBiosphere/azul/pull/1668#discussion_r405216252) on this was wrong but that doesn't mean there is nothing to be be addressed. You can always ask for clarification on Slack.

test/test_openapi.py
```diff
@@ -1,91 +1,112 @@
...
-        @openapi_spec('/foo', ['GET'], path_spec={'a': 'b'})
-        def route2():
+        with self.assertRaises(AssertionError):
+            @app.route('/foo', methods=['GET'], method_spec={'a': 'b'})
+            @app.route('/foo', methods=['GET'], method_spec={'a': 'XXX'})
+            def route():
+                pass
+
+        self.assertEqual(mock_route.call_count, 2)
+        # Decorators are applied from the bottom up
```
hannes-ucsc, line 69: I do not understand this comment.

Issue #1670, opened 2020-04-02
Test logs are too noisy


hannes-ucsc:
There simply too many of these: https://travis-ci.org/github/DataBiosphere/azul/builds/670311272#L757

Either combine them in one log statement like https://travis-ci.org/github/DataBiosphere/azul/builds/670311272#L756 or demote them to DEBUG.

amarjandu:
This might be useful. https://github.com/HumanCellAtlas/data-store/pull/2561

PR #1682, opened 2020-04-07
Reduce noisy indexer aggregation logs (#1670)

danielsotirhos:


src/azul/indexer/__init__.py
```diff
@@ -353,6 +351,12 @@ def _aggregate(self, contributions: List[Contribution]) -> List[Aggregate]:
                                   num_contributions=tallies[entity])
             aggregates.append(aggregate)
 
+        log.info('Selected %i contribution(s) to be aggregated. Breakdown by entity: %s',
```
achave11, line 354: Maybe you want to consider moving this to line No. 343, before the aggregation happens such that 
"... to be aggregated" is factual. Or perhaps wording it in a way that indicates that the aggregation has happened.

danielsotirhos: Good point. I think having the message logged before the action is performed makes sense. Fixup incoming.

src/azul/indexer/__init__.py
```diff
@@ -338,11 +338,16 @@ def _aggregate(self, contributions: List[Contribution]) -> List[Aggregate]:
...
         transformers = {t.entity_type(): t for t in self.transformers() if isinstance(t, AggregatingTransformer)}
 
+        log.info('Selected %i contribution(s) to be aggregated. Breakdown by entity: %s',
+                 sum(len(contributions) for contributions in contributions_by_entity.values()),
+                 {
+                     f'{entity.entity_type}/{entity.entity_id}': len(contributions)
+                     for entity, contributions in sorted(contributions_by_entity.items())
+                 })
+
         # Aggregate contributions for the same entity
```
achave11, line 348: Since the log statement was move up, I am wondering if this comment is still necessary. I'll leave it to your discretion.

danielsotirhos: Although the log message makes this comment slightly redundant, I think it is worth keeping as it matches with the other comments in the function.

src/azul/indexer/__init__.py
```diff
@@ -338,11 +338,16 @@ def _aggregate(self, contributions: List[Contribution]) -> List[Aggregate]:
         # Create lookup for transformer by entity type
         transformers = {t.entity_type(): t for t in self.transformers() if isinstance(t, AggregatingTransformer)}
 
+        log.info('Selected %i contribution(s) to be aggregated. Breakdown by entity: %s',
+                 sum(len(contributions) for contributions in contributions_by_entity.values()),
+                 {
+                     f'{entity.entity_type}/{entity.entity_id}': len(contributions)
```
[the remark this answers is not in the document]
danielsotirhos: The slash was used to match the similar compact format used in the log entry that this one is following. https://travis-ci.org/github/DataBiosphere/azul/builds/670311272#L756

I'm happy to change both to a different format however I'm unsure what the preferred syntax would be as there doesn't seem to be another precedent.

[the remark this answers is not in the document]
hannes-ucsc: Makes sense.

src/azul/indexer/__init__.py
```diff
@@ -303,14 +303,14 @@ def _read_contributions(self, tallies: Tallies) -> List[Contribution]:
...
-        log.info('Read %i contribution(s). Breakdown by entity: %s',
-                 len(contributions),
-                 {
-                     f'{entity.entity_type}/{entity.entity_id}': sum(1 for _ in contribution_group)
-                     for entity, contribution_group in groupby(sorted(contributions, key=entity_key), key=entity_key)
-                 })
+        log.info('Read %i contribution(s). ', len(contributions))
+        if log.isEnabledFor(logging.DEBUG):
+            entity_key = attrgetter('entity')
+            log.debug({
```
hannes-ucsc, line 309: This needs a human readable message to provide context.

src/azul/indexer/__init__.py
```diff
@@ -334,15 +334,20 @@ def _aggregate(self, contributions: List[Contribution]) -> List[Aggregate]:
                     assert entity == contribution.entity
                     contributions_by_entity[entity].append(contribution)
                     break
+        log.info('Selected %i contribution(s) to be aggregated.',
+                 sum(len(contributions) for contributions in contributions_by_entity.values()))
+        if log.isEnabledFor(logging.DEBUG):
+            log.debug({
```
hannes-ucsc, line 340: Same here.

Issue #1561, opened 2020-02-10
Do not hard-code account number of assumed role in permissions boundary


hannes-ucsc:
Follow-up from https://github.com/databiosphere/azul/issues/1536.

The ARN of the role that is assumed for direct access is mentioned in the assuming lambda's role policy. The boundary must therefore include that statement. While the ARN is configurable per deployment, the boundary must cover all deployments. Individual Azul deployments could target different DSS deployments, assuming roles in different accounts but the boundary must include them all.

hannes-ucsc:
See above commit for relevant code sections.

PR #1684, opened 2020-04-07
Generalize permissions boundary account id (#1561)

danielsotirhos:


src/azul/__init__.py
```diff
@@ -162,6 +162,13 @@ def dss_direct_access(self) -> bool:
     def dss_direct_access_role(self, lambda_name: str) -> Optional[str]:
         return os.environ.get('AZUL_DSS_DIRECT_ACCESS_ROLE').format(lambda_name=lambda_name)
 
+    @property
+    def dss_direct_access_account(self) -> Optional[str]:
+        """
+        The direct access account id extracted from the direct access role value
+        """
+        return os.environ.get('AZUL_DSS_DIRECT_ACCESS_ROLE').split(':')[4]
```
hannes-ucsc, line 170: What if `.get()` returns `None`?

hannes-ucsc: @jessebrennan, this one is easy to catch, wouldn't you agree?

danielsotirhos: The env variable `AZUL_DSS_DIRECT_ACCESS_ROLE` is already expected to be required by `dss_direct_access_role()`.
I agree however that if a key is required then `[]` should be used not `.get()`. I will update along with `dss_direct_access_role()` to match.

danielsotirhos:
Upon closer inspection it seems like `config.dss_direct_access_role()` and `config.dss_direct_access_account` should only be required if direct access is enabled.

@jessebrennan  What are your thoughts on this approach with `_get_direct_access_related_value()`? I'm currently unsure if this approach is correct for the case where direct access is not enabled and the related variables (role & account) are defined in the environment, or if they should always be returned as None when direct access is not enabled.

src/azul/__init__.py
hannes-ucsc: Yes, it seems that I made the same mistake in config.dss_direct_access_role.

hannes-ucsc:
Looks pretty complicated to me.

danielsotirhos:
> Looks pretty complicated to me.

Simplified, though with the same assumptions. Let me know if you disagree that these config properties related to DSS direct access should only be required if DSS direct access is enabled.

src/azul/__init__.py
jessebrennan: Yes, I should have seen this.

Issue #1678, opened 2020-04-07
TypeError: '<' not supported between instances of 'str' and 'NoneType' during reindexing


jessebrennan:
Here is a stack trace:
```
Traceback (most recent call last):
  File ""/var/task/app.py"", line 202, in index
    contributions = indexer.transform(notification, delete=False)
  File ""/var/task/azul/indexer.py"", line 151, in transform
    metadata_files=metadata_files))
  File ""/var/task/azul/project/hca/transformers.py"", line 586, in transform
    cell_suspensions=list(map(self._cell_suspension, visitor.cell_suspensions.values())),
  File ""/var/task/azul/project/hca/transformers.py"", line 259, in _cell_suspension
    'organ_part': sorted(organ_parts)
TypeError: '<' not supported between instances of 'NoneType' and 'str'"
```

The following is a dump of all 40 occurrences of this error during [this reindexing](https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1524):
[logs-insights-results.csv.txt](https://github.com/DataBiosphere/azul/files/4441388/logs-insights-results.csv.txt)

PR #1689, opened 2020-04-08
Fix TypeError sorting None on reindex (#1678)

danielsotirhos:


test/indexer/test_hca_indexer.py
```diff
@@ -870,6 +870,25 @@ def test_cell_line_sample(self):
...
+        self._index_canned_bundle(('1b6d8348-d6e9-406a-aa6a-7ee886e52bf9', '2019-10-03T105524.911627Z'))
+        hits = self._get_all_hits()
+        for hit in hits:
+            contents = hit['_source']['contents']
+            entity_type, aggregate = config.parse_es_index_name(hit['_index'])
+            cell_suspension = one(contents['cell_suspensions'])
+            self.assertEqual(cell_suspension['organ'], ['embryo', 'immune system'])
+            self.assertEqual(cell_suspension['organ_part'], ['skin epidermis', self.translated_str_null])
+            if aggregate and entity_type != 'samples':
+                self.assertEqual(one(contents['samples'])['entity_type'], [
```
achave11, line 887: I am wondering if this could benefit from `self.subTest` I'll leave it to your discretion.

danielsotirhos: The main purpose of the test is for the `cell_suspension['organ_part']` value, however since I am including a verification of the `contents['samples'][]['entity_type']` value I agree it should be checked in more cases than the current `if` block. I'll add the missing `else`.

hannes-ucsc: Yes, I don't think a subtest is needed here.

PR #1691, opened 2020-04-08
Fix: Manifest content hash is computed even when 'token' is provided

achave11:


achave11:
@hannes-ucsc it seems like your suggestion was not addressing the case in which the cached manifest isn't available, so a new one needs to be generated.

lambdas/service/app.py
```diff
@@ -1112,24 +1112,25 @@ def handle_manifest_generation_request():
...
+        return async_service.start_or_inspect_manifest_generation(retry_url,
+                                                                  token=token,
+                                                                  format_=format_,
+                                                                  filters=filters)
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'ExecutionDoesNotExist':
+            raise BadRequestError('Invalid token given')
+        raise
+    except ValueError as e:
+        raise BadRequestError(e.args)
```
achave11, line 1133: @hannes-ucsc 
Optionally, this could be extracted into a method and be utilize as an else branch for when the `token` is  available as well as another else branch for when the `presigned_url` is None.

hannes-ucsc: I think it's fine for now. The diff is pretty small when ignoring white space.

Issue #1688, opened 2020-04-08
If share ES domain is set to false in a new deployment `make deploy` fails


achave11:
Reproduce:

1) Configure new deployment as specified in README.md
2) In the `environment.py` file associated with the new deployment, set the variable `AZUL_SHARE_ES_DOMAIN` to 0 (Do not want to share ES domain with `dev`)
3) After all other necessary steps have been completed run `make deploy`

The following is observed:
```
$ make deploy
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C lambdas
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C indexer deploy
python .chalice/config.json.template.py .chalice/config.json
Traceback (most recent call last):
  File ".chalice/config.json.template.py", line 14, in <module>
    "environment_variables": aws.lambda_env,
  File "/Users/abrahamchavez/PycharmProjects/Azul/azul.sc/src/azul/deployment.py", line 105, in lambda_env
    return config.lambda_env(self.es_endpoint, self.es_instance_count)
  File "/Users/abrahamchavez/PycharmProjects/Azul/azul.sc/src/azul/deployment.py", line 92, in es_endpoint
    es_domain_status = self.es.describe_elasticsearch_domain(DomainName=config.es_domain)
  File "/Users/abrahamchavez/PycharmProjects/Azul/azul.sc/.venv-es/lib/python3.6/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/abrahamchavez/PycharmProjects/Azul/azul.sc/.venv-es/lib/python3.6/site-packages/botocore/client.py", line 661, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the DescribeElasticsearchDomain operation: Domain not found: azul-index-achave11es
make[2]: *** [.chalice/config.json] Error 1
make[1]: *** [indexer] Error 2
make: *** [deploy] Error 2

```

PR #1699, opened 2020-04-10
 Fix: `make deploy` requires ES resources to be provisioned (#1688)

achave11:


README.md
```diff
@@ -344,7 +344,7 @@ module is part of a rarely used feature that can be disabled by unchecking
 ## 3.1 One-time provisioning of shared cloud resources
 
 Most of the cloud resources used by a particular deployment (personal or shared) 
-are provisioned automatically by `make terraform` and `make deploy`. A handful 
+are provisioned automatically by `make deploy` and `make terraform`. A handful
```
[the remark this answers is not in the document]
achave11: I though it would be a good idea to have the sequence of commands be described in the same order. In any case, when `make package` comes along it is already ordered or we can remove it. But I could drop this commit message if you believe we would not gain any value.

achave11:
> Terraform Data Sources might need to be used for deployments that use anothers' ES instance
I don't quite understand this sentence @jessebrennan

jessebrennan:
Basically I was saying that if we used Terraform to obtain the ES endpoint, then deployments that use the ES instance of different deployment will be unable to access the Terraform field that defines the ES endpoint. They won't be able to access it because the Terraform configuration that defines the ES instance is part of the other deployment's Terraform config.

One could work around this problem by using Terraform Data Sources.

achave11:
> ...
One could work around this problem by using Terraform Data Sources.

I see, that is a great option to reduce the intervention of scripts in the provisioning of cloud resources but I think we are trying to achieve the opposite. To have a resilient deployment that does not require resources to be provided when they have not been created. Specifically the section that makes the lambda packages (`make deploy`).

README.md
```diff
@@ -399,6 +399,15 @@ make terraform
 which will display a plan and ask you to confirm it. Please consult the 
 Terraform documentation for details.
 
+If an environment is using Elasticsearch resources that are not shared by other
+environments, and you are deploying this environment for the first time, you
+will need to run
+```
```
hannes-ucsc, line 405: Newline before and after preformatted blocks.

src/azul/deployment.py
```diff
@@ -85,20 +85,48 @@ def api_gateway_export(self, gateway_id):
...
         else:
-            es_domain_status = self.es.describe_elasticsearch_domain(DomainName=config.es_domain)
-            return es_domain_status['DomainStatus']['Endpoint'], 443
+            log.warning('''
+                ES endpoint is not configured statically and needs to be
+                determined dynamically. This is expected when deploying for the
+                first time but should not occur inside the lambda. If it occurs
+                in the lambda, the fix is to run make deploy and make terraform
+                again.
+            ''')
```
hannes-ucsc, line 98:
```suggestion
            log.warning('''
The Elasticseach endpoint is not configured statically. This is normal during a deployment 
but if it occurs in a Lambda function, frequent dynamic look-ups are required. To eliminate 
that inefficiency, simply run `make deploy` followed by `make terraform` again.
            ''')
```

But fix wrapping.

src/azul/deployment.py
```diff
@@ -85,20 +85,48 @@ def api_gateway_export(self, gateway_id):
...
     @property
-    def es_instance_count(self) -> int:
+    def es_instance_count(self) -> Optional[int]:
         if config.es_endpoint:
             return config.es_instance_count
         else:
-            es_domain_status = self.es.describe_elasticsearch_domain(DomainName=config.es_domain)
-            return es_domain_status['DomainStatus']['ElasticsearchClusterConfig']['InstanceCount']
+            log.warning('''
+                ES instance count is not configured statically and needs to be
```
hannes-ucsc, line 115: Likewise.

src/azul/deployment.py
```diff
@@ -85,20 +85,48 @@ def api_gateway_export(self, gateway_id):
         return json.load(response['body'])
 
     @property
-    def es_endpoint(self) -> Netloc:
+    def es_endpoint(self) -> Optional[Netloc]:
```
hannes-ucsc, line 88: This method and the one below are 95% identical. We should extract the common parts into a method. This should have occurred to either of us earlier.

src/azul/deployment.py
```diff
@@ -85,20 +86,45 @@ def api_gateway_export(self, gateway_id):
...
 
     @property
-    def es_instance_count(self) -> int:
+    def es_instance_count(self) -> Optional[int]:
         if config.es_endpoint:
             return config.es_instance_count
         else:
+            return self._describe_es_domain(config.es_instance_count)
+
+    def _describe_es_domain(self, es_config_param) -> Union[Netloc, int, None]:
```
hannes-ucsc, line 102: Too complicated. I don't want to give too much away because I want you to think harder about this. I think you'll learn more from that than if I held your hand through the refactoring.

README.md
```diff
@@ -399,6 +399,17 @@ make terraform
 which will display a plan and ask you to confirm it. Please consult the 
 Terraform documentation for details.
 
+If an environment is using Elasticsearch resources that are not shared by other
+environments, and you are deploying this environment for the first time, you
+will need to run
```
hannes-ucsc, line 404: The word is *deployment*, not *environment*.

hannes-ucsc:
```suggestion
Unless a deployment uses the Elasticsearch domain of another one (see AZUL_SHARE_ES_DOMAIN), you need to run
```

README.md
```diff
@@ -399,6 +399,17 @@ make terraform
...
+
+```
+make deploy
+make terraform
```
hannes-ucsc, line 408:
```suggestion
make deploy terraform
```

README.md
```diff
@@ -399,6 +399,17 @@ make terraform
...
+```
+
+again.
```
hannes-ucsc, line 411:
```suggestion
again. If you don't, the deployment will work but it will run slower and incur extra cost. Running the `deploy` and `terraform` targets twice is only necessary when the deployment is first created.
```

hannes-ucsc:
Not bumping review count since I could have seen the issue earlier.

hannes-ucsc:
This is labelled as `chain` but no dependency PR is set.

achave11:
> This is labelled as `chain` but no dependency PR is set.

I did not remember about the blocking relationship.

Issue #1645, opened 2020-03-18
Rethink template config variable mechanism and make `resolve_env` transitive


hannes-ucsc:
Before #1637, `AZUL_SUBDOMAIN_TEMPLATE` contains `{lambda_name}`. In #1637, this construct needed to be escaped as `"{{lambda_name}}"` in `environment.py` to protect it from being interpreted as a variable reference by `export_environment.resolve_env`. We'd like that function to be transitive such that we can resolve references in referenced variables but the choice of curly braces to delimit the `lambda_name` template variable prevents us from doing so. To support transitivity, we'd have to repeatedly apply the resolution until nothing gets substituted anymore. After the first iteration, the escaped `"{{lambda_name}}"` would become `"{lambda_name}"` , which would become `""` after the second iteration. That's clearly not what we want.

We should either use different delimiters, or use string concatenation instead of template string substitution, or eliminate the variable.

Then we can change the body of `resolve_env` to

```
    while True:
        resolved_env = {k: v.format_map(env) for k, v in env.items()}
        if env == resolved_env:
            return resolved_env
        else:
            env = resolved_env
```

PR #1705, opened 2020-04-15
Redesign template config variable mechanism (#1645)

danielsotirhos:


danielsotirhos:
@hannes-ucsc Requesting a preliminary WIP review.

UPGRADING.rst
```diff
@@ -10,6 +10,28 @@ branch that does not have the listed changes, the steps would need to be
 reverted. This is all fairly informal and loosely defined. Hopefully we won't
 have too many entries in this file.
 
+#1645 Rethink template config variable mechanism
```
hannes-ucsc, line 13: Glad you thought of this.

environment.py
```diff
@@ -91,17 +90,17 @@ def env() -> Mapping[str, Optional[str]]:
         # deployment's indexer to assume for direct access. Presumably that other
         # indexer has sufficient privileges to directly access the DSS buckets.
         #
-        # The string {lambda_name} will be replaced with the name of the lambda
-        # wishing to gain access. This parameterization can be used to have the
+        # The character '*' will be replaced with the name of the lambda
+        # wishing to gain access. This parametrization can be used to have the
```
hannes-ucsc, line 94:
```suggestion
        # wishing to gain access. This parameterization can be used to have the
```

deployments/sandbox/environment.py
```diff
@@ -21,27 +21,23 @@ def env() -> Mapping[str, Optional[str]]:
...
-        #
         # Only modify this file if you intend to commit those changes. To change the
         # environment with a setting that's specific to you AND the deployment, create a
-        # environment.local right next to this file and make your changes there. In that
-        # file the same rules apply with regard to `_set` vs `export`. Settings
-        # applicable to all environments but specific to you go into environment.local
-        # at the project root.
+        # environment.local right next to this file and make your changes there.
+        # Settings applicable to all environments but specific to you go into
+        # environment.local at the project root.
```
hannes-ucsc, line 28:
```suggestion
        # environment.local.py at the project root.
```

Here and in the other copies. It's my bug but since you touched this …

scripts/convert_environment.py
```diff
@@ -121,19 +121,21 @@ def env() -> Mapping[str, Optional[str]]:
 
 
 def convert_value(value: str) -> Optional[str]:
```
hannes-ucsc, line 123: Don't understand why you need to touch this function …

scripts/convert_environment.py
```diff
@@ -121,19 +121,21 @@ def env() -> Mapping[str, Optional[str]]:
...
+    """
+    Convert shell-style interpolations to Python str.format() templates.
+    Translate `$foo` and `${foo}` to `{foo}`.
+    Translate special case '{{lambda_name}}' to '*'.
```
hannes-ucsc, line 127: … but I'm pretty sure it should not implement this special case. The case should not occur. The string `{lambda_name}` should be gone from the code base.

danielsotirhos:
Reverted change to `convert_environment.py` and rebased on latest develop.
Unsure why coveralls report never came through, currently it is sitting with status
```
coverage/coveralls Expected — Waiting for status to be reported
```

danielsotirhos:
Taking ticket out of `Review/QA` status to rebase on latest `develop` to resolve brach conflicts.

scripts/export_environment.py
```diff
@@ -170,32 +160,45 @@ def resolve_env(env: Environment) -> Environment:
     >>> resolve_env({'x': '{y}', 'y': '42'})
     {'x': '42', 'y': '42'}
 
-    Literal curly braces need to be escaped:
```
hannes-ucsc, line 173: Where did this case go?

scripts/export_environment.py
```diff
@@ -170,32 +160,45 @@ def resolve_env(env: Environment) -> Environment:
...
 
-    Transitive resolution would yield
+    Compared to:
 
-    {'x': '42', 'y': '42', 'z': '42'}
+    >>> resolve_env({'x': '{{y}}', 'y': 'z', 'z': '42'})
+    {'x': 'z', 'y': 'z', 'z': '42'}
 
-    Does not support generative ones (references formed during resolution):
+    References to a missing key yields an empty string replacement:
```
hannes-ucsc, line 179:
```suggestion
    A reference to a missing key yields an empty string replacement:
```

scripts/export_environment.py
```diff
@@ -170,32 +160,45 @@ def resolve_env(env: Environment) -> Environment:
...
 
-    >>> resolve_env({'x': '{y}}}', 'y': '{{z', 'z': '42'})
-    {'x': '{{z}', 'y': '{z', 'z': '42'}
+    >>> resolve_env({'x': 'y is {y}'})
+    {'x': 'y is '}
 
-    which could be
+    Unmatched curly braces are not allowed:
```
hannes-ucsc, line 184: So how _do_ I embed a single, literal, unmatched curly brace in a value?

scripts/export_environment.py
```diff
@@ -170,32 +160,45 @@ def resolve_env(env: Environment) -> Environment:
...
 
-    {'x': '42', 'y': '{z', 'z': '42'}
+    >>> resolve_env({'x': '{y'})
+    Traceback (most recent call last):
+    ...
+    ValueError: expected '}' before end of string
     """
-    # `{missing_var}` references in values should yield the empty string,
```
hannes-ucsc, line 196: Why did you think it necessary to remove this? I see you added a test but the motivation is lost.

scripts/export_environment.py
```diff
@@ -170,32 +160,45 @@ def resolve_env(env: Environment) -> Environment:
...
-    # consistent with $missing_var in the shell.
-    return {k: v.format_map(defaultdict(str, env)) for k, v in env.items()}
+    while True:
+        resolved_env = {}
+        for k, v in env.items():
+            try:
+                resolved_env[k] = v.format_map(env)
+            except KeyError as e:
+                resolved_env[k] = v.replace('{' + e.args[0] + '}', '')
```
hannes-ucsc, line 197: This smells a little because it is emulating format_map's behaviour, which is tends to be brittle or incomplete.

Consider `'{a!s}'.format_map(dict(b=1))`

What's wrong with the defaultdict approach?

scripts/export_environment.py
danielsotirhos: This test case was removed because making `resolve_env()` a recursive function to support transitive references removed the ability to support unmatched curly braces as they cause `format_map()` to raise a ValueError. 

That said, I've given this some more thought and come up with a solution that allows for the transitive & generative properties desired while also allowing for unmatched curly braces.

scripts/export_environment.py
danielsotirhos: In my original solution this was not possible. In my current implementation any unmatched curly braces or curly braces that do not wrap a valid syntax key in the mapping will be ignored.

scripts/export_environment.py
danielsotirhos: The comment was removed as it seemed repetitive now that the behavior is documented in the docstring test. It was an oversight on my part to assume the the motivation of the behavior wasn't necessary to document.

scripts/export_environment.py
danielsotirhos: Ah, the ticket description had a recommended change for the function that omitted the `defaultdict` so I thought it was desired to remove it. https://github.com/DataBiosphere/azul/issues/1645#issue-583472613

I've reimplemented `defaultdict` in my current solution.

hannes-ucsc: I think I added the defaultdict after I created #1645.

scripts/export_environment.py
```diff
@@ -170,32 +162,67 @@ def resolve_env(env: Environment) -> Environment:
...
 
-    Transitive resolution would yield
+    Generative references formed during resolution are supported:
 
-    {'x': '42', 'y': '42', 'z': '42'}
+    >>> resolve_env({'x': '{y}}', 'y': '{z', 'z': '42'})
+    {'x': '42', 'y': '{z', 'z': '42'}
 
-    Does not support generative ones (references formed during resolution):
+    Consistent with a $missing_var in the shell a references to a missing key
```
hannes-ucsc, line 181:
```suggestion
    Similar to how Unix shells replace references to non-existing variables with the empty string, a reference to a missing key
```

scripts/export_environment.py
```diff
@@ -170,32 +162,67 @@ def resolve_env(env: Environment) -> Environment:
...
-    return {k: v.format_map(defaultdict(str, env)) for k, v in env.items()}
+    while True:
+        resolved_env = {k: _custom_format_map(k, v, env) for k, v in env.items()}
+        if env == resolved_env:
+            return resolved_env
+        else:
+            env = resolved_env
+
+
+def _custom_format_map(key: str, val: str, mapping: Mapping[str, str]) -> str:
```
hannes-ucsc, line 202: Not exactly an improvement because it feels like it's adding more assumptions about format_map. Also, this is complex enough to warrant dedicated doctests. Those would also help to more accurately document the behavior (the docstring below is somewhat ambiguous). For example, what is a "valid syntax keyword"? (rhetorical question)

We always have to weight the gain in functionality against the complexity of the approach and this feels off balance. I think the breadth first approach in the calling function is what gets us into trouble. I think I have an idea of how this can be simplified at an acceptable cut in functionality.

PR #1709, opened 2020-04-16
Allow detached head for personal deployments in make check_branch

noah-aviel-dove:


scripts/check_branch.py
```diff
@@ -72,11 +72,14 @@ def current_branch():
...
         # (no active branch). Extract the branch name from the runner environment.
-        branch = os.environ['CI_COMMIT_REF_NAME']
+        branch_name = os.environ['CI_COMMIT_REF_NAME']
     except KeyError:
+        # Detached head may also occur outside of the gitlab environment, in
+        # which case it is only allowed for personal deployments.
         repo = git.Repo(config.project_root)
-        branch = repo.active_branch.name
-    return branch
+        branch_name = 'DETACHED HEAD' if repo.head.is_detached else repo.active_branch.name
```
hannes-ucsc, line 80: Before inventing special magic values, you should always consider if the type system can be used to express special cases. A detached head implies that there is no branch checked out so `None` is much more adequate value, even if it requires more changes downstream of this function.

noah-aviel-dove: I chose this value to produce informative error messages.

hannes-ucsc: I understand. I didn't mean to imply that you chose this value without consideration. My version also produces informative messages but doesn't have the flaw of overloading the branch name when there really is no branch.

Issue #1522, opened 2020-01-30
Upgrade Elasticsearch version to 6.8 


jessebrennan:
what breaks if we upgrade to a supported version? the latest 7.1 and the oldest one 6.8 still in support.

PR #1719, opened 2020-04-20
Upgrade Elasticsearch version to 6.8 (#1522)

achave11:


terraform/elasticsearch.tf.json.template.py
```diff
@@ -91,7 +91,7 @@
                         "volume_size": config.es_volume_size,
                         "volume_type": "gp2"
                     },
-                    "elasticsearch_version": "5.5",
+                    "elasticsearch_version": "6.8",
```
danielsotirhos, line 94: Now that the version specified in requrirements.txt is the full 'major.minor.patch', should this be more specific to match?
Also, probably not worth it since how infrequently we change the version but did you consider reading this value out of the package?
```
>>> from elasticsearch import VERSION as ES_VERSION
>>> '.'.join(map(str, ES_VERSION))
'6.8.1'
```

achave11: >Now that the version specified in requrirements.txt is the full 'major.minor.patch', should this be more specific to match?

I am not sure how that would interfere with how AWS defines the version of its [Elasticsearch service](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/what-is-amazon-elasticsearch-service.html#aes-choosing-version).

>Also, probably not worth it since how infrequently we change the version but did you consider reading this value out of the package?

I had not consider that possibility, I think we could benefit from it. But there could be a benefit in not being too strict given that Elasticsearch servers might be compatible with minor & patched client libraries without necessarily going through the trouble of updating the server. But I am not sure how best practices would disallow us to do that. @hannes-ucsc ?

danielsotirhos: Looks like you are right that AWS only lets you specify a major & minor version and not a full major.minor.patch syntax. I unfortunately can't seem to find any documentation that explicitly mentions this though.
Screen Shot 2020-04-30 at 11 12 46 AM

We could still possibly pull the version out of the package by only using the first two values in the tuple, although this was just a thought I had to make the terraform config more dynamic and not a requirement in any way.

```
>>> from elasticsearch import VERSION as ES_VERSION
>>> '.'.join(map(str, ES_VERSION[:2]))
'6.8'
```

hannes-ucsc: One is the client version the other is the server. They don't need to match, just be compatible and that can't be expressed easily. It's entirely possible that a 6.7.3 client should be able to read from a 6.8.1 server, or even a 7.x server. It all depends. So I wouldn't bother tying them together.

requirements.txt
```diff
@@ -7,13 +7,14 @@ botocore==1.13.19
...
-elasticsearch>=5.0.0,<6.0.0
-elasticsearch-dsl>=5.0.0,<6.0.0
+elasticsearch==6.8.1
+elasticsearch-dsl==6.4.0
 jsonobject==0.9.9
 Faker==0.8.11
 furl==2.0.0
 future==0.16.0
 hca==7.0.1
+httplib2==0.17.2
```
hannes-ucsc, line 17: Is this a transitive dependency?

test/es_test_case.py
```diff
@@ -26,7 +26,7 @@ class ElasticsearchTestCase(DockerContainerTestCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        es_endpoint = cls._create_container('docker.elastic.co/elasticsearch/elasticsearch:5.5.3',
+        es_endpoint = cls._create_container('docker.elastic.co/elasticsearch/elasticsearch:6.8.1',
```
hannes-ucsc, line 29: This is patch level 1. What patch level do we get on AWS?

test/es_test_case.py
```diff
@@ -40,16 +40,6 @@ def setUpClass(cls):
...
             cls._kill_containers()
             raise
-        else:
-            # By default `max_compilations_per_minute` is set to 15, we must increase that threshold in order to execute
-            # dynamic header row and HashCode generation for test_full_hash_validity.
-            # If this also happens in production code, we should move this code to be run as part of `make terraform`,
-            # as a provisioner, for example. Or we can precompile the Painless scripts as part of `make deploy`.
-            cls.es_client.cluster.put_settings(body={
-                'persistent': {
-                    'script.max_compilations_per_minute': 60
```
hannes-ucsc, line 50: It appears the setting has been renamed, but still has the same default of 15/min (well, 75/5min to be precise, which seems to be equivalent). How is it possible that we can completely remove this override even though the same default rate applies?

test/service/test_request_builder.py
```diff
@@ -78,9 +78,6 @@ def test_create_request(self):
                         }
                     ]
                 }
-            },
```
hannes-ucsc, line 81: Can you talk a little about why these can be removed?

achave11: Yes, since we do not actually use the `query` route of the search request for this particular cases, it is not being specified in the request. It seem to have been a bug, you can read further [here](https://github.com/elastic/elasticsearch-dsl-py/issues/924).

test/es_test_case.py
achave11: You could technically say they are equivalent but they are different non the less. The period for circuit breaker is now more ample for invoking the script, compile it and run it without exhausting the allowed rate. The compiling of the script is only done for manifest generation requests, and since that portion of our code does not run constantly we can get by with removing it. Besides, if you recall this was a problem that would manifest it self occasionally when running manifest object key generation tests. Which to me indicates that we might have been right on the limit of the exhaustion, extending the period benefited us.

test/es_test_case.py
achave11: I am unsure as to what patch level is given by AWS, since I was unable to find that information from AWS. If you look at the [elasticsearch repo](https://github.com/elastic/elasticsearch), the branches they have available for listing do not show a patch level either.

requirements.txt
achave11: That is correct, it seem to have played a role in the test that was failing.

achave11: This was removed, since it was not actually the root of the problem. The issues has been addressed.

test/es_test_case.py
achave11: That is true. It is also true that 
```
60/5 + 10/5 + 0/5 + 2/5 + 3/5 == 75/5
```
The concern isn't mathematical but rather temporal. The average of 75/5min over a period of a minute is 15/1min, however we are not constrained to 1 minute but instead we are constrained to 5 minutes. We must compile our scripts less than 75 times before the period of 5 minutes. Taking into account that we compile scripts occasionally when Service is required to obtain a content-hash from Elasticsearch for a given manifest, we know that this process is not continuous e.i. it does not happen more than 75 times in 5 minutes. Unless of course Service is bombarded with more than 75 manifest request within a period of 5 minutes.
So if each of the following fraction represents the number of compilations in one minute over a span of five minute e.i. 5  fractions.
```
60/5, 0/5, 0/5, 10/5, 5/5
```
We can see that it is possible to compile at least 60 times in one minute leaving a pool of 15 compilations for the rest of the 4 minutes. Therefore, not requiring an expansion on the default limit of compilations. That is my reasoning, perhaps there is something that I am missing, 🤔 in which case I would appreciate if we discuss it in a teleconference.

test/es_test_case.py
```diff
@@ -26,7 +26,7 @@ class ElasticsearchTestCase(DockerContainerTestCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        es_endpoint = cls._create_container('docker.elastic.co/elasticsearch/elasticsearch:5.5.3',
```
achave11, line 29: It is worth mentioning that the Elasticsearch instance in AWS uses version `5.5.1` in its nodes.

test/es_test_case.py
hannes-ucsc: Nope, that makes sense.

test/indexer/test_hca_indexer.py
```diff
@@ -96,6 +97,8 @@ def test_indexing(self):
         hits = self._get_all_hits()
         self.assertElasticsearchResultsEqual(expected_hits, hits)
 
+    http.client._MAXHEADERS = 1000  # Prevents `test_deletion` to fail; https://stackoverflow.com/questions/23055378
```
hannes-ucsc, line 100: I understand the motivation (the comment makes that clear) but think about why it isn't a good idea to do this here.

test/es_test_case.py
hannes-ucsc: Interesting. But I still think we should attempt to match in the future even if we currently don't match them.

test/indexer/test_hca_indexer.py
achave11: Yea my conception here was to propagate the patch at the highest context possible in this test suite. But as you hinted, only `test_deletion` requires the patch, so it is not necessary to have it at the class level 👍. Right?

hannes-ucsc: For one, this isn't a test suite. It's a test *case*. But the way this patch is done here has consequences beyond the test case. What are they? And why are they bad?

achave11: > What are they? And why are they bad?

Since `http` is the module that defines classes that implement the client of the `HTTP` and `HTTPS` protocols it would change the default configuration.  And since this module is used by other modules/libraries this change would be propagated through the modules that use `http` and those modules would also be affected by this change. Allowing for unwanted behavior to manifest it self in other parts of the codebase unrelated to this test case.

test/indexer/test_hca_indexer.py
```diff
@@ -101,6 +102,8 @@ def test_deletion(self):
         Delete a bundle and check that the index contains the appropriate flags
         """
 
+        http.client._MAXHEADERS = 1000  # Prevents test from failing; https://stackoverflow.com/questions/23055378
```
hannes-ucsc, line 105: Still wrong.

achave11:
Integration test summary:

\----------------------------------------------------------------------
Ran 6 tests in 922.799s

OK

hannes-ucsc:
When I looked over the logs for the complete reindexing, I found 9,431 ECEs and only one ScanError. No other exceptions.

achave11:
[Upgrading Doc](https://docs.google.com/document/d/17ZP0nPTkpvoC9TJsoiiXCVzZDlnG72TvokmY-Od_Y48/edit?usp=sharing)

hannes-ucsc:
I don't seem to be able to make comments or suggest changes to that document. Could you share it in a way that allows me to do so?

achave11:
> I don't seem to be able to make comments or suggest changes to that document. Could you share it in a way that allows me to do so?

@hannes-ucsc your permissions have been modified. That was an overlook on my part.

achave11:
> Did one pass over the document. Some minor wording changes and request to provide specifics.
> 
> And there are conflicts to be resolved.

✅ Resolve conflicts
✅ Upgrade via AWS console
❌ Document modifications

Seems like it's ready for sandbox.

Issue #1712, opened 2020-04-16
`dev.sc` should not specify role assumptors


hannes-ucsc:

melainalegaspi:
@achave11 ask @danielsotirhos for details

PR #1721, opened 2020-04-21
Remove role assumptors for dev.sc & sandbox deployments (#1712)

achave11:


src/azul/__init__.py
```diff
@@ -190,12 +190,12 @@ def num_dss_workers(self) -> int:
...
     @property
     def external_lambda_role_assumptors(self) -> MutableMapping[str, List[str]]:
-        try:
-            accounts = os.environ['AZUL_EXTERNAL_LAMBDA_ROLE_ASSUMPTORS']
-        except KeyError:
-            return {}
-        else:
+        accounts = os.environ.get('AZUL_EXTERNAL_LAMBDA_ROLE_ASSUMPTORS')
+        # This env variable should only be set for HCA deployments
+        if accounts is not None and config.domain_name.endswith('explore.data.humancellatlas.org'):
```
jessebrennan, line 195: Is there a way to avoid hard-coding the URL here? What if you just left this file the way it was except for the comment on line 194. The HCA deployments will still have this environment variable set AFAICT, and everything should just work. What do you think of that approach?

Otherwise, this looks good.

src/azul/__init__.py
```diff
@@ -190,12 +190,12 @@ def num_dss_workers(self) -> int:
 
     @property
     def external_lambda_role_assumptors(self) -> MutableMapping[str, List[str]]:
-        accounts = os.environ.get('AZUL_EXTERNAL_LAMBDA_ROLE_ASSUMPTORS')
-        # This env variable should only be set for HCA deployments
```
jessebrennan: I said this in my last review, but I think this comment is useful and should be preserved.

src/azul/__init__.py
```diff
@@ -191,6 +191,7 @@ def num_dss_workers(self) -> int:
     @property
     def external_lambda_role_assumptors(self) -> MutableMapping[str, List[str]]:
         try:
+            # This env variable should only be set for HCA deployments
```
hannes-ucsc, line 194: I wouldn't go that far. Currently only the HCA deployments have this set. But that could change in the future in which case this comment would become incorrect.

Issue #1609, opened 2020-03-02
Some endpoints may return a 500 when mandatory arguments are omitted


hannes-ucsc:
@noah-aviel-dove found that 

https://github.com/DataBiosphere/azul/blob/1859a365ee51529d3a8cd08fd00f5a41ca759f2f/lambdas/service/app.py#L1097

was lacking `or {}` at the end. That particular occurrence will be fixed in https://github.com/DataBiosphere/azul/pull/1586 but there may be other occurrences. The response should either be a 4xx status response if some or all parameters are required or a successful response otherwise.

hannes-ucsc:
Make a pass over all endpoints to verify if this is an issue anywhere. Then fix as needed.

noah-aviel-dove:
There are no remaining endpoints that produce 500 errors when arguments are missing.

noah-aviel-dove:
`/dss/files` and `/fetch/dss/files` however do produce ugly stack traces when the `replica` parameter is omitted:

```
Traceback (most recent call last):
  File "/var/task/chalice/app.py", line 1104, in _get_view_function_response
    response = view_function(**function_args)
  File "/var/task/app.py", line 1282, in fetch_dss_files
    body = _dss_files(uuid, fetch=True)
  File "/var/task/app.py", line 1348, in _dss_files
    dss_response.raise_for_status()
  File "/var/task/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://dss.staging.data.humancellatlas.org/v1/files/ee6a75bd-3252-41ee-b253-425bbd377f0c
```

However Azul doesn't actually require this parameter, it just forwards it to DSS, which does require it. `replica` is documented as required in the Swagger UI but is not checked during parameter validation. Given that we'll soon be migrating away from DSS I suspect it's not worth the bother of filing a PR at this.

hannes-ucsc:
Unless the fix is complicated, I would like use to move forward with this.

noah-aviel-dove:
The fix is trivial. I'll take care of it.

PR #1728, opened 2020-04-24
Validate mandatory param for DSS (#1609)

noah-aviel-dove:
See [comments on the ticket](https://github.com/DataBiosphere/azul/issues/1609#issuecomment-619245458).

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -652,7 +652,7 @@ def validate_params(query_params: Mapping[str, str],
     >>> validate_params({'size': 'foo'}, size=int)
     Traceback (most recent call last):
         ...
-    chalice.app.BadRequestError: BadRequestError: Invalid input type for `size`
+    chalice.app.BadRequestError: BadRequestError: Invalid input value for `size`
 
     >>> validate_params({'order': 'asc', 'foo': 'bar'}, order=str)
     Traceback (most recent call last):
@@ -697,7 +697,7 @@ def validate_params(query_params: Mapping[str, str],
         try:
             validator(param_value)
         except (TypeError, ValueError):
-            raise BadRequestError(msg=f'Invalid input type for `{param_name}`')
+            raise BadRequestError(msg=f'Invalid input value for `{param_name}`')
 
 
 @app.route('/integrations', methods=['GET'], cors=True)
@@ -1376,7 +1376,17 @@ def fetch_dss_files(uuid):
 
 def _dss_files(uuid, fetch=True):
     query_params = app.current_request.query_params or {}
-    validate_params(query_params, True, fileName=str, wait=int, requestIndex=int)
+
+    def validate_replica(replica):
+        if replica != 'aws':
+            raise ValueError
+
+    validate_params(query_params,
+                    allow_extra_params=True,
+                    fileName=str,
+                    wait=int,
+                    requestIndex=int,
+                    replica=Mandatory(validate_replica))
     dss_endpoint = config.dss_endpoint
     url = dss_endpoint + '/files/' + urllib.parse.quote(uuid, safe='')
     file_name = query_params.pop('fileName', None)


lambdas/service/app.py
```diff
@@ -1283,7 +1285,18 @@ def fetch_dss_files(uuid):
...
+    def validate_replica(replica):
+        if replica != 'aws':
+            raise ValueError
+
+    validate_params(query_params,
+                    allow_extra_params=True,
+                    fileName=str,
+                    wait=int,
+                    requestIndex=int,
+                    # See method spec for the DSS endpoints
```
achave11, line 1298: Not sure the comment is necessary. See other uses of `Mandatory`

lambdas/service/app.py
```diff
@@ -1283,7 +1285,18 @@ def fetch_dss_files(uuid):
...
+                    replica=Mandatory(validate_replica))
```
achave11, line 1299: Not sure the comment is necessary. See other uses of `Mandatory`.

noah-aviel-dove:
Requesting review with incomplete checks because codecov appears to be indefinitely stuck.

lambdas/service/app.py
```diff
@@ -1436,8 +1449,12 @@ def _dss_files(uuid, fetch=True):
                                                      Params=params)
         response = {"Status": 302, "Location": location}
     else:
-        dss_response.raise_for_status()
-        assert False
+        try:
+            dss_response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
```
hannes-ucsc, line 1454: What's the intent here? It appears that any non-success HTTP status code (4xx 5xx etc) from DSS is translated to a 400. That can't be right.

lambdas/service/app.py
```diff
@@ -696,8 +696,10 @@ def fmt_error(err_description, params):
         validator = validators[param_name]
         try:
             validator(param_value)
-        except (TypeError, ValueError):
+        except TypeError:
             raise BadRequestError(msg=f'Invalid input type for `{param_name}`')
+        except ValueError:
```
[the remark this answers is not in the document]
noah-aviel-dove: https://github.com/DataBiosphere/azul/issues/1761

lambdas/service/app.py
noah-aviel-dove: `raise_for_status` raises `HTTPError` which isn't caught by chalice resulting in a stacktrace being printed instead of a JSON response showing the error message. I was under the impression we were trying to avoid that.

lambdas/service/app.py
```diff
@@ -1436,8 +1446,12 @@ def _dss_files(uuid, fetch=True):
                                                      Params=params)
         response = {"Status": 302, "Location": location}
     else:
-        dss_response.raise_for_status()
-        assert False
+        response = {
+            "Status": dss_response.status_code,
+            "Message": {
```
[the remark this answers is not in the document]
noah-aviel-dove: Apologies. I've replied to your previous comments directly. I will undo my changes to this section as it appears I have been allocating effort ineffectively.

lambdas/service/app.py
hannes-ucsc: I see. I'm assuming by "printed" you mean sent back to the client as part of the response. That bhaviour is controlled by AZUL_DEBUG. See AzulChaliceApp. If this is the only reason this change was made, we can revert the change.

noah-aviel-dove:
> Can you make sure to use correct spelling in commit titles? DSS is an acronym. Spelling and grammatical errors are, at best, a distraction to me (I can't help but notice) and a source for confusion and ambiguity at worst.

Assuming you're referring to capitalization, I guess I'm just accustomed to seeing DSS in lower case in the source code. I understand PR's are different though and I'll do my best to remember.

hannes-ucsc:
> I understand PR's are different though 

Not PRs. Prose, English. Commit & PR titles, issue titles, code comments, documentation etc.

>  guess I'm just accustomed to seeing DSS in lower case in the source code

That's why we use back quotes in prose to distinguish a source code artefact from the concept.

Issue #1646, opened 2020-03-18
app.route `method_spec` param value overwritten by another app.route `method_spec`


danielsotirhos:
The `method_spec` param of one `@app.route` is being overwritten by another `@app.route`'s `method_spec` when both `@app.route` have a shared `path_spec`.

Example:
```
swagger_test_path_spec = {
    'parameters': [
        params.query('foo', schema.optional({'type': 'string'})),
    ]
}


@app.route('/swagger-test-1', methods=['GET'], cors=True, path_spec=swagger_test_path_spec, method_spec={
    'summary': 'Swagger test 1',
    'description': 'Swagger test 1',
    'responses': {'201': {'description': 'OK'}}
})
def swagger_test_1():
    pass


@app.route('/swagger-test-2', methods=['GET'], cors=True, path_spec=swagger_test_path_spec, method_spec={
    'summary': 'Swagger test 2',
    'description': 'Swagger test 2',
    'responses': {'202': {'description': 'OK'}}
})
def swagger_test_2():
    pass
```

Expected Result: In the generated Swagger documentation both endpoints would have their unique `method_spec` values.

Acutal Result: The value of the `method_spec` from the `@app.route` last to occur in the source code is applied to both `@app.route`.

<img width="715" alt="Screen Shot 2020-03-18 at 10 34 10 AM" src="https://user-images.githubusercontent.com/46407464/76989944-685e0800-6904-11ea-9477-001bd3f75838.png">

Note that the effect can be negated with a shallow copy of the path_spec:
```
@app.route('/swagger-test-1', methods=['GET'], cors=True, path_spec=copy.copy(swagger_test_path_spec), ...
```

hannes-ucsc:
Residual FIXMEs

https://github.com/DataBiosphere/azul/blob/65970a0947f38fae439a3bf8fd960d351787b7a3/lambdas/service/app.py#L1395
https://github.com/DataBiosphere/azul/blob/65970a0947f38fae439a3bf8fd960d351787b7a3/lambdas/service/app.py#L1460

PR #1731, opened 2020-04-27
Add test proving path_spec can be shared (#1646)

jessebrennan:


jessebrennan:
I'm going to reopen this because I'm assuming it was closed unintentionally. @hannes-ucsc if this was closed intentionally, can you leave a comment explaining that?

test/test_openapi.py
```diff
@@ -120,6 +121,67 @@ def route2():
...
 
+    def test_shared_path_spec(self):
+        app = AzulChaliceApp('testing', spec={'foo': 'bar'})
+        swagger_test_path_spec = {
+            'parameters': [
+                params.query('foo', schema.optional({'type': 'string'})),
+            ]
+        }
+
+        @app.route('/swagger-test-1', methods=['GET'], cors=True, path_spec=swagger_test_path_spec, method_spec={
```
hannes-ucsc, line 132: Duplicative.

test/test_openapi.py
```diff
@@ -120,6 +121,67 @@ def route2():
...
+                    }],
+                    'get': {
+                        'summary': 'Swagger test 1',
+                        'description': 'Swagger test 1',
+                        'responses': {
+                            '201': {'description': 'OK'}
+                        }
+                    }
+                },
+                '/swagger-test-2': {
```
hannes-ucsc, line 166: Duplicative.

test/test_openapi.py
```diff
@@ -120,6 +121,67 @@ def route2():
...
+
+        @app.route('/swagger-test-2', methods=['GET'], cors=True, path_spec=swagger_test_path_spec, method_spec={
+            'summary': 'Swagger test 2',
+            'description': 'Swagger test 2',
+            'responses': {'202': {'description': 'OK'}}
+        })
+        def swagger_test_2():
+            pass
+
+        expected_spec = {
```
hannes-ucsc, line 148: If possible the assertion should not cover the other parts of the output, only that the `foo` path parameter is the same in the two routes.

test/test_openapi.py
jessebrennan: I removed some unnecessary content from the method spec so that this takes up less space, but for function definitions, I'm having a hard time thinking of a way to avoid at least _some_ duplication.

test/test_openapi.py
jessebrennan: I'm interpreting this comment to mean that I should consolidate this JSON into a dict comprehension.

test/test_openapi.py
jessebrennan: We want the assertion to cover all parts of the output. The original issue was that the method spec from the later route overwrote the method spec of the first one, therefore we need to assert the method spec as well.

jessebrennan: I added a docstring that hopefully makes that more clear.

hannes-ucsc: It does.

test/test_openapi.py
hannes-ucsc: Would a loop work?

image

But note how both functions observe i at its final value. The decorator doesn't though which should work for you here.

test/test_openapi.py
```diff
@@ -120,6 +121,48 @@ def route2():
...
+            ]
+        }
+
+        @app.route('/swagger-test-1', methods=['GET'], cors=True, path_spec=swagger_test_path_spec, method_spec={
+            'summary': 'Swagger test 1',
+        })
+        def swagger_test_1():
+            pass
+
+        @app.route('/swagger-test-2', methods=['GET'], cors=True, path_spec=swagger_test_path_spec, method_spec={
```
hannes-ucsc, line 142: I'd wrap this line. The amount of duplicated code is the same if one line or multiple.

jessebrennan: > Would a loop work?

The loop works! I don't know why I didn't think to try this.

> I'd wrap this line.

This line is wrapped already. Do you want

test/test_openapi.py
hannes-ucsc: I take that back. If the goal of the test is to ensure that method specs are not overwritten—which would cause two things to be the *same* that shouldn't be—then the assertion should use assertNotEqual on the method specs.

jessebrennan: Good point.

jessebrennan:
This rebased onto #1789 with no conflicts, so for the sake of simplicity, I'm going to leave the PR based on develop.

test/test_openapi.py
```diff
@@ -120,6 +121,27 @@ def route2():
...
+            ]
+        }
+        for i in range(2):
+            @app.route(f'/swagger-test-{i}', methods=['GET'], cors=True, path_spec=shared_path_spec, method_spec={
+                'summary': f'Swagger test {i}',
+            })
+            def swagger_test():
+                pass
+
+        method_specs = (spec for spec in app.specs['paths'].values())
```
hannes-ucsc, line 142:
```suggestion
        method_specs = app.specs['paths'].values()
```

Right?

jessebrennan: Absolutely.

test/test_openapi.py
```diff
@@ -120,6 +121,27 @@ def route2():
...
+        other's properties.
+        """
+        app = AzulChaliceApp('testing', spec={'foo': 'bar'})
+        shared_path_spec = {
+            'parameters': [
+                params.query('foo', schema.optional({'type': 'string'})),
+            ]
+        }
+        for i in range(2):
+            @app.route(f'/swagger-test-{i}', methods=['GET'], cors=True, path_spec=shared_path_spec, method_spec={
```
hannes-ucsc, line 136: Wrap all or none. In this case all. Sorry for not catching this earlier.

hannes-ucsc:
Not bumping review count.

test/test_openapi.py
```diff
@@ -120,6 +121,31 @@ def route2():
...
+        """
+        app = AzulChaliceApp('testing', spec={'foo': 'bar'})
+        shared_path_spec = {
+            'parameters': [
+                params.query('foo', schema.optional({'type': 'string'})),
+            ]
+        }
+        for i in range(2):
+            @app.route(
+                f'/swagger-test-{i}',
```
hannes-ucsc, line 137:
```suggestion
            @app.route(f'/swagger-test-{i}',
```

Aligned indent until the lines get too long.

jessebrennan: Sorry about that... I confused myself when I read "wrap all"

PR #1736, opened 2020-04-28
Fix: `_describe_es_domain_status` is subscribed when None is returned

achave11:


src/azul/deployment.py
```diff
@@ -91,18 +91,24 @@ def es_endpoint(self) -> Optional[Netloc]:
             return config.es_endpoint
         else:
             self._es_config_warning('endpoint')
-            return self._describe_es_domain_status['Endpoint'], 443
+            if self._describe_es_domain_status is None:
+                return None
+            else:
+                return self._describe_es_domain_status['Endpoint'], 443
```
hannes-ucsc, line 97: This makes two round-trips to ES which we should avoid.

src/azul/deployment.py
```diff
@@ -91,18 +91,26 @@ def es_endpoint(self) -> Optional[Netloc]:
             return config.es_endpoint
         else:
             self._es_config_warning('endpoint')
-            return self._describe_es_domain_status['Endpoint'], 443
+            es_domain_status = self._describe_es_domain_status
```
hannes-ucsc, line 94:
```suggestion
            status = self._describe_es_domain_status
```

Small scope, short names.

src/azul/deployment.py
```diff
@@ -91,18 +91,26 @@ def es_endpoint(self) -> Optional[Netloc]:
...
+            if es_domain_status is None:
```
hannes-ucsc, line 95: In this case an if-expression would be more appropriate.

src/azul/deployment.py
```diff
@@ -91,18 +91,26 @@ def es_endpoint(self) -> Optional[Netloc]:
...
+                return None
+            else:
+                return es_domain_status['Endpoint'], 443
 
     @property
     def es_instance_count(self) -> Optional[int]:
```
hannes-ucsc, line 101: Same comments apply to this method.

src/azul/deployment.py
```diff
@@ -91,18 +91,20 @@ def es_endpoint(self) -> Optional[Netloc]:
             return config.es_endpoint
         else:
             self._es_config_warning('endpoint')
-            return self._describe_es_domain_status['Endpoint'], 443
+            status = self._describe_es_domain_status
+            return (status['Endpoint'], 443) if status else status
```
hannes-ucsc, line 95:
```suggestion
            return None if status is None else (status['Endpoint'], 443)
```

hannes-ucsc: If we wanted to be less strict (which we don't), then your version could be abbreviated as `status and (status['Endpoint'], 443)`

PR #1743, opened 2020-04-30
Rename `deploy` target to `package` & `terraform` to `deploy` (#1654)

achave11:


CHANGELOG.yaml
```diff
@@ -4,7 +4,7 @@ meta:
   issues: a list of ticket URLs (GitHub, JIRA, ...) related to the change
   upgrade: a list of actions required to upgrade to a version that includes this change
   It may be any combination of the following values:
-    - deploy: (requires `make deploy`)
+    - package: (requires `make package`)
```
jessebrennan, line 7: Except for this instance, I'm not sure it makes sense to retroactively change this CHANGELOG entries. Note that I'm not requesting you to change this, but I'm just expressing my doubt.

UPGRADING.rst
```diff
@@ -25,7 +25,7 @@ to ::
     'AZUL_SUBDOMAIN_TEMPLATE': '*.{AZUL_DEPLOYMENT_STAGE}',
 
 
-#1272 Use Lambda layers to speed up ``make deploy``
+#1272 Use Lambda layers to speed up ``make package``
 ===================================================
```
jessebrennan, line 29: Extend the heading underline by one more character.

common.mk
```diff
@@ -24,7 +24,7 @@ endif
 
 ifeq ($(shell which terraform),)
 $(warning Looks like TerraForm is not installed. This is ok as long as you're not trying to create a new deployment. \
-          Deploying new lambdas is still possible with `make deploy` but `make terraform` will not work.)
+          Deploying new lambdas is still possible with `make package` but `make terraform` will not work.)
```
jessebrennan, line 27: This warning statement is actually not even correct, since `make package` doesn't actually deploy the lambda. Do you think you could fix it?

environment.py
```diff
@@ -213,7 +213,7 @@ def env() -> Mapping[str, Optional[str]]:
         'AZUL_DSS_WORKERS': '8',
 
         # Whether to create a subscription to DSS during deployment. Set this
-        # variable to 1 to enable `make deploy` to subscribe the indexer in the
+        # variable to 1 to enable `make package` to subscribe the indexer in the
```
jessebrennan, line 216: Same as above: this statement is actually not even correct, since make package doesn't actually deploy the lambda. Do you think you could fix it?

.gitlab-ci.yml
```diff
@@ -2,7 +2,7 @@ stages:
   - build_image
   - setup
   - test
-  - deploy
```
hannes-ucsc, line 5: The GL stage and job names should remain `deploy`.

achave11:
> There are a few text substitutions that don't make sense. I'm not gonna reveal where, to ensure that you do a full pass of all substitutions.

I did multiple passes over the substitutions, and I was unable to identify a non-sensical change. By the lack of the fix it commit you can rest assured there were no changes made unrelated to the rebase. Meaning the non pertinent changes you pointed out remain in this set of changes, if I can request that you at least point out the file path if you don't want to revel the exact location it would benefit me to understand why it does not make sense.

README.md
```diff
@@ -1482,7 +1482,7 @@ will **not** require redeployment. Any other changes **will**, such as
 
 To redeploy, run
 ```
-make -C lambdas/service deploy  # deploy to service only
+make -C lambdas/service package  # package to service only
```
hannes-ucsc, line 1485:
```suggestion
make -C lambdas/service package  # package the service only
```

Makefile
```diff
@@ -23,12 +23,12 @@ requirements: check_venv
...
 
-.PHONY: terraform
-terraform: check_env
-	$(MAKE) -C terraform
-
 .PHONY: deploy
 deploy: check_env
+	$(MAKE) -C terraform
+
+.PHONY: package
```
hannes-ucsc, line 30: The `package` target should precede the `deploy` target. That matches the natural sequence in which they should be run and would also make this diff hunk more legible.

README.md
```diff
@@ -318,7 +318,7 @@ module is part of a rarely used feature that can be disabled by unchecking
 ## 3.1 One-time provisioning of shared cloud resources
 
 Most of the cloud resources used by a particular deployment (personal or shared) 
-are provisioned automatically by `make deploy` and `make terraform`. A handful 
+are provisioned automatically by `make package` and `make deploy`. A handful
```
hannes-ucsc, line 321:
```suggestion
are provisioned automatically by `make deploy`. A handful 
```

And rewrap at 80.

UPGRADING.rst
```diff
@@ -25,8 +25,8 @@ to ::
     'AZUL_SUBDOMAIN_TEMPLATE': '*.{AZUL_DEPLOYMENT_STAGE}',
 
 
-#1272 Use Lambda layers to speed up ``make deploy``
```
hannes-ucsc, line 28: Revert the changes to this file. These instructions should match source at the commit that introduced them. We shouldn't retroactively change them.

environment.py
```diff
@@ -213,12 +213,12 @@ def env() -> Mapping[str, Optional[str]]:
         'AZUL_DSS_WORKERS': '8',
 
         # Whether to create a subscription to DSS during deployment. Set this
-        # variable to 1 to enable `make deploy` to subscribe the indexer in the
+        # variable to 1 to enable `make subscribe` to subscribe the indexer in the
```
hannes-ucsc, line 216: Good catch.

achave11: @jessebrennan and I were able to address this together.

Issue #1741, opened 2020-04-29
IndexNotFoundError has deceiving message


jessebrennan:
The message: `f'{missing_index} is not a valid uuid.')` in deceiving because an index name should generally not be a UUID.

PR #1750, opened 2020-05-01
Reword IndexNotFoundError message (#1741)

danielsotirhos:


danielsotirhos:
> What about something slightly more explicit, like, "Index `{missing_index}` was not found"

I originally went with a terse exception message to avoid repetition as it is an `IndexNotFoundError` however I like your suggestion, keeping the message brief while descriptive of the problem.

hannes-ucsc:
Skipping sandbox on this one.

Issue #1761, opened 2020-05-04
`load_app_module` breaks doctest addition


noah-aviel-dove:
When an app module is imported using `azul.modules.load_app_module`, `doctest.DocTestSuite` cannot find the doctests in the resulting module object. This means that our doctests in the app modules are currently not being executed by the `test_doctests` module.

When importing "normally" from project root:
```
>>> import doctest
>>> import lambdas.service.app as app
>>> doctest.DocTestSuite(app)
<doctest._DocTestSuite tests=[validate_facet (lambdas.service.app), validate_filters (lambdas.service.app), validate_params (lambdas.service.app), validate_size (lambdas.service.app)]>
```

When using `load_app_module`:
```
>>> import doctest
>>> from azul.modules import load_app_module
>>> app = load_app_module('service')
>>> doctest.DocTestSuite(app)
<doctest._DocTestSuite tests=[]>
```

noah-aviel-dove:
Debugging progress: in `doctests.DoctestFinder._from_module`: 

`return module is inspect.getmodule(object)` 

fails because the module attributes from `load_app_module` have their `__module__` attribute set to `'__main__'` instead of `'lambdas.service.app'`

PR #1774, opened 2020-05-05
Enable app module doctest discovery (#1761)

noah-aviel-dove:
See my comments on [the issue](https://github.com/DataBiosphere/azul/issues/1761) for an explanation of how this addresses the issue.

### DIFF
--- a/src/azul/modules.py
+++ b/src/azul/modules.py
@@ -39,4 +39,5 @@ def load_module(path: str,
 
 def load_app_module(lambda_name, **module_attributes):
     path = os.path.join(config.project_root, 'lambdas', lambda_name, 'app.py')
-    return load_module(path, '__main__', module_attributes)
+    # Changing the module name here will break doctest discoverability
+    return load_module(path, f'lambdas.{lambda_name}.app', module_attributes)
--- a/test/test_doctests.py
+++ b/test/test_doctests.py
@@ -4,6 +4,7 @@ import unittest
 import azul
 import azul.azulclient
 import azul.collections
+import azul.doctests
 import azul.dss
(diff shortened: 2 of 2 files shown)


noah-aviel-dove:
@hannes-ucsc's original instructions were to leave the module name unspecified to allow the default to persist, however `importlib.util.spec_from_file_location` requires an explicitly provided name for the module, so to keep things simple I'm just explicitly providing the "default" name

src/azul/modules.py
```diff
@@ -39,4 +39,5 @@ def load_module(path: str,
 
 def load_app_module(lambda_name, **module_attributes):
     path = os.path.join(config.project_root, 'lambdas', lambda_name, 'app.py')
-    return load_module(path, '__main__', module_attributes)
+    # Changing the module name here will break doctest disscoverability
```
hannes-ucsc, line 42:
```suggestion
    # Changing the module name here will break doctest discoverability
```

noah-aviel-dove:
Commit show out of order for some reason. The correct sequence is:
```
$ git log --oneline
0fd70216 (HEAD -> issues/noah-aviel-dove/1761-app-module-doctests, origin/issues/noah-aviel-dove/1761-app-module-doctests) drop! Build failure in service app doctests
baee4f08 Use correct __name__ when loading app module (#1761)
e3fca547 Remove modules with no doctests from test_doctests
9ad3038d Verify doctest discovery
```

jessebrennan:
> Commits display out of order for some reason
Github UI orders commits by author date instead of topologically for some reason. See: https://github.com/isaacs/github/issues/386

PR #1786, opened 2020-05-09
Represent manifest format as enum (#1723)

noah-aviel-dove:


src/azul/service/manifest_service.py
```diff
@@ -70,15 +74,30 @@
 logger = logging.getLogger(__name__)
 
 
+class ManifestFormat(Enum):
+    COMPACT = 'compact'
```
hannes-ucsc, line 78: The precedent in Azul is to use lower-case, even for enum members. There are no true constants in Python so I see no point in pretending.

src/azul/service/manifest_service.py
```diff
@@ -70,15 +74,30 @@
...
+    FULL = 'full'
+    TERRA_BDBAG = 'terra.bdbag'
+
+    @property
+    def generator_class(self) -> Type['ManifestGenerator']:
+        return {
+            'compact': CompactManifestGenerator,
```
hannes-ucsc, line 85: Duplicating the strings here is exactly what enums were designed to avoid. You could just use the actual enum values.

Besides, I also don't want this enum to be smart. The service should have the authority on which class to use for which enum value. That's a more powerful pattern than hardcoding the mapping here. It allows potential service subclasses to override the mapping, for example.

src/azul/service/manifest_service.py
```diff
@@ -366,14 +385,7 @@ def for_format(cls,
         :return: a ManifestGenerator instance. Note that the protocol used for
                  consuming the generator output is defined in subclasses.
         """
-        if format_ == 'compact':
```
hannes-ucsc, line 369: I'd be open to considering the dictionary lookup approach, if you can show me (via evidence, reason or an authoritative reference) that it isn't considerable slower. It does look more concise and yes, lookup is O(1), but my rule of thumb has always been that the constant factor dominates until n=10. 1 * 100 is still more than 9 * 10. For example, is there any evidence that a `dict` literal in a function body doesn't create a new `dict` instance in every call? Is the bytecode compiler smart enough to realize that this `dict` is never modified and that it can be hoisted into a constant? We could of course extract the literal into an class attribute but then this approach loses some of its attraction.

src/azul/service/manifest_service.py
```diff
@@ -350,13 +369,13 @@ def source_filter(self) -> SourceFilters:
...
     def for_format(cls,
-                   format_: str,
+                   format_: ManifestFormat,
                    service: ManifestService,
                    filters: Filters) -> 'ManifestGenerator':
         """
         Return a generator instance for the given format and filters.
 
-        :param format_: a string describing the format
+        :param format_: a member of ManifestFormat
```
hannes-ucsc, line 378:
```suggestion
        :param format_: a member of ManifestFormat
```

Both fragments, the original and the changed version don't tell me anything I don't already know from looking at the signature.

test/service/test_manifest.py
```diff
@@ -73,7 +74,7 @@ def tearDown(self):
 
     def _setup_git_commit(self):
         """
-        Set git variables required to generate the object key of format='full' type manifests
+        Set git variables required to generate the object key of format=ManifestFormat.FULL type manifests
```
hannes-ucsc, line 77: The existing text is awkward. This makes it worse.

src/azul/service/manifest_service.py
noah-aviel-dove: The bytecode compiler is not that smart and does indeed build the dictionary anew on each function invocation. This is about half as fast as if/elif/else. However if the dictionary is built once ahead of time and the function only needs to perform the lookup, it is actually about 18% faster than if/elif/else.

Source: `timeit.timeit` and `dis.dis`.

noah-aviel-dove: Extraction into a class attribute is complicated because the subclasses haven't been defined defined when the class body of `ManifestService` is being processed. I'll sick with if/elif/else for now.

test/service/test_manifest.py
```diff
@@ -73,7 +74,7 @@ def tearDown(self):
 
     def _setup_git_commit(self):
         """
-        Set git variables required to generate the object key of format='full' type manifests
+        Set git variables required to generate the object key of full-type manifests
```
danielsotirhos, line 77:
```suggestion
        Set git variables required to derive the manifest object key
```
Not your fault, but when originally added this was only needed for the test of full format manifests. Now things have changed a bit and the object key is used in the generation of all manifests.

noah-aviel-dove:
I spotted a few mistakes on my own while looking over the code again.

Issue #1753, opened 2020-05-04
Create DSS adapter script 


hannes-ucsc:
https://docs.google.com/document/d/1NsibP8g-NeLnksxlcBWQsSj5Zg_uimCDaAF1qB_qkjg/edit?usp=sharing

hannes-ucsc:
scripts/dss_v2_adapter.py

Command line arguments should be the URL to the source DSS and the name of the v2 staging bucket as the destination.

hannes-ucsc:
For now we can use the `travis-test` project for Google but as soon as #1766 is solved, we can switch to the v2-specific Google project.

hannes-ucsc:
As an extension we should make the target (2nd) script argument a full `gs://bucket/prefix` URL so multiple adapter invocation can share a bucket in isolated prefixes. The standard invocation would be `gs://bucket/` (empty prefix).

danielsotirhos:
Making this ticket a freebie of #1834

Issue #1792, opened 2020-05-12
Add multi-threading to DSS adapter


theathorn:

danielsotirhos:
Making this ticket a freebie of #1834

Issue #1791, opened 2020-05-12
Add idempotency to DSS adapter


theathorn:

danielsotirhos:
Making this ticket a freebie of #1834

PR #1794, opened 2020-05-12
Create DSS adapter script for v2 (#1753)

danielsotirhos:


scripts/dss_v2_apdapter.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        Generate ascending hex prefixes starting with given value.
+
+        >>> list(cls.ascending_prefixes('7'))
+        ['7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
+
+        >>> list(cls.ascending_prefixes('aaa'))
+        ['aaa', 'aab', 'aac', 'aad', 'aae', 'aaf', 'ab', 'ac', 'ad', 'ae', 'af', 'b', 'c', 'd', 'e', 'f']
+        """
+        chars = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
```
hannes-ucsc, line 298: Note that this is essentially counting up in base 16. Given that your implementation struck me as a little complicated:

image

hannes-ucsc: But it doesn't have the `aaf` to `ab` shortening. How do you know you don't shorten beyond the original prefix length?

danielsotirhos: > How do you know you don't shorten beyond the original prefix length?

I'm not sure I understand this question. The original intent of the "start prefix" argument was to allow the user to resume processing bundles from where they last left off after a crash or ctrl-c. While `aaf` to `ab` shortening isn't necessary for this occur, it does allow for a very slight performance increase by reducing the number of calls to get a list of bundles from the DSS `/bundles/all` endpoint.

scripts/dss_v2_apdapter.py
```diff
@@ -0,0 +1,316 @@
...
+        return isinstance(dst_blob, Blob)
+
+    def log_error(self, bundle_uuid: str, file_name: str, msg: str):
+        """
+        Log an error message and save a copy for re-reporting at end of script
+        """
+        self.errors.append((bundle_uuid, file_name, msg))
+        log.error(msg)
+
+    @classmethod
```
hannes-ucsc, line 259: I believe we've had this discussion many times before. There is NO point in declaring every method that doesn't use `self` as classmethod. Then your just typing extra text `@classmethod` and replacing an unused `self` argument with an unused `cls` argument. A method should be declared `@classmethod` only if the call site doesn't actually have an instance available. This is typically only true for methods that CREATE an instance, so called factory methods. Every other case of classmethod is questionable and deserves extra scrutiny.

This is especially annoying here because this class is a singleton so there isn't ANY substantial difference between a class/static method and an instance method.

scripts/dss_v2_apdapter.py
danielsotirhos: I see. The two bundle prefix arguments were intended to be mutually exclusive. My mistake was in not specifying this with argparse.

danielsotirhos:
Applied 1 suggested change (move the google-cloud-storage requirement to requirements.dev.txt), and will now stop development on this branch which is the head of a chain of PRs, and instead commit all new changes to the current tail of the PR chain (#1814)

For reference the current chain of tickets is:
#1753 (PR #1794) Add DSS adapter script
#1791 (PR #1803) Add idempotency
#1792 (PR #1804) Add multi-threading
#1809 (PR #1816) Migrate manifest properties (md5)
#1814 (PR #1826) Rewrite links.json

hannes-ucsc:
@danielsotirhos, yes let's use one branch and PR for all issues. This implies that the other issues are converted to freebies.

Issue #1782, opened 2020-05-08
Determine/enforce default sorting behavior for repository searches


noah-aviel-dove:
It is unclear what, if any sorting behavior of the hits occurs on the repository endpoints when the `sort` parameter is not provided.

Hannes: "If there is a default, we should document it. If there isn't, we should enforce one."

noah-aviel-dove:
Now that I know what the solution will entail I'm instead going to resolve this as a commit in #1405 .

PR #1798, opened 2020-05-13
Document default sorting behavior (#1782)

noah-aviel-dove:


lambdas/service/app.py
```diff
@@ -838,11 +839,11 @@ def repository_search_spec(entity_type):
                 description='The number of hits included per page.'),
             params.query(
                 'sort',
-                schema.optional(schema.enum(*facets)),
+                schema.optional(schema.with_default(type_=schema.enum(*facets), default=sort_default)),
```
hannes-ucsc, line 842: `with_foo(foo=)` is redundant.

noah-aviel-dove:
It worked in apidev but failed IT due to flake8 insisting on space around the `/` "operator." Adding the space however broke another rule (space before comma) so I had to just suppress the error.

Issue #1793, opened 2020-05-12
Rearrange tags in Swagger UI


hannes-ucsc:
Repository
Manifest
Auxiliary

noah-aviel-dove:
@hannes-ucsc What about DSS?

PR #1799, opened 2020-05-13
Rearrange tags in Swagger UI (#1793)

noah-aviel-dove:
Also add previously missing tag definition for DSS


hannes-ucsc:
Skipping sandbox.

PR #1811, opened 2020-05-18
Move `json-object` wheel from lambda packages to layer package (#1770)

jessebrennan:


jessebrennan:
I couldn't get `test_concurrent_portal_db_crud()` to pass, but all of the other integration tests passed. I think the failure is unrelated because I couldn't get that test to pass on my personal deployment with develop either. I'm debugging why that is, but I think this PR is safe to review.

Issue #1777, opened 2020-05-05
Links in general API documentation open in new window


jessebrennan:
The links to the Manifest and Repository sections open in a new tab, but this can be fixed by using the anchor from the HTML id tag.

hannes-ucsc:
"Index" link still doesn't work. Moving back to sprint backlog.

Issue #1724, opened 2020-04-22
Rename `repository` in service routes


jessebrennan:
Potentially to `index`, but open to something better.

PR #1813, opened 2020-05-19
Fix links in OpenAPI description (#1777)

jessebrennan:


lambdas/service/app.py
```diff
@@ -118,12 +118,13 @@
...
-            [DCP CLI](https://github.com/HumanCellAtlas/dcp-cli) to download the
-            listed files. Manifests can be generated for a selection of files
-            using filters. These filters are interchangeable with the filters
-            used by the [Repository](#/Repository) endpoints.
+            the [Manifests](#operations-tag-Manifests) endpoints. The resulting
+            manifests include links to associated data files and can be used by
+            the [DCP CLI](https://github.com/HumanCellAtlas/dcp-cli) to download
+            the listed files. Manifests can be generated for a selection of
+            files using filters. These filters are interchangeable with the
+            filters used by the [Repository](#operations-tag-Repository)
```
danielsotirhos, line 126:
```suggestion
            filters used by the [Index](#operations-tag-Index)
```
Based on in progress ticket https://github.com/DataBiosphere/azul/issues/1724, though I could see leaving it `Repository` for now and updating once that lands.

jessebrennan: Since #1647 will likely land before #1724, I think it makes sense to leave it at Repository.

hannes-ucsc: Yes, a PR should be internally consistent and not anticipate changes in other PRs. We use chaining for that.

Issue #1673, opened 2020-04-06
`make terraform` is not stable on Gitlab


hannes-ucsc:
I'm using the term `stable` in the way it is used to describe stable sorts. Don't know if there is better term.

Two consecutive retries of the `deploy` job on  https://github.com/DataBiosphere/azul/pull/1635:

https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1546
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/1547

Even though there were no changes to code or infrastructure between these two jobs, the second job redeployed the lambda and the gateway.

This is currently only reproducible on that PR branch but I am leaning towards merging that PR despite this issue so it will shortly be reproducible on the `develop` branch as well.

jessebrennan:
I don't know how I overlooked this before, but Chalice now ensures that the ZIP that is generated is deterministc. https://github.com/aws/chalice/commit/635de7f2561357bbe0539b7da59173ef5c876962 

This appears in 13.1 at the earliest. I think the most sensible approach is just to update the version of Chalice that we use.

hannes-ucsc:
#1725 was merged but `make terraform` still appears to be unstable:

https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/2911

PR #1839, opened 2020-05-28
Force compiling of source files during deployment (#1673)

jessebrennan:
We have to do "force", because if a `pyc` file is generated before the tests run, it will be generated with a timestamp for validation instead of a hash. This likely is happening when the script to generate the changelog is run. Running compileall won't override this file. The timestamp validation will change between deployments and therefore the package hash won't match between deployments.

### DIFF
--- a/lambdas/indexer/Makefile
+++ b/lambdas/indexer/Makefile
@@ -5,11 +5,7 @@ include ../../common.mk
 include ../lambdas.mk
 
 .PHONY: package
-package: check_branch check_python check_aws config
+package: check_branch check_python check_aws config compile
 	python -m azul.changelog vendor
-	# Ensure a .pyc file is present for every .py file. Set literals will
-	# compile in a non-deterministic order unless PYTHONHASHSEED is set. For a
-	# full explanation see http://benno.id.au/blog/2013/01/15/python-determinism
-	PYTHONHASHSEED=0 python -m compileall -q --invalidation-mode checked-hash vendor vendor/azul
 	chalice package --stage $(AZUL_DEPLOYMENT_STAGE) --pkg-format terraform .chalice/terraform
 	python $(project_root)/scripts/prepare_lambda_deployment.py indexer
--- a/lambdas/indexer/Makefile
+++ b/lambdas/indexer/Makefile
@@ -2,9 +2,7 @@
 all: package
 
 include ../../common.mk
-
-.PHONY: config
-config: .chalice/config.json
+include ../lambdas.mk
 
 .PHONY: package
 package: check_branch check_python check_aws config
@@ -15,11 +13,3 @@ package: check_branch check_python check_aws config
 	PYTHONHASHSEED=0 python -m compileall -q --invalidation-mode checked-hash vendor vendor/azul
 	chalice package --stage $(AZUL_DEPLOYMENT_STAGE) --pkg-format terraform .chalice/terraform
 	python $(project_root)/scripts/prepare_lambda_deployment.py indexer
-
-.PHONY: local
-local: check_python config
-	chalice local
-
-.PHONY: clean
-clean: check_env
-	git clean -Xdf
--- a/lambdas/lambdas.mk
+++ b/lambdas/lambdas.mk
@@ -1,4 +1,33 @@
 
+# The compile target is used during packaging of lambdas. The target ensures
+# that a .pyc file is present for every .py file in the package.
+#
+# One reason to compile before deploying is to reduce lambda start-up time. But
+# more importantly, it ensures that the same files are always included in the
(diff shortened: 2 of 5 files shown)


jessebrennan:
I'm skipping peer review because this is a small change and the only way to validate it is on Gitlab.

lambdas/indexer/Makefile
```diff
@@ -12,7 +12,7 @@ package: check_branch check_python check_aws config
 	# Ensure a .pyc file is present for every .py file. Set literals will
 	# compile in a non-deterministic order unless PYTHONHASHSEED is set. For a
 	# full explanation see http://benno.id.au/blog/2013/01/15/python-determinism
-	PYTHONHASHSEED=0 python -m compileall -q --invalidation-mode checked-hash vendor vendor/azul
+	PYTHONHASHSEED=0 python -m compileall -f -q --invalidation-mode checked-hash vendor vendor/azul
```
hannes-ucsc, line 15: The use of `-f` should be motivated in a comment. The PR description would be a good start. It might make sense to factor this line out into a separate `compile` target so the explanatory comments don't overwhelm the rest of the recipe.

That said, I don't quite understand the current PR description wrt `timestamp` vs `hash` so even the description might need a clarification.

lambdas/service/Makefile
```diff
@@ -12,7 +12,7 @@ package: check_branch check_python check_aws config
 	# Ensure a .pyc file is present for every .py file. Set literals will
 	# compile in a non-deterministic order unless PYTHONHASHSEED is set. For a
 	# full explanation see http://benno.id.au/blog/2013/01/15/python-determinism
-	PYTHONHASHSEED=0 python -m compileall -q --invalidation-mode checked-hash vendor vendor/azul
+	PYTHONHASHSEED=0 python -m compileall -f -q --invalidation-mode checked-hash vendor vendor/azul
```
hannes-ucsc, line 15: Same here.

lambdas/indexer/Makefile
jessebrennan: I added a more thorough description.

jessebrennan:
@hannes-ucsc I've made two commits here, one fixup and another refactor. They could be squashed together, but I wanted to keep them separate in case you weren't a fan of the refactor. Then it would be easy to reverse. I'm sure you'll have your own ideas as well for how to consolidate the duplication in the Makefiles. Let me know what you think.

lambdas/lambdas.mk
```diff
@@ -1,4 +1,33 @@
 
+.PHONY: compile
+compile: check_env
+	PYTHONHASHSEED=0 python -m compileall -f -q --invalidation-mode checked-hash vendor vendor/azul
+
+# The compile target is used during packaging of lambdas. The target ensures
+# that a .pyc file is present for every .py file in the package.
+#
+# One reason to compile before deploying is to ensure that the same files are
```
hannes-ucsc, line 9:
```suggestion
# One reason to compile before deploying is to reduce Lambda start-up time. But more importantly, it ensures that the same files are
```

lambdas/lambdas.mk
```diff
@@ -1,4 +1,33 @@
 
+.PHONY: compile
+compile: check_env
+	PYTHONHASHSEED=0 python -m compileall -f -q --invalidation-mode checked-hash vendor vendor/azul
+
+# The compile target is used during packaging of lambdas. The target ensures
```
hannes-ucsc, line 6: I think the comment should precede the target. That's what we do elsewhere. The length of the comment makes that a bit awkward but I prefer the consistency.

jessebrennan: Sure. I was following the pattern here https://github.com/DataBiosphere/azul/blob/develop/common.mk#L89 (since I had originally put this target in common.mk). Should I move that comment or leave it as it is?

hannes-ucsc: That one comment is an exception because of the fact that .FORCE has to succeed the recipe. So I though it better to have the comment succeed it.

test/integration_test.py
```diff
@@ -145,7 +145,7 @@ def tearDown(self):
         super().tearDown()
 
     def test(self):
-        if config.deployment_stage != 'prod':
+        if config.deployment_stage != 'prod' and False:
```
hannes-ucsc, line 148: ?

jessebrennan: I added this while debugging... I definitely didn't intend to commit this.

hannes-ucsc:
This appears to work as expected:

https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/3509

image

https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/3516

image

Issue #1796, opened 2020-05-13
Switch to functools.cached_properties


hannes-ucsc:
Change every use of `boltons.cacheutils.cachedproperty` to `functools.cached_property` that's included Python 3.8 and remove the dependency on `boltons`.

PyCharm has trouble inferring the return value of a method decorated with `boltons.cacheutils.cachedproperty` which leads to misses during refactoring.

hannes-ucsc:
Blocked by #1673 because the resolution for that includes an upgrade to Python 3.8.

PR #1843, opened 2020-05-28
Switch to functools.cached_properties (#1796)

jessebrennan:


src/azul/objects.py
```diff
@@ -64,3 +64,17 @@ def __new__(_cls, *args, **kwargs):
             return _self
 
         cls.__new__ = __new__
+
+
+# Vendored from boltons==19.3.0
```
hannes-ucsc, line 69: I would call this *theft*, not vendoring.

```suggestion
# Stolen from boltons==19.3.0
```

src/azul/objects.py
```diff
@@ -64,3 +64,17 @@ def __new__(_cls, *args, **kwargs):
...
+class classproperty(object):
```
hannes-ucsc, line 70: Does PyCharm infer the return type correctly with this?

Does [this](https://github.com/python/mypy/issues/2563#issuecomment-523837469) technique have any advantages?

jessebrennan: > Does PyCharm infer the return type correctly with this?

No, not for my test:

```
class C:
    @classproperty
    def foo(cls) -> int:
        return 42
    

c = "a" + C.foo
d = "a" + 42
```

> Does this technique have any advantages?

PyCharm does not correctly infer types with this form either. Since this form requires two decorators, I think it is inferior. Also the two decorators raise a different, "incompatible decorator" warning in pycharm.

[This implementation](https://github.com/python/mypy/issues/2563#issue-195305499) is functionally the same as the boltons one, but with better type annotations. Even though the type annotations don't allow inference with PyCharm, I'll _steal_ this one instead.

hannes-ucsc: The main reason we switch to functools.cached_property is to enable PyCharm type inference.

I think we should not encourage the use of `@classproperty` since there is no implementation that we know of that allows PyCharm type inference of the return type. For that reason, I think we should completely remove the decorator and eliminate the one use in `health.py`.

hannes-ucsc:
> History was borked again.

Bumping review count for that reason.

Issue #1704, opened 2020-04-14
In new main deployments, lambdas need to dynamically look up ES endpoint


achave11:
It is possible that Terraform might allow for the provisioning of the Chalice-Terraform lambdas with references to resources that have not been created using the Terraform syntax.
For example:
```
${aws_elasticsearch_domain.elasticsearch.endpoint}
```
Can be used instead to replace `AZUL_ES_ENDPOINT` when provisioning the lambdas.

This would allow for Terraform to replace references with valid values at creation time. Allowing for `make deploy` (command name might be replaced soon) and `make terraform` to be ran only once. 

A forceable problem is that deployments sharing an Elasticsearch domain would have difficulties provisioning the resources with the appropriate references. Whoever, this could be solved by using Terraform Data Sources. Allowing for multiple deployments to fetch/provision references to the shared resources.

jessebrennan:
When provisioning a new deployment that will not share its ES instance, lambdas need to infer the ES endpoint dynamically. We can work around this inefficiency by deploying the lambdas again (i.e. running `make deploy` and `make terraform` a second time).

Now that the lambdas are provisioned via Terraform we can reference the [ES endpoint](https://www.terraform.io/docs/providers/aws/r/elasticsearch_domain.html#endpoint) property within the lambdas' Terraform configuration. This way, the lambdas won't need to do a dynamic look-up when provisioned the first time.

Deployments that use the already provisioned ES instance of another deployment won't be able to reference the necessary Terraform property. Terraform Data Sources could be used in those cases since they give you access to the Terraform properties of cloud resource that is provisioned elsewhere.

hannes-ucsc:
> This way, the lambdas won't need to do a dynamic look-up when provisioned the first time.

To be precise, the dynamic lookup of the ES endpoint happens *every* time the lambda needs to connect to ES, and will have to be done until the lambda is deployed (by running `make deploy && make terraform`) a *second* time.

We should also mention that this behaviour is a quirk of the fix for #1688.

PR #1856, opened 2020-06-01
Use Terraform to determine ES env variables (#1704)

achave11:


terraform/modules.tf.json.template.py
```diff
@@ -1,6 +1,14 @@
 from azul import config
 from azul.deployment import emit_tf
 
+
+def get_es_domain():
+    if not config.share_es_domain:
```
jessebrennan, line 6: Kind of a nit, but could you remove the `not` and switch the order?

jessebrennan: Also I think Hannes prefers methods that don't start with `get`.

Finally, consider (very optional) inlining this as an expression.

src/azul/deployment.py
```diff
@@ -91,18 +91,15 @@ def es_endpoint(self) -> Optional[Netloc]:
...
-        if config.es_endpoint:
-            return config.es_instance_count
+        return config.es_instance_count
 
     @property
     def lambda_env(self) -> Mapping[str, str]:
         if config.share_es_domain:
-            endpoint = "${data.aws_elasticsearch_domain.azul_es_domain.endpoint}"
-            instance_count = "${data.aws_elasticsearch_domain.azul_es_domain.cluster_config[0].instance_count}"
+            instance_count = "${var.azul_es_domain.cluster_config[0].instance_count}"
```
jessebrennan, line 99: Since the variable is now always set, I suspect that `instance_count` can just _always_ be set to this value.

src/azul/deployment.py
```diff
@@ -91,18 +91,15 @@ def es_endpoint(self) -> Optional[Netloc]:
...
         else:
-            endpoint = "${var.azul_es_domain.endpoint}"
             instance_count = config.es_instance_count
```
jessebrennan, line 101: If the comment above is true then this can just be removed. 

Otherwise... I'm wondering why this is using `config` directly instead of using the property here: https://github.com/DataBiosphere/azul/pull/1856/commits/edf1c22e8f1f10841d4003281704c6f0ea6c4022#diff-747ee2f562a8dc659d6f60c2b431abbeR93. Do you know why the property above exists / can it be removed?

src/azul/deployment.py
```diff
@@ -91,18 +91,15 @@ def es_endpoint(self) -> Optional[Netloc]:
 
     @property
     def es_instance_count(self) -> Optional[int]:
```
jessebrennan, line 93: Now this seems to be identical to the `config` property with the same name. Does it make sense to remove this proxy in favour of the `config` attribute?

src/azul/deployment.py
```diff
@@ -91,18 +91,15 @@ def es_endpoint(self) -> Optional[Netloc]:
...
     def lambda_env(self) -> Mapping[str, str]:
         if config.share_es_domain:
-            endpoint = "${data.aws_elasticsearch_domain.azul_es_domain.endpoint}"
-            instance_count = "${data.aws_elasticsearch_domain.azul_es_domain.cluster_config[0].instance_count}"
+            instance_count = "${var.azul_es_domain.cluster_config[0].instance_count}"
         else:
-            endpoint = "${var.azul_es_domain.endpoint}"
             instance_count = config.es_instance_count
-        return config.lambda_env((endpoint, 443), instance_count)
+        return config.lambda_env(("${var.azul_es_domain.endpoint}", 443), instance_count)
```
jessebrennan, line 102: Finally, if all of the parameters to `config.lambda_env` turn out to be constants, then I think the constants should be factored down into that method. Then _this_ `lambda_env` property can be removed and the one from config can always just be used instead.

src/azul/deployment.py
```diff
@@ -91,18 +91,15 @@ def es_endpoint(self) -> Optional[Netloc]:
```
jessebrennan, line 91: Since the property `es_endpoint` is now just a proxy for `config.es_endpoint` shouldn't it just be removed?

src/azul/deployment.py
achave11: It makes sense since ES instance count gets called directly in the `ES.template` so it would be provisioned by terraform either way. Good catch.

achave11:
I did not address the rest of your suggestions since I have a hunch regarding the utility of the AWS class. Deferring to @hannes-ucsc on those.

achave11:
> Looks great! Nice work. I'm surprised and impressed that this PR is actually a reduction in code.

@jessebrennan Well you should also take credit for that since we both contributed! 😃

src/azul/__init__.py
```diff
@@ -57,15 +57,19 @@ def es_endpoint(self) -> Optional[Netloc]:
...
 
-    def es_endpoint_env(self, es_endpoint: Optional[Netloc], es_instance_count: Optional[int]) -> Mapping[str, str]:
+    def es_endpoint_env(self,
+                        es_endpoint: Optional[Netloc] = None,
+                        es_instance_count: Optional[int] = None
+                        ) -> Mapping[str, str]:
         if es_endpoint is None:
-            return {}
+            host, port = "${var.azul_es_domain.endpoint}", 443
+            es_instance_count = "${var.azul_es_domain.cluster_config[0].instance_count}"
```
hannes-ucsc, line 66: Brittle and hacky. An acceptable solution to this ticket should not embed TF syntax in `azul.config`, especially not TF syntax that hard-codes TF variable names. TF config can depend on `azul.config` but not the other way around.

lambdas/indexer/.chalice/config.json.template.py
```diff
@@ -11,7 +10,8 @@
     "api_gateway_stage": config.deployment_stage,
     "manage_iam_role": False,
     "iam_role_arn": "${var.role_arn}",
-    "environment_variables": aws.lambda_env,
+    "environment_variables": config.lambda_env(("${var.azul_es_domain.endpoint}", 443),
```
hannes-ucsc, line 13: This would be more readable if the args were passed as keywords with hanging indent.

achave11:
Summary of functional changes:
         
- Included an integration test `_test_check_all_indices` which uses the ESClientFactory to very all indices for available entity types and their aggregates are available via `get()` which failed initially before any changes were made.
- Solved the problem observed [here](https://github.com/DataBiosphere/azul/pull/1856#pullrequestreview-424893916) and ran `make integration_test` with a successful outcome.
- Ran `make package && make deploy` and confirmed targeted env values are present in update plan for each lambda.
- Ran `make reindex` on personal deployment, with a successful reindex.
- Created a new deployment, not sharing ES-domain successfully.

test/integration_test.py
```diff
@@ -484,6 +487,12 @@ def run(thread_count):
         # Reset to pre-test state.
         portal_service.overwrite(old_db)
 
+    def _test_check_all_indices(self):
```
hannes-ucsc, line 490: This looks confusing. _test_indexing wouldn't pass if indices were missing. The reason I asked for this test was not so much to assert that all indices exist (even though that is useful) but to assert that ESClientFactory.get() works. I am ok keeping this test and even the name under two conditions: it is invoked before _test_indexing and a comment is added to emphasize the dual purpose of this test. Before asking me for another review, have someone help you with the comment, by asking for a peer review.

test/integration_test.py
```diff
@@ -489,9 +489,9 @@ def run(thread_count):
 
     def _test_check_all_indices(self):
         """
-        Confirms valid configuration of ESClientFactory client, obtaining
-        all Azul indices in Elasticsearch for the deployment and validating them
-        confirms a successful connection and configuration of ESClientFactory.
+        This test serves two purposes. First, by calling ESClientFActory.get()
```
jessebrennan, line 492:
```suggestion
        This test serves two purposes. First, by calling ESClientFActory.get(),
```

test/integration_test.py
```diff
@@ -167,6 +170,7 @@ def test(self):
                 raise e
             else:
                 self._test_manifest()
+                self._test_check_all_indices()
```
hannes-ucsc, line 173: Why is this called twice?

test/integration_test.py
```diff
@@ -487,6 +491,19 @@ def run(thread_count):
...
         portal_service.overwrite(old_db)
 
+    def _test_check_all_indices(self):
+        """
+        This test serves two purposes. First, by calling ESClientFActory.get(),
+        it validates ES client configuration. Second, it tests the existence of
+        all of the expected indices.
+        """
+
+        log.info("Confirming valid configuration of ESClientFactory ...")
```
hannes-ucsc, line 501: Remove. There is no point in logging this. It doesn't take long and if it fails it is clear what's going on.

test/integration_test.py
```diff
@@ -487,6 +491,19 @@ def run(thread_count):
         # Reset to pre-test state.
         portal_service.overwrite(old_db)
 
+    def _test_check_all_indices(self):
+        """
+        This test serves two purposes. First, by calling ESClientFActory.get(),
+        it validates ES client configuration. Second, it tests the existence of
+        all of the expected indices.
```
hannes-ucsc, line 498:
```suggestion
        Aside from checking that all indices exist this method also asserts that we can instantiate a local ES client pointing at a real, remote ES domain.  
```

test/integration_test.py
achave11: Bad conflict resolution.

terraform/elasticsearch.tf.json.template.py
```diff
@@ -11,7 +11,15 @@
 
 domain = config.es_domain
 
-emit_tf(None if config.share_es_domain else {
+emit_tf({
+            "data": {
```
hannes-ucsc, line 15: Since you reintroduced the methods in `deployment.py`, we can use those instead of the data source, I think.

lambdas/indexer/.chalice/config.json.template.py
```diff
@@ -11,7 +10,9 @@
     "api_gateway_stage": config.deployment_stage,
     "manage_iam_role": False,
     "iam_role_arn": "${var.role_arn}",
-    "environment_variables": aws.lambda_env,
+    "environment_variables":
+        config.lambda_env(es_endpoint=("${var.azul_es_domain.endpoint}", 443),
```
hannes-ucsc, line 14: To see TF syntax in this file still irks me. I think we should inject those two variables in `prepare_lambda_deployment` and eliminate the arguments to lambda_env

terraform/modules.tf.json.template.py
```diff
@@ -8,12 +9,19 @@
...
+
         },
         "chalice_service": {
             "source": "./service",
             "role_arn": "${aws_iam_role.service.arn}",
-            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}"
+            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}",
+            "azul_es_domain":
+                "${data.aws_elasticsearch_domain.azul_es_domain}" if config.share_es_domain else
+                "${aws_elasticsearch_domain.elasticsearch[0]}"
```
hannes-ucsc, line 24:
```suggestion
            "azul_es_domain": 
                "${data.aws_elasticsearch_domain.azul_es_domain}" 
                if config.share_es_domain else
                "${aws_elasticsearch_domain.elasticsearch[0]}"
```

terraform/elasticsearch.tf.json.template.py
achave11: True, we could go back to using boto for provisioning the ES-Endpoint for deployments sharing ES instance with `dev`. This would require adding conditional logic in `prepare_lambda_deployment` to determine if it should use Terraform syntax or the appropriate method in `deployment.py`. This approach will add additional web requests when packaging lambdas, when it could all be done when Terraform is updating/publishing resources. For those reasons I think we should delegate that task to Terraform instead, since it is the authoritative entity for managing the ES resource. However, if you don't think this reasons are convincing enough I can revert to using `deployment.py` instead, as you emphasize.

scripts/prepare_lambda_deployment.py
```diff
@@ -34,6 +35,12 @@ def transform_tf(input_json):
         assert 'layers' not in func
         func['layers'] = ["${var.layer_arn}"]
 
+        # Inject runtime ES environment variables provisioned by Terraform
+        es_endpoint = ('${var.azul_es_domain.endpoint}', 443)
+        es_instance_count = '${var.azul_es_domain.cluster_config[0].instance_count}'
```
hannes-ucsc, line 40: This `.cluster_config[0].instance_count` still irks me as it reveals an intimate dependency on TF resources or data sources and what properties they have. I wonder if it be better to pass the endpoint and instance count as separate variables to the module rather than the entire domain status object as one variable.

terraform/modules.tf.json.template.py
```diff
@@ -8,12 +9,21 @@
...
         },
         "chalice_service": {
             "source": "./service",
             "role_arn": "${aws_iam_role.service.arn}",
-            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}"
+            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}",
+            "azul_es_domain":
+                "${data.aws_elasticsearch_domain.azul_es_domain}"
+                if config.share_es_domain else
+                "${aws_elasticsearch_domain.elasticsearch[0]}"
```
hannes-ucsc, line 26: Since you have added code, the fact that these two dicts are more or less duplicates of each other becomes more evident. Can you try to eliminate the duplication either via a list comprehension over ['service', 'indexer'] or by extracting the common dict entries into a separate dict which is ** into each of the concrete dicts.

terraform/modules.tf.json.template.py
```diff
@@ -1,19 +1,27 @@
...
-        "chalice_service": {
-            "source": "./service",
-            "role_arn": "${aws_iam_role.service.arn}",
-            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}"
-        }
+        f"chalice_{lambda_name}": {
+            "source": f"./{lambda_name}",
+            "role_arn": "${aws_iam_role." + lambda_name + ".arn}",
+            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}",
+            "azul_es_domain_endpoint":
```
hannes-ucsc, line 16:
```suggestion
            "es_endpoint":
```

terraform/modules.tf.json.template.py
```diff
@@ -1,19 +1,27 @@
...
+                aws.es_endpoint[0]  # Returns host and port but only host is required
+                if config.share_es_domain else
+                "${aws_elasticsearch_domain.elasticsearch[0].endpoint}",
+            "azul_es_domain_instance_count":
```
hannes-ucsc, line 20:
```suggestion
            "es_instance_count":
```

scripts/prepare_lambda_deployment.py
```diff
@@ -34,6 +36,11 @@ def transform_tf(input_json):
         assert 'layers' not in func
         func['layers'] = ["${var.layer_arn}"]
 
+        # Inject runtime ES environment variables provisioned by Terraform
+        for var, val in config.es_endpoint_env(es_endpoint=('${var.azul_es_domain_endpoint}', 443),
```
hannes-ucsc, line 40: Don't hardcode 443 here

terraform/modules.tf.json.template.py
```diff
@@ -1,19 +1,27 @@
...
-            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}"
-        }
+        f"chalice_{lambda_name}": {
+            "source": f"./{lambda_name}",
+            "role_arn": "${aws_iam_role." + lambda_name + ".arn}",
+            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}",
+            "azul_es_domain_endpoint":
+                aws.es_endpoint[0]  # Returns host and port but only host is required
+                if config.share_es_domain else
+                "${aws_elasticsearch_domain.elasticsearch[0].endpoint}",
```
hannes-ucsc, line 19: hard-code 443 here

terraform/modules.tf.json.template.py
```diff
@@ -1,19 +1,27 @@
...
-            "source": "./service",
-            "role_arn": "${aws_iam_role.service.arn}",
-            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}"
-        }
+        f"chalice_{lambda_name}": {
+            "source": f"./{lambda_name}",
+            "role_arn": "${aws_iam_role." + lambda_name + ".arn}",
+            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}",
+            "azul_es_domain_endpoint":
+                aws.es_endpoint[0]  # Returns host and port but only host is required
```
hannes-ucsc, line 17: and eliminate the tuple extraction

terraform/modules.tf.json.template.py
```diff
@@ -1,19 +1,27 @@
...
+            "layer_arn": "${aws_lambda_layer_version.dependencies.arn}",
+            "es_endpoint":
+                aws.es_endpoint
+                if config.share_es_domain else
+                ("${aws_elasticsearch_domain.elasticsearch[0].endpoint}", 443),
+            "es_instance_count":
+                aws.es_instance_count
+                if config.share_es_domain else
+                "${aws_elasticsearch_domain.elasticsearch[0].cluster_config[0].instance_count}",
+
```
hannes-ucsc, line 24:
```suggestion
```

scripts/prepare_lambda_deployment.py
```diff
@@ -34,6 +36,11 @@ def transform_tf(input_json):
         assert 'layers' not in func
         func['layers'] = ["${var.layer_arn}"]
 
+        # Inject runtime ES environment variables provisioned by Terraform
```
hannes-ucsc, line 39:
```suggestion
        # Inject ES-specific environment from variables set by Terraform.
        # See  terraform/modules.tf.json.template.py for their definition.
```

scripts/prepare_lambda_deployment.py
```diff
@@ -34,6 +36,11 @@ def transform_tf(input_json):
...
+        for var, val in config.es_endpoint_env(es_endpoint=('${var.es_endpoint[0]}', '${var.es_endpoint[1]}'),
+                                               es_instance_count='${var.es_instance_count}').items():
```
hannes-ucsc, line 41:
```suggestion
        for var, val in config.es_endpoint_env(
            es_endpoint=('${var.es_endpoint[0]}', '${var.es_endpoint[1]}'),
            es_instance_count='${var.es_instance_count}'
        ).items():
```

scripts/prepare_lambda_deployment.py
```diff
@@ -34,6 +36,11 @@ def transform_tf(input_json):
...
+            func['environment']['variables'][var] = val
```
hannes-ucsc, line 42: Assert that the variable isn't already set before setting it.

achave11: Asserting that 'AZUL_ES_INSTANCE_COUNT' variable isn't present for deployments not sharing ES domain with `dev` might fail. That variable is required to be present in `newdeployment.local/environment.py` in order to create a ES instance for a new deployment. Should the assertion be selective on whether the deployment shares an ES domain with `dev` ?

hannes-ucsc: It is a bit awkward but I have no better idea. Scratch the assert.

Issue #1762, opened 2020-05-04
Enumerate links.json files in TDR snapshot


hannes-ucsc:

hannes-ucsc:
This was demoed as part of the DCP/2 demo.

PR #1857, opened 2020-06-01
Implement list_bundles for TDR plugin (#1762)

noah-aviel-dove:


noah-aviel-dove:
Requesting preliminary review.

src/azul/tdr.py
```diff
@@ -0,0 +1,72 @@
...
+log = logging.getLogger(__name__)
+
+
+class Snapshot(NamedTuple):
+    uuid: str
+    name: str
+
+
+class AzulTDRClient:
+    valid_uuid_re = re.compile(r'[-0-9a-z]*')
```
hannes-ucsc, line 22: Could be stricter. Then again, the best validation is to actually parse the UUID. Look for sites in our code that raise InvalidUUIDError and extract a utility method for validation of UUIDs.

Since many people erroneously compare UUIDs as strings, which is incorrect because of the case ambiguity, the validation method should also optionally reject UUIDs with upper-case characters in it. Maybe that should even be the default behaviour such that callers need to explicitly opt out to allow upper-case UUIDs.

src/azul/tdr.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @cachedproperty
+    def query_service(self) -> GoogleQueryAdapter:
```
hannes-ucsc, line 25: Scratch the previous comment. This is what you should patch for tests.

src/azul/tdr.py
```diff
@@ -0,0 +1,72 @@
...
+        return GoogleQueryAdapter.create_adaptee()
+
+    def __init__(self, terra_endoint: str):
+        self.endpoint = terra_endoint
+
+    def get_snapshot(self) -> Snapshot:
+        response = requests.get(self.endpoint +
```
hannes-ucsc, line 32: I want us to move away from requests and use urllib3 directly.

And see https://urllib3.readthedocs.io/en/latest/user-guide.html#query-parameters for a easier to parameterize way to pass query args.

src/azul/tdr.py
```diff
@@ -0,0 +1,72 @@
...
+import requests
+
+from azul.bigquery import GoogleQueryAdapter
+from azul.indexer import BundleFQID
+
+log = logging.getLogger(__name__)
+
+
+class Snapshot(NamedTuple):
+    uuid: str
```
hannes-ucsc, line 17: We should probably stick with their naming scheme and call this `id`

hannes-ucsc:
Prelim review. Not bumping review count.

noah-aviel-dove:
@achave11 Let me know if you have any questions about how bundles are represented in TDR. There's quite a bit of context needed for this one.

noah-aviel-dove:
@achave11 I need to move one change from the 2nd commit in this PR into the 1st. Shouldn't change the final diff. Just a head up that IT is going to be re-triggered.

achave11:
> @achave11 I need to move one change from the 2nd commit in this PR into the 1st. Shouldn't change the final diff. Just a head up that IT is going to be re-triggered.

No problem!

noah-aviel-dove:
> I did not see any ticket mentions/references in the commits of this PR. Was that intentional?

Nope, that was an oversight on my part. Thanks for catching it.

src/azul/bigquery.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @abstractmethod
+    def assert_table_exists(self, dataset_name: str, table_name: str) -> None:
+        """
+        Raise exception if the specified table does not exist.
+        """
+        raise NotImplementedError
+
+
+class BigQueryAdaptee(GoogleQueryAdapter):
```
hannes-ucsc, line 33:
```suggestion
class BigQueryAdapter(GoogleQueryAdapter):
```

test/service/test_tdr.py
```diff
@@ -0,0 +1,124 @@
...
+        kwargs.update(fields)
+        return [
+            dict(name=k,
+                 type=self._bq_types.get(v, v),
+                 mode='NULLABLE')
+            for k, v in kwargs.items()
+        ]
+
+
+class TinyQueryAdaptee(GoogleQueryAdapter):
```
hannes-ucsc, line 97:
```suggestion
class TinyBigQueryAdapter(GoogleQueryAdapter):
```

src/azul/uuids.py
```diff
@@ -0,0 +1,77 @@
+import uuid
```
hannes-ucsc, line 1: The extraction of this could be a separate commit.

src/azul/tdr.py
```diff
@@ -0,0 +1,45 @@
...
+    GoogleQueryAdapter,
+)
+from azul.indexer import BundleFQID
+from azul.uuids import validate_uuid_prefix
+
+
+class BigQueryDataset(NamedTuple):
+    id: str
+    name: str
+    is_snapshot: bool
```
hannes-ucsc, line 18: Looking at the usages of this attribute, there seems to be no enforcement of the uniqueness constraint in snapshots. Like any other table, the links table in a TDR dataset can contain multiple versions of the same subgraph. Should that not be asserted or dealt with somehow?

src/azul/uuids.py
hannes-ucsc: I mean the extraction of the UUID validation functionality into a separate azul module.

requirements.txt
```diff
@@ -26,4 +26,5 @@ rsa==3.4.2
 typing-extensions==3.7.4
 urllib3==1.24.3
 werkzeug==0.15.5
+google-cloud-bigquery==1.24.0
```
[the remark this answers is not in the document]
noah-aviel-dove: TinyQuery is causing problem with this. 
```
$ make requirements_update
...
  File "scripts/manage_requirements.py", line 222, in <module>
    main.run()
  File "scripts/manage_requirements.py", line 161, in run
    direct_build_reqs = self.get_direct_reqs(self.build)
  File "scripts/manage_requirements.py", line 204, in get_direct_reqs
    return self.parse_reqs(f)
  File "scripts/manage_requirements.py", line 185, in parse_reqs
    parsed_reqs = set(map(PinnedRequirement.create, parsed_reqs))
  File "scripts/manage_requirements.py", line 63, in create
    return cls(name=req.name.lower())
AttributeError: 'NoneType' object has no attribute 'lower'
Makefile:56: recipe for target 'requirements_update' failed
make: *** [requirements_update] Error 1
```

hannes-ucsc:
Also noticed that the PR was assigned to me before the test had passed. Can you please stick to the agreed upon protocol?

src/azul/tdr.py
```diff
@@ -0,0 +1,52 @@
...
+        iter_rows = self.big_query_adapter.run_sql(f'''
+            SELECT links_id, version
+            FROM {self.target.bigquery_dataset_name}.links
+            WHERE STARTS_WITH(links_id, "{prefix}")
+        ''')
+        bundles_fqids = [BundleFQID(uuid=row['links_id'],
+                                    version=row['version'].strftime(self.timestamp_format))
+                         for row in iter_rows]
+
+        if self.target.is_snapshot:
```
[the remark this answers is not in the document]
noah-aviel-dove: Looking at the `azulclient` module I now realize that only the most recent version should be included in the list of bundles when multiple versions are present. I'm refactoring this check to a separate method since it's the same pattern that's used extensively in the `fetch_bundle` implementation.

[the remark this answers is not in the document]
noah-aviel-dove: Actually, this is redundant since `AzulClient` still filters out non-current bundles, but my refactor has the added advantage of verifying bundle UUID uniqueness, which Terra only enforces during ingest due to lack of proper keys in BigQuery. Might need to revert though if there's a reason why `list_bundles` should include outdated versions.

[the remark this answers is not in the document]
noah-aviel-dove: Added a unit test to cover the separate snapshot and dataset cases

requirements.dev.txt
```diff
@@ -17,5 +17,6 @@ pyyaml==5.1
 git+git://github.com/hannes-ucsc/requirements-parser@v0.2.0+1#egg=requirements-parser
 responses==0.10.2
 watchdog==0.9.0
+git+git://github.com/DataBiosphere/azul-tinyquery@1.0#egg=azul-tinyquery
```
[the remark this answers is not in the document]
noah-aviel-dove: Fixed egg name and specified version as a tag.

environment.py
```diff
@@ -350,4 +350,7 @@ def env() -> Mapping[str, Optional[str]]:
 
         # This is set only during integration tests. Never override this to 1.
         'AZUL_TEST_MODE': '0',
+
+        # GCP project containing the TDR BigQuery datasets
```
hannes-ucsc, line 354:
```suggestion
        # Name of the Google Cloud project containing the TDR BigQuery datasets
```

requirements.dev.txt
```diff
@@ -17,5 +17,6 @@ pyyaml==5.1
 git+git://github.com/hannes-ucsc/requirements-parser@v0.2.0+1#egg=requirements-parser
 responses==0.10.2
 watchdog==0.9.0
+git+git://github.com/DataBiosphere/azul-tinyquery@tags/1.0#egg=Tinyquery
```
hannes-ucsc, line 20: Tag name is still inconsistent. See my previous comment on this line.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+log = logging.getLogger(__name__)
+
+
+class Plugin(RepositoryPlugin):
+
+    @classmethod
+    def get_dataset(cls) -> BigQueryDataset:
+        return BigQueryDataset('8c6633e3-ce9c-4aa4-b0ae-3fba122386a1', 'hca_mvp', False)
```
hannes-ucsc, line 32: If we make the project name configurable, we should also make these parameters configurable via the environment. To avoid proliferation of environment variables, we should compact multiple values into a single variable using an ARN-like syntax. Something like

`arn:google:bigquery:US:${project_id}:dataset/${dataset_name}`

The TDR endpoint should probably go into another variable but it looks like you don't need that here.

I'm confused why we need both name and ID. If we really do, we need to figure something out because the ARN syntax doesnt lend itself well to specifying conceptually redundant information.

src/azul/tdr.py
```diff
@@ -0,0 +1,68 @@
...
+from azul.indexer import BundleFQID
+from azul.types import (
+    JSON,
+    JSONs,
+)
+from azul.uuids import validate_uuid_prefix
+
+
+class BigQueryDataset(NamedTuple):
+    id: str
```
hannes-ucsc, line 27: Looks like this isn't being used anywhere. If so, remove.

environment.py
hannes-ucsc: But see other comment on ARNs

requirements.dev.txt
```diff
@@ -17,5 +17,6 @@ pyyaml==5.1
 git+git://github.com/hannes-ucsc/requirements-parser@v0.2.0+1#egg=requirements-parser
 responses==0.10.2
 watchdog==0.9.0
+git+git://github.com/DataBiosphere/azul-tinyquery@tags/1.2+1#egg=Tinyquery
```
hannes-ucsc, line 20:
```suggestion
git+git://github.com/DataBiosphere/azul-tinyquery@tags/1.2+1#egg=tinyquery
```

Any idea why you need `tags/` but the other "git+git:" dependencies don't?

deployments/dev/environment.py
```diff
@@ -33,6 +33,8 @@ def env() -> Mapping[str, Optional[str]]:
         'AZUL_DSS_DIRECT_ACCESS_ROLE': 'arn:aws:iam::109067257620:role/azul-sc',
         'AZUL_SUBSCRIBE_TO_DSS': '0',
 
+        'AZUL_TDR_BIGQUERY_DATASET': 'gbq:broad-jade-dev-data:dataset/datarepo_hca_mvp',
```
hannes-ucsc, line 36:
```suggestion
        'AZUL_TDR_DATASET': 'gbq:broad-jade-dev-data:dataset/datarepo_hca_mvp',
```

src/azul/__init__.py
```diff
@@ -140,6 +141,20 @@ def data_portal_name(self):
     def dss_endpoint(self) -> str:
         return os.environ['AZUL_DSS_ENDPOINT']
 
+    @property
+    def tdr_bigquery_dataset(self) -> BigQueryDataset:
```
hannes-ucsc, line 145: Add `@property def tdr_dataset(self) -> str:`

for just the env var access.

src/azul/__init__.py
```diff
@@ -140,6 +141,20 @@ def data_portal_name(self):
...
+        service, project, dataset = os.environ['AZUL_TDR_BIGQUERY_PROJECT'].split(':')
```
hannes-ucsc, line 146:
```suggestion
        service, project, dataset = os.environ['AZUL_TDR_BIGQUERY_DATASET'].split(':')
```

This worries me. Did you not tests this?

src/azul/uuids.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def validate_uuid(uuid_str: str, allow_uppercase: bool = False) -> uuid.UUID:
+    """
+    >>> validate_uuid('foo')
+    Traceback (most recent call last):
+    ...
+    azul.uuids.InvalidUUIDError: foo is not a valid uuid.
+
+    >>> validate_uuid('aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa')
```
hannes-ucsc, line 23: This is a version 10 UUID which hasn't been standardized. We should be stricter and only allow UUID versions that have been formally specified.

src/azul/uuids.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__(f'{entity_id} is not a valid uuid.')
+
+
+class InvalidUUIDPrefixError(Exception):
+
+    def __init__(self, prefix: str):
+        super().__init__(f'{prefix} is not a valid uuid prefix.')
+
+
+def validate_uuid(uuid_str: str, allow_uppercase: bool = False) -> uuid.UUID:
```
hannes-ucsc, line 16:
```suggestion
def validate_uuid(uuid_str: str) -> uuid.UUID:
```

Unless we actually need this.

test/service/test_tdr.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class TestTDRClient(AzulTestCase):
+    _gbq_adapter_mock = None
+
+    def setUp(self) -> None:
+        self.query_adapter = TinyBigQueryAdapter()
+        self._gbq_adapter_mock = mock.patch.object(AzulTDRClient,
+                                                   'big_query_adapter',
+                                                   new=property(lambda client: self.query_adapter))
```
hannes-ucsc, line 37:
```suggestion
                                                   new=self.query_adapter)
```

doesn't work?

image

test/service/test_tdr.py
```diff
@@ -0,0 +1,131 @@
...
+        columns = self.client.evaluate_query(query).columns
+        num_rows = one(set(map(lambda c: len(c.values), columns.values())))
+        for i in range(num_rows):
+            yield {k[1]: v.values[i] for k, v in columns.items()}
+
+    def assert_table_exists(self, dataset_name: str, table_name: str) -> None:
+        self.client.get_table(dataset_name, table_name)
+
+    def create_table(self, dataset_name: str, table_name: str, schema: JSONs, rows: JSONs) -> None:
+        # TinyQuery's errors are typically not helpful in catching these mismatches
```
hannes-ucsc, line 120: What mismatches? Comment is not clear to me.

requirements.dev.txt
noah-aviel-dove: It works without `tags/` now that the version is correct.

src/azul/__init__.py
noah-aviel-dove: Apparently not. Adding variable access to the unit tests.

test/service/test_tdr.py
noah-aviel-dove: Between row and table schemas. Updated.

environment.py
```diff
@@ -350,4 +350,13 @@ def env() -> Mapping[str, Optional[str]]:
 
         # This is set only during integration tests. Never override this to 1.
         'AZUL_TEST_MODE': '0',
+
+        # ARN-style configuration for The Google Cloud project and BigQuery
+        # dataset name for the dataset/snapshot to be indexed from the
+        # Terra Data Repository.
+        # Syntax is gbq:{project-name}:{"dataset"|"snapshot"}/{dataset-name}
```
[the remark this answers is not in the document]
noah-aviel-dove: [BigQuery does not allow colons or slashes in dataset names](https://cloud.google.com/bigquery/docs/datasets#dataset-naming).

deployments/dev/environment.py
noah-aviel-dove: I don't believe this is accurate given your previous comment. Sticking to TDR naming conventions, this variable could be used to specify a snapshot, not a dataset. A TDR snapshot is a BigQuery dataset, but it is not a TDR dataset.

environment.py
```diff
@@ -350,4 +350,13 @@ def env() -> Mapping[str, Optional[str]]:
...
+        # Note that if the target is a TDR "dataset" (as opposed to a snaphot),
+        # the name of the coresponding BigQuery dataset (used here) will not
+        # match the name used by Terra.
```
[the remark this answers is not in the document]
noah-aviel-dove: This will force us to rely on the undocumented convention for TDR to prepend `datarepo_` to the names of the BigQuery datasets corresponding to TDR datasets.

deployments/dev/environment.py
noah-aviel-dove: I will remove the reference to BigQuery, but I think "TARGET" is the most accurate one-word solution (unless you'd prefer "DATASET_OR_SNAPSHOT").

src/azul/__init__.py
noah-aviel-dove: Implementing as `@property def tdr_target(self) -> str:` 
See my comments above

src/azul/__init__.py
```diff
@@ -138,6 +139,23 @@ def data_portal_name(self):
     def dss_endpoint(self) -> str:
         return os.environ['AZUL_DSS_ENDPOINT']
 
+    @property
+    def tdr_target(self) -> str:
+        return os.environ['AZUL_TDR_TARGET']
```
hannes-ucsc, line 144: TARGET is good.

src/azul/uuids.py
```diff
@@ -0,0 +1,73 @@
...
+    ...
+    azul.uuids.InvalidUUIDError: foo is not a valid UUID.
+
+    >>> validate_uuid('8f53d355-b2fa-4bab-a2f2-6852d852d2ec')
+    UUID('8f53d355-b2fa-4bab-a2f2-6852d852d2ec')
+
+    >>> validate_uuid('aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa')
+    Traceback (most recent call last):
+    ...
+    azul.uuids.InvalidUUIDVersionError: aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa is a version 10 UUID (not standardized).
```
hannes-ucsc, line 35:
```suggestion
    azul.uuids.InvalidUUIDVersionError: aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa is not a valid RFC-4122 UUID (undefined version 10).
```

src/azul/uuids.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    try:
+        formatted_uuid = uuid.UUID(uuid_str)
+    except ValueError:
+        raise InvalidUUIDError(uuid_str)
+    else:
+        if str(formatted_uuid) != uuid_str:
+            raise InvalidUUIDError(uuid_str)
+        if formatted_uuid.version is None or not (1 <= formatted_uuid.version <= 4):
```
hannes-ucsc, line 44:
```suggestion
        if formatted_uuid.version not in (1, 3, 4, 5):
```

First paragraph in https://docs.python.org/3/library/uuid.html

environment.py
```diff
@@ -350,4 +350,12 @@ def env() -> Mapping[str, Optional[str]]:
 
         # This is set only during integration tests. Never override this to 1.
         'AZUL_TEST_MODE': '0',
+
+        # Identifies the Terra Data Repository dataset or snapshot to index.
+        # The syntax in EBNF is:
+        # 'tdr:', Google Cloud project name, ':', ( 'dataset' | 'snapshot' ), '/', 'TDR dataset or snaphot name'
+        # Examples:
+        # tdr:broad-jade-dev-data:dataset/hca_mvp
+        # tdr:broad-jade-dev-data:snapshot/hca_mvp
```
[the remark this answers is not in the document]
noah-aviel-dove: Unfortunately the names of the TDR datasets returned by the TDR REST API do not match the names of the corresponding BigQuery datasets. `datarepo_hca_mvp` is just listed as `hca_mvp`. I cannot find any way to recover the `datarepo_` prefix programatically. 

Here is the endpoint for listing datasets, filtered by the text "mvp":

Screenshot from 2020-06-30 15-41-43

And here is the endpoint for retrieving more detailed information on a dataset. It's cut off but the remainder of the response only describes the table schemas and provides the creation date. Note that we _could_ retrieve the Google Cloud project for a dataset/snapshot* from this endpoint instead of using the ARN syntax, if you'd prefer.

Screenshot from 2020-06-30 15-43-34

*separate endpoint for snapshots but the response is very similar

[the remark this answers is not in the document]
noah-aviel-dove: (this limitation was part of my motivation for initially using the BigQuery nomenclature instead of Terra, so that the `datarepo_` prefix would be part of the environment configuration rather than the resolution logic).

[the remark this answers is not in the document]
noah-aviel-dove: From slack:
Noah: "@dan Is the "convention" of using the datarepo_ prefix recoverable from any part of the TDR REST API, or perhaps another source? @Hannes has concerns about hardcoding this into this configuration logic."

Dan: "Not that I’m aware of. datrepo_ prefixes should only apply at the dataset level, so the most common flow of indexing snapshots shouldn’t need to worry about it"

Hannes: "I guess it's ok to hard-code because the worst that can happen is that we'll fail should the prefix ever change."

Issue #1738, opened 2020-04-28
Pin transitive dependencies


jessebrennan:
The dependencies lambda layer is content addressed by the hash of the `requirements.txt`.

Since transitive dependencies aren't pinned, it's possible that an update to a transitive dependency could cause a newly generated lambda layer to have different content event though it would be cached by the same `requirements.txt` hash.

Personally, I (Jesse) don't think this will be an issue. Consider the cases. Either an update in a transitive dependency can cause a problem, or cause no problems.
1. **An update to a transitive dependency causes a new problem.**  We will only notice the new problem if, locally, we remove, then rebuild our virtual environment. This kind of problem will only be visible in code running locally. This kind of problem won't show up in a deployment, unless we force the dependencies layer to be rebuilt. This kind of problem can be solved by pinning the version of the transitive dependency.
2. **An update to a transitive dependency causes no problems.** It's possible that the requirements installed in a fresh virtual environment will be different from those in a deployment, but in this case it isn't causing any issues.

If we ever need to force the dependencies lambda layer to be rebuild we can run `python scripts/stage_layer.py --force`.

hannes-ucsc:
We can't rely on the layer caching as a means to provide reproducibility. Being a cache, and not a primary build artefact, I should be able to delete it and still get a reproducible build. The important invariant we are pursuing (with some success) is: same source tree == same behaviour. Pinning all dependencies is a necessary prerequisite for that, independently of Lambda layer caching.

Imagine these two scenarios:

Scenario 1)

1) `prod` layer cache at hash 123, and hash of layer.zip is 234
1) `dev` layer cache at hash 123, and hash of layer.zip is 234
1) PR to `dev` is merged which updates `requirements.txt`
1) `dev` layer cache now 345, and hash of layer.zip is 456
1) a minor update to a transitive dependency is published
1) `dev` is promoted to `prod`, layer cache will be rebuilt due to change to requirements.txt
1) Update to transitive dependency is picked up 
1) `prod` layer cache now at 345, but hash of layer.zip is not 456 as in dev, but 567
1) `prod` breaks due to bug in transitive dependency

Scenario 2)

1) `prod` layer cache at hash 123, and hash of layer.zip is 234
1) `dev` layer cache at hash 123, and hash of layer.zip is 234
1) PR to `dev` is merged which does NOT update `requirements.txt`
1) `dev` layer cache still at 123, and hash of layer.zip is 234
1) a minor update to a transitive dependency is published
[...]

PR #1864, opened 2020-06-03
Pin transitive dependencies (#1738)

achave11:


requirements.txt
```diff
@@ -24,6 +24,40 @@ rsa==3.4.2
 urllib3==1.24.3
 more-itertools==4.3.0
 typing-extensions==3.7.4
-werkzeug==0.15.5
+Werkzeug==0.15.5
```
jessebrennan, line 27: Why the case change?

requirements.dev.txt
```diff
@@ -5,15 +5,57 @@ locustio==0.9.0
...
 pytest==3.2.3
 coverage==5.1
-gitpython==2.1.11
-gitdb2==3.0.1  # Not a direct dependency but https://github.com/gitpython-developers/GitPython/issues/983
+GitPython==2.1.11
 google-api-python-client==1.7.8
 google-auth==1.14.1
 moto==1.3.13  # 1.3.14 does not work, possibly related to https://github.com/spulec/moto/issues/2567
-pyyaml==5.1
+PyYAML==5.1
```
jessebrennan, line 12: Why the case change?

requirements.dev.txt
```diff
@@ -5,15 +5,57 @@ locustio==0.9.0
 gevent==20.5.0
 pytest==3.2.3
 coverage==5.1
-gitpython==2.1.11
-gitdb2==3.0.1  # Not a direct dependency but https://github.com/gitpython-developers/GitPython/issues/983
+GitPython==2.1.11
```
jessebrennan, line 8: Why the case change?

requirements.txt
achave11: Note the name on the url and the package name on PyPI
https://pypi.org/project/werkzeug/

requirements.dev.txt
achave11: https://pypi.org/project/pyyaml/

requirements.dev.txt
achave11: https://pypi.org/project/gitpython/

achave11:
> It would be nice if there was a convenient way to update dependencies... I suppose that should be a different issue though.
> 
> I want to be sure the change of the case is not a mistake before I approve.

Note that I introduced a new commit with the renaming. I believe they should be listed as they are displayed in the PyPI library.

hannes-ucsc:
Superseded by #1882

Issue #1739, opened 2020-04-28
Log when manifest content hash is computed


achave11:
This would indicate that the content-hash generated by the service is being recycled and used in the step-function. This is done in order to generate a manifest if one isn't already available from cache.

hannes-ucsc:
It's fine to log the fact that a provided object key is reused, but more importantly, the code should emit a log statement before any long-running operation, and this includes the ES request that produces the content hash. I've edited the title accordingly.

PR #1868, opened 2020-06-03
Log when manifest content hash is being computed (#1739)

achave11:


src/azul/service/manifest_service.py
```diff
@@ -135,8 +135,10 @@ def _compute_object_key(self,
                             generator: 'ManifestGenerator',
                             format_: ManifestFormat,
                             filters: Filters) -> str:
+        logger.info('Computing manifest object key ...')
```
danielsotirhos, line 138:
```suggestion
        logger.info('Computing object key for %s format manifest with filters %s', format_.value, filters)
```

achave11:
> This misses the point. The ticket refers to logging _content hash_ not _object key_. The point is to log when we are about to perform a computationally expensive operation. Since `generator.manifest_content_hash` is a cached property, we don't really know in _compute_object_key if it has already been computed or not.

I missed that.  😅

Issue #447, opened 2018-10-16
Disable hourly snapshots on the Elasticsearch domains


hannes-ucsc:
Except for prod, we don't need those backups because we can recreate the index by reindexing.  In prod that might take too long for a live site so we'll keep the backups there.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-297)
┆Project Name: azul
┆Issue Number: AZUL-297

hannes-ucsc:
Note that the commit references here are all wrong because I messed up the ticket reference in the commit message. I fixed that by force pushing but the bad refs here stuck.

hannes-ucsc:
I've opened a support case and AWS has responded:

> Dear Hannes,
>
> Thank you for your continued patience in this case while the internal team was working on disabling the hourly snapshots in this account.
>
> I got an update from the internal team. The internal team mentioned that the disabling of the automated snapshot is a per region per account process.
>
> Thus, as informed in the previous correspondence, since there was only 1 domain in the account which is "azul-index-dev" in the region "us-east-1", the internal team has for now disabled the hourly automated snapshots for any domain that is present and that would be created in the "us-east-1" region.
>
> However, if you would like to get this feature disabled for any other region, the internal team has requested you to kindly provide the region details so that the ElasticSearch internal team would go ahead and apply the same for the specified region.
>
> I hope the above information helps. However if you have further concerns/queries, please let me know and I'll be glad to assist your further.
>
> We value your feedback. Please share your experience by rating this correspondence using the AWS Support Center link at the end of this correspondence. Each correspondence can also be rated by selecting the stars in top right corner of each correspondence within the AWS Support Center.
>
Best regards,
REDACTED
Amazon Web Services

hannes-ucsc:
Demo by showing hourly snapshot on hca/prod instance and absence of any snapshots on the `dev` instance created today around 10am.

PR #1871, opened 2020-06-04
Create ES indices eagerly during `make deploy` (#447)

hannes-ucsc:


hannes-ucsc:
Note branch name is wrong.

Issue #1853, opened 2020-06-01
Add sequencing process ID to `terra.bdbag` manifests


hannes-ucsc:
Follow-up on #1805. As agreed in meeting with @kbergin, @ruchim and @jessicaway, the ID of the sequencing process is a better primary key for workflows than `bundle.sequencing_input[0].document_id` (metadata) or the equivalent `sample.document_id` (Azul index) because the latter would be duplicated if the same sample is sequenced twice.

During graph traversal in the indexer, record the `document_id` of every process associated with a `sequencing_protocol` and track that in the index as `content.sequencing_process.document_id`. Use a set accumulator with a size limit of 10 for that field. For now, don't expose that field in `/repository` (or `/index`). Only include the field in the `terra.bdbag` manifest as a new column called `sequencing_process.provenance.document_id`.

hannes-ucsc:
Implement first as a PR against `hca/develop` and once merged, forward-port to `develop`.

jessicaway:
wondering if there is a status update on this, would love to get this change merged! Thanks!

hannes-ucsc:
It's been merged and can be observed on https://dev.data.humancellatlas.org/explore/, @jessicaway, but we need to promote it to `prod`.

The final column name is `sample__provenance__document_id` (we have to use double underscore because period isn't allowed by Terrra).

jessicaway:
Thanks for the update and all the work on this!

melainalegaspi:
Request @hannes-ucsc to reindex

hannes-ucsc:
@jessicaway, @kbergin I made two edits, one to my [comment](https://github.com/DataBiosphere/azul/issues/1853#issuecomment-644922179) and one to the ticket description, fixing two mistakes I made.

PR #1872, opened 2020-06-04
Add sequencing process ID to terra.bdbag manifests (#1853)

danielsotirhos:
Note that this PR is for the `hca/develop` branch.


src/azul/project/hca/transformers.py
```diff
@@ -877,3 +907,9 @@ def _get_accumulator(self, field) -> Optional[Accumulator]:
             return FrequencySetAccumulator(max_size=100)
         else:
             return SetAccumulator()
+
+
+class SequencingProcessAggregator(SimpleAggregator):
+
+    def _get_accumulator(self, field) -> Optional[Accumulator]:
+        return SetAccumulator(max_size=10)
```
noah-aviel-dove, line 915: Is 10 really the appropriate `max_size` value here? Every other usage of `max_size` in this file it set to 100, so this looks like a missing zero. If 10 really is the correct value, consider adding a comment explaining why.

hannes-ucsc: I did specify 10 in the ticket. The reason for it being so low is that we don't expect more than 10 processes (inner entitiy) being linked to each file (outer entity). For other outer entities there may hundreds or thousands of processes but we don't care about them for those types, only the `files` type is important because it is used to generate the manifests. It might make sense to capture this reasoning in a comment.

hannes-ucsc:
To demo, show new column in a freshly generated `terra.bdbag` manifest.

hannes-ucsc:
https://gitlab.dev.explore.data.humancellatlas.org/azul/azul/-/jobs/5777

hannes-ucsc:
Reindex in progress: https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/3865

Issue #1783, opened 2020-05-08
Moto's dependency on `future` gets in the way


hannes-ucsc:
It pulls in some urllib shim that pretends urlparse returns bytes. 

![image](https://user-images.githubusercontent.com/5143256/81436300-18d6d400-911e-11ea-8bec-2ae7ba4c14b9.png)

I think we should try to prevent `future` from being installed. Don't know if that's possible.

melainalegaspi:
@achave11 sees a connection to the need for patching the maximum header count in test_deletions #1719

hannes-ucsc:
I can't repro this anymore. It may have been fixed inadvertently by upgrading future to in PR #1882 for #1738. I used to have future 0.14.0 in my virtualenv but now I have 0.16.0. Gitlab builds have been reporting `future` to be at 0.16.0 for months.

The associated PR should just remove the FIXME and a demonstration should show that this can't be reproduced.

hannes-ucsc:
Ignore previous comment. I am still able to [repro](https://github.com/DataBiosphere/azul/pull/1875#issuecomment-643478925) this.

I don't think this is `future`'s fault and I think the ticket description is wrong.

PR #1875, opened 2020-06-04
Prevent `future` related warnings (#1783)

achave11:


achave11:
Screen Shot 2020-06-12 at 1 13 09 PM


@hannes-ucsc I seem to still have the warning while having the same version of future. When testing did you also remove [this comment](https://github.com/DataBiosphere/azul/pull/1875/commits/9d54dff79155a20d2f40a0cf1389ee38b261f9a9#diff-2cd9b199c176e414cd3bf292f423f764L183)?

hannes-ucsc:
I did not and the warning does pop up again when I remove it. So we're back to square one.

Unfortunately, `future` is declared a direct runtime dependency and the `--no-deps` approach wouldn't have worked. Furthermore, I don't think it should even be a runtime dependency (https://github.com/DataBiosphere/azul/issues/1898).

I am also not even sure if `future` was the culprit here. I think this is a PyCharm bug.

This patch also fixes the warning: 

```diff
Index: src/azul/plugins/metadata/hca/full_metadata.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/azul/plugins/metadata/hca/full_metadata.py	(revision ad83a8156b929d3c462fb8dc11d22d96c8a5695c)
+++ src/azul/plugins/metadata/hca/full_metadata.py	(date 1591994867946)
@@ -178,9 +178,9 @@
                 'file_format': self._deep_get(metadata_file, 'file_core', 'file_format'),
             }
 
-            described_by = parse.urlparse(self._deep_get(metadata_file, 'describedBy'))
-            # FIXME: https://github.com/DataBiosphere/azul/issues/1783
-            # noinspection PyTypeChecker
+            described_by = self._deep_get(metadata_file, 'describedBy')
+            assert isinstance(described_by, str)
+            described_by = parse.urlparse(described_by)
             _, _, schema_type = described_by.path.rpartition('/')
             if schema_type in self.blocked_file_type:
                 continue
```

PR #1877, opened 2020-06-05
Improve HEAD method spec for index endpoints

noah-aviel-dove:


noah-aviel-dove:
Patch is not verbatim due to inconsistent indentation on line 867.

Issue #1499, opened 2019-12-05
Remove pagination elements made obsolete by next/previous links


hannes-ucsc:
Remove 

* `pagination.search_after`
* `pagination.search_after_uid`
* `pagination.search_before`
* `pagination.search_before_uid`

from all service responses. 

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/AZUL-922)
┆Project Name: azul
┆Issue Number: AZUL-922

PR #1878, opened 2020-06-05
Remove pagination elements made obsolete by next/previous links (#1499)

amarjandu:


amarjandu:
@hannes-ucsc do you have an opinion if these tests should be refactored or removed?

```
test/service/test_pagination:assert_page1_correct
test/service/test_pagination:assert_page2_correct
test/service/test_pagination:test_search_after_page1
test/service/test_pagination:test_search_after_page1_explicit_from
test/service/test_pagination:test_search_after_page2
test/service/test_pagination:test_search_after_page2_explicit_from
test/service/test_pagination:test_search_after_last_page
```

amarjandu:
@hannes-ucsc : "the tests check edge cases, we should retain/refactor them with the data present within .previous .next"

hannes-ucsc:
Yes, we don't want to lose coverage of either state or code. If those tests contribute coverage of either we need to retain them.

test/service/test_pagination.py
```diff
@@ -13,6 +13,42 @@ def setUpModule():
...
+
+
+def parse_url_qs(url):
+    url_parts = parse.urlparse(url)
+    query_dict = dict(parse.parse_qsl(url_parts.query, keep_blank_values=True))
+    return query_dict
+
+
+def extract_pagination_params_from_url(url: str):
+    """Helper method that extracts parameters for pagination from a url"""
```
danielsotirhos, line 26:
```suggestion
    """
    Helper method that extracts parameters for pagination from a url
    """
```

danielsotirhos, line 26:
```suggestion
    """
    Helper method that extracts parameters for pagination from a url
    """
```

src/azul/pagination.py
```diff
@@ -0,0 +1,38 @@
+from azul.requests import parse_url_qs
+
+obsolete_pagination_parameters = ('search_after', 'search_after_uid', 'search_before', 'search_before_uid')
```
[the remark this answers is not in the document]
amarjandu: `internal_pagination_parameters` sounds good.

src/azul/pagination.py
```diff
@@ -0,0 +1,38 @@
...
+    found_params = dict()
+    for x in obsolete_pagination_parameters:
+        if x in query_params:
+            found_params[x] = query_params[x]
+        else:
+            found_params[x] = None
+    return found_params
+
+
+def extract_pagination_params_from_response(json_response):
```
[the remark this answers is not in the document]
amarjandu: I thought it would be strange to modify the response that was coming from the API. Hence the use of 2 data structures. 
I think renaming the function would be `ok`, how about something like `extract_internal_pagination_params_from_response`..... too wordy?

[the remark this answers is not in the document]
danielsotirhos: A little wordy but acceptable. Alternatively I think `extract_internal_pagination_params(json_response)` works just as well.

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
+from azul.requests import parse_url_qs
+
+internal_pagination_parameters = ('search_after', 'search_after_uid', 'search_before', 'search_before_uid')
+'''
+    These pagination parameters are used internally, but should removed from the response that a client receives.
```
hannes-ucsc, line 5: If you looked around, you'd see that we don't indent docstrings like that. And docstrings are always triple double quoted.

hannes-ucsc: Wrap documentation on 80.

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
+from azul.requests import parse_url_qs
+
+internal_pagination_parameters = ('search_after', 'search_after_uid', 'search_before', 'search_before_uid')
```
hannes-ucsc, line 3: wrap all

scripts/count-bundles.py
```diff
@@ -47,8 +48,10 @@ def get_project_bundle_count(project_name):
 
                 all_bundle_ids.add(f"{bundle['bundleUuid']}.{bundle['bundleVersion']}")
 
-        search_after = response_json['pagination']['search_after']
-        search_after_uid = response_json['pagination']['search_after_uid']
+        pagination_params = extract_internal_pagination_params(response_json)
```
hannes-ucsc, line 51: This script is a client to Azul and so should just be using the next prev links and does not need to access the internals. If you eliminate the helper calls here the helper functions in `pagination.py` (the existence of which is a bad smell, btw) are only used in tests and can therefore be added as methods of a TestCase subclass.

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+'''
+    These pagination parameters are used internally, but should removed from the response that a client receives.
+'''
+
+
+def extract_internal_pagination_params_from_url(url: str):
+    """
+       Helper method that extracts parameters for pagination from a url
```
hannes-ucsc, line 11: unindent

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    query_params = parse_url_qs(url)
```
hannes-ucsc, line 13: A dict comprehension would make this a one liner.

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+    found_params = dict()
+    for x in internal_pagination_parameters:
+        found_params[x] = query_params.get(x)
+    return found_params
+
+
+def extract_internal_pagination_params(json_response):
+    """
+       Helper method that extracts parameters for pagination from a response
```
hannes-ucsc, line 22: unindent

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    next_url = json_response['pagination'].get('next')
```
hannes-ucsc, line 24: This doing the same thing twice. Try to refactor that somehow.

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+    previous_url = json_response['pagination'].get('previous')
+    joint_url_params = {x: None for x in internal_pagination_parameters}
+    if next_url:
+        for x, y in extract_internal_pagination_params_from_url(next_url).items():
```
hannes-ucsc, line 28:
```suggestion
        for k, v in extract_internal_pagination_params_from_url(next_url).items():
```

Lots of precedence in existing source code. Try to pick up more on precedence.

src/azul/pagination.py
```diff
@@ -0,0 +1,35 @@
...
+    return found_params
+
+
+def extract_internal_pagination_params(json_response):
+    """
+       Helper method that extracts parameters for pagination from a response
+    """
+    next_url = json_response['pagination'].get('next')
+    previous_url = json_response['pagination'].get('previous')
+    joint_url_params = {x: None for x in internal_pagination_parameters}
```
hannes-ucsc, line 26: Don't use None values. Absence and use of .get() is preferred.

src/azul/requests.py
```diff
@@ -20,3 +21,9 @@ def requests_session_with_retry_after() -> requests.Session:
     session.mount('http://', adapter)
     session.mount('https://', adapter)
     return session
+
+
+def parse_url_qs(url):
```
hannes-ucsc, line 26: If this has universal value it should get a docstring and maybe some doctests. But I doubt it does have universal value.

hannes-ucsc:
Some of my comments are a bit terse. If you have any questions we can discuss at stand-up.

src/azul/pagination.py
amarjandu: `None` is used here to allow for easier merging of the extracted parameter values.
The code will become more complex if there is a hard requirement for its removal.

src/azul/requests.py
amarjandu: It did not seem to have universal value, i place it back within the `test_pagination.py` file.

amarjandu:
@hannes-ucsc 
I had to modify the `integration_test` to account for the removal of the `internal_pagination` parameters; the `dont_test_concurrent_portal_db_crud` test failed after running for ~45min; the test was unable to complete due to the MFA token expiring, I've been having issues with this test running for extended periods of time.  

If possible I would like to run the IT in the sandbox.

src/azul/service/elasticsearch_service.py
```diff
@@ -535,6 +531,7 @@ def transform_request(self,
             facets = es_response_dict['aggregations'] if 'aggregations' in es_response_dict else {}
             pagination['sort'] = inverse_translation[pagination['sort']]
             paging = self._generate_paging_dict(filters, es_response_dict, pagination)
+
```
hannes-ucsc, line 534: Remove this change.

test/integration_test.py
```diff
@@ -417,20 +417,17 @@ def _get_entities_by_project(self, entity_type, project_short_name):
         filters = json.dumps({'project': {'is': [project_short_name]}})
         entities = []
         size = 100
-        params = dict(filters=filters, size=str(size))
+        url = f'{config.service_endpoint()}/index/{entity_type}?filters={filters}&size={size}'
```
hannes-ucsc, line 420: This hurts my eyes. You actually reintroduced the broken way of composing the URL. Think about why this doesn't generally work.

test/service/test_pagination.py
```diff
@@ -13,6 +12,31 @@ def setUpModule():
     configure_test_logging()
 
 
+internal_pagination_parameters = ('search_after',
+                                  'search_after_uid',
+                                  'search_before',
+                                  'search_before_uid')
+"""
+These pagination parameters are used internally,
```
hannes-ucsc, line 20: This is wrapped weirdly. It should be wrapped on 80.

test/service/test_pagination.py
```diff
@@ -13,6 +12,31 @@ def setUpModule():
...
+"""
+
+
+def extract_url_params(json_response):
+    """
+    Helper method that extracts parameters for pagination from a response
+    """
+    next_url = json_response['pagination'].get('next', '')
+    previous_url = json_response['pagination'].get('previous', '')
+    return {**parse_url_qs(next_url), **parse_url_qs(previous_url)}
```
hannes-ucsc, line 31: We can't have this method. If `next` had `&search_before=…&search_after_uid=…` in it and `prev` had `&search_after=…&search_before_uid=…` next and prev would be broken but any test using it would not detect that since it eliminates the separation and merges everything into a single dict, hiding the criss-crossed URL params.

test/integration_test.py
amarjandu: Yeah this composition makes future extensions of additional parameters to be a pain. Moved to composing with furl()

test/service/test_pagination.py
```diff
@@ -35,22 +52,16 @@ def assert_page1_correct(self, json_response):
...
-            self.assertIsNone(json_response['pagination']['search_before'],
-                              "search_before != null on first page of results")
-            self.assertIsNone(json_response['pagination']['search_before_uid'],
-                              "search_before != null on first page of results")
-        else:
-            self.fail("search_before not set to null on first page of results")
+        pagination = json_response['pagination']
+        next_ = parse_url_qs(pagination['next'])
+
+        self.assertIsNone(pagination['previous'])
```
hannes-ucsc, line 58:
```suggestion
        self.assertIsNone(pagination['previous'])
        next_ = parse_url_qs(pagination['next'])
```

test/service/test_pagination.py
```diff
@@ -62,25 +73,25 @@ def assert_page2_correct(self, json_response, json_response_second, sort_order):
...
-
-        if 'search_after' in json_response['pagination']:
-            self.assertEqual(json_response_second['pagination']['search_after'],
-                             json.dumps(json_response_second['hits'][num_hits_second - 1]['entryId']),
-                             "search_after on second page not set to last returned document on second page, "
-                             + "order=" + sort_order)
-            self.assertIsNotNone(json_response['pagination']['search_after_uid'],
-                                 "No search_after_uid returned on second page")
-        else:
-            self.fail("search_after not set on second page of results, sortOrder=" + sort_order)
```
hannes-ucsc, line 83: We should add a `with self.subTest(sort_order=sort_order)` instead of interpolating `sort_order` into all of those assertion failure messages.

test/integration_test.py
```diff
@@ -418,19 +418,21 @@ def _get_entities_by_project(self, entity_type, project_short_name):
         entities = []
         size = 100
         params = dict(filters=filters, size=str(size))
+        url = furl(
+            url=f'{config.service_endpoint()}',
+            path=f'/index/{entity_type}',
```
hannes-ucsc, line 423:
```suggestion
            path=('index', entity_type),
```

test/service/test_pagination.py
```diff
@@ -13,6 +12,24 @@ def setUpModule():
...
 
+internal_pagination_parameters = (
+    'search_after',
+    'search_after_uid',
+    'search_before',
+    'search_before_uid'
+)
+"""
+These pagination parameters are used internally,they are should not be
+included in the `pagination` response field.
```
hannes-ucsc, line 23:
```suggestion
These pagination parameters are used internally, they should not be
included in the `pagination` response field.
```

and rewrap on 80

test/service/test_pagination.py
```diff
@@ -154,45 +151,34 @@ def test_search_after_last_page(self):
         Tests that the last page returned in search_after pagination mode is correct.
         :return:
         """
+
```
hannes-ucsc, line 154:
```suggestion
```

test/service/test_pagination.py
amarjandu: I opted to remove the variable, it was not being used.

Issue #1592, opened 2020-02-20
Rename `disease` facet to `sampleDisease`


hannes-ucsc:
Once DB switches to `specimenDisease` we can remove the `disease` facet.

Edit: We decided to rename the facet to `sampleDisease` instead of removing it.

achave11:
@hannes-ucsc : "Instead of removing the `disease` facet, we should rename it to `sampleDisease` it might still be useful to filter by the disease of only those specimens that are samples"

PR #1881, opened 2020-06-08
Rename `disease` facet to `sampleDisease` (#1592)

achave11:


src/azul/plugins/metadata/hca/__init__.py
```diff
@@ -273,7 +273,6 @@ def service_config(self) -> ServiceConfig:
                 "organismAge",
                 "organismAgeUnit",
                 "biologicalSex",
-                "disease",
```
hannes-ucsc, line 276: We should  have an term facet entry for `sampleDisease`.

PR #1900, opened 2020-06-15
Add retry to integration test notifications

amarjandu:


amarjandu:
@achave11 i plan to change the commit title for the first commit `Remove exception warning from azulclient._index ` to `Remove log.warning for unhanded exceptions from azulclient._index`
I'll wait for the review to complete before doing so.

src/azul/azulclient.py
```diff
@@ -123,22 +123,18 @@ def handle_future(future):
                 nonlocal indexed
                 # @formatter:on
                 # Block until future raises or succeeds
```
achave11, line 125: Wondering if this comment still applies after the removal?

amarjandu: removed.

amarjandu:
> I would have personally done it in one commit given that you are just making a delete in the first commit. I'll leave it at your discretion, I do not think there would be a penalty for having it in 1 or 2 commits.

Noted, I merged them together, and added a fixup to address the comment removal.

PR #1901, opened 2020-06-16
DSS adapter get schema type from describedBy

danielsotirhos:
PR base branch 'issues/danielsotirhos/1753-dss_adapter_for_v2'

### DIFF
--- a/scripts/dss_v2_apdapter.py
+++ b/scripts/dss_v2_apdapter.py
@@ -200,6 +200,13 @@ class DSSv2Adapter:
                 except CancelledError:
                     pass  # ignore exception raised when a future is cancelled
                 except BaseException as e:
+                    # Bundle's are pre-checked and skipped if an error is found,
+                    # so an exception here means some of the bundle's content
+                    # was probably transfered, leaving the destination bucket in
+                    # a polluted state. To prevent further pollution we do a
+                    # graceful exit canceling all unbegun futures. Should any
+                    # in-progress future also fail it will be caught here and
+                    # added to self.errors to report at the end of the script.
                     self.errors[bundle_fqid] = e
                     for f in future_to_bundle:
                         f.cancel()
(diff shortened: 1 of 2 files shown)


scripts/dss_v2_apdapter.py
```diff
@@ -388,18 +386,12 @@ def _setup(self):
...
-            elif manifest_entry['indexed']:  # Metadata files
-                # FIXME: get schema type from 'describeBy' https://github.com/databiosphere/azul/issues/1831
-                match = self.file_name_re.match(file_name)
-                if match:
-                    self.schema_types[file_uuid] = match.group(1)
-                else:
-                    raise Exception(f'Indexed file {file_name} has unknown file name format.')
+            if manifest_entry['indexed']:  # Metadata files
+                described_by = self.indexed_files[file_name]['describedBy']
+                _, schema_type = described_by.rsplit('/', maxsplit=1)
```
hannes-ucsc, line 391:
```suggestion
                _, _, schema_type = described_by.rpartition('/')
```

scripts/dss_v2_apdapter.py
```diff
@@ -395,18 +393,12 @@ def _setup(self):
...
-                if match:
-                    self.schema_types[file_uuid] = match.group(1)
-                else:
-                    raise Exception(f'Indexed file {file_name} has unknown file name format.')
+            if manifest_entry['indexed']:  # Metadata files
+                described_by = self.indexed_files[file_name]['describedBy']
+                _, _, schema_type = described_by.rpartition('/')
+                self.schema_types[file_uuid] = schema_type
+                if schema_type == 'project':
+                    self.project_uuid = manifest_entry['uuid']
```
hannes-ucsc, line 401: If there is a way to get the project UUID out of the `provenance.document_id` metadata property that would be more natural but I would also assert that it is `== manifest_entry['uuid']`

Issue #1715, opened 2020-04-17
Unable to deploy from develop branch to a personal deployment


amarjandu:
Active Branch: Develop
Stage Sourced: amar

```
(venv-azul) ubuntu@ip-172-31-12-34:~/dev/hca/azul$ make integration_test
python /home/ubuntu/dev/hca/azul/scripts/check_branch.py
Traceback (most recent call last):
  File "/home/ubuntu/dev/hca/azul/scripts/check_branch.py", line 112, in <module>
    main(sys.argv[1:])
  File "/home/ubuntu/dev/hca/azul/scripts/check_branch.py", line 105, in main
    check_branch(branch, stage)
  File "/home/ubuntu/dev/hca/azul/scripts/check_branch.py", line 59, in check_branch
    raise RuntimeError(f"Protected branch '{branch}' should be deployed to '{expected_stage}', not '{stage}'")
RuntimeError: Protected branch 'develop' should be deployed to 'dev', not 'amar'
common.mk:37: recipe for target 'check_branch' failed
make: *** [check_branch] Error 1
```

Unable to run `make` commands from the deploy branch if the `stage` is set to a personal deployment.

PR #1918, opened 2020-06-23
Allow personal deployment from any branch (#1715)

natanlao:
Changes the behavior of check_branch to permit deploying protected
branches to a personal stage. A deployment to an improper protected
stage will still fail (e.g., trying to deploy `develop` to `prod`, or
something unfortunate like that).

### DIFF
--- a/scripts/check_branch.py
+++ b/scripts/check_branch.py
@@ -21,9 +21,8 @@ def check_branch(branch: Optional[str], stage: str) -> None:
     RuntimeError: Non-protected branch 'feature/foo' can't be deployed to main deployment 'prod'
 
     >>> check_branch('staging', 'hannes.local')
-    Traceback (most recent call last):
-    ...
-    RuntimeError: Protected branch 'staging' should be deployed to 'staging', not 'hannes.local'
+
+    >>> check_branch('develop', 'hannes.local')
 
     >>> check_branch('staging', 'integration')
     Traceback (most recent call last):
@@ -47,7 +46,7 @@ def check_branch(branch: Optional[str], stage: str) -> None:
             )
     else:
         assert branch is not None
-        if stage != expected_stage:
+        if stage != expected_stage and config.is_main_deployment(stage):
             raise RuntimeError(f"Protected branch '{branch}' should be deployed to '{expected_stage}', not '{stage}'")
 
 
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -465,9 +465,10 @@ class Config:
         'prod': 'prod'
     }
 
-    @property
-    def is_main_deployment(self):
-        return self.deployment_stage in self.main_deployments_by_branch.values()
+    def is_main_deployment(self, stage=None):
+        if stage is None:
+            stage = self.deployment_stage
+        return stage in self.main_deployments_by_branch.values()
 
     @property
     def _git_status(self) -> Mapping[str, str]:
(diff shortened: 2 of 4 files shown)


scripts/check_branch.py
```diff
@@ -47,7 +44,7 @@ def check_branch(branch: Optional[str], stage: str) -> None:
             )
     else:
         assert branch is not None
-        if stage != expected_stage:
+        if stage != expected_stage and stage in stage_by_branch.values():
```
jessebrennan, line 47: I noticed that `config.is_main_deployment` is basically doing the same thing as the check you added.

Consider (this is completely optional and only at your discretion) consolidating the two.

natanlao: I think it makes sense to keep it as is since there's no guarantee that `stage` is `self.deployment_stage` (which `config.is_main_deployment`) checks. But I didn't know about `config.is_main_deployment` and I'll remember it for next time!

hannes-ucsc: Agree with @jessebrennan. There is room for consolidation.

```diff
Index: src/azul/__init__.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/azul/__init__.py	(revision b8fff54d12d4b6e18cf659996f6674bd9e1e0058)
+++ src/azul/__init__.py	(date 1593133817469)
@@ -465,9 +465,10 @@
         'prod': 'prod'
     }
 
-    @property
-    def is_main_deployment(self):
-        return self.deployment_stage in self.main_deployments_by_branch.values()
+    def is_main_deployment(self, stage=None):
+        if stage is None:
+            stage = self.deployment_stage
+        return stage in self.main_deployments_by_branch.values()
 
     @property
     def _git_status(self) -> Mapping[str, str]:
Index: terraform/s3.tf.json.template.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- terraform/s3.tf.json.template.py	(revision b8fff54d12d4b6e18cf659996f6674bd9e1e0058)
+++ terraform/s3.tf.json.template.py	(date 1593131564372)
@@ -20,7 +20,7 @@
                 },
                 "url_bucket": {
                     "bucket": config.url_redirect_full_domain_name,
-                    "force_destroy": not config.is_main_deployment,
+                    "force_destroy": not config.is_main_deployment(),
                     "acl": "public-read",
                     "website": {
                         # index_document is required; pointing to a non-existent file to return a 404
Index: test/integration_test.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- test/integration_test.py	(revision b8fff54d12d4b6e18cf659996f6674bd9e1e0058)
+++ test/integration_test.py	(date 1593131580944)
@@ -441,7 +441,7 @@
         notifications = [invalid_notification]
         self.assertRaises(AzulClientNotificationError, self.azul_client._index, notifications)
 
-    @unittest.skipIf(config.is_main_deployment, 'Test would pollute portal DB')
+    @unittest.skipIf(config.is_main_deployment(), 'Test would pollute portal DB')
     def test_concurrent_portal_db_crud(self):
         """
         Use multithreading to simulate multiple users simultaneously modifying
```

scripts/check_branch.py
```diff
@@ -21,9 +21,6 @@ def check_branch(branch: Optional[str], stage: str) -> None:
     RuntimeError: Non-protected branch 'feature/foo' can't be deployed to main deployment 'prod'
 
     >>> check_branch('staging', 'hannes.local')
```
hannes-ucsc, line 23:
```suggestion
    >>> check_branch('develop', 'hannes.local')
```

Since that's our main use case. Or maybe have one for develop and one for staging.

PR #1926, opened 2020-06-25
Fix: HEAD /index/summary spec declares a body (#1916)

noah-aviel-dove:
I refactored many of the `repository_spec` methods to accomplish this change while introducing minimal redundancy. This involved changing a lot of variables storing singular words to plural, since "summary" is not pluralized at any point in the spec. Reviewers should be careful to check that there were no unintentional changes from singular to plural or vice versa in the existing spec. However, the pluralization of the entity type in the generic HEAD spec on line 959 IS intentional, since index names are plural.

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -835,9 +835,9 @@ def filters_param_spec(facets):
     )
 
 
-def repository_seach_params_spec(entity_type):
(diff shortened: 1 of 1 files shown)


noah-aviel-dove:
Found broken link, returning to peer review.

Issue #1927, opened 2020-06-25
openapi size schema missing min/max


amarjandu:
min/max values that the `size` parameter can use is missing from the openapi spec schema. 
![Screen Shot 2020-06-25 at 10 12 29 AM](https://user-images.githubusercontent.com/1947534/85767036-745c2180-b6cc-11ea-9840-8e274d738606.png)

see `Default Parameter Values` on https://swagger.io/docs/specification/describing-parameters/

PR #1928, opened 2020-06-25
Add minimum / maximum to openapi size spec (#1927)

amarjandu:


src/azul/openapi/schema.py
```diff
@@ -220,10 +220,19 @@ def with_default(default: PrimitiveJSON, /, type_: Optional[TYPE] = None) -> JSO
         "format": "double",
         "default": 0
     }
+    >>> assert_json(with_default(10, min=10,max=20))
```
hannes-ucsc, line 223:
```suggestion
    >>> assert_json(with_default(10, min=10, max=20))
```

hannes-ucsc: min max vs minimum maximum?

src/azul/openapi/schema.py
```diff
@@ -203,7 +203,7 @@ def enum(*items: PrimitiveJSON, type_: TYPE = None) -> JSON:
     }
 
 
-def with_default(default: PrimitiveJSON, /, type_: Optional[TYPE] = None) -> JSON:
+def with_default(default: PrimitiveJSON, /, type_: Optional[TYPE] = None, **kwargs) -> JSON:
```
hannes-ucsc, line 206:
```suggestion
def with_default(default: PrimitiveJSON, /, type_: Optional[TYPE] = None, **kwargs: AnyJSON) -> JSON:
```

src/azul/openapi/schema.py
amarjandu: I assumed these are just doc tests, and is to just showcase how it works? I'll change them.

hannes-ucsc: In that case I would have used something obviously meaningless like "foo" and "bar". Min and max were in the twilight zone between real and fantasy, which could be confusing. But this is fine.

lambdas/service/app.py
```diff
@@ -842,7 +842,7 @@ def repository_seach_params_spec(entity_type):
         filters_param_spec(facets),
         params.query(
             'size',
-            schema.optional(schema.with_default(10)),
+            schema.optional(schema.with_default(10, minimum=1, maximum=1000)),
```
hannes-ucsc, line 845: For now, these limits need to be tied to where they are actually enforced.

lambdas/service/app.py
```diff
@@ -540,6 +540,10 @@ def validate_repository_search(params, **validators):
     })
 
 
+max_page_size = 1000
```
hannes-ucsc, line 543: min before max, low before high, left before right etc.

amarjandu: @hannes-ucsc would it be appropriate to make the change after a hold-warm tag has been placed on the PR?

hannes-ucsc: Not unless I relinquish the PR back to you. I probably won't get around to merging it today, so I'll do just that.

Thanks for asking.

hannes-ucsc:
https://github.com/DataBiosphere/azul/pull/1928#discussion_r447440205

lambdas/service/app.py
```diff
@@ -877,7 +880,7 @@ def repository_seach_params_spec(index_name):
         filters_param_spec(facets),
         params.query(
             'size',
-            schema.optional(schema.with_default(10)),
+            schema.optional(schema.with_default(10, minimum=min_page_size, maximum=max_page_size)),
```
hannes-ucsc, line 883: It bugged me that we can only specify ranges for schemas that have a default so I generalized this a bit.

Issue #1888, opened 2020-06-10
Enable error logs on Elasticsearch domain


hannes-ucsc:
Error logs are essential in diagnosing issues like the recently stuck snapshot. See third section in screen shot below. I think the logs will be voluminous so we should make sure the log group has a short retention of, say one month.

![image](https://user-images.githubusercontent.com/5143256/84298470-0fb59a00-ab04-11ea-83b1-e4e1e3830fe4.png)

PR #1933, opened 2020-06-25
Enable error logs on Elasticsearch domain (#1888)

achave11:


achave11:
> You must have tested this. Can you point me at the error log group for the deployment you tried this on?
> 
> Not bumping review count.

Here is the requested [CloudWatch Log Group](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Faes$252Fdomains$252Fazul-index-achave11es$252Ferror-logs), let me know when it will be a good time to take down the deployment.

hannes-ucsc:
You can take the domain down, now. Thank you.

hannes-ucsc:
Try to stick with issue title for commits and PRs unless there is a good reason not to do so. I noticed that your title choices often carry the same meaning as the issue but are still different in insignificant ways.

Issue #1932, opened 2020-06-25
Optionally skip secondary bundles in DSS adapter for v2


theathorn:
The Broad will be running updated pipelines to produce new secondary analysis outputs for DCP/1 data.
As a result we should not transfer any existing analysis files from DCP/1 to DCP/2 (decision confirmed with Kathleen Tibbetts).

PR #1950, opened 2020-06-30
Optionally skip secondary bundles in DSS adapter for v2 (#1932)

danielsotirhos:


scripts/dss_v2_apdapter.py
```diff
@@ -287,6 +297,12 @@ def _process_bundle(self, bundle_fqid: BundleFQID):
                 except Exception as e:
                     raise Exception(bundle_fqid, file_name, 'Failed to upload file contents') from e
 
+    def _is_analysis_bundle(self, bundle_manifest: JSON) -> bool:
+        for manifest_entry in bundle_manifest['files']:
+            if manifest_entry['indexed'] and manifest_entry['name'].startswith('analysis_file_'):
```
hannes-ucsc, line 302: The presence of `analysis_process` is a more reliable criterion. And I'd prefer the use the schema_type from as determined from `describedBy` inside the metadata document.

Would also be good to know if this change actually results in a different set of bundles being skipped.

danielsotirhos: Confirmed with first `fixup` https://github.com/DataBiosphere/azul/pull/1950/commits/f18bf517c2341d5ae91c472a19c1a0b572a6d533 and ten runs with a random 2-character prefix that both approaches (schema_type vs filename) found the exact same result set of analysis bundles.

hannes-ucsc: Nice.

danielsotirhos:
Agreed, much simpler. Patch applied.

Issue #1977, opened 2020-07-02
version query parameter schema missing from /dss/ endpoints


amarjandu:
the `version` parameter can be used in the /dss/ endpoints as a query parameter however it is missing from the openapi schema.

PR #1978, opened 2020-07-02
Add `version` parameter to spec of /dss endpoints (#1977)

amarjandu:


amarjandu:
@danielsotirhos fixed

lambdas/service/app.py
```diff
@@ -1379,7 +1379,22 @@ def generate_manifest(event, context):
                 more information refer to https://dss.data.humancellatlas.org
                 under `GET /files/{uuid}`. The only mandatory forwarded
                 parameter is `replica`, for which Azul only supports AWS.
-            ''')
+            '''),
+        params.query(
+            'version',
+            schema.optional(str),
+            description='''
+                A timestamp that generally follows
```
hannes-ucsc, line 1387: The format of the timestamp is an implementation detail of the DSS, really. Clients should treat the version as an opaque string that is lexicographically sorted by time. We should only document that property. Can you check the DRS/DSS endpoints for this issue, too?

lambdas/service/app.py
```diff
@@ -1349,6 +1349,18 @@ def generate_manifest(event, context):
...
 
+dss_version_parameter = params.query(
+    'version',
+    schema.optional(str),
+    description='''
+        The version of the file to be checked out. Versions can be
+        lexicographically compared to determine which is more recent.
+        If this parameter is omitted then the most recent version
+        of the file is returned. For more information refer to
+        https://dss.data.humancellatlas.org under `GET /files/{uuid}`.
```
hannes-ucsc, line 1360:
```suggestion
        The version of the file to be returned. File versions are opaque strings with only one documented property: they can be
        lexicographically compared with each other in order to determine which version is more recent.
        If this parameter is omitted then the most recent version
        of the file is returned.`.
```

This will soon need to work with TDR, too.

lambdas/service/app.py
```diff
@@ -2390,9 +2406,8 @@ def get_data_object(file_uuid):
         the DSS to do a checkout.
     ''') + drs_description,
     'parameters': [
-        params.path('file_uuid', str, description='UUID of the file to be fetched'),
         params.path('access_id', str, description='Access ID returned from a previous request'),
-    ],
+        *drs_dss_parameters],
```
hannes-ucsc, line 2410:
```suggestion
        *drs_dss_parameters
    ],
```

lambdas/service/app.py
```diff
@@ -1349,6 +1349,18 @@ def generate_manifest(event, context):
     return {'Location': presigned_url}
 
 
+dss_version_parameter = params.query(
```
hannes-ucsc, line 1352:
```suggestion
file_version_parameter = params.query(
```

lambdas/service/app.py
```diff
@@ -1392,6 +1392,16 @@ def generate_manifest(event, context):
     return {'Location': presigned_url}
 
 
+file_version_parameter = params.query(
+    'version',
+    schema.optional(str),
+    description='''
+        The version of the file to be returned. File versions are opaque strings with only one documented property:
```
hannes-ucsc, line 1399: Wrap on 80 and use `format_description`, please.

amarjandu: Q: Should we be using `format_description` with triple strings in documentation? There appears to be mixed usage in the final spec for `/dss/files/{uuid}`; the parameter spec `dss_files_spec` not using the `format_description`, while response spec using `format_description`.

Issue #1985, opened 2020-07-07
`token` parameter  spec missing from `/dss` proxy endpoints


amarjandu:
`token` parameter is used within testing and is not documented in the openapi spec.
https://github.com/DataBiosphere/azul/blob/cf5404c93ea244fabfad56bcc2667d39d4f27fd6/test/service/test_dss_proxy.py#L155

PR #1988, opened 2020-07-07
Add `token` spec to `/dss` endpoints (#1985)

amarjandu:


lambdas/service/app.py
```diff
@@ -1384,7 +1384,12 @@ def generate_manifest(event, context):
                 more information refer to https://dss.data.humancellatlas.org
                 under `GET /files/{uuid}`. The only mandatory forwarded
                 parameter is `replica`, for which Azul only supports AWS.
-            ''')
+            '''),
+        params.query(
+            'token',
+            schema.optional({'type': 'string'}),
```
hannes-ucsc, line 1390: There is a shortcut for strings. If you borrowed this from somewhere else, that place should also use the shortcut.

amarjandu:
@hannes-ucsc 
Q: I was wondering about the usefulness of having a central registry for the service-api query parameters. By centralizing them we can de-clutter the variables that we are making at the global namespace (`filters_param_spec`, `catalog_param_spec`, `...` ) , and have an easier time sharing query parameters across endpoint paths.

hannes-ucsc:
> Q: I was wondering about the usefulness of having a central registry for the service-api query parameters. By centralizing them we can de-clutter the variables that we are making at the global namespace (filters_param_spec, catalog_param_spec, ... ) , and have an easier time sharing query parameters across endpoint paths.

@amarjandu and I talked. To summarize, I don't think centralizing helps. The default principle should be that tightly coupled constructs should be in close proximity. Centralizing all parameter specs would put parameter specs that aren't shared at a great distance from the endpoint specs that use them. This seems too high a price to pay for having all shared parameter specs in one place. We should define shared parameter specs in proximity of their first use. Subsequent uses may be further away from the definition but that could be fixed by reordering the endpoints such that all endpoint specs sharing a parameter spec are in close proximity to each other and therefore to the definition of the shared parameter spec.

Issue #2002, opened 2020-07-12
IT doesn't print missing bundles if it fails


hannes-ucsc:
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/4386

only shows that there is a difference between the collections of actual and expected bundles, not what the actually difference is.

PR #2020, opened 2020-07-15
Print missing bundles in `IndexingIntegrationTest` (#2002)

amarjandu:


amarjandu:
Bundles differences show as 
```
AssertionError: Items in the second set but not the first:
BundleFQID(uuid='6f674f83-afc2-4b2c-83fa-4d619f7b576a', version='2019-08-01T200147.035627Z')
BundleFQID(uuid='6f52f10e-2aa3-4ecf-97d1-cd521aa2c889', version='2019-08-01T200147.934245Z')
BundleFQID(uuid='6fb3d637-6b10-4a00-8701-0fc1500dda4b', version='2019-05-15T230845.945000Z')
BundleFQID(uuid='6f100e38-d057-477d-8bb1-d002d06f582c', version='2019-08-01T200147.139512Z')
BundleFQID(uuid='6f532dbc-a2ea-4f3b-b84c-98c7de98b09f', version='2019-08-01T200147.191452Z')
```

rather than 
`AssertionError: {BundleFQID(uuid='d6015595-a613-4211-804e[5876 chars]8Z')} != dict_keys([BundleFQID(uuid='d607a428-67a8[5981 chars]Z')])`

amarjandu:
> Looks good. Consider also setting [maxDiff](https://docs.python.org/3/library/unittest.html#unittest.TestCase.maxDiff) so that if there is a large discrepancy, you will see everything.

I attempted to do that on a preliminary revision but from the docs 
>> `Assert methods affected by this attribute are assertSequenceEqual() (including all the sequence comparison methods that delegate to it), assertDictEqual() and assertMultiLineEqual().`

When i ran the test with `self.maxdiff = None` the full diff was not printed out :(

hannes-ucsc:
I'll run this in the sandbox and then remove the drop commit.

hannes-ucsc:
IT fails as expected with correct assertion message:

https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/4622

Issue #2025, opened 2020-07-16
Register indexer SA with Broad's SAM during deployment


hannes-ucsc:
In order to be allow-listed for access to TDR, an SA needs to be registered with SAM.

This is how we register an account using `gcloud`:

`curl https://sam.dsde-dev.broadinstitute.org/register/user/v1 -d ""  -H "Authorization: Bearer $(gcloud auth --account $service_account_name print-access-token)"`

For the real solution, the access token should not be obtained using gcloud, but programmatically somehow, probably using https://google-auth-oauthlib.readthedocs.io/en/latest/reference/google_auth_oauthlib.flow.html#google_auth_oauthlib.flow.InstalledAppFlow.fetch_token

We still need to allow-list the account here: https://bvdp-saturn-dev.appspot.com/#groups/JadeStewards-dev but maybe we can automate that as well if we make all developers and Gitlab admin for the JadeStewards-dev group.

noah-aviel-dove:
This is what I was able to work out when we discussed this previously:
```
from google.oauth2.service_account import Credentials
from google.auth.transport.requests import Request

def get_access_token():
    creds = Credentials.from_service_account_file(os.environ['GOOGLE_APPLICATION_CREDENTIALS'])
    creds = creds.with_scopes(['email'])
    creds.refresh(Request())
    return creds.token
```

PR #2029, opened 2020-07-17
Programatically register service accounts with SAM (#2025)

noah-aviel-dove:


noah-aviel-dove:
@jessebrennan When I followed the instructions in UPGRADING.rst, I got a 409 error because my account has already been manually registered. So I'd appreciate it if you could run that step and report whether your registration was successful or not, since I can't observe a success myself.

UPGRADING.rst
```diff
@@ -10,6 +10,17 @@ branch that does not have the listed changes, the steps would need to be
...
+================================================================
+
+Google service accounts are now registered with SAM (needed for TDR
+authorization) as a part of the provisioning process during deployment. However,
+previously provisioned credentials are not retroactively registered by default.
+`_select` your personal deployment and run::
+
+     python3 scripts/provision_credentials.py google-key --build azul-ucsc-indexer-"$AZUL_DEPLOYMENT_STAGE"@"$GOOGLE_PROJECT".iam.gserviceaccount.com --register-existing
+
+To register your existing personal deployment's indexer service account.
```
hannes-ucsc, line 23: We should strive for idempotent, self-healing solutions wherever possible. This current implementation gives us one shot at registering with SAM, at the time the SA is provisioned. If something goes wrong during that, manual intervention is required. It would be more elegant, if running `make deploy` again would fix the issue. The requests to Google and SAM are cheap so I don't mind doing them every time `make deploy` is run.

IIRC, SAM returns a 409 when attempting to register an already registered user so we could make this idempotent and not require this manual intervention.

scripts/provision_credentials.py
```diff
@@ -29,12 +35,13 @@ def get_google_service():
 
 
 class CredentialsProvisioner:
+    sam_endpoint = 'https://sam.dsde-dev.broadinstitute.org/register/user/v1'
```
hannes-ucsc, line 38: We can't hard-code this here obviously as it is specific to `dev`

scripts/provision_credentials.py
```diff
@@ -143,6 +156,22 @@ def _destroy_service_account_creds(self, service_account_email):
             logger.info("Successfully deleted service account key with id '%s' for user '%s'",
                         key_id, service_account_email)
 
+    def _register_sam(self, google_key):
```
hannes-ucsc, line 159:
```suggestion
    def _register_sam(self, google_key: JSON):
```

scripts/provision_credentials.py
```diff
@@ -143,6 +156,22 @@ def _destroy_service_account_creds(self, service_account_email):
...
+        credentials = service_account.Credentials.from_service_account_info(google_key)
+        token = self._get_access_token(credentials)
+        response = pool.request('POST',
+                                self.sam_endpoint,
+                                body='',
+                                headers={'Authorization': f'Bearer {token}'})
+        if response.status != 200:
+            raise Exception(f'SAM registration failed: status {response.status}, {response.msg}')
+
+    def _get_access_token(self, credentials: service_account.Credentials):
```
hannes-ucsc, line 170:
```suggestion
    def _get_access_token(self, credentials: service_account.Credentials) -> str:
```

scripts/provision_credentials.py
```diff
@@ -143,6 +156,22 @@ def _destroy_service_account_creds(self, service_account_email):
...
 
+    def _register_sam(self, google_key):
+        pool = urllib3.PoolManager(ca_certs=certifi.where())
+        credentials = service_account.Credentials.from_service_account_info(google_key)
+        token = self._get_access_token(credentials)
+        response = pool.request('POST',
+                                self.sam_endpoint,
+                                body='',
+                                headers={'Authorization': f'Bearer {token}'})
+        if response.status != 200:
```
hannes-ucsc, line 167: It looks like SAM returns a 201 in the success case which would trigger a false error.

That seems to indicate that this wasn't tested properly.

hannes-ucsc: For idempotency, the expected error would be 409.

hannes-ucsc: I guess it's hard to test without first deregistering the SA. Does SAM have the ability to deregister a user?

noah-aviel-dove: Not as far as I can tell. https://sam.dsde-prod.broadinstitute.org/

noah-aviel-dove: Better URL: https://sam.dsde-prod.broadinstitute.org/#/Users

scripts/register_sam.py
```diff
@@ -0,0 +1,51 @@
...
+        token = self.get_access_token()
+        response = self.http.request('POST',
+                                     f'{config.sam_service_url}/register/user/v1',
+                                     body='',
+                                     headers={'Authorization': f'Bearer {token}'})
+
+        if response.status == 201:
+            log.info('Google service account successfully registered with SAM')
+        elif response.status == 409:
+            log.info('Google service account previously registered with SAM')
```
jessebrennan, line 35: Glad to see you added this. I was going to suggest it otherwise.

hannes-ucsc:
Upgrade instructions should include asking me to add SA to steward group in SAM.

noah-aviel-dove:
@hannes-ucsc How about having the registration script make a TDR API call to test authorization, and if it fails, print an explanation? Or would that get lost too easily in the terraform output?

hannes-ucsc:
If the warning is visible in the last, say, 20 lines of terminal output it would be ok, I think. We could simply add a command invocation after `terraform apply` in the `apply` recipe of the terraform/Makefile.

UPGRADING.rst
```diff
@@ -10,6 +10,13 @@ branch that does not have the listed changes, the steps would need to be
 reverted. This is all fairly informal and loosely defined. Hopefully we won't
 have too many entries in this file.
 
+#2025 Register indexer service accounts with SAM
+================================================
+
+Service accounts are now registered with SAM upon deployment.
```
hannes-ucsc, line 16: Why plural? Isn't there just one SA?

UPGRADING.rst
```diff
@@ -10,6 +10,13 @@ branch that does not have the listed changes, the steps would need to be
...
+They can also manually be registered via ``make -C terraform sam``.
+However, the accounts must still be manually authorized for steward access by
+the team lead.
```
hannes-ucsc, line 19: This should go in the README since the need to get authorized does not only occur during upgrades but for every new deployment.

UPGRADING.rst
```diff
@@ -10,6 +10,13 @@ branch that does not have the listed changes, the steps would need to be
 reverted. This is all fairly informal and loosely defined. Hopefully we won't
 have too many entries in this file.
 
+#2025 Register indexer service accounts with SAM
+================================================
+
+Service accounts are now registered with SAM upon deployment.
+They can also manually be registered via ``make -C terraform sam``.
+However, the accounts must still be manually authorized for steward access by
```
hannes-ucsc, line 18:
```suggestion
However, the account must still be authorized for repository read access (datasets and snapshot).
This is currently done manually by someone on the Azul team. Ask on Slack to find out more.
```

UPGRADING.rst
hannes-ucsc, line 19:
```suggestion

```

scripts/register_sam.py
```diff
@@ -0,0 +1,72 @@
...
+        else:
+            raise RuntimeError(f'Unexpected response during SAM registration: {response.data}')
+
+    def get_access_token(self) -> str:
+        credentials = self.credentials.with_scopes(['email', 'openid'])
+        credentials.refresh(Request())
+        return credentials.token
+
+    def test_access(self):
+        response = self.http.request('GET',
```
hannes-ucsc, line 56: Why isn't this in the TDR client?

src/azul/tdr.py
```diff
@@ -162,12 +181,50 @@ def add_entity(self, entity_key, entity_type: str, entity_row: JSON) -> None:
...
+
 class AzulTDRClient:
     timestamp_format = '%Y-%m-%dT%H:%M:%S.%fZ'
 
     def __init__(self, dataset: BigQueryDataset):
         self.target = dataset
-        self.big_query_adapter.assert_table_exists(dataset.name, 'links')
+        self.big_query_adapter.assert_table_exists(dataset.bq_name, 'links')
+
+    def get_target_info(self) -> JSON:
```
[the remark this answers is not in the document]
hannes-ucsc: The caller would need to know what the structure of returned JSON is. That's a leaky abstraction. The AzulTDRClient class should hide as much of the TDR details as possible. Only expose exactly what's needed by calling code.

src/azul/tdr.py
```diff
@@ -162,12 +181,50 @@ def add_entity(self, entity_key, entity_type: str, entity_row: JSON) -> None:
...
 
 
+class TerraDataRepoException(HTTPError):
+
+    def __init__(self, response: urllib3.response.HTTPResponse):
+        super().__init__(response.geturl(), response.status, response.msg, response.headers, None)
+
+
+@lru_cache
+def http() -> AuthorizedHttp:
```
hannes-ucsc, line 191: This shouldn't be so public. I'd prefer it to be a method on the AzulTDRClient class.

hannes-ucsc, line 191:
```suggestion
def oauthed_http() -> AuthorizedHttp:
```

hannes-ucsc: Then you can use our cached_property again.

src/azul/tdr.py
```diff
@@ -46,21 +65,21 @@
...
-        assert service == 'tdr'
-        if target_type == 'snapshot':
-            return cls(project=project, name=target_name, is_snapshot=True)
-        elif target_type == 'dataset':
-            return cls(project=project, name=f'datarepo_{target_name}', is_snapshot=False)
-        else:
-            assert False, target_type
+        assert service == 'tdr', service
+        assert target_type in ('dataset', 'snapshot'), target_type
+        return BigQueryDataset(project=project, tdr_name=target_name, is_snapshot=target_type == 'snapshot')
```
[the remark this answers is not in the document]
noah-aviel-dove: With the reversion to the `if`-based approach, the line is now exactly 80 chars not counting indentation, which I'm guessing removes the need for this change.

src/azul/tdr.py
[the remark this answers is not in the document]
noah-aviel-dove: This commit was inserted here from [another branch](https://github.com/DataBiosphere/azul/tree/issues/noah-aviel-dove/1850-drs-urls-broad-resolver). I'll split it up so that only the needed parts are included here, and your feedback on the other parts will appear in the other branch.

UPGRADING.rst
noah-aviel-dove: Per deployment, yes. I was thinking in terms of the entire team. I'll change to singular since it seems less prone to confusion.

src/azul/tdr.py
noah-aviel-dove: I kept it outside the class because much like `azul.dss.shared_credentials` I wanted it to available in cases where an actual client instance wasn't readily available (keep in mind a valid dataset/snapshot is needed at initialization). Also didn't want to cache on the `self` parameter. So, for now at least, I'm sticking with `lru_cache` and moving to a classmethod.

scripts/register_sam.py
noah-aviel-dove: Didn't occur to me. Moved and refactored a bit so it's purpose is more clear in the new context.

UPGRADING.rst
noah-aviel-dove: Do datasets and snapshots have separate access permissions? If so, shouldn't both be tested?

README.md
```diff
@@ -374,6 +374,20 @@ If you intend to set up a Gitlab instance for CI/CD of your Azul deployments, an
 EBS volume needs to be created as well. See [gitlab.tf.json.template.py] and the 
 [section on CI/CD](#9-continuous-deployment-and-integration) and for details. 
 
+In order to index data stored in the
```
hannes-ucsc, line 377: wrong section

README.md
```diff
@@ -374,6 +374,20 @@ If you intend to set up a Gitlab instance for CI/CD of your Azul deployments, an
...
+[Terra Data Repository](https://jade.datarepo-dev.broadinstitute.org/) (TDR),
+the Google service account provisioned by Terraform must be
+
+1. [Registered](https://github.com/DataBiosphere/jade-data-repo/blob/develop/docs/register-sa-with-sam.md)
```
hannes-ucsc, line 381: TMI

README.md
```diff
@@ -374,6 +374,20 @@ If you intend to set up a Gitlab instance for CI/CD of your Azul deployments, an
...
+ with [SAM](https://github.com/broadinstitute/sam).
+
+2. Granted respoitory read access for datasets and snapshots.
```
hannes-ucsc, line 384: typo

README.md
```diff
@@ -374,6 +374,20 @@ If you intend to set up a Gitlab instance for CI/CD of your Azul deployments, an
...
+
+Step 1 is handled automatically during ``make deploy`` by a helper script.
```
hannes-ucsc, line 386: Listing two steps and then referring to them like this is awkward

src/azul/tdr.py
```diff
@@ -169,6 +186,34 @@ def __init__(self, dataset: BigQueryDataset):
...
+        Raise RequirementError if an authorization error occurs during a simple API call to TDR.
+        """
+        # List snapshots
+        response = cls.oauthed_http().request('GET',
+                                              f'{config.tdr_service_url}/api/repository/v1/snapshots')
+        if response.status == 200:
+            log.info('Google service account is authorized for TDR access.')
+        elif response.status == 401:
+            raise RequirementError('Google service account is not authorized for TDR access.'
+                                   ' Make sure that the SA is registered with SAM and has been'
```
hannes-ucsc, line 212: Nowhere in the code base do we wrap strings before the space like this. Please stick with precedent or propose a change.

src/azul/tdr.py
```diff
@@ -169,6 +186,34 @@ def __init__(self, dataset: BigQueryDataset):
...
+        """
+        with shared_credentials():
+            credentials = Credentials.from_service_account_file(os.environ['GOOGLE_APPLICATION_CREDENTIALS'])
+        return AuthorizedHttp(credentials.with_scopes(['email', 'openid']),
+                              urllib3.PoolManager(ca_certs=certifi.where()))
+
+    @classmethod
+    def verify_authorization(cls) -> None:
+        """
+        Raise RequirementError if an authorization error occurs during a simple API call to TDR.
```
hannes-ucsc, line 203: That's obvious. Document the intent.

src/azul/tdr.py
hannes-ucsc: Didn't like the way this looked in the code and renamed it back.

hannes-ucsc: I don't think you hit the nail on the head though. I've mentioned this before: when you add a static/class method that is not a factory, you are doing something wrong in 90% of the cases. In this case there are some needless duplication, lack of abstraction.

Read https://en.wikipedia.org/wiki/Information_hiding

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -43,9 +43,7 @@ def __init__(self, *args, **kwargs):
 
     @cached_property
     def client(self):
-        import azul.dss
-        with azul.dss.shared_credentials():
```
hannes-ucsc, line 47: This was bugging me like crazy. I think it only worked because the AzulTDRClient constructor made a BQ round-trip, otherwise the AuthorizedHttp instance would have been initialized outside of the context manager which would have failed.

src/azul/dss.py
```diff
@@ -234,7 +234,7 @@ def shared_credentials():
         f.write(creds['SecretString'])
         f.flush()
         with patch.dict(os.environ, GOOGLE_APPLICATION_CREDENTIALS=f.name):
-            yield
+            yield f.name
```
hannes-ucsc, line 237: Note how this avoids the duplication of the env var name.

src/azul/tdr.py
```diff
@@ -162,6 +178,71 @@ def add_entity(self, entity_key, entity_type: str, entity_row: BigQueryRow) -> N
...
 
+class SAMClient:
+    """
+    A client to Broad's SAM (https://github.com/broadinstitute/sam). TDR uses
+    SAM for authorization, and SAM uses Google OAuth for authentication.
+    """
+
+    @cached_property
+    def credentials(self) -> Credentials:
+        with shared_credentials() as file_name:
```
hannes-ucsc, line 189: Note how this is now the only call to the CM.

src/azul/tdr.py
```diff
@@ -162,6 +178,71 @@ def add_entity(self, entity_key, entity_type: str, entity_row: BigQueryRow) -> N
...
+            return Credentials.from_service_account_file(file_name)
+
+    oauth_scopes = ['email', 'openid']
```
hannes-ucsc, line 192: Note how this is deduplicated.

src/azul/tdr.py
```diff
@@ -162,6 +178,71 @@ def add_entity(self, entity_key, entity_type: str, entity_row: BigQueryRow) -> N
...
+        A urllib3 HTTP client with OAuth credentials.
+        """
+        return AuthorizedHttp(self.credentials.with_scopes(self.oauth_scopes),
+                              urllib3.PoolManager(ca_certs=certifi.where()))
+
+    def register_with_sam(self) -> None:
+        """
+        Register the current service account with SAM.
+
+        https://github.com/DataBiosphere/jade-data-repo/blob/develop/docs/register-sa-with-sam.md
```
hannes-ucsc, line 206: The URL you added to the README makes much more sense here.

src/azul/tdr.py
```diff
@@ -162,6 +178,71 @@ def add_entity(self, entity_key, entity_type: str, entity_row: BigQueryRow) -> N
...
+        response = self.oauthed_http.request('POST',
+                                             f'{config.sam_service_url}/register/user/v1',
+                                             body='',
+                                             headers={'Authorization': f'Bearer {token}'})
+        if response.status == 201:
+            log.info('Google service account successfully registered with SAM.')
+        elif response.status == 409:
+            log.info('Google service account previously registered with SAM.')
+        else:
+            raise RuntimeError('Unexpected response during SAM registration', response.data)
```
hannes-ucsc, line 218: Note the 2nd arg instead of f-string. I prefer that in most cases.

src/azul/dss.py
noah-aviel-dove: Nice

Issue #2021, opened 2020-07-15
TDR repository plugin is slow-ish


hannes-ucsc:
Sandbox: TDR plugin

![image](https://user-images.githubusercontent.com/5143256/87610468-43af4c80-c6ba-11ea-96f9-c88b21bf63c1.png)

Sandbox, DSS plugin:

![image](https://user-images.githubusercontent.com/5143256/87610390-1498db00-c6ba-11ea-8af2-293cc6476f80.png)

hannes-ucsc:
We fetch the different tables concurrently.

PR #2030, opened 2020-07-17
Use multithreading when reading TDR bigquery tables (#2021)

noah-aviel-dove:
Currently seeing a 4-5x speedup on fetching one bundle on the initial commit

### DIFF
--- a/environment.py
+++ b/environment.py
@@ -230,9 +230,9 @@ def env() -> Mapping[str, Optional[str]]:
         #
         'AZUL_VERSIONED_BUCKET': 'edu-ucsc-gi-singlecell-azul-config-dev.{AWS_DEFAULT_REGION}',
 
-        # The number of workers pulling files from DSS. There is one such set of DSS
-        # workers per index worker!
-        'AZUL_DSS_WORKERS': '8',
+        # The number of workers pulling files from the repository (DSS or TDR).
+        # There is one such set of repository workers per index worker!
+        'AZUL_REPO_WORKERS': '8',
 
         # Whether to create a subscription to DSS during deployment. Set this
         # variable to 1 to enable `make subscribe` to subscribe the indexer in
--- a/src/azul/tdr.py
+++ b/src/azul/tdr.py
@@ -299,9 +299,9 @@ class AzulTDRClient:
             link_type = link['link_type']
             if link_type == 'process_link':
                 entities[link['process_type']].add(link['process_id'])
-                for catgeory in ('input', 'output', 'protocol'):
-                    for entity_ref in link[catgeory + 's']:
-                        entities[entity_ref[catgeory + '_type']].add(entity_ref[catgeory + '_id'])
+                for category in ('input', 'output', 'protocol'):
+                    for entity_ref in link[category + 's']:
+                        entities[entity_ref[category + '_type']].add(entity_ref[category + '_id'])
             elif link_type == 'supplementary_file_link':
                 # For MVP, only project entities can have associated supplementary files.
                 entity = link['entity']
--- a/src/azul/tdr.py
+++ b/src/azul/tdr.py
@@ -35,6 +35,7 @@ from azul import (
     RequirementError,
     cached_property,
     config,
+    require,
 )
 from azul.bigquery import (
     AbstractBigQueryAdapter,
@@ -290,11 +291,12 @@ class AzulTDRClient:
                 AND version = TIMESTAMP('{bundle_fqid.version}')
         '''))
         links_json = json.loads(links_row['content'])
+        bundle_project_id = links_row['project_id']
         log.info('Retrieved links content, %s top-level links', len(links_json['links']))
         bundler.add_entity('links.json', 'links', links_row)
 
         entities = defaultdict(set)
-        entities['project'].add(links_row['project_id'])
+        entities['project'].add(bundle_project_id)
         for link in links_json['links']:
             link_type = link['link_type']
             if link_type == 'process_link':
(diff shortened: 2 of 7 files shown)


src/azul/tdr.py
```diff
@@ -202,32 +203,37 @@ def emulate_bundle(self,
                 # For MVP, only project entities can have associated supplementary files.
                 entity = link['entity']
                 if entity['entity_type'] != 'project' or entity['entity_id'] != links_row['project_id']:
-                    raise ValueError(f'Supplementary file not associated with bundle project: {entity}')
+                    raise Exception(f'Supplementary file not associated with bundle project: {entity}')
```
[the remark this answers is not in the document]
noah-aviel-dove: `or` should `and` according to DeMorgan's laws, no?

src/azul/tdr.py
```diff
@@ -202,32 +203,37 @@ def emulate_bundle(self,
...
                 for supp_file in link['files']:
                     entities['supplementary_file'].add(supp_file['file_id'])
             else:
-                raise ValueError(f'Unexpected link_type: {link_type}')
+                raise Exception(f'Unexpected link_type: {link_type}')
```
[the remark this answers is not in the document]
noah-aviel-dove: Yes! Thank you, I knew there had to be a better exception to use here but couldn't think of what it was.

src/azul/tdr.py
```diff
@@ -70,17 +71,17 @@ class ManifestEntry(NamedTuple):
     @property
     def entry(self):
         return {
-            "name": self.name,
```
hannes-ucsc, line 73: separate commit

src/azul/tdr.py
```diff
@@ -195,39 +200,44 @@ def emulate_bundle(self,
             link_type = link['link_type']
             if link_type == 'process_link':
                 entities[link['process_type']].add(link['process_id'])
-                for catgeory in ('input', 'output', 'protocol'):
```
hannes-ucsc, line 198: yet another separate commit

src/azul/tdr.py
```diff
@@ -195,39 +200,44 @@ def emulate_bundle(self,
...
             uuid_in_list = ' OR '.join(f'{pk_column} = "{entity_id}"' for entity_id in entity_ids)
             rows = self._query_latest_version(f'''
                 SELECT {columns}
                 FROM {self.target.name}.{entity_type}
                 WHERE {uuid_in_list}
             ''', group_by=pk_column)
             log.info('Retrieved %s %s entities', len(rows), entity_type)
+            return rows
+
+        with ThreadPoolExecutor(max_workers=len(entities)) as executor:
```
hannes-ucsc, line 229: Use a TP with a configurable maximum size like the DSS repo plugin does.

We should probably use the same env variable for both. It may need to be renamed.

src/azul/tdr.py
```diff
@@ -176,14 +177,19 @@ def _choose_one_version(self, versioned_items: Iterable[JSON]) -> JSON:
     def emulate_bundle(self,
                        bundle_fqid: BundleFQID,
                        manifest_only: bool = False) -> Bundle:
+        from azul import (
```
hannes-ucsc, line 180: No local imports if possible, please.

environment.py
```diff
@@ -230,9 +230,9 @@ def env() -> Mapping[str, Optional[str]]:
         #
         'AZUL_VERSIONED_BUCKET': 'edu-ucsc-gi-singlecell-azul-config-dev.{AWS_DEFAULT_REGION}',
 
-        # The number of workers pulling files from DSS. There is one such set of DSS
-        # workers per index worker!
-        'AZUL_DSS_WORKERS': '8',
+        # The number of workers pulling files from the repository (DSS or TDR).
+        # There is one such set of DSS workers per index worker!
```
hannes-ucsc, line 234:
```suggestion
        # There is one such set of repository workers per index worker!
```

src/azul/tdr.py
```diff
@@ -224,12 +226,12 @@ def emulate_bundle(self,
             elif link_type == 'supplementary_file_link':
                 # For MVP, only project entities can have associated supplementary files.
                 entity = link['entity']
-                if entity['entity_type'] != 'project' or entity['entity_id'] != links_row['project_id']:
-                    raise ValueError(f'Supplementary file not associated with bundle project: {entity}')
+                require(entity['entity_type'] == 'project' and entity['entity_id'] == links_row['project_id'],
+                        f'Supplmentary files may only be associated with the bundle project: {entity}')
```
hannes-ucsc, line 234:
```suggestion
                        f'Supplementary files may only be associated with the bundle project: {entity}')
```

hannes-ucsc: I would split this into two calls to require(), one for each side of the `and`. The the error messages can be more specific about the violation. As in `'… must be associated with the current project ({…}, not {…}).'`

src/azul/tdr.py
```diff
@@ -233,24 +236,30 @@ def emulate_bundle(self,
...
+                entity_type: executor.submit(retrieve_rows, entity_type, entity_ids)
+                for entity_type, entity_ids in entities.items()
+            }
+        for entity_type, future in futures.items():
+            rows = future.result()
+            entity_ids = entities[entity_type]
             for i, row in enumerate(rows):
                 bundler.add_entity(f'{entity_type}_{i}.json', entity_type, row)
-                entity_ids.remove(row[pk_column])
+                entity_ids.remove(row[f'{entity_type}_id'])
```
hannes-ucsc, line 263: I think `+` would look better here.

Issue #1693, opened 2020-04-09
Research most recent index failures


hannes-ucsc:
Five messages in total:

[azul-fail-dev.zip](https://github.com/DataBiosphere/azul/files/4457513/azul-fail-dev.zip)

danielsotirhos:
@hannes-ucsc The fails appear to be caused by #1678 which is currently in QA

I was able to match given errors by bundle uuid to the full error messages in azul-indexer-dev-index and all were due to the same issue trying to sort the `organ_part` list when it containied both `str` and `None` type values

```
[WARNING]	2020-04-09T07:33:17.401Z	ee7d2cd9-ee02-5a8d-b5bb-7bd13fe277de	Worker failed to handle message {'action': 'add', 'notification': {'query': {'query': {'bool': {'must_not': [{'term': {'admin_deleted': True}}], 'must': [{'exists': {'field': 'files.project_json'}}, {'prefix': {'uuid': '52'}}]}}}, 'subscription_id': 'cafebabe-feed-4bad-dead-beaf8badf00d', 'transaction_id': 'd53364fb-e233-4beb-8d34-dc4992c8eb8e', 'match': {'bundle_uuid': '52245757-73ba-43a2-a78d-0eba01cb4ce1', 'bundle_version': '2019-09-26T022557.424404Z'}}}.
Traceback (most recent call last):
  File "/var/task/app.py", line 202, in index
    contributions = indexer.transform(notification, delete=False)
  File "/var/task/azul/indexer/__init__.py", line 151, in transform
    metadata_files=metadata_files))
  File "/var/task/azul/project/hca/transformers.py", line 586, in transform
    cell_suspensions=list(map(self._cell_suspension, visitor.cell_suspensions.values())),
  File "/var/task/azul/project/hca/transformers.py", line 259, in _cell_suspension
    'organ_part': sorted(organ_parts)
TypeError: '<' not supported between instances of 'str' and 'NoneType'
```

```
[...]

Issue #2013, opened 2020-07-14
One project failed to be aggregated for 'hca/prod' deployment


achave11:
The following logs …
```logs
[INFO]	2020-07-11T05:58:45.705Z	78135580-57f6-5444-acbe-020300b432f5	Attempt 10 of handling 99460 contribution(s) for entity projects/e0009214-c0a0-4a7b-96e2-d6a83e966ce0
[INFO]	2020-07-11T05:58:45.836Z	78135580-57f6-5444-acbe-020300b432f5	Reading 174593 expected contribution(s) using scan().
REPORT RequestId: 78135580-57f6-5444-acbe-020300b432f5	Duration: 300068.02 ms	Billed Duration: 300000 ms	Memory Size: 2048 MB	Max Memory Used: 2048 MB
2020-07-11T06:03:26.447Z 78135580-57f6-5444-acbe-020300b432f5 Task timed out after 300.07 seconds
```

… indicate that project UUID  `e0009214-c0a0-4a7b-96e2-d6a83e966ce0` tallies in `hca/prod` failed to be aggregated in its totality due to a memory exhaustion failure "... Max Memory Used: 2048 MB"

Backporting #1732 into `hca` deployments could be a solution.

hannes-ucsc:
I don't think #1732 is the solution here. It is about optimizing cost and was fairy complex and invasive. IIRC, we had a ticket that just increased lambda memory for the lambda that's timing out in this case and that's what we need to backport. @achave11 can you try to find that ticket?

achave11:
It seems like the lambda memory upgrade was introduced by the work on ticket #1693 PR #1716. #1732 seems to be follow up work to #1693 such that memory is optimally used.

PR #1716, opened 2020-04-17
Debug indexing assertion errors (#1693)

danielsotirhos:

PR #2032, opened 2020-07-17
Prevent failure due to `write` AWS lambda memory exhaustion (#2013)

achave11:
* Fix obtained from contribution related to ticket (#1693, PR #1716).


achave11:
> I had trouble understanding the body of the commit message. I found the bullet at the beginning confusing because generally commit bodies are paragraph-like text. Also the commit body is not a complete sentence.
> 
> I think it would read more cleanly if rewritten something like "The fix was copied from PR #1716 which resolved #1693."

Thank @jessebrennan, I took your suggestion!

achave11:
@hannes-ucsc Perhaps it was my mistake for trying to create a PR against the `hca/prod` branch, which seems to be the only one presenting this problem. Should this be promoted through the deployments starting with `hca/dev`? That way we can get the consequent Gitlab builds. I am unsure as to how to proceed.

hannes-ucsc:
I'm not sure either but this is a problem you need to solve. The base branch should be `hca/dev`.

achave11:
Invalid PR base, closing.

Issue #2034, opened 2020-07-17
TDR unit test failure due to environment variable


amarjandu:
When the environment variable for `AZUL_TDR_TARGET` is set to `None` failures occur within `test_tdr_dataset_config`
```
Error
Traceback (most recent call last):
  File "/Users/amar/.pyenv/versions/3.8.2/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/Users/amar/.pyenv/versions/3.8.2/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/Users/amar/.pyenv/versions/3.8.2/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/Users/amar/dev/hca/azul/test/service/test_tdr.py", line 92, in test_tdr_dataset_config
    dataset = config.tdr_bigquery_dataset
  File "/Users/amar/dev/hca/azul/src/azul/__init__.py", line 154, in tdr_bigquery_dataset
    service, google_project, target = self.tdr_target.split(':')
  File "/Users/amar/dev/hca/azul/src/azul/__init__.py", line 149, in tdr_target
    return os.environ['AZUL_TDR_TARGET']
  File "/Users/amar/.pyenv/versions/3.8.2/lib/python3.8/os.py", line 675, in __getitem__
    raise KeyError(key) from None
KeyError: 'AZUL_TDR_TARGET'
```
To resolve this issue the value for `AZUL_TDR_TARGET` had to be copied from the `dev` deployment, ideally the unit test would mock this value.

PR #2045, opened 2020-07-20
Mock config property for TDR unit test (#2034)

danielsotirhos:


test/service/test_tdr.py
```diff
@@ -89,8 +89,10 @@ def test(dataset: BigQueryDataset):
             test(BigQueryDataset('test-project', 'name', False))
 
     def test_tdr_dataset_config(self):
-        dataset = config.tdr_bigquery_dataset
-        self.assertNotEqual(dataset.is_snapshot, dataset.name.startswith('datarepo_'))
+        from deployments.sandbox.environment import env as sandbox_env
```
hannes-ucsc, line 92: I wouldn't use a real snapshot. It shouldn't matter if this is a unit test.

Also, you can't import like this. It's a fluke that the project root is on the python path.

test/service/test_tdr.py
```diff
@@ -89,8 +89,10 @@ def test(dataset: BigQueryDataset):
...
+        with mock.patch.object(type(config), 'tdr_target', sandbox_env()['AZUL_TDR_TARGET']):
```
hannes-ucsc, line 93: I don't think this is how you patch a property. This only passes because AZUL_TDR_TARGET is already set on Travis.

I'd like to see a drop! commit that patches this to the wrong value causing the assertion below to fail. That's proof that patch works.

hannes-ucsc: You could also simply patch the environment variable.

test/service/test_tdr.py
```diff
@@ -105,8 +106,11 @@ def test(dataset: BigQueryDataset):
             test(BigQueryDataset(project='test-project', name='name', is_snapshot=False))
 
     def test_tdr_dataset_config(self):
-        dataset = BigQueryDataset.parse(config.tdr_target)
-        self.assertNotEqual(dataset.is_snapshot, dataset.name.startswith('datarepo_'))
+        for target_type in ('snapshot', 'dataset'):
+            with mock.patch.dict(os.environ, AZUL_TDR_TARGET=f'tdr:project_name:{target_type}/target_name'):
+                with self.subTest(target_type=target_type):
+                    dataset = BigQueryDataset.parse(config.tdr_target)
+                    self.assertNotEqual(dataset.is_snapshot, dataset.name.startswith('datarepo_'))
```
hannes-ucsc, line 113: This is a really odd assertion. There is no guarantee that is_snapshot is True in one sub-test case and False in the other.

test/service/test_tdr.py
```diff
@@ -105,8 +106,11 @@ def test(dataset: BigQueryDataset):
             test(BigQueryDataset(project='test-project', name='name', is_snapshot=False))
 
     def test_tdr_dataset_config(self):
-        dataset = BigQueryDataset.parse(config.tdr_target)
-        self.assertNotEqual(dataset.is_snapshot, dataset.name.startswith('datarepo_'))
+        for target_type in ('snapshot', 'dataset'):
+            with mock.patch.dict(os.environ, AZUL_TDR_TARGET=f'tdr:project_name:{target_type}/target_name'):
+                with self.subTest(target_type=target_type):
+                    dataset = BigQueryDataset.parse(config.tdr_target)
```
hannes-ucsc, line 112: The elaborate mocking above is only to make config.tdr_target return a particular value. It seems like overkill.

danielsotirhos:
Closing PR for alternate solution PR #2088

Issue #1802, opened 2020-05-14
TestHCAIndexer.test_deletion indexes two bundles under same FQID


hannes-ucsc:
If it is intentional it should be explained in a comment, otherwise it should be fixed.

melainalegaspi:
@hannes-ucsc : "There's a FIXME in test_hca_indexer.py"

PR #2046, opened 2020-07-21
Fix test indexing bundles using same FQID (#1802)

danielsotirhos:


hannes-ucsc:
I hope you got this right. I'm surprised that the solution is so trivial.

Issue #2060, opened 2020-07-23
Type check warning in `tdr.py` and `test_tdr.py`


hannes-ucsc:
![image](https://user-images.githubusercontent.com/5143256/88310193-3864ae00-ccc4-11ea-9f20-684fcc56dcab.png)

![image](https://user-images.githubusercontent.com/5143256/88310417-811c6700-ccc4-11ea-8f97-52ea2c518f34.png)

PR #2065, opened 2020-07-23
Define BigQuery types (#2060)

noah-aviel-dove:


test/service/test_tdr.py
```diff
@@ -357,7 +359,7 @@ def _make_mock_entity_table(self,
                                                                **additional_columns),
                                         rows=rows)
 
-    _bq_types = {
+    _bq_types: Dict[type, str] = {
         str: 'STRING',
```
amarjandu, line 363: nit: alpha sort keys.

(ps. its a super nit)

test/service/test_tdr.py
```diff
@@ -369,7 +371,7 @@ def _bq_schema(self, fields: Mapping[str, Union[str, type]] = frozendict(), **kw
         kwargs.update(fields)
         return [
             dict(name=k,
-                 type=self._bq_types.get(v, v),
+                 type=(self._bq_types[v] if isinstance(v, type) else v),
```
amarjandu, line 374: were the outer parenthesis intentional here?

test/service/test_tdr.py
```diff
@@ -357,7 +359,7 @@ def _make_mock_entity_table(self,
                                                                **additional_columns),
                                         rows=rows)
 
-    _bq_types = {
+    _bq_types: Dict[type, str] = {
         str: 'STRING',
         bytes: 'BYTES',
         int: 'INTEGER',
```
amarjandu, line 365: is `DATETIME` (or `TIMESTAMP`) and `NULL` intentionally missing here?  `datetime` and `None` are present within `BigQueryValue`.

test/service/test_tdr.py
noah-aviel-dove: Yes, I acknowledge that they are semantically superfluous but my personal experience is that they make the line easier to visually scan.

test/service/test_tdr.py
noah-aviel-dove: Nullable columns allow `NULL`/`None` as a value but it makes no sense to define a column with data type `NULL`.

noah-aviel-dove: I intentionally omitted the time-related types because the python types are only supported as shorthand (`type=int` is translated to `type="INTEGER"`) and supporting this for non-primitive types didn't feel worth the hassle of importing them. If the inconsistency is considered a problem I'd opt to remove this dictionary altogether.

noah-aviel-dove: @hannes-ucsc will probably want to weigh in on this.

hannes-ucsc: `datetime` is already being imported. And there is one reference to a TIMESTAMP column IIRC.

test/service/test_tdr.py
```diff
@@ -373,19 +373,19 @@ def _make_mock_entity_table(self,
...
+        bool: 'BOOL',
         bytes: 'BYTES',
-        int: 'INTEGER',
         float: 'FLOAT',
-        bool: 'BOOL'
+        int: 'INTEGER',
+        str: 'STRING'
     }
 
     def _bq_schema(self, fields: Mapping[str, Union[str, type]] = frozendict(), **kwargs: Union[str, type]) -> JSONs:
```
hannes-ucsc, line 384: line too long

hannes-ucsc: not a fan of the complex signature. doesn't buy us much

hannes-ucsc:
For demo, just show that both files (`tdr.py` and `test_tdr.py`) are free of warnings.

Issue #2064, opened 2020-07-23
Terra handover fails with invalid characters in bundle version


jessebrennan:
Here's the error as reported in Terra.
```
{
  "statusCode": 400,
  "source": "rawls",
  "timestamp": 1595530086136,
  "causes": [],
  "stackTrace": [],
  "message": "Invalid input: a0ca0624-ddf5-4b80-8ba6-a481002731be.2019-05-14T10:50:20.494000Z. Input may only contain alphanumeric characters, underscores, dashes, and periods."
}
```

For this export I only had the "Single-cell RNA-seq analysis of human pancreas from healthy individuals and type 2 diabetes patients" project selected.

jessebrennan:
![terra-broke](https://user-images.githubusercontent.com/10734740/88328408-3e1abd80-ccdd-11ea-9f9d-e8fe715bdbb3.png)

PR #2067, opened 2020-07-24
Remove `:` from versions in BDBag manifest (#2064)

jessebrennan:
~I wasn't sure how fine grained of an approach to take so I just stripped out all `:` in all regex matches that look like a bundle version.~

~I intended to recreate the problem on my deployment and add a test that recreates this, but I realized that this involved indexing data in TDR, so I've postponed. I believe this fix will work, but it's not a very precise solution.~

### DIFF
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -844,7 +844,8 @@ class BDBagManifestGenerator(FileBasedManifestGenerator):
             # For each bundle containing the current file …
             doc_bundle: JSON
             for doc_bundle in doc['bundles']:
-                bundle_fqid: FQID = (doc_bundle['uuid'], doc_bundle['version'])
+                # Versions indexed by TDR contain ':', but Terra won't allow ':' in the 'entity:participant_id' field
+                bundle_fqid: FQID = doc_bundle['uuid'], doc_bundle['version'].replace(':', '')
 
                 bundle_cells = {'entity:participant_id': '.'.join(bundle_fqid)}
                 self._extract_fields([doc_bundle], bundle_column_mapping, bundle_cells)
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -844,7 +844,8 @@ class BDBagManifestGenerator(FileBasedManifestGenerator):
             # For each bundle containing the current file …
             doc_bundle: JSON
             for doc_bundle in doc['bundles']:
-                # Versions indexed by TDR contain ':', but Terra won't allow ':' in the 'entity:participant_id' field
+                # Versions indexed by TDR contain ':', but Terra won't allow ':'
+                # in the 'entity:participant_id' field
                 bundle_fqid: FQID = doc_bundle['uuid'], doc_bundle['version'].replace(':', '')
 
                 bundle_cells = {'entity:participant_id': '.'.join(bundle_fqid)}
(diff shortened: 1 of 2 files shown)


jessebrennan:
Yes. I tested this to be sure, but other fields contain URLs too which
definitely have ':'s in them.

src/azul/service/manifest_service.py
```diff
@@ -919,4 +919,27 @@ def qualify(qualifier, column_name, index=None):
                             assert False
             # Join concatenated values using the joiner
             row = {k: self.column_joiner.join(sorted(v)) if isinstance(v, set) else v for k, v in row.items()}
+            sanitize_participant_id(row)
             bundle_tsv_writer.writerow(row)
+
+
+def sanitize_participant_id(row: MutableMapping[str, str]):
+    """
```
hannes-ucsc, line 927: I think this doctest is overkill.

src/azul/service/manifest_service.py
```diff
@@ -919,4 +919,27 @@ def qualify(qualifier, column_name, index=None):
...
+    >>> sanitize_participant_id(d)
+    >>> from azul.doctests import assert_json
+    >>> assert_json(d)
+    {
+        "entity:participant_id": "587d74b4-1075-4bbf-b96a-4d1ede0481b2.2018-09-14T133314.453337Z",
+        "bundle_uuid": "587d74b4-1075-4bbf-b96a-4d1ede0481b2",
+        "bundle_version": "2018-09-14T13:33:14.453337Z"
+    }
+    """
+    pieces = row['entity:participant_id'].split(':')
```
hannes-ucsc, line 944: Can't you use str.replace()? Reviewer should have caught this.

src/azul/service/manifest_service.py
```diff
@@ -919,4 +919,27 @@ def qualify(qualifier, column_name, index=None):
                             assert False
             # Join concatenated values using the joiner
             row = {k: self.column_joiner.join(sorted(v)) if isinstance(v, set) else v for k, v in row.items()}
+            sanitize_participant_id(row)
```
hannes-ucsc, line 922: Don't like the "patch" after the fact approach and that it would even remove colons from the UUID part. This should be more surgical in only translating the version part and it should explain where the colons come from. It should also be applied where the row is composed, not after everything is done.

jessebrennan:
@hannes-ucsc and I discussed why adding a unit test would be more complicated. It doesn't just involve switching the catalog, it also would require creating a new canned bundle. Hannes deemed the change to the integration test sufficient.

src/azul/service/manifest_service.py
jessebrennan: I've removed this whole function.

hannes-ucsc:
Don't know why I added `sandbox` label.

src/azul/service/manifest_service.py
```diff
@@ -836,7 +836,9 @@ def _samples_tsv(self, bundle_tsv: IO[str]) -> None:
             # For each bundle containing the current file …
             doc_bundle: JSON
             for doc_bundle in doc['bundles']:
-                bundle_fqid: FQID = (doc_bundle['uuid'], doc_bundle['version'])
+                # Versions indexed by TDR contain ':', but Terra won't allow ':'
```
hannes-ucsc, line 839: line was too long

test/integration_test.py
```diff
@@ -310,9 +310,13 @@ def _check_terra_bdbag(self, response: bytes):
             data_path = os.path.join(os.path.dirname(first(zip_fh.namelist())), 'data')
             file_path = os.path.join(data_path, 'participants.tsv')
             with zip_fh.open(file_path) as file:
-                self.__check_manifest(file, 'bundle_uuid')
+                rows = self.__check_manifest(file, 'bundle_uuid')
+                for row in rows:
```
hannes-ucsc, line 314: I don't like modal args in general but I also didn't like how in your verison the control flow forks for bdbag and compact and then rejoins and then forks again.

hannes-ucsc:
For demo, show Terra handover working.

Issue #2051, opened 2020-07-22
Add `cell_suspension.biomaterial_id` to terra.bdbag


hannes-ucsc:
In order to be able to at least have a shot at joining or comparing DCP-generated matrices with contributor generated ones, the DCP-generated matrices should contain the contributor-allocated `biomaterial_id` of the `cell_suspension` entity in addition to its document_id (an ingest-allocated UUID). 

To enable the DCP workflows to add that label to the matrices, we should add a `cell_suspension__biomaterial_core__biomaterial_id` column to the `terra.bdbag` format.

cc: @kbergin

PR #2079, opened 2020-07-28
Add cell_suspension.biomaterial_id to terra.bdbag (#2051)

danielsotirhos:


hannes-ucsc:
For demo, just show presence of column in `terra.bdbag` manifest.

PR #2088, opened 2020-07-29
Add doctest to BigQueryDataset.parse (#2034)

danielsotirhos:


danielsotirhos:
This PR is an alternate solution to #2034. Instead of mocking the config property `tdr_target` for the test `test_tdr_dataset_config` (as done in the first solution PR #2045), this PR removes `test_tdr_dataset_config` and ensures functionality of `BigQueryDataset.parse()` through doctests.

src/azul/tdr.py
```diff
@@ -50,17 +50,39 @@ class BigQueryDataset:
...
+
+        >>> BigQueryDataset.parse('tdr:my_project:package/package1')
+        Traceback (most recent call last):
+        ...
+        AssertionError: Invalid target type: 'package'
+        """
+        service, project, target = tdr_target.split(':')
         target_type, target_name = target.split('/')
-        assert service == 'tdr'
+        assert service == 'tdr', f'{service} != tdr'
```
hannes-ucsc: I think we talked about this before. For assertions, this is overkill. When the assertion fires, the trace back will include this very source line. Assertions are for us, not the user so it's ok if the output is terse and the source code is need to make sense of them.

src/azul/tdr.py
```diff
@@ -50,17 +50,39 @@ class BigQueryDataset:
...
         if target_type == 'snapshot':
             return cls(project=project, name=target_name, is_snapshot=True)
         elif target_type == 'dataset':
             return cls(project=project, name=f'datarepo_{target_name}', is_snapshot=False)
         else:
-            assert False, target_type
+            assert False, f"Invalid target type: '{target_type}'"
```
hannes-ucsc:
```suggestion
            assert False, f'Invalid target type: {target_type!r}'
```

See contributions guide. Then again, see above.

Issue #2108, opened 2020-08-03
DSS adapter fails on prefixes starting with 0s


danielsotirhos:
Due to a bug in `get_prefix_list()`, the DSS Adapter will process an incorrect list of bundles when given a prefix starting with `0`. The problem stems from the prefix list being generated using the `hex()` function which ignores prefixed `0`'s

```
>>> from scripts.dss_v2_apdapter import DSSv2Adapter

# Correct examples:
>>> DSSv2Adapter.get_prefix_list(DSSv2Adapter, 'a')
['a']
>>> DSSv2Adapter.get_prefix_list(DSSv2Adapter, 'aa')
['aa']
>>> DSSv2Adapter.get_prefix_list(DSSv2Adapter, '0')
['0']

# Incorrect examples:
>>> DSSv2Adapter.get_prefix_list(DSSv2Adapter, '00')
['0']
>>> DSSv2Adapter.get_prefix_list(DSSv2Adapter, '00a')
['a']
```

PR #2111, opened 2020-08-04
Fix: DSS adapter fails on prefixes starting with 0s (#2108)

danielsotirhos:


hannes-ucsc:
Bypassing sandbox since this does not touch production code or code that tests production code.

PR #2116, opened 2020-08-05
Verify BigQuery access during deployment (#2025)

noah-aviel-dove:


noah-aviel-dove:
I'll add this change as a separate (non-fixup) commit. I'd also be open to any ideas on how to speed up the BigQuery check. Unfortunately the "Forbidden" exception isn't raised until the `result()` method is called. Getting table or dataset metadata would probably be faster than queuing a job to retrieve actual data [but this involves a separate set of permissions](https://cloud.google.com/bigquery/docs/access-control).

noah-aviel-dove:
Correction, simply calling `.result()` in insufficient as his only returns an iterable. The actual rows must be accessed to trigger the exception.

UPGRADING.rst
```diff
@@ -15,7 +15,12 @@ have too many entries in this file.
 
 This PR introduces two new deployment-specific environment variables,
 ``AZUL_TDR_SERVICE_URL`` and ``AZUL_SAM_SERVICE_URL``. Copy the settings for
-these variables from the sandbox deployment to your personal deployment.
+these variables from the example deployment to your personal deployment.
+
+Service accounts must be registered and authorized with SAM for integration
+tests to pass. See the last two paragraphs of section 3.2 of the README for
```
hannes-ucsc, line 21: Make subsection in README and hyperlink to it if possible.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -221,6 +225,18 @@ def retrieve_rows(entity_type: str, entity_ids: Set[str]):
...
 
+    def verify_authorization(self):
+        """
+        Verify that the current service account is authorized to read from the
+        TDR BigQuery tables.
+        """
+        try:
+            one(self._run_sql(f'SELECT * FROM {self._source.bq_name}.links LIMIT 1'))
+        except Forbidden:
+            tdr.on_auth_failure('the TDR BigQuery tables')
```
hannes-ucsc, line 236: The message is an implementation detail of on_auth_failure. Whether the string fits is brittle. Replace with boolean `bigquery` argument, True here, False below.

src/azul/tdr.py
```diff
@@ -139,24 +139,28 @@ def get_access_token(self) -> str:
         return credentials.token
 
 
+def on_auth_failure(resource):
+    raise RequirementError(f'Google service account is not authorized to access {resource}. '
```
hannes-ucsc, line 143: Should say which SA.

hannes-ucsc:
And the other two commit should refer to respective issue.

UPGRADING.rst
noah-aviel-dove: Don't think there's any way to link to .md from .rst

UPGRADING.rst
```diff
@@ -10,6 +10,16 @@ branch that does not have the listed changes, the steps would need to be
...
 
+#2025 Register indexer SA with Broad's SAM during deployment
+============================================================
+
+This PR introduces two new deployment-specific environment variables,
+``AZUL_TDR_SERVICE_URL`` and ``AZUL_SAM_SERVICE_URL``. Copy the settings for
+these variables from the example deployment to your personal deployment.
+
+Service accounts must be registered and authorized with SAM for integration
+tests to pass. See section 3.2.1 of the README for resgistration instructions.
```
hannes-ucsc, line 21: hyperlink (something that can be clicked) 

The link should be relative https://github.blog/2013-01-31-relative-links-in-markup-files/ and deep (point at the section heading).

hannes-ucsc, line 21:
```suggestion
tests to pass. See section 3.2.1 of the README for registration instructions.
```

noah-aviel-dove: `UPGRADING.rst` is reStructuredText, not markup. The syntax in your link does not work. I can create a relative link to the README using a different syntax, but I see no way to create a _deep_ link, as rst does not understand markup document structure.

hannes-ucsc: Think this over one more time. The concept of relative links and deep links are HTTP concepts and have nothing to to with how the document is marked up. If you still can't figure it out, I'll do it during the review.

UPGRADING.rst
```diff
@@ -18,7 +18,7 @@ This PR introduces two new deployment-specific environment variables,
 these variables from the example deployment to your personal deployment.
 
 Service accounts must be registered and authorized with SAM for integration
-tests to pass. See section 3.2.1 of the README for resgistration instructions.
+tests to pass. See section 3.2.1 of the README for registration instructions.
```
hannes-ucsc, line 21: The fix for `resgistration` shouldn't be in the commit it is currently in but in a fixup on an earlier commit.

UPGRADING.rst
```diff
@@ -10,6 +10,18 @@ branch that does not have the listed changes, the steps would need to be
...
+============================================================
+
+This PR introduces two new deployment-specific environment variables,
+``AZUL_TDR_SERVICE_URL`` and ``AZUL_SAM_SERVICE_URL``. Copy the settings for
+these variables from the example deployment to your personal deployment.
+
+Service accounts must be registered and authorized with SAM for integration
+tests to pass. See `section 3.2.1`_ of the README for registration instructions.
+
+.. _section 3.2.1: ./README.md#321-tdr-and-sam
```
hannes-ucsc, line 23: Note the link. Relative and deep.

Issue #1995, opened 2020-07-09
Service returns 500 status in deployments with newly created indices


hannes-ucsc:
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/4201 was the first build with the new v2 index names. The `deploy` stage created the indices but apparently they are missing mapping information and the service returns a 500 responses:

```
hannes@hannes-xps:~$ http https://service.dev.singlecell.gi.ucsc.edu/index/projects
HTTP/1.1 500 Internal Server Error
Access-Control-Allow-Headers: Authorization,Content-Type,X-Amz-Date,X-Amz-Security-Token,X-Api-Key
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 1955
Content-Type: text/plain
Date: Thu, 09 Jul 2020 17:17:06 GMT
Via: 1.1 cfdda2377366ad140c308c7f141eb744.cloudfront.net (CloudFront)
X-Amz-Cf-Id: FKBf751EeRO4yajjRZA0KXWfo1XQI10IZlc3NZVv33x91mVA4mj3Og==
X-Amz-Cf-Pop: LAX3-C3
X-Amzn-Trace-Id: Root=1-5f075112-ccec13f68ba96b98f7ed69cf;Sampled=0
X-Cache: Error from cloudfront
x-amz-apigw-id: Pama9Gj2oAMF_Ug=
x-amzn-RequestId: b87931bf-2c32-4347-80ce-16390115c763

Traceback (most recent call last):
  File "/var/task/chalice/app.py", line 1112, in _get_view_function_response
    response = view_function(**function_args)
  File "/var/task/app.py", line 1041, in get_project_data
    return repository_search('projects', project_id)
  File "/var/task/app.py", line 794, in repository_search
    return service.get_data(catalog, entity_type, pagination, filters, item_id, file_url)
  File "/var/task/azul/service/repository_service.py", line 86, in get_data
    return self._get_items(catalog, entity_type, pagination, filters, file_url_func)
  File "/var/task/azul/service/repository_service.py", line 62, in _get_items
    response = self._get_data(catalog, entity_type, pagination, filters, file_url_func)
  File "/var/task/azul/service/repository_service.py", line 37, in _get_data
    response = self.transform_request(catalog=catalog,
  File "/var/task/azul/service/elasticsearch_service.py", line 537, in transform_request
    es_response = es_search.execute(ignore_cache=True)
[...]

PR #2119, opened 2020-08-05
Return empty 200 service response for unpopulated ES indices (#1995)

achave11:


src/azul/service/elasticsearch_service.py
```diff
@@ -587,6 +587,9 @@ def transform_request(self,
                 es_response = es_search.execute(ignore_cache=True)
             except elasticsearch.NotFoundError as e:
                 raise IndexNotFoundError(e.info["error"]["index"])
+            except elasticsearch.RequestError:
+                final_response = FileSearchResponse({}, {}, {}, entity_type)
```
danielsotirhos, line 591:
```suggestion
                final_response = FileSearchResponse(hits={},
                                                    pagination={},
                                                    facets={},
                                                    entity_type=entity_type)
```
Just because (almost) everywhere else a `FileSearchResponse` is constructed the keyword arguments are specified.

src/azul/service/elasticsearch_service.py
```diff
@@ -587,6 +587,12 @@ def transform_request(self,
                 es_response = es_search.execute(ignore_cache=True)
             except elasticsearch.NotFoundError as e:
                 raise IndexNotFoundError(e.info["error"]["index"])
+            except elasticsearch.RequestError:
```
hannes-ucsc, line 590: This needs to be more specific to the type of error we're expecting in the case of an empty index.

achave11: I might be misunderstanding the context in which error type is expressed here but `RequestError` is the type of error for an empty index.

src/azul/service/elasticsearch_service.py
```diff
@@ -587,6 +587,15 @@ def transform_request(self,
                 es_response = es_search.execute(ignore_cache=True)
             except elasticsearch.NotFoundError as e:
                 raise IndexNotFoundError(e.info["error"]["index"])
+            except elasticsearch.RequestError as e:
+                if one(e.info['error']['root_cause'])['reason'].startswith('No mapping found for'):
+                    final_response = FileSearchResponse(hits={},
```
hannes-ucsc, line 592: I would like to add a round-trip to ES here, the `_count` API to confirm that the index exists but is indeed empty.

hannes-ucsc: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-count.html

test/service/test_response.py
```diff
@@ -1589,6 +1589,30 @@ def test_summary_filter_none(self):
...
+        super().setUpClass()
+        cls.index_service.create_indices(cls.catalog)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.index_service.delete_indices(cls.catalog)
+        super().tearDownClass()
+
+    def test_empty_response(self):
+        for entity_type in 'projects', 'samples', 'files', 'bundles':
```
hannes-ucsc, line 1605: don't hard-code we have a list of those somewhere

src/azul/service/elasticsearch_service.py
```diff
@@ -586,6 +586,16 @@ def transform_request(self,
                 es_response = es_search.execute(ignore_cache=True)
             except elasticsearch.NotFoundError as e:
                 raise IndexNotFoundError(e.info["error"]["index"])
+            except elasticsearch.RequestError as e:
+                if one(e.info['error']['root_cause'])['reason'].startswith('No mapping found for'):
+                    # Index exists but it is empty
+                    if es_search.count() == 0:
```
hannes-ucsc, line 592: If you look at the implemetaion of .count, it reuses the previous response if it exists. Worse though, even if it round-trips again, it reuses the same query from the original request so this may be conflating "no documents match the filter" with "the index is completely empty".

test/service/test_response.py
hannes-ucsc: In fact, I don't see the point of repating this for every entity type. Pick one.

test/service/test_response.py
```diff
@@ -1589,6 +1589,30 @@ def test_summary_filter_none(self):
...
+            with self.subTest(entity_type=entity_type):
+                url = self.base_url + "/index/" + entity_type
+                response = requests.get(url)
+                response.raise_for_status()
+                response_json = response.json()
```
hannes-ucsc, line 1610:
```suggestion
                response = response.json()
```

test/service/test_response.py
```diff
@@ -1589,6 +1589,30 @@ def test_summary_filter_none(self):
...
+                self.assertEqual(response_json['hits'], [])
+                self.assertEqual({v for v in response_json['pagination'].values()}, {None})
```
hannes-ucsc, line 1612: no comprehension necessary

test/service/test_response.py
```diff
@@ -1589,6 +1589,30 @@ def test_summary_filter_none(self):
...
+                self.assertEqual(response_json['termFacets'], {})
```
hannes-ucsc, line 1613: expected value should be 1st argument to assert methods here and everywhere

src/azul/service/elasticsearch_service.py
```diff
@@ -586,6 +589,18 @@ def transform_request(self,
                 es_response = es_search.execute(ignore_cache=True)
             except elasticsearch.NotFoundError as e:
                 raise IndexNotFoundError(e.info["error"]["index"])
+            except elasticsearch.RequestError as e:
+                if one(e.info['error']['root_cause'])['reason'].startswith('No mapping found for'):
+                    es_response = self.es_client.count(index=config.es_index_name(catalog=catalog,
+                                                                                  entity_type=entity_type,
+                                                                                  aggregate=True))
+                    if not es_response['count']:  # Count is zero for empty index
```
hannes-ucsc, line 597: conflates `None` and `0`

src/azul/service/elasticsearch_service.py
```diff
@@ -586,6 +589,18 @@ def transform_request(self,
...
+                        final_response = FileSearchResponse(hits={},
```
hannes-ucsc, line 598: reassign to es_response

hannes-ucsc (changes requested):

src/azul/service/elasticsearch_service.py
achave11: I was using precedence for the assignation, as used in the non empty response.

hannes-ucsc: I don't understand.

hannes-ucsc (changes requested):
Please thumb-up those of my comments that you addressed. See CONTRIBUTING.rst.

I don't understand one of your responses to my comments. Please word your response differently.

src/azul/service/elasticsearch_service.py
achave11: At the end of `transform_request` the variable `final_response` is used to store the outgoing request's response. I was using that as precedence for naming the variable you requested I change.
It feels odd to use `es_response` to store the request's response. I think it is because `es_response` is used internally to store the response coming from Elasticsearch, which isn't representative of an outgoing response. But if you prefer it that way, I have no objection.

achave11: @hannes-ucsc: "I now understand, go back to `final_response`."

hannes-ucsc (approved):

hannes-ucsc:
For demo attempt to reproduce issue #1995 after `make delete create`.

Issue #1589, opened 2020-02-19
SQS processing code in indexer/app.py is uncovered


jessebrennan:
Currently this code can only be debugged via log messages.

PR #2121, opened 2020-08-06
Add testing for index_controller (#1589)

amarjandu:

- [x] Sanity-checked history
- [x] Build passes in `sandbox` <sub>or commented that sandbox will be skipped</sub>

- [ ] Moved freebies to Merged column <sub>or there are no freebies in this PR</sub> 
- [ ] Pushed merge commit to Gitlab

### DIFF
--- a/src/azul/queues.py
+++ b/src/azul/queues.py
@@ -79,16 +79,23 @@ class Queues:
 
     def _dump(self, queue, path):
         logger.info('Writing messages from queue "%s" to file "%s"', queue.url, path)
+        messages = self._get_messages(queue)
+        self._dump_messages(messages, queue.url, path)
+        logger.info(f'Finished writing {path !r}')
+
+    def _get_messages(self, queue):
         messages = []
         while True:
             message_batch = queue.receive_messages(AttributeNames=['All'],
                                                    MaxNumberOfMessages=10,
                                                    VisibilityTimeout=300)
             if not message_batch:  # Nothing left in queue
-                break
+                return messages
             else:
                 messages.extend(message_batch)
-        self._dump_messages(messages, queue.url, path)
+
+    def read_messages(self, queue):
+        messages = self._get_messages(queue)
         message_batches = list(more_itertools.chunked(messages, 10))
         if self._delete:
             logger.info('Removing messages from queue "%s"', queue.url)
@@ -96,7 +103,7 @@ class Queues:
         else:
             logger.info('Returning messages to queue "%s"', queue.url)
             self._return_messages(message_batches, queue)
-        logger.info(f'Finished writing {path !r}')
+        return messages
 
     def _dump_messages(self, messages, queue_url, path):
         messages = [self._condense(message) for message in messages]
--- a/src/azul/queues.py
+++ b/src/azul/queues.py
@@ -78,10 +78,10 @@ class Queues:
             self._dump(queue, queue_name + '.json')
 
     def _dump(self, queue, path):
-        logger.info('Writing messages from queue "%s" to file "%s"', queue.url, path)
+        logger.info('Writing messages from queue %r to file %r', queue.url, path)
         messages = self._get_messages(queue)
         self._dump_messages(messages, queue.url, path)
-        logger.info(f'Finished writing {path !r}')
+        logger.info(f'Finished writing {path!r}')
 
     def _get_messages(self, queue):
         messages = []
@@ -98,10 +98,10 @@ class Queues:
         messages = self._get_messages(queue)
         message_batches = list(more_itertools.chunked(messages, 10))
         if self._delete:
-            logger.info('Removing messages from queue "%s"', queue.url)
+            logger.info('Removing messages from queue %r', queue.url)
             self._delete_messages(message_batches, queue)
         else:
-            logger.info('Returning messages to queue "%s"', queue.url)
+            logger.info('Returning messages to queue %r', queue.url)
             self._return_messages(message_batches, queue)
         return messages
 
@@ -294,14 +294,14 @@ class Queues:
             orig_queue = content['queue']
             messages = content['messages']
         queue = self.sqs.get_queue_by_name(QueueName=queue_name)
-        logger.info('Writing messages from file "%s" to queue "%s"', path, queue.url)
+        logger.info('Writing messages from file %r to queue %r', path, queue.url)
         if orig_queue != queue.url:
             if force:
-                logger.warning('Messages originating from queue "%s" are being fed into queue "%s"',
+                logger.warning('Messages originating from queue %r are being fed into queue %r',
                                orig_queue, queue.url)
             else:
-                raise RuntimeError(f'Cannot feed messages originating from "{orig_queue}" to "{queue.url}". '
-                                   f'Use --force to override')
+                raise RuntimeError(f'Cannot feed messages originating from {orig_queue!r} to {queue.url!r}. '
+                                   f'Use --force to override.')
         message_batches = list(more_itertools.chunked(messages, 10))
 
         def _cleanup():
@@ -311,7 +311,7 @@ class Queues:
                     self._dump_messages(messages, orig_queue, path)
                 else:
                     assert len(remaining_messages) == len(messages)
-                    logger.info('No messages were submitted, not touching local file "%s"', path)
+                    logger.info('No messages were submitted, not touching local file %r', path)
 
         while message_batches:
             message_batch = message_batches[0]
@@ -328,7 +328,7 @@ class Queues:
             if message_batches:
                 _cleanup()
             else:
-                logger.info('All messages were submitted, removing local file "%s"', path)
+                logger.info('All messages were submitted, removing local file %r', path)
                 os.unlink(path)
 
     def purge(self, queue_name):
@@ -349,7 +349,7 @@ class Queues:
             self._handle_futures(futures)
 
     def _purge_queue(self, queue: Queue):
-        logger.info('Purging queue "%s"', queue.url)
+        logger.info('Purging queue %r', queue.url)
         queue.purge()
         self._wait_for_queue_empty(queue)
 
@@ -358,7 +358,7 @@ class Queues:
             num_inflight_messages = int(queue.attributes['ApproximateNumberOfMessagesNotVisible'])
             if num_inflight_messages == 0:
                 break
-            logger.info('Queue "%s" has %i in-flight messages', queue.url, num_inflight_messages)
+            logger.info('Queue %r has %i in-flight messages', queue.url, num_inflight_messages)
             time.sleep(3)
             queue.reload()
 
@@ -369,7 +369,7 @@ class Queues:
             num_messages = sum(map(int, map(queue.attributes.get, attribute_names)))
             if num_messages == 0:
                 break
-            logger.info('Queue "%s" still has %i messages', queue.url, num_messages)
+            logger.info('Queue %r still has %i messages', queue.url, num_messages)
             time.sleep(3)
             queue.reload()
 
@@ -380,14 +380,14 @@ class Queues:
         mapping = one(response['EventSourceMappings'])
 
         def update_():
-            logger.info('%s push from "%s" to lambda function "%s"',
+            logger.info('%s push from %r to lambda function %r',
                         'Enabling' if enable else 'Disabling', queue.url, function_name)
             lambda_.update_event_source_mapping(UUID=mapping['UUID'],
                                                 Enabled=enable)
 
         while True:
             state = mapping['State']
-            logger.info('Push from "%s" to lambda function "%s" is in state "%s".',
+            logger.info('Push from %r to lambda function %r is in state %r.',
                         queue.url, function_name, state)
             if state in ('Disabling', 'Enabling', 'Updating'):
                 pass
--- a/test/indexer/test_indexer_controller.py
+++ b/test/indexer/test_indexer_controller.py
@@ -1,0 +1,174 @@
+import json
+import logging
+from typing import (
+    List,
+    Sequence,
+)
+
+import boto3
+from chalice.app import (
+    SQSRecord,
+)
+import mock
+from moto import (
+    mock_sqs,
+    mock_sts,
+)
+
+from azul import (
+    config,
+    queues,
+)
+from azul.azulclient import (
+    AzulClient,
+)
+from azul.indexer import (
+    BundleFQID,
+)
+from azul.indexer.document import (
+    Contribution,
+)
+from azul.indexer.index_controller import (
+    IndexController,
+)
+from azul.indexer.index_service import (
+    CataloguedTallies,
+)
+from azul.logging import (
+    configure_test_logging,
+)
+from indexer import (
+    ForcedRefreshIndexService,
+    IndexerTestCase,
+)
+
+log = logging.getLogger(__name__)
+
+
+# noinspection PyPep8Naming
+def setUpModule():
+    configure_test_logging(log)
+
+
+Contributions = Sequence[Contribution]
+
+
+@mock_sts
+@mock_sqs
+class TestIndexController(IndexerTestCase):
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.index_service.create_indices(self.catalog)
+        self.client = AzulClient()
+        self.controller = IndexController()
+        self.queue_manager = queues.Queues(delete=True)
+
+    def tearDown(self):
+        self.index_service.delete_indices(self.catalog)
+        super().tearDown()
+
+    def _create_mock_queues(self):
+        sqs = boto3.resource('sqs')
+        for queue_name in config.all_queue_names:
+            sqs.create_queue(QueueName=queue_name)
+
+    def _mock_sqs_record(self, **body):
+        event_dict = {
+            'body': json.dumps(body),
+            'receiptHandle': 'ThisWasARandomString',
+            'attributes': {'ApproximateReceiveCount': 1}
+        }
+        return SQSRecord(event_dict=event_dict, context='controller_test')
+
+    def test_contribution_action_invalid(self):
+        event = [
+            self._mock_sqs_record(action='banana',
+                                  notification='slug',
+                                  catalog=self.catalog)
+        ]
+        self.assertRaises(AssertionError, self.controller.contribute, event)
+
+    def test_contribution_action_reindex(self):
+        event = [self._mock_sqs_record(action='reindex', prefix='ff')]
+        with mock.patch('azul.azulclient.AzulClient.do_remote_reindex') as mock_reindex:
+            mock_reindex.return_value = True
+            self.controller.contribute(event)
+            self.assertEqual(1, mock_reindex.call_count)
+
+    def test_contribute_and_aggregate(self):
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        event = []
+        bundles = []
+        contributions_per_bundle: List[Contributions] = []
+        bundle_fqids = [
+            BundleFQID('56a338fe-7554-4b5d-96a2-7df127a7640b', '2018-03-28T151023.074974Z'),
+            BundleFQID('b2216048-7eaa-45f4-8077-5a3fb4204953', '2018-03-29T104041.822717Z')
+        ]
+        for bundle_fqid in bundle_fqids:
+            notification = self.client.synthesize_notification(self.catalog, bundle_fqid)
+            event.append(self._mock_sqs_record(action='add',
+                                               catalog=self.catalog,
+                                               notification=notification))
+            bundle = self._load_canned_bundle(bundle_fqid)
+            bundles.append(bundle)
+            contributions_per_bundle.append(self.index_service.transform(self.catalog, bundle=bundle, delete=False))
+        repository_patch = mock.patch('azul.indexer.index_controller.IndexController.repository_plugin')
+        service_patch = mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                                   new_callable=mock.PropertyMock)
+        with repository_patch as mock_repository, service_patch as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_repository.return_value = mock.MagicMock()
+            mock_repository.return_value.fetch_bundle.side_effect = bundles
+            # Test contributions from two bundles with from the same project
+            tallies = self._assert_notification_contributions(event, contributions_per_bundle)
+            # Test aggregation for tallies, inspect for deferred tallies
+            project_tally = self._assert_aggregation_creates_deferral(tallies)
+            # Test aggregation for deferred tallies
+            self._assert_aggregation_handles_deferral(project_tally)
+
+    def _assert_notification_contributions(self,
+                                           event: Sequence[SQSRecord],
+                                           contributions_per_bundle: Sequence[Contributions]):
+        self.controller.contribute(event)
+        tallies = [
+            json.loads(m.body)
+            for m in self.queue_manager.read_messages(self.controller._tallies_queue())
+        ]
+        entities_from_tallies = {(t['entity_id'], t['entity_type']) for t in tallies}
+        entities_from_contributions = {
+            (c.entity.entity_id, c.entity.entity_type)
+            for cs in contributions_per_bundle for c in cs
+        }
+        self.assertSetEqual(entities_from_tallies, entities_from_contributions)
+        return tallies
+
+    def _assert_aggregation_creates_deferral(self, tallies: Sequence[CataloguedTallies]):
+        event = []
+        for tally in tallies:
+            event.append(self._mock_sqs_record(**tally))
+        self.controller.aggregate(event)
+        messages = self.queue_manager.read_messages(self.controller._tallies_queue())
+        # Check that the deferral for the project entity was returned to the tallies queue.
+        self.assertEqual(1, len(messages))
+        project_tally = json.loads(messages[0].body)
+        expected_tally = {
+            'catalog': 'test',
+            'entity_type': 'projects',
+            'entity_id': '93f6a42f-1790-4af4-b5d1-8c436cb6feae',
+            'num_contributions': 2
+        }
+        self.assertDictEqual(project_tally, expected_tally)
+        return project_tally
+
+    def _assert_aggregation_handles_deferral(self, deffered_tally: CataloguedTallies):
+        event = [self._mock_sqs_record(**deffered_tally)]
+        self.controller.aggregate(event)
+        messages = self.queue_manager.read_messages(self.controller._tallies_queue())
+        self.assertEqual(0, len(messages), )
--- a/test/indexer/test_indexer_controller.py
+++ b/test/indexer/test_indexer_controller.py
@@ -74,8 +74,8 @@ class TestIndexController(IndexerTestCase):
 
     def test_invalid_notification(self):
         event = [
-            self._mock_sqs_record(action='banana',
-                                  notification='slug',
+            self._mock_sqs_record(action='foo',
+                                  notification='bar',
                                   catalog=self.catalog)
         ]
         self.assertRaises(AssertionError, self.controller.contribute, event)
@@ -85,7 +85,7 @@ class TestIndexController(IndexerTestCase):
         with mock.patch.object(AzulClient, 'do_remote_reindex') as mock_reindex:
             mock_reindex.return_value = True
             self.controller.contribute(event)
-            self.assertEqual(1, mock_reindex.call_count)
+            mock_reindex.assert_called_once_with(dict(action='reindex', prefix='ff'))
 
     def test_contribute_and_aggregate(self):
         """
@@ -136,7 +136,8 @@ class TestIndexController(IndexerTestCase):
                 for t in tallies
             }
             self.assertSetEqual(expected_entities, entities_from_tallies)
-            self.assertListEqual([mock.call(f) for f in bundle_fqids], mock_plugin.fetch_bundle.mock_calls)
+            self.assertListEqual([mock.call(f) for f in bundle_fqids],
+                                 mock_plugin.fetch_bundle.mock_calls)
 
             # Test aggregation for tallies, inspect for deferred tallies
             event = [self._mock_sqs_record(**t) for t in tallies]
--- a/test/indexer/test_indexer_controller.py
+++ b/test/indexer/test_indexer_controller.py
@@ -1,15 +1,14 @@
 import json
 import logging
-from typing import (
-    List,
-    Sequence,
-)
 
 import boto3
 from chalice.app import (
     SQSRecord,
 )
 import mock
+from more_itertools import (
+    one,
+)
 from moto import (
     mock_sqs,
     mock_sts,
@@ -25,20 +24,13 @@ from azul.azulclient import (
 from azul.indexer import (
     BundleFQID,
 )
-from azul.indexer.document import (
-    Contribution,
-)
 from azul.indexer.index_controller import (
     IndexController,
-)
-from azul.indexer.index_service import (
-    CataloguedTallies,
 )
 from azul.logging import (
     configure_test_logging,
 )
 from indexer import (
-    ForcedRefreshIndexService,
     IndexerTestCase,
 )
 
@@ -50,9 +42,6 @@ def setUpModule():
     configure_test_logging(log)
 
 
-Contributions = Sequence[Contribution]
-
-
 @mock_sts
 @mock_sqs
 class TestIndexController(IndexerTestCase):
@@ -62,6 +51,8 @@ class TestIndexController(IndexerTestCase):
         self.index_service.create_indices(self.catalog)
         self.client = AzulClient()
         self.controller = IndexController()
+        # noinspection PyPropertyAccess
+        self.controller.index_service = self.index_service
         self.queue_manager = queues.Queues(delete=True)
 
     def tearDown(self):
@@ -81,7 +72,7 @@ class TestIndexController(IndexerTestCase):
         }
         return SQSRecord(event_dict=event_dict, context='controller_test')
 
-    def test_contribution_action_invalid(self):
+    def test_invalid_notification(self):
         event = [
             self._mock_sqs_record(action='banana',
                                   notification='slug',
@@ -89,24 +80,26 @@ class TestIndexController(IndexerTestCase):
         ]
         self.assertRaises(AssertionError, self.controller.contribute, event)
 
-    def test_contribution_action_reindex(self):
+    def test_remote_reindex(self):
         event = [self._mock_sqs_record(action='reindex', prefix='ff')]
-        with mock.patch('azul.azulclient.AzulClient.do_remote_reindex') as mock_reindex:
+        with mock.patch.object(AzulClient, 'do_remote_reindex') as mock_reindex:
             mock_reindex.return_value = True
             self.controller.contribute(event)
             self.assertEqual(1, mock_reindex.call_count)
 
     def test_contribute_and_aggregate(self):
         """
-        Index two bundles that make contributions to the same project.
-        Inspect that the contributions match the tallies that are returned to
-        SQS. During aggregation only the project entity is deferred due to
+        Contribution and aggregation of two bundles
+
+        Index two bundles that make contributions to the same project. Inspect
+        that the contributions match the tallies that are returned to SQS.
+        During aggregation only the project entity is deferred due to
         multiple contributions.
         """
         self._create_mock_queues()
         event = []
         bundles = []
-        contributions_per_bundle: List[Contributions] = []
+        expected_entities = set()
         bundle_fqids = [
             BundleFQID('56a338fe-7554-4b5d-96a2-7df127a7640b', '2018-03-28T151023.074974Z'),
             BundleFQID('b2216048-7eaa-45f4-8077-5a3fb4204953', '2018-03-29T104041.822717Z')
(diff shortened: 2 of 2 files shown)


test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,137 @@
...
+)
+from indexer import (
+    ForcedRefreshIndexService,
+    IndexerTestCase,
+)
+
+log = logging.getLogger(__name__)
+
+
+class Record():
```
danielsotirhos, line 32:
```suggestion
class Record:
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,137 @@
...
+            mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                       new_callable=mock.PropertyMock) as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_transform.side_effect = list_of_contributions
+            self.controller.contribute(event)
+            tallies = list(json.loads(x.body) for x in
+                           queues.Queues(delete=True).read_messages(self.controller._tallies_queue()))
+            entities_from_tallies = set((x["entity_id"], x["entity_type"]) for x in tallies)
+            entities_from_contributions = set((contributions.entity.entity_id, contributions.entity.entity_type)
+                                              for x in list_of_contributions for contributions in x)
```
danielsotirhos, line 119:
```suggestion
                                              for cl in contribution_lists for contribution in cl)
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,137 @@
...
+                event.append(self.generate_record(**tally))
+            self.controller.aggregate(event)
+            messages = queues.Queues(delete=True).read_messages(self.controller._tallies_queue())
+            self.assertEqual(len(messages), 1)
+            project_tally = json.loads(messages[0].body)
+            self.assertDictEqual(project_tally,
+                                 {"catalog": "test",
+                                  "entity_type": "projects",
+                                  "entity_id": "93f6a42f-1790-4af4-b5d1-8c436cb6feae",
+                                  "num_contributions": 2})
```
danielsotirhos, line 132:
```suggestion
            expected_tally = {
                "catalog": "test",
                "entity_type": "projects",
                "entity_id": "93f6a42f-1790-4af4-b5d1-8c436cb6feae",
                "num_contributions": 2
            }
            self.assertDictEqual(project_tally, expected_tally)
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        event = list()
+        bundles = list()
+        list_of_contributions = list()  # rename this to something more meaningful, (ss) for multiple contributions
```
danielsotirhos, line 96:
```suggestion
        contribution_lists = list()
```
Not sure what `(ss)` is referring to here.

amarjandu: eeep thats the secret sauce ill take it out.

src/azul/queues.py
```diff
@@ -108,6 +100,27 @@ def _dump_messages(self, messages, queue_url, path):
...
         logger.info('Wrote %i messages', len(messages))
 
+    def _get_messages(self, queue):
+        messages = []
+        while True:
+            message_batch = queue.receive_messages(AttributeNames=['All'],
+                                                   MaxNumberOfMessages=10,
+                                                   VisibilityTimeout=300)
+            if not message_batch:  # Nothing left in queue
+                break
```
hannes-ucsc, line 110:
```suggestion
                return messages
```

src/azul/queues.py
```diff
@@ -108,6 +100,27 @@ def _dump_messages(self, messages, queue_url, path):
...
+            else:
+                messages.extend(message_batch)
+        return messages
```
hannes-ucsc, line 113:
```suggestion
```

src/azul/queues.py
```diff
@@ -108,6 +100,27 @@ def _dump_messages(self, messages, queue_url, path):
...
+
+    def read_messages(self, queue):
+        messages = self._get_messages(queue)
+        message_batches = list(more_itertools.chunked(messages, 10))
```
hannes-ucsc, line 117: This duplicates code in _dump()

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+            event.append(self.generate_record(
+                action='add',
+                catalog=self.catalog,
+                notification=self.client.synthesize_notification(self.catalog, bundle_fqid)))
+            bundle = self._load_canned_bundle(bundle_fqid)
+            bundles.append(bundle)
+            list_of_contributions.append(self.index_service.transform(self.catalog,
+                                                                      bundle=bundle,
+                                                                      delete=False))
+        with mock.patch('azul.indexer.index_controller.IndexController.transform') as mock_transform, \
```
hannes-ucsc, line 109: No line continuations with backslashes. If that's not mentioned in the C guide, please add it.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+        event = list()
+        bundles = list()
+        list_of_contributions = list()
+        bundle_fqids = [BundleFQID('56a338fe-7554-4b5d-96a2-7df127a7640b', '2018-03-28T151023.074974Z'),
+                        BundleFQID('b2216048-7eaa-45f4-8077-5a3fb4204953', '2018-03-29T104041.822717Z')]
+        for bundle_fqid in bundle_fqids:
+            event.append(self.generate_record(
+                action='add',
+                catalog=self.catalog,
+                notification=self.client.synthesize_notification(self.catalog, bundle_fqid)))
```
hannes-ucsc, line 103: extract the notification into a variable and use aligned indent  for the args

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+)
+from indexer import (
+    ForcedRefreshIndexService,
+    IndexerTestCase,
+)
+
+log = logging.getLogger(__name__)
+
+
+class Record:
```
hannes-ucsc, line 32: Any reason we can't use SQSRecord from CHalice?

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+        super().setUp()
+        self.index_service.create_indices(self.catalog)
+        self.client = AzulClient()
+        self.controller = IndexController()
+
+    def tearDown(self):
+        self.index_service.delete_indices(self.catalog)
+        super().tearDown()
+
+    def generate_events(self):
```
hannes-ucsc, line 56: What's this?

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_transform.side_effect = list_of_contributions
+            self.controller.contribute(event)
+            tallies = list(json.loads(x.body) for x in
+                           queues.Queues(delete=True).read_messages(self.controller._tallies_queue()))
+            entities_from_tallies = set((x["entity_id"], x["entity_type"]) for x in tallies)
+            entities_from_contributions = set((contributions.entity.entity_id, contributions.entity.entity_type)
+                                              for cl in list_of_contributions for contributions in cl)
+            self.assertSetEqual(entities_from_tallies, entities_from_contributions)
+            event.clear()
```
hannes-ucsc, line 121:
```suggestion
            event = []
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+            list_of_contributions.append(self.index_service.transform(self.catalog,
+                                                                      bundle=bundle,
+                                                                      delete=False))
+        with mock.patch('azul.indexer.index_controller.IndexController.transform') as mock_transform, \
+            mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                       new_callable=mock.PropertyMock) as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_transform.side_effect = list_of_contributions
+            self.controller.contribute(event)
+            tallies = list(json.loads(x.body) for x in
```
hannes-ucsc, line 115: use list comprehension

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+    @mock_sqs
+    def _test_contribute_and_aggregate(self):
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        event = list()
```
hannes-ucsc, line 94:
```suggestion
        event = []
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+        bundles = list()
+        list_of_contributions = list()
+        bundle_fqids = [BundleFQID('56a338fe-7554-4b5d-96a2-7df127a7640b', '2018-03-28T151023.074974Z'),
```
hannes-ucsc, line 97: this is not how he wrap literals see C guide

test/indexer/test_indexer_controller.py
hannes-ucsc: I think you should inspect some of my recent commits to get a better feel for how we wrap things.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+                                                                      delete=False))
+        with mock.patch('azul.indexer.index_controller.IndexController.transform') as mock_transform, \
+            mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                       new_callable=mock.PropertyMock) as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_transform.side_effect = list_of_contributions
+            self.controller.contribute(event)
+            tallies = list(json.loads(x.body) for x in
+                           queues.Queues(delete=True).read_messages(self.controller._tallies_queue()))
+            entities_from_tallies = set((x["entity_id"], x["entity_type"]) for x in tallies)
```
hannes-ucsc, line 117: quotes

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,139 @@
...
+            entities_from_contributions = set((contributions.entity.entity_id, contributions.entity.entity_type)
+                                              for cl in list_of_contributions for contributions in cl)
+            self.assertSetEqual(entities_from_tallies, entities_from_contributions)
+            event.clear()
+            for tally in tallies:
+                event.append(self.generate_record(**tally))
+            self.controller.aggregate(event)
+            messages = queues.Queues(delete=True).read_messages(self.controller._tallies_queue())
```
hannes-ucsc, line 125: extract the Queues instance into a variable

src/azul/queues.py
amarjandu: was trying to minimize the changes in this file, ill refactor it.

test/indexer/test_indexer_controller.py
amarjandu: was easier when prototyping to mock this functionality out. refactored to SQSRecord.

test/indexer/test_indexer_controller.py
amarjandu: It was a method that assisted with event generation, its been refactored to assist with creation of SQSRecord objects.

test/indexer/test_indexer_controller.py
amarjandu: Is there a reason to prefer one over the other in this context?

amarjandu:
@jessebrennan Please focus your attention on the requests that @hannes-ucsc made regarding matching formatting specifications from the contributing guidelines.

src/azul/queues.py
amarjandu: I added the changes to this file as a fix-up to the commit that added the `read_message` functionality to the Queue class. see c7e02846bc6954be4cc9e9d6999bba3acdd149ca

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+        ]
+        for bundle_fqid in bundle_fqids:
+            notification = self.client.synthesize_notification(self.catalog, bundle_fqid)
+            event.append(self.generate_sqs_record(action='add',
+                                                  catalog=self.catalog,
+                                                  notification=notification))
+            bundle = self._load_canned_bundle(bundle_fqid)
+            bundles.append(bundle)
+            list_of_contributions.append(self.index_service.transform(self.catalog, bundle=bundle, delete=False))
+        with mock.patch('azul.indexer.index_controller.IndexController.transform') as mock_transform, (
```
jessebrennan, line 116: I'm not sure of a better way to format this line, but it's pretty huge. Also I suspect parentheses for the second patch call are redundant.

jessebrennan: I suspect that this mocking is at the wrong layer of abstraction. Since this test is for the controller, I think you should be mocking `fetch_bundle` instead of `transform`. Otherwise it looks like the `index_controller.transform` will never be covered.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+            self.assertEqual(len(messages), 1)
+            project_tally = json.loads(messages[0].body)
+            expected_tally = {
+                'catalog': 'test',
+                'entity_type': 'projects',
+                'entity_id': '93f6a42f-1790-4af4-b5d1-8c436cb6feae',
+                'num_contributions': 2
+            }
+            self.assertDictEqual(project_tally, expected_tally)
+            event = []
```
jessebrennan, line 146:
```suggestion
            events = []
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+        for queue_name in config.all_queue_names:
+            sqs.create_queue(QueueName=queue_name)
+
+    def generate_sqs_record(self, **kwargs):
+        event_dict = {
+            'body': json.dumps(kwargs),
+            'receiptHandle': 'ThisWasARandomString',
+            'attributes': {'ApproximateReceiveCount': 1}
+        }
+        context = 'contorller_test'
```
jessebrennan, line 70:
```suggestion
        context = 'controller_test'
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+        self.queue_manager = queues.Queues(delete=True)
+
+    def tearDown(self):
+        self.index_service.delete_indices(self.catalog)
+        super().tearDown()
+
+    def generate_events(self):
+        pass
+
+    def _create_mock_queues(self):
```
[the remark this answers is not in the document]
amarjandu: Is there a reason why the decorator would not be operational? I was running into this issue when trying to add @mock_sqs and @mock_sts to the `_create_mock_queues` definition and it would never mock the sts functionality correctly, it would try to using an invalid token to perform the creation of the mock queues.

I do agree that the positioning of the `@mock_x` would be better if it was closer to the method definition, i was forced to follow precedence from another file, https://github.com/DataBiosphere/azul/blob/e0b7febd31003f3555a322ca4354bdca18b14056/test/health_check_test_case.py#L282 which also lead me to the naming decision, but also does the mocking the same way.

test/indexer/test_indexer_controller.py
amarjandu: it was refactored  see https://github.com/DataBiosphere/azul/blob/e6b3a14a9624c24977675de3a52571c3e6a4be52/test/indexer/test_indexer_controller.py#L64

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+        return SQSRecord(event_dict=event_dict, context=context)
+
+    def test_contributions(self):
+        self._test_contribution_action_invalid()
+        self._test_contribution_action_reindex()
+        self._test_contribute_and_aggregate()
+
+    @mock_sts
+    def _test_contribution_action_invalid(self):
+        event = [self.generate_sqs_record(action='banana', notification='slug', catalog=self.catalog)]
```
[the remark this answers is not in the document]
amarjandu: `event` was used as the argument to the index_controller.contribute() method, it made sense keeping it the same. I can change it.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+    def generate_sqs_record(self, **kwargs):
+        event_dict = {
+            'body': json.dumps(kwargs),
+            'receiptHandle': 'ThisWasARandomString',
+            'attributes': {'ApproximateReceiveCount': 1}
+        }
+        context = 'contorller_test'
+        return SQSRecord(event_dict=event_dict, context=context)
+
+    def test_contributions(self):
```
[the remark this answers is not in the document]
amarjandu: originally i was going to test out the contributions separately, then test out the aggregations after. I was using it as a grouping. I'll remove it.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+        event = [self.generate_sqs_record(action='banana', notification='slug', catalog=self.catalog)]
+        self.assertRaises(AssertionError, self.controller.contribute, event)
+
+    @mock_sts
+    def _test_contribution_action_reindex(self):
+        event = [self.generate_sqs_record(action='reindex', prefix='ff')]
+        with mock.patch('azul.azulclient.AzulClient.do_remote_reindex') as mock_reindex:
+            mock_reindex.return_value = True
+            self.controller.contribute(event)
+            self.assertTrue(mock_reindex.called)
```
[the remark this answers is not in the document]
amarjandu: added it

test/indexer/test_indexer_controller.py
amarjandu: I think there are a limited number of ways we can format a long context manager like this, i was using the parenthesis to get around the usage of `\`.

From the contributing guidelines 
```
* We avoid the use of backslash for continuing statements beyond one line.
  Instead, we exploit the fact that Python can infer continuation if they
  occur in balanced constructs like brackets or parentheses. If necessary we
  introduce a a pair of parentheses around the wrapping expression.

  With some keywords it is impossible to add semantically insignificant
  parentheses. For example, ``assert foo, "bad"`` is not equivalent to ``assert
  (foo, "bad")``. In these exceptional situations it is permissible to use
  backslash for line continuation.
```

Another alternative is:
```
with mock.patch(
    "azul.indexer.index_controller.IndexController.transform"
) as mock_transform, mock.patch(
    "azul.indexer.index_controller.IndexController.index_service",
    new_callable=mock.PropertyMock,
) as mock_index_service:
    pass

```
I dont think the alternative fits the contribution guidelines for align indented wrapped literals.  (although it says we prefer, not we demand; maybe an exception could be made)

There might be another alternative I have not thought of, I'm all for an alternative solution.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+                for m in self.queue_manager.read_messages(self.controller._tallies_queue())
+            ]
+            entities_from_tallies = {(t['entity_id'], t['entity_type']) for t in tallies}
+            entities_from_contributions = {
+                (contributions.entity.entity_id, contributions.entity.entity_type)
+                for cl in contributions_per_bundle for contributions in cl
+            }
+            self.assertSetEqual(entities_from_tallies, entities_from_contributions)
+            events = []
+            for tally in tallies:
```
jessebrennan, line 136: I think it would make sense to split this into two separate test methods, starting the second one here. This isn't necessary, but there is a logical divide here between testing `contribute` and `aggregate`, and the test method is already pretty long.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+                events.append(self.generate_sqs_record(**tally))
+            self.controller.aggregate(events)
+            messages = self.queue_manager.read_messages(self.controller._tallies_queue())
+            self.assertEqual(len(messages), 1)
```
jessebrennan, line 140: This is testing the deferral, right? It could be helpful to make that more explicit with a comment here too.

test/indexer/test_indexer_controller.py
amarjandu: @jessebrennan 
The original plan was to test `contribute` and `aggregate` separately, but @hannes-ucsc  suggested to  just perform the aggregation because we already have the contributions ready to go within this test. 

Perhaps a subtest can be used show distinction from the two different operations?

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+                                                   notification=notification))
+            bundle = self._load_canned_bundle(bundle_fqid)
+            bundles.append(bundle)
+            contributions_per_bundle.append(self.index_service.transform(self.catalog, bundle=bundle, delete=False))
+        with mock.patch('azul.indexer.index_controller.IndexController.repository_plugin') as mock_repository, (
+            mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                       new_callable=mock.PropertyMock)) as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_repository.return_value = mock.MagicMock()
+            mock_repository.return_value.fetch_bundle.return_value = yield contributions_per_bundle
```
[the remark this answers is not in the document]
amarjandu: everytime that a fetch_bundle is called, we would like to yield a different bundle that was loaded from the canned bundles, this allows us to not actually use the data-plugins to fetch data.

I think I can change this to a `fetch_bundle.side_effect`, it should behave in a similar manor to what we need.
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

I'll try to change this.

test/indexer/test_indexer_controller.py
amarjandu: that was stupidity on my part, i did not see that the method was still present. Its been removed. 😬

test/indexer/test_indexer_controller.py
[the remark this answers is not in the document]
amarjandu: I like the idea of wrapping the class, and not having to add decorators everywhere. 👍🏽  got it working thx.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+    @mock_sqs
+    def test_contribute_and_aggregate(self):
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        events = []
```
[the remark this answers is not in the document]
amarjandu: I think this was a link to the other discussion https://github.com/DataBiosphere/azul/pull/2121#discussion_r468231291

I reverted it back to `event` I like using what the controller had.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+        ]
+        for bundle_fqid in bundle_fqids:
+            notification = self.client.synthesize_notification(self.catalog, bundle_fqid)
+            events.append(self.generate_sqs_record(action='add',
+                                                   catalog=self.catalog,
+                                                   notification=notification))
+            bundle = self._load_canned_bundle(bundle_fqid)
+            bundles.append(bundle)
+            contributions_per_bundle.append(self.index_service.transform(self.catalog, bundle=bundle, delete=False))
+        with mock.patch('azul.indexer.index_controller.IndexController.repository_plugin') as mock_repository, (
```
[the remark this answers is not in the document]
amarjandu: Cleaner, I'm going to borrow this. thx

test/indexer/test_indexer_controller.py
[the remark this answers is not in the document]
amarjandu: It was a no-go, by changing this to 
` mock_repository.return_value.fetch_bundle.side_effect = contributions_per_bundle`
The entire list of bundles are returned to where the method call to `fetch_bundle` was made in the `transform`.

I think that we could get around this by doing a more intensive mockup of the `repository_plugin`, but I would like to be precise with the mocks/patches that are made.

test/indexer/test_indexer_controller.py
amarjandu: Added!

test/indexer/test_indexer_controller.py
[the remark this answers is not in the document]
amarjandu: alright after too much time staring at this problem, i found the issue.
```
mock_repository.return_value = mock.MagicMock()
mock_repository.return_value.fetch_bundle.side_effect = bundles
```
resolves the issue, it was getting a `list` because `contributions_per_bundle` is a list of lists and not what we want to return back to the transform method. >.> 

For some reason when i placed a yield as the return value, the test never really ran, is that what you meant by?

> putting yield here would turn the entire test into a generator and that would definitely not be the behaviour we want.

This was only caught when trying to break the test into sub-tests to show the distinction between the parts of the test, the test output did not show the msg from the sub-tests, upon investigation the tests were passing without running.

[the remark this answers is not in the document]
jessebrennan: > For some reason when i placed a yield as the return value, the test never really ran, is that what you meant

Yes, that's what I mean. This SO post give a pretty good explanation of what was going on: https://stackoverflow.com/questions/32128412/python-yield-and-yield-assignment

test/indexer/test_indexer_controller.py
jessebrennan: Personally I'm not a fan of using subtest this way, but keep it if you prefer and/or Hannes will decide.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        event = []
+        bundles = []
+        contributions_per_bundle: Sequence[Sequence[Contribution]] = []
```
jessebrennan, line 99: There's a type error because one of these "Sequence"s should be a "MutableSequence".

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,151 @@
...
+            with self.subTest('Test contributions from 2 bundles with matching project_ids'):
+                self.controller.contribute(event)
+                tallies = [
+                    json.loads(m.body)
+                    for m in self.queue_manager.read_messages(self.controller._tallies_queue())
+                ]
+                entities_from_tallies = {(t['entity_id'], t['entity_type']) for t in tallies}
+                entities_from_contributions = {
+                    (contributions.entity.entity_id, contributions.entity.entity_type)
+                    for cl in contributions_per_bundle for contributions in cl
```
jessebrennan, line 128: I don't know what `cl` stands for. Also for `contributions`, isn't it just a single contribution at that point?

If I'm correct, then I think
```suggestion
                    for cs in contributions_per_bundle for c in cs
```
makes more sense.

amarjandu:
@hannes-ucsc  threads that need your 👀  (and opinion) 

https://github.com/DataBiosphere/azul/pull/2121#discussion_r468193791
https://github.com/DataBiosphere/azul/pull/2121#discussion_r468183740
https://github.com/DataBiosphere/azul/pull/2121#discussion_r469453453

src/azul/queues.py
```diff
@@ -79,34 +79,41 @@ def dump_all(self):
...
         message_batches = list(more_itertools.chunked(messages, 10))
         if self._delete:
             logger.info('Removing messages from queue "%s"', queue.url)
             self._delete_messages(message_batches, queue)
         else:
             logger.info('Returning messages to queue "%s"', queue.url)
             self._return_messages(message_batches, queue)
-        logger.info(f'Finished writing {path !r}')
-
-    def _dump_messages(self, messages, queue_url, path):
```
hannes-ucsc, line 101: Try to minimize the diff. The diff  looks like the _dump_messages method was moved unnecessarily.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+)
+from azul.logging import (
+    configure_test_logging,
+)
+from indexer import (
+    ForcedRefreshIndexService,
+    IndexerTestCase,
+)
+
+logger = logging.getLogger(__name__)
```
hannes-ucsc, line 42:
```suggestion
log = logging.getLogger(__name__)
```

hannes-ucsc: See C guide

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+    def tearDown(self):
+        self.index_service.delete_indices(self.catalog)
+        super().tearDown()
+
+    def _create_mock_queues(self):
+        sqs = boto3.resource('sqs')
+        for queue_name in config.all_queue_names:
+            sqs.create_queue(QueueName=queue_name)
+
+    def generate_sqs_record(self, **kwargs):
```
hannes-ucsc, line 70:
```suggestion
    def generate_sqs_record(self, **body):
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+        event_dict = {
+            'body': json.dumps(kwargs),
+            'receiptHandle': 'ThisWasARandomString',
+            'attributes': {'ApproximateReceiveCount': 1}
+        }
+        context = 'controller_test'
```
hannes-ucsc, line 76: inline

test/indexer/test_indexer_controller.py
hannes-ucsc, line 70:
```suggestion
    def _mock_sqs_record(self, **kwargs):
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+        return SQSRecord(event_dict=event_dict, context=context)
+
+    def test_contribution_action_invalid(self):
+        event = [self.generate_sqs_record(action='banana', notification='slug', catalog=self.catalog)]
```
hannes-ucsc, line 80: too long, see C guide

test/indexer/test_indexer_controller.py
hannes-ucsc: type hint return value

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+            bundles.append(bundle)
+            contributions_per_bundle.append(self.index_service.transform(self.catalog, bundle=bundle, delete=False))
+        repository_patch = mock.patch('azul.indexer.index_controller.IndexController.repository_plugin')
+        service_patch = mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                                   new_callable=mock.PropertyMock)
+        with repository_patch as mock_repository, service_patch as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_repository.return_value = mock.MagicMock()
+            mock_repository.return_value.fetch_bundle.side_effect = bundles
+            with self.subTest('Test contributions from 2 bundles with matching project_ids'):
```
hannes-ucsc, line 120:
```suggestion
            with self.subTest('Test contributions from two bundles from the same project'):
```

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+                    json.loads(m.body)
+                    for m in self.queue_manager.read_messages(self.controller._tallies_queue())
+                ]
+                entities_from_tallies = {(t['entity_id'], t['entity_type']) for t in tallies}
+                entities_from_contributions = {
+                    (c.entity.entity_id, c.entity.entity_type)
+                    for cs in contributions_per_bundle for c in cs
+                }
+                self.assertSetEqual(entities_from_tallies, entities_from_contributions)
+            with self.subTest('Test aggregation for tallies, inspect for deferred tallies'):
```
hannes-ucsc, line 132: Let's say the subtest above fails. Does it make sense to attempt this subtest or is this one bound to fail in that case? If so, this shouldn't be a subtest.

hannes-ucsc: With or without subtests, consider extracting the three parts into separate methods. The subtest `with` statement should be part of the extracted method.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,152 @@
...
+                event = []
+                for tally in tallies:
+                    event.append(self.generate_sqs_record(**tally))
+                self.controller.aggregate(event)
+                messages = self.queue_manager.read_messages(self.controller._tallies_queue())
+                # Check that the deferral for the project entity was returned to the tallies queue.
+                self.assertEqual(len(messages), 1)
```
hannes-ucsc, line 139: expected value first

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        event = list()
+        bundles = list()
+        list_of_contributions = list()
```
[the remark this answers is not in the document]
hannes-ucsc: Yes

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,150 @@
...
+            mock.patch('azul.indexer.index_controller.IndexController.index_service',
+                       new_callable=mock.PropertyMock)) as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_transform.side_effect = list_of_contributions
+            self.controller.contribute(event)
+            tallies = [
+                json.loads(x.body)
+                for x in self.queue_manager.read_messages(self.controller._tallies_queue())
+            ]
+            entities_from_tallies = {(x['entity_id'], x['entity_type']) for x in tallies}
```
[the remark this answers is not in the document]
hannes-ucsc: Yes

test/indexer/test_indexer_controller.py
hannes-ucsc: Subtests cause the test code to resume after the CM body upon failure in the CM body. So subtests should be used to mark section in the progression. Here I think it makes sense to test contribution and aggregation in the same method. If it were done in a separate method, the method for testing aggregation would duplicate a lot of the code from the method for testing contributions.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,168 @@
...
+            # Test contributions from two bundles with from the same project
+            tallies = self._assert_notification_contributions(event, contributions_per_bundle)
+            # Test aggregation for tallies, inspect for deferred tallies
+            project_tally = self._assert_aggregation_creates_defferal(tallies)
+            # Test aggregation for deferred tallies
+            self._assert_aggregation_handles_defferal(project_tally)
+
+    def _assert_notification_contributions(self,
+                                           event: Sequence[SQSRecord],
+                                           contributions_per_bundle: MutableSequence[Sequence[Contribution]]):
```
hannes-ucsc, line 132: I don't think you need mutability here.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,168 @@
...
+        ]
+        entities_from_tallies = {(t['entity_id'], t['entity_type']) for t in tallies}
+        entities_from_contributions = {
+            (c.entity.entity_id, c.entity.entity_type)
+            for cs in contributions_per_bundle for c in cs
+        }
+        self.assertSetEqual(entities_from_tallies, entities_from_contributions)
+        return tallies
+
+    def _assert_aggregation_creates_defferal(self, tallies):
```
hannes-ucsc, line 146: Try to add more type hints. On `tallies`, for example.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,168 @@
...
+        with repository_patch as mock_repository, service_patch as mock_index_service:
+            mock_index_service.return_value = ForcedRefreshIndexService()
+            mock_repository.return_value = mock.MagicMock()
+            mock_repository.return_value.fetch_bundle.side_effect = bundles
+            # Test contributions from two bundles with from the same project
+            tallies = self._assert_notification_contributions(event, contributions_per_bundle)
+            # Test aggregation for tallies, inspect for deferred tallies
+            project_tally = self._assert_aggregation_creates_defferal(tallies)
+            # Test aggregation for deferred tallies
+            self._assert_aggregation_handles_defferal(project_tally)
```
hannes-ucsc, line 128:
```suggestion
            self._assert_aggregation_handles_deferral(project_tally)
```

hannes-ucsc (changes requested):

src/azul/indexer/document.py
```diff
@@ -582,4 +583,6 @@ def delete(self):
         return super().delete or not self.contents
 
 
+Contributions = Sequence[Contribution]
```
hannes-ucsc, line 586: Not here. In the test module.

hannes-ucsc (changes requested):

hannes-ucsc:
No bump.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        Index two bundles that make contributions to the same project.
+        Inspect that the contributions match the tallies that are returned to
+        SQS. During aggregation only the project entity is deferred due to
+        multiple contributions.
+        """
+        self._create_mock_queues()
+        event = []
+        bundles = []
+        contributions_per_bundle: MutableSequence[Contributions] = []
```
hannes-ucsc, line 109:
```suggestion
        contributions_per_bundle: List[Contributions] = []
```

hannes-ucsc (changes requested):
Sorry, didn't see this earlier.

No bump.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def test_contribution_action_invalid(self):
+        event = [
+            self._mock_sqs_record(action='banana',
+                                  notification='slug',
+                                  catalog=self.catalog)
+        ]
+        self.assertRaises(AssertionError, self.controller.contribute, event)
+
+    def test_contribution_action_reindex(self):
```
hannes-ucsc: Misleading name.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def _mock_sqs_record(self, **body):
+        event_dict = {
+            'body': json.dumps(body),
+            'receiptHandle': 'ThisWasARandomString',
+            'attributes': {'ApproximateReceiveCount': 1}
+        }
+        return SQSRecord(event_dict=event_dict, context='controller_test')
+
+    def test_contribution_action_invalid(self):
```
hannes-ucsc: Awkward name.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_contribution_action_reindex(self):
+        event = [self._mock_sqs_record(action='reindex', prefix='ff')]
+        with mock.patch('azul.azulclient.AzulClient.do_remote_reindex') as mock_reindex:
+            mock_reindex.return_value = True
+            self.controller.contribute(event)
+            self.assertEqual(1, mock_reindex.call_count)
+
+    def test_contribute_and_aggregate(self):
+        """
+        Index two bundles that make contributions to the same project.
```
hannes-ucsc: Note that the first line of the docstring is used for diagnostic output by unit test.

hannes-ucsc: image

hannes-ucsc:
@jessebrennan: Given that you logged three hours of review time, I didn't get much bang for the buck. I engaged with this for a couple of hours and I came up with a lot of stuff.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,160 @@
...
+                'entity_id': '93f6a42f-1790-4af4-b5d1-8c436cb6feae',
+                'num_contributions': 2
+            }
+            self.assertDictEqual(project_tally, expected_tally)
+
+            # Test aggregation of deferred project entity
+            event = [self._mock_sqs_record(**project_tally)]
+            self.controller.aggregate(event)
+            messages = self.queue_manager.read_messages(self.controller._tallies_queue())
+            self.assertEqual(0, len(messages))
```
hannes-ucsc, line 160: There's probably more you can assert about the message.

test/indexer/test_indexer_controller.py
```diff
@@ -0,0 +1,160 @@
...
+                                  catalog=self.catalog)
+        ]
+        self.assertRaises(AssertionError, self.controller.contribute, event)
+
+    def test_remote_reindex(self):
+        event = [self._mock_sqs_record(action='reindex', prefix='ff')]
+        with mock.patch.object(AzulClient, 'do_remote_reindex') as mock_reindex:
+            mock_reindex.return_value = True
+            self.controller.contribute(event)
+            self.assertEqual(1, mock_reindex.call_count)
```
hannes-ucsc, line 88: There's more you can assert about the mock calls. For reference see how this is handled for `mock_plugin`

hannes-ucsc (changes requested):

test/indexer/test_indexer_controller.py
amarjandu: As there are no messages returned after the deferral was processed what other assertions were you thinking about?

Should we inspect the index for the number of total contributions made?

test/indexer/test_indexer_controller.py
amarjandu: i dont see much in the repo under `mock_plugin` https://github.com/DataBiosphere/azul/search?q=mock_plugin&type=Code is there a link/file you can direct me to? 

I'll look around for other assertion that are used to mocked objects around the repo.

test/indexer/test_indexer_controller.py
hannes-ucsc: Nah. I was confused.

hannes-ucsc (changes requested):
You don't want to fixup my commit because that would either make my commit appear has having been authored by you or vice versa. You probably want to make a `[3/3] …` commit and rename the other two accordingly.

hannes-ucsc (changes requested):
Commits are numbered incorrectly.

hannes-ucsc (approved):

hannes-ucsc:
For demo, show coverage gains in PyCharm.

Issue #2120, opened 2020-08-06
SAM registration should use dev instance, not prod


noah-aviel-dove:
Currently `AZUL_SAM_SERVICE_URL` points to `'https://sam.dsde-prod.broadinstitute.org/'`.

Noah: "Is this a mistake, then?"

Hannes: "I think so.  The TDR instance is dev so it should be using the dev instance of SAM."

hannes-ucsc:
That variable was introduced for #2025 which is still open because the demo revealed issues with it. We should either close this or make it a freebie of PR #2096.

noah-aviel-dove:
Added as a freebie to https://github.com/DataBiosphere/azul/pull/2116. Sorry for jumping on this without asking, I didn't notice it was still in Triage.

PR #2140, opened 2020-08-10
Use dev SAM instance (#2120)

noah-aviel-dove:


deployments/.example.local/environment.py
```diff
@@ -32,6 +32,8 @@ def env() -> Mapping[str, Optional[str]]:
         'AZUL_DSS_DIRECT_ACCESS': '1',
         'AZUL_DSS_DIRECT_ACCESS_ROLE': 'arn:aws:iam::109067257620:role/azul-sc',
 
+        'AZUL_SAM_SERVICE_URL': 'https://sam.dsde-dev.broadinstitute.org/',
```
hannes-ucsc, line 35: AZUL_TDR_SERVICE_URL should be defined, too.

deployments/.example.local/environment.py
```diff
@@ -32,6 +32,9 @@ def env() -> Mapping[str, Optional[str]]:
...
+        'AZUL_TDR_SERVICE_URL': 'https://jade.datarepo-dev.broadinstitute.org',
```
hannes-ucsc, line 36: In the other places, this comes first.

Issue #2150, opened 2020-08-12
Generalize DRS client


danielsotirhos:
Refactor `azul.drs.Client`:
- Constructor should not take in a url
- `Client.get_object` should get a `drs_uri` not `file_uuid`.

Issue #2147, opened 2020-08-12
`make reindex` times out when indexing all catalogs; IT often does, too


hannes-ucsc:
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/5316

The reindexing eventually finished without any errors.

```
$ python scripts/manage_queues.py list
2020-08-12 09:54:49,636 INFO    MainThread: Listing queues

Queue Name                          Messages Available  Messages In Flight  Messages Delayed 

2020-08-12 09:54:49,640 INFO    MainThread: Found credentials in shared credentials file: ~/.aws/credentials
azul-notifications-dev                      0                   0                  0         
azul-tallies-dev.fifo                       0                   0                  0         
azul-tallies_retry-dev.fifo                 0                   0                  0         
azul-tallies_fail-dev.fifo                  0                   0                  0         
azul-notifications_fail-dev                 0                   0                  0         
```

hannes-ucsc:
We'll probably just need to adjust the assumed bundles/minute rate of processing.

hannes-ucsc:
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/5340

hannes-ucsc:
Note the FIXME in the referenced commits.

hannes-ucsc:
Superseded by https://github.com/DataBiosphere/azul/issues/2617.

PR #2094, opened 2020-07-30
Clean up Terraform resource names (#2033)

natanlao:
Rename Terraform resources for consistency, and add a script that
renames existing resources when `make deploy` is called.

Issue #1764, opened 2020-05-04
Adapt /dss/files proxy to work with v2 and TDR


hannes-ucsc:
Used for direct file download in the Data Browser.

hannes-ucsc:
To get a signed URL (or an unsigned URL plus an auth header) we can use TDR's DRS implementation.

That's why this is now blocked on #1850.

hannes-ucsc:
This is urgent because it's required for DCP/2 MVP.

PR #2157, opened 2020-08-13
Generalize DRS client (#2150)

danielsotirhos:

- [x] Added `HCA` label <sub>or this PR does not target an `hca/*` branch</sub>
- [x] Added `chain` label <sub>or this PR is not the base of another PR</sub>
- [x] Made this PR a blocker of next PR in chain <sub>or this PR is not the base of another PR</sub>

- [ ] Checked `reindex` label and `[r]` prefix of commit title
- [ ] Rebased and squashed branch
- [ ] Sanity-checked history
- [ ] Build passes in `sandbox` <sub>or commented that sandbox will be skipped</sub>
- [ ] Reindexed `sandbox` <sub>or this PR does not require reindexing</sub>

- [ ] Pushed merge commit to Github and Gitlab

### DIFF
--- a/src/azul/drs.py
+++ b/src/azul/drs.py
@@ -36,6 +36,8 @@ from azul import (
     CatalogName,
     config,
     dss,
+    reject,
+    require,
 )
 from azul.types import (
     JSON,
@@ -254,13 +256,10 @@ def access_url(url):
     return {'url': url}
 
 
-class Client:
-
-    def __init__(self, url: str):
-        self.url = furl(url)
+class DRSClient:
 
     def get_object(self,
-                   object_id: str,
+                   drs_uri: str,
                    access_id: Optional[str] = None,
                    access_method: AccessMethod = AccessMethod.https
                    ) -> str:
@@ -269,12 +268,36 @@ class Client:
         depends on the access method specified.
         """
         if access_id is None:
-            return self._get_object(object_id, access_method=access_method)
+            return self._get_object(drs_uri, access_method)
         else:
-            return self._get_object_access(object_id, access_id, access_method=access_method)
+            return self._get_object_access(drs_uri, access_id, access_method)
 
-    def _get_object(self, object_id: str, access_method: AccessMethod) -> str:
-        url = self.url / object_id
+    def _get_drs_url(self, drs_uri: str, access_id: Optional[str] = None) -> str:
+        """
+        Translate a DRS URI into a DRS URL. All query params included in the DRS
+        URI (eg '{drs_uri}?version=123') will be carried over to the DRS URL.
+        Only hostname-based DRS URIs (drs://<hostname>/<id>) are supported while
+        compact, identifier-based URIs (drs://[provider_code/]namespace:accession)
+        are not.
+        """
+        parsed = furl(drs_uri)
+        require(parsed.scheme == 'drs',
+                f'The DRS URI {drs_uri} does not have the correct scheme')
+        # "The colon character is not allowed in a hostname-based DRS URI".
+        # https://ga4gh.github.io/data-repository-service-schemas/preview/develop/docs/#_drs_uris
+        # It is worth noting that compact identifier-based URI can be hard to
+        # parse when following RFC3986, with the 'namespace:accession' part
+        # matching either the heir-part or path production depending if the
+        # optional provider code and following slash is included.
+        reject(':' in parsed.netloc or ':' in str(parsed.path),
+               f'The value {drs_uri} is not a valid hostname-based DRS URI')
+        parsed.scheme = 'https'
+        object_id = one(parsed.path.segments)
+        parsed.path.set(http_object_path(object_id, access_id))
+        return parsed.url
+
+    def _get_object(self, drs_uri: str, access_method: AccessMethod) -> str:
+        url = self._get_drs_url(drs_uri)
         while True:
             response = requests.get(url)
             if response.status_code == 200:
@@ -282,8 +305,7 @@ class Client:
                 access_methods = response.json()['access_methods']
                 method = one(m for m in access_methods if m['type'] == access_method.scheme)
                 if 'access_id' in method:
-                    return self._get_object_access(object_id, method['access_id'],
-                                                   access_method=access_method)
+                    return self._get_object_access(drs_uri, method['access_id'], access_method)
                 elif 'access_url' in method:
                     return method['access_url']['url']
                 else:
@@ -296,7 +318,7 @@ class Client:
                 raise DRSError(response.status_code, response.text)
 
     def _get_object_access(self,
-                           object_id: str,
+                           drs_uri: str,
                            access_id: str,
                            access_method: AccessMethod
                            ) -> Union[requests.Response, str]:
@@ -304,7 +326,7 @@ class Client:
         For AccessMethod.gs, the gs:// URL is returned. Otherwise for
         AccessMethod.https, the object's bytes will be returned.
         """
-        url = self.url / object_id / 'access' / access_id
+        url = self._get_drs_url(drs_uri, access_id=access_id)
         while True:
             response = requests.get(url, allow_redirects=False)
             if response.status_code == 200:
--- a/test/integration_test.py
+++ b/test/integration_test.py
@@ -77,8 +77,7 @@ from azul.azulclient import (
 )
 from azul.drs import (
     AccessMethod,
-    Client,
-    http_object_path,
+    DRSClient,
 )
 import azul.dss
 from azul.es import (
@@ -328,12 +327,12 @@ class IndexingIntegrationTest(IntegrationTestCase, AlwaysTearDownTestCase):
         return rows
 
     def _test_drs(self, file_uuid: str):
-        base_url = config.service_endpoint() + http_object_path('')
-        client = Client(base_url)
+        drs = DRSClient()
         for access_method in AccessMethod:
             with self.subTest('drs', access_method=AccessMethod.https):
                 log.info('Resolving file %r with DRS using %r ...', file_uuid, access_method)
-                file_url = client.get_object(file_uuid, access_method=access_method)
+                drs_uri = f'drs://{config.api_lambda_domain("service")}/{file_uuid}'
+                file_url = drs.get_object(drs_uri, access_method=access_method)
                 log.info('Downloading file from %s ...', file_url)
                 if access_method is AccessMethod.https:
                     content = self._get_url_content(file_url)


src/azul/drs.py
```diff
@@ -254,36 +254,49 @@ def access_url(url):
     return {'url': url}
 
 
-class Client:
+@dataclass
+class ClientObject:
```
jessebrennan, line 258: I'm not necessarily objecting to it, but what exactly is gained by bundling all of the arguments into an object? The URL parsing is done here, but that could also be done in a separate method called within `get_object`.

src/azul/drs.py
```diff
@@ -254,36 +254,49 @@ def access_url(url):
...
+    drs_uri: str
+    access_id: Optional[str] = None
+    access_method: AccessMethod = AccessMethod.https
+
+    @property
+    def url(self) -> str:
```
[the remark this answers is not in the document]
danielsotirhos: Hm. Thanks I'll consider that. The trick is the URL needs to be built possibly twice if first an `access_id` is not given and then obtained from the response.

src/azul/drs.py
```diff
@@ -254,36 +254,49 @@ def access_url(url):
...
+        parsed = furl(self.drs_uri)
+        file_uuid = str(parsed.path).lstrip('/')
+        parsed.scheme = 'https'
```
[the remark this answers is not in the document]
danielsotirhos: Yes, great idea. I will add that in.

src/azul/drs.py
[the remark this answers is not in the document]
jessebrennan: Duh :facepalm:

src/azul/drs.py
danielsotirhos: The `ClientObject` class idea was born from the need to extract the netloc and `file_uuid` from the `drs_uri` and then pass those values along with the `access_id` and `access_method` variables to the private methods `_get_object()` and `_get_object_access()`. When first approaching this I stored these as instance variables in the `Client` class itself, however it started to seem a bit messy to me to be using `get_object()` as a constructor and method to return a resulting value.
I'll experiment with this again and/or see if just passing the variables to the functions separately is really that bad.

jessebrennan: > see if just passing the variables to the functions separately

 that's more what I was imagining. I can see how storing them in the client would get messy.

src/azul/drs.py
```diff
@@ -269,21 +266,30 @@ def get_object(self,
...
-        url = self.url / object_id
+            return self._get_object_access(drs_uri, access_id, access_method)
+
+    def _get_url(self, drs_uri: str, access_id: Optional[str] = None) -> str:
+        # Note the desire to include any query params found in the DRS URI
+        # (eg 'drs://{netloc}/{id}?version=123') when building the URL.
+        parsed = furl(drs_uri)
+        assert parsed.scheme == 'drs', parsed.scheme
+        parsed.scheme = 'https'
+        file_uuid = str(parsed.path).lstrip('/')
```
jessebrennan, line 279: Nitpick: while "we all know" this is a file UUID, since we're in DRS land here I think it makes more sense to call this `drs_id` or `object_id` or something like that.

src/azul/drs.py
```diff
@@ -269,21 +266,30 @@ def get_object(self,
...
-            return self._get_object(object_id, access_method=access_method)
+            return self._get_object(drs_uri, access_method)
         else:
-            return self._get_object_access(object_id, access_id, access_method=access_method)
-
-    def _get_object(self, object_id: str, access_method: AccessMethod) -> str:
-        url = self.url / object_id
+            return self._get_object_access(drs_uri, access_id, access_method)
+
+    def _get_url(self, drs_uri: str, access_id: Optional[str] = None) -> str:
```
jessebrennan, line 273: I was just reading the DRS spec, and I realized that even with your improvements, we're still only support one type of DRS URI---we're not supporting https://ga4gh.github.io/data-repository-service-schemas/preview/develop/docs/#_compact_identifier_based_drs_uris.

I think we should leave this unsupported for now, but we _may_ want to add a comment or error message if the compact identifier based URIs are used.

Feel free to leave this out though, since our DRS client is just for internal use anyway.

src/azul/drs.py
```diff
@@ -269,21 +266,30 @@ def get_object(self,
         depends on the access method specified.
         """
         if access_id is None:
```
jessebrennan, line 268: Consider calling `_get_url` here and pass the parsed URL instead. This would save 1 line of code :)

jessebrennan: To be totally clear, I mean calling `_get_url` before the `if`-`else`

danielsotirhos: Unfortunately both `_get_object()` and `_get_object_access()` need to call `_get_url()` individually since `_get_object()` can obtain an `access_id` and then call `_get_object_access()` with it which requires a new URL being generated.

danielsotirhos:
Retracting review request to fix an issue.

src/azul/drs.py
jessebrennan: Good catch.

danielsotirhos:
Taking ticket out of review to rebase on latest develop.

src/azul/drs.py
```diff
@@ -269,21 +268,38 @@ def get_object(self,
...
-            return self._get_object(object_id, access_method=access_method)
+            return self._get_object(drs_uri, access_method)
         else:
-            return self._get_object_access(object_id, access_id, access_method=access_method)
+            return self._get_object_access(drs_uri, access_id, access_method)
 
-    def _get_object(self, object_id: str, access_method: AccessMethod) -> str:
-        url = self.url / object_id
+    @classmethod
```
hannes-ucsc, line 275: Does this have to be a class method?

src/azul/drs.py
```diff
@@ -256,11 +258,8 @@ def access_url(url):
 
 class Client:
```
hannes-ucsc, line 259:
```suggestion
class DRSClient:
```

test/integration_test.py
```diff
@@ -328,12 +327,12 @@ def __check_manifest(self, file: IO[bytes], uuid_field_name: str) -> List[Mappin
         return rows
 
     def _test_drs(self, file_uuid: str):
-        base_url = config.service_endpoint() + http_object_path('')
-        client = Client(base_url)
+        client = Client()
```
hannes-ucsc, line 330:
```suggestion
        drs = DRSClient()
```

src/azul/drs.py
hannes-ucsc: And be public?

hannes-ucsc: If the motivation is in the chained PR, make it a protected instance method in this PR and then make the necessary modification in the other PR.

src/azul/drs.py
```diff
@@ -269,21 +268,38 @@ def get_object(self,
...
+            return self._get_object_access(drs_uri, access_id, access_method)
 
-    def _get_object(self, object_id: str, access_method: AccessMethod) -> str:
-        url = self.url / object_id
+    @classmethod
+    def drs_uri_to_url(cls, drs_uri: str, access_id: Optional[str] = None) -> str:
+        """
+        Translate a DRS URI into a DRS URL. All query params included in the DRS
+        URI (eg '{drs_uri}?version=123') will be carried over to the DRS URL.
+        Only Hostname-based DRS URIs (drs://<hostname>/<id>) are supported and
```
hannes-ucsc, line 280:
```suggestion
        Only hostname-based DRS URIs (drs://<hostname>/<id>) are supported while
```

src/azul/drs.py
```diff
@@ -269,21 +268,38 @@ def get_object(self,
...
+        not Compact Identifier-based URIs (drs://[provider_code/]namespace:accession)
```
hannes-ucsc, line 281:
```suggestion
        compact, identifier-based URIs (drs://[provider_code/]namespace:accession) are not.
```

hannes-ucsc (changes requested):

hannes-ucsc (approved):

hannes-ucsc (changes requested):
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/5704

hannes-ucsc:
Please also remove these commits from your branch:

* Temporarily double queue waiting time (#2147, #2189)
* Match Gitlab job timeout against that of the queue waiter
* Log # of expected bundles when waiting for queues

They landed as part of #2094.

src/azul/drs.py
```diff
@@ -269,21 +268,38 @@ def get_object(self,
...
+        compact, identifier-based URIs (drs://[provider_code/]namespace:accession)
+        are not.
+        """
+        parsed = furl(drs_uri)
+        require(parsed.scheme == 'drs',
+                f'The DRS URI {drs_uri} does not have the correct scheme')
+        reject(':' in parsed.netloc or ':' in str(parsed.path),
+               f'The value {drs_uri} is not a valid Hostname-based DRS URI')
+        parsed.scheme = 'https'
+        object_id = str(parsed.path).lstrip('/')
```
hannes-ucsc, line 289: I think you should assert that the path has only one element. Then you can avoid the string conversion and the weird lstrip.

src/azul/drs.py
```diff
@@ -269,21 +268,38 @@ def get_object(self,
...
+        Translate a DRS URI into a DRS URL. All query params included in the DRS
+        URI (eg '{drs_uri}?version=123') will be carried over to the DRS URL.
+        Only hostname-based DRS URIs (drs://<hostname>/<id>) are supported while
+        compact, identifier-based URIs (drs://[provider_code/]namespace:accession)
+        are not.
+        """
+        parsed = furl(drs_uri)
+        require(parsed.scheme == 'drs',
+                f'The DRS URI {drs_uri} does not have the correct scheme')
+        reject(':' in parsed.netloc or ':' in str(parsed.path),
```
hannes-ucsc, line 286: Why does a colon in netloc indicate the URI is not hostname based?

Same question for path.

src/azul/drs.py
```diff
@@ -269,21 +268,38 @@ def get_object(self,
...
+               f'The value {drs_uri} is not a valid Hostname-based DRS URI')
```
hannes-ucsc, line 287:
```suggestion
               f'The value {drs_uri} is not a valid hostname-based DRS URI')
```

When I requested the corresponding change to the docstring, you should have picked this up on your own.

hannes-ucsc (changes requested):

src/azul/drs.py
danielsotirhos: From: https://ga4gh.github.io/data-repository-service-schemas/preview/develop/docs/#_for_drs_clients
"The colon character is not allowed in a Hostname-based DRS URI, making it easy to tell them apart."

Hostname-based URI: `drs://<hostname>/<id>`
Compact Identifier-based URI: `drs://[provider_code/]namespace:accession`

The reason both `netloc` and `path` need to be checked for a colon is because the `provider_code` and following slash is optional in CI-based URI, which means when parsed by `furl` the colon could end up in the `netloc` or `path` depending if a `provider_code` was included or not.

hannes-ucsc: That's actually an issue with the DRS standard, making it hard to parse when following RFC3986.

https://tools.ietf.org/html/rfc3986#section-3

In `drs://foo:bar` (the variant without provider code), `foo:bar` matches the `hier-part` production of the RFC grammar.

In `drs://fubaz/foo:bar` (the variant with provider code), `foo:bar` matches the `path` production. Both are valid URIs but the fact that the `namespace` and `accession` URI components end up matching different production in the RFC grammar, depending on the case, is problematic. We may want to raise this with the author if the standard. A better alternative standard would be 

`drs://[provider_code]/namespace:accession`

hannes-ucsc (commented):
Let's discuss after stand-up.

hannes-ucsc (approved):

hannes-ucsc:
Not much to demo here, will be demoed as part of PR for #1764.

Issue #2189, opened 2020-08-19
Add throttling to BigQuery requests to prevent 403 rateLimitExceeded errors


danielsotirhos:
From https://cloud.google.com/bigquery/troubleshooting-errors:
```
rateLimitExceeded: 403

This error returns if your project exceeds a short-term rate limit by sending too many requests too quickly. For example, see the rate limits for query jobs and rate limits for API requests.

https://cloud.google.com/bigquery/quotas#query_jobs

Concurrent rate limit for interactive queries — 100 concurrent queries
Concurrent rate limit for interactive queries against Bigtable external data sources — 4 concurrent queries

https://cloud.google.com/bigquery/quotas#api_requests

API requests per second, per user — 100
Concurrent API requests, per user: 300
```

From CloudWatch `dev` logs:
```
[ERROR] Forbidden: 403 Exceeded rate limits: too many concurrent queries for this project_and_region. For more information, see https://cloud.google.com/bigquery/troubleshooting-errors
(job ID: b5e9ff04-a0c6-4b32-939a-8506eab6d534)
-----Query Job SQL Follows-----
1:
2: SELECT content, analysis_process_id, version, BYTE_LENGTH(content) AS content_size, JSON_EXTRACT_SCALAR(content, "$.schema_type") AS content_type
3: FROM hca_dev_20200817_dssPrimaryOnly.analysis_process
4: WHERE analysis_process_id = "bfd9e10b-5fa9-405c-ba1f-dc54aa296029"
5:
Traceback (most recent call last):
  File "/var/task/app.py", line 112, in new_handler
    return old_handler(self, event, context)
  File "/var/task/chalice/app.py", line 1430, in __call__
    return self.func(event_obj)
  File "/var/task/app.py", line 121, in contribute
    app.index_controller.contribute(event)
  File "/var/task/azul/indexer/index_controller.py", line 145, in contribute
    contributions = self.transform(catalog, notification, delete)
  File "/var/task/azul/indexer/index_controller.py", line 172, in transform
    bundle = self.repository_plugin(catalog).fetch_bundle(bundle_fqid)
  File "/var/task/azul/plugins/repository/tdr/__init__.py", line 101, in fetch_bundle
    bundle = self.emulate_bundle(bundle_fqid)
  File "/var/task/azul/plugins/repository/tdr/__init__.py", line 213, in emulate_bundle
    rows = future.result()
  File "/var/lang/lib/python3.8/concurrent/futures/_base.py", line 432, in result
[...]

PR #2317, opened 2020-10-02
Add explicit retries to TDR queries (#2189)

amarjandu:

- [x] Added `HCA` label <sub>or this PR does not target an `hca/*` branch</sub>
- [x] This PR is blocked by previous PR in chain <sub>or this PR is not chained to another PR</sub>
- [x] Added `chain` label to the blocking PR <sub>or this PR is not chained to another PR</sub>
- [x] Ran `make requirements_update` <sub>or this PR does not change requirements*.txt</sub>
- [x] Mentioned `make requirements` in UPGRADING.rst <sub>or this PR does not change requirements*.txt</sub>
- [x] Documented upgrade of personal deployments in UPGRADING.rst <sub>or this PR does not require upgrading</sub>

- [x] Sanity-checked history
- [x] Build passes in `sandbox` <sub>or commented that sandbox will be skipped</sub>

- [x] Moved freebies to Merged column <sub>or there are no freebies in this PR</sub> 
- [ ] Pushed merge commit to Gitlab

### DIFF
--- a/src/azul/plugins/repository/tdr/__init__.py
+++ b/src/azul/plugins/repository/tdr/__init__.py
@@ -139,7 +139,25 @@ class Plugin(RepositoryPlugin):
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        delays = (10, 20, 40, 80, None)
+        max_attempts = len(delays)
+        assert sum(delays[:-1]) < config.contribution_lambda_timeout
+        for attempt, delay in enumerate(delays, start=1):
+            query_job = self._bigquery.query(query)
+            try:
+                return query_job.result()
+            except Forbidden as e:
+                if 'Exceeded rate limits' in e.message:
+                    msg = 'Exceeded BigQuery rate limit during attempt %i/%i.'
+                    if delay is not None:
+                        log.warning(msg + ' Retrying in %is.', attempt, max_attempts, delay, exc_info=e)
+                        time.sleep(delay)
+                    else:
+                        log.warning(msg, attempt, max_attempts, exc_info=e)
+                else:
+                    raise e
+        raise RuntimeError(f'Unable to complete query after {max_attempts} '
+                           f'attempted retries: {query!r}. ')
 
     def list_links_ids(self, prefix: str) -> List[BundleFQID]:
         validate_uuid_prefix(prefix)
--- a/src/azul/plugins/repository/tdr/__init__.py
+++ b/src/azul/plugins/repository/tdr/__init__.py
@@ -139,25 +139,20 @@ class Plugin(RepositoryPlugin):
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        delays = (10, 20, 40, 80, None)
-        max_attempts = len(delays)
-        assert sum(delays[:-1]) < config.contribution_lambda_timeout
-        for attempt, delay in enumerate(delays, start=1):
-            query_job = self._bigquery.query(query)
+        delays = (10, 20, 40, 80)
+        assert sum(delays) < config.contribution_lambda_timeout
+        for attempt, delay in enumerate((*delays, None)):
+            job = self._bigquery.query(query)
             try:
-                return query_job.result()
+                return job.result()
             except Forbidden as e:
-                if 'Exceeded rate limits' in e.message:
-                    msg = 'Exceeded BigQuery rate limit during attempt %i/%i.'
-                    if delay is not None:
-                        log.warning(msg + ' Retrying in %is.', attempt, max_attempts, delay, exc_info=e)
-                        time.sleep(delay)
-                    else:
-                        log.warning(msg, attempt, max_attempts, exc_info=e)
+                if 'Exceeded rate limits' in e.message and delay is not None:
+                    log.warning('Exceeded BigQuery rate limit during attempt %i/%i. Retrying in %is.',
+                                attempt + 1, len(delays), delay, exc_info=e)
+                    time.sleep(delay)
                 else:
                     raise e
-        raise RuntimeError(f'Unable to complete query after {max_attempts} '
-                           f'attempted retries: {query!r}. ')
+        assert False
 
     def list_links_ids(self, prefix: str) -> List[BundleFQID]:
         validate_uuid_prefix(prefix)
--- a/src/azul/plugins/repository/tdr/__init__.py
+++ b/src/azul/plugins/repository/tdr/__init__.py
@@ -148,7 +148,7 @@ class Plugin(RepositoryPlugin):
             except Forbidden as e:
                 if 'Exceeded rate limits' in e.message and delay is not None:
                     log.warning('Exceeded BigQuery rate limit during attempt %i/%i. Retrying in %is.',
-                                attempt + 1, len(delays), delay, exc_info=e)
+                                attempt + 1, len(delays) + 1, delay, exc_info=e)
                     time.sleep(delay)
                 else:
                     raise e


amarjandu:
Tested out on my deployment with prefix 0 to observe the retries.
Confirmed IT passes on personal deployment.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -140,7 +140,27 @@ def _bigquery(self) -> bigquery.Client:
...
+                    msg = ('Exceeded BigQuery rate limit, attempting explicit retry after %.3f '
+                           'seconds. Retry %s of %s.')
+                    log.warning(msg, delay, attempt, len(backoff_intervals), exc_info=e)
+                    time.sleep(delay)
+                else:
+                    raise e
+            else:
+                return query_job
+        raise Exception(f'Unable to complete requested query: {query!r}. '
+                        f'Attempted to retry {len(backoff_intervals)} times.')
```
noah-aviel-dove, line 163: If the query is very long, it may be difficult to correlate the first part of this message with the second. I suggest:

```suggestion
        raise RuntimeError(f'Unable to complete query after {len(backoff_intervals)} '
                           f'attempted retries: {query!r}. ')
```

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -140,7 +140,27 @@ def _bigquery(self) -> bigquery.Client:
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        attempt = 0
+        backoff_intervals = (10, 20, 40, 80)
```
noah-aviel-dove, line 144: You have the option of storing the length in a variable, which could be re-used throughout the method on place of the calls to `len`. This would also facilitate expressing the backoffs formulaically as `tuple(10*2**i for i in range(num_retries))`. But I'll leave it up to you, I might be over-engineering here.

noah-aviel-dove (changes requested):

src/azul/plugins/repository/tdr/__init__.py
amarjandu: i was suggested to switch over from a formula to a clearly defined set of tuples, but I'll take up your suggestion of storing the len in a var.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -140,7 +140,27 @@ def _bigquery(self) -> bigquery.Client:
...
+        assert (sum(backoff_intervals) <= config.contribution_lambda_timeout)
+        while attempt < len(backoff_intervals):
+            query_job = self._bigquery.query(query)
+            try:
+                query_job.result()
+            except Forbidden as e:
+                if 'Exceeded rate limits' in e.message:
+                    delay = backoff_intervals[attempt]
+                    attempt += 1
```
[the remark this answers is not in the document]
amarjandu: This is much cleaner thanks for the suggestion :)

src/azul/plugins/repository/tdr/__init__.py
amarjandu: I was planning on using the suggestion of a variable for the length, but after incorporating the suggestions below, it would only be used twice. I'm going to keep it out.

noah-aviel-dove: Yeah, var isn't worth it if you're sticking with the tuple literal. All good!

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -140,7 +140,23 @@ def _bigquery(self) -> bigquery.Client:
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        backoff_intervals = (10, 20, 40, 80)
+        assert (sum(backoff_intervals) < config.contribution_lambda_timeout)
```
noah-aviel-dove, line 144: Whoops, I thought I had already mentioned this. Redundant parentheses:
```suggestion
        assert sum(backoff_intervals) < config.contribution_lambda_timeout
```

noah-aviel-dove (changes requested):

noah-aviel-dove (approved):
Looks good!

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -140,7 +140,23 @@ def _bigquery(self) -> bigquery.Client:
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        backoff_intervals = (10, 20, 40, 80)
```
hannes-ucsc, line 143: variable name too long

hannes-ucsc (changes requested):
I've looked at `.result()` for a bit and I am unsure if this implementation eagerly reads all the rows or if it's still lazy as we'd like it to be. I'd like yours and @noah-aviel-dove's opinion on this. I think debugging this might be the best option.

```diff
Index: src/azul/plugins/repository/tdr/__init__.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/azul/plugins/repository/tdr/__init__.py	(revision 0defc7d32e3885f6cca0663b39edf0e143cd4374)
+++ src/azul/plugins/repository/tdr/__init__.py	(date 1602291792820)
@@ -131,22 +131,22 @@
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        backoff_intervals = (10, 20, 40, 80)
-        assert sum(backoff_intervals) < config.contribution_lambda_timeout
-        for attempt, delay in enumerate(backoff_intervals, start=1):
+        delays = (10, 20, 40, 80)
+        assert sum(delays) < config.contribution_lambda_timeout
+        for attempt, delay in enumerate(delays, start=1):
             query_job = self._bigquery.query(query)
             try:
                 query_job.result()
             except Forbidden as e:
                 if 'Exceeded rate limits' in e.message:
-                    log.warning('Exceeded BigQuery rate limit, attempting explicit retry after %i '
-                                'seconds. Retry %i of %i.', delay, attempt, len(backoff_intervals), exc_info=e)
+                    log.warning('Exceeded BigQuery rate limit during attempt %i/%i. '
+                                'Retrying in %is.', attempt, len(delays), delay, exc_info=e)
                     time.sleep(delay)
                 else:
                     raise e
             else:
                 return query_job
-        raise RuntimeError(f'Unable to complete query after {len(backoff_intervals)} '
+        raise RuntimeError(f'Unable to complete query after {len(delays)} '
                            f'attempted retries: {query!r}. ')
 
     def list_links_ids(self, prefix: str) -> List[BundleFQID]:
```

noah-aviel-dove:
@hannes-ucsc `iter(query_job)` just calls `iter(query_job.result())` so there's no change in lazyness.

https://github.com/googleapis/python-bigquery/blob/35627d145a41d57768f19d4392ef235928e00f72/google/cloud/bigquery/job.py#L3409

amarjandu:
By debugging the code we have confirmed that this change does not effect laziness.

hannes-ucsc:
> By debugging the code we have confirmed that this change does not effect laziness.

And that it IS actually lazy. We also identified one more improvement: that we should return `query_job.result()` instead of `query_job` to avoid the one unnecessary instantiation of a iterator.

hannes-ucsc (changes requested):
Assume BQ is really overloaded and always returns with the `Exceeded rate limits` error. The following steps will be taken by this implementation: 

```
make attempt, log 1/4, wait 10s
make attempt, log 2/4, wait 20s
make attempt, log 3/4, wait 40s
make attempt, log 4/4, wait 80s
```

Note that no attempt is made after waiting 80s. There is no point in waiting 80s only to then exit the loop and raise an exception.

hannes-ucsc:
Ideally, peer reviewer should have caught this.

amarjandu:
@hannes-ucsc I reverted back to the while loop to ensure the nth attempt is being executed. The failure to catch this was on my part as it was a suggestion, i should have confirmed that the changes behave the same way as we wanted.  

Due to this change in the number of attempts that are made I do think there is a potential source of confusion when presenting the warnings `Exceeded BigQuery rate limit during attempt %i/%i`. 
Currently the last run (`n/n`) would not be logged as the previous attempts due to how the control flow breaks. The last run will fail with a `[ERROR] Unable to complete query after {n} attempted retries... `. 

```
running something, Failed attempt 1/5, retrying in 10
running something, Failed attempt 2/5, retrying in 20
running something, Failed attempt 3/5, retrying in 40
running something, Failed attempt 4/5, retrying in 80
running something, raise error
```

I thought it might be annoying/confusing to not see the last message in the logs, so i opted to rework the warning messages to include the `nth` warning message. If you think this is not something to be concerned about I can drop the commit.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -131,7 +131,33 @@ def _bigquery(self) -> bigquery.Client:
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        attempt = 0
+        delays = (10, 20, 40, 80)
```
amarjandu, line 135: some of the weirdness can be addressed if this is prepended with a 0.

amarjandu: added place holder of 0 in the first value, removes need increment by 1 in some locations.

amarjandu: Moving back to an equation that calculates the delay can simply the code more.

hannes-ucsc: I prefer sticking with the tuple and reworking the control flow instead.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -131,7 +131,31 @@ def _bigquery(self) -> bigquery.Client:
...
+            try:
+                return query_job.result()
+            except Forbidden as e:
+                if 'Exceeded rate limits' in e.message:
+                    msg = 'Exceeded BigQuery rate limit during attempt %i/%i.'
+                    if attempt == max_attempts:
+                        log.warning(msg, attempt, max_attempts, exc_info=e)
+                    else:
+                        log.warning(msg + ' Retrying in %is.',
+                                    attempt,
```
hannes-ucsc, line 150: Don't need to wrap all in log calls.

hannes-ucsc (changes requested):
Try to simplify. Consider a `None` sentinel at the end of `delays` and stick with original approach of iterating over `enumerate(delays)`.

noah-aviel-dove:
> Ideally, peer reviewer should have caught this.

I'm confused. Did my suggestion of switching from a `while` to a `for` loop accidentally break Amar's original implementation? Or are you saying it was my responsibility to decide that the loop should behave differently than it was originally implemented?

hannes-ucsc (commented):
I'll take a stab at this as my fun activity of the day.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -131,7 +131,25 @@ def _bigquery(self) -> bigquery.Client:
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        delays = (10, 20, 40, 80, None)
+        max_attempts = len(delays)
+        assert sum(delays[:-1]) < config.contribution_lambda_timeout
+        for attempt, delay in enumerate(delays, start=1):
```
hannes-ucsc, line 137: A 1-based loop variable is cardinal sin.

hannes-ucsc (commented):
How's this?

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -131,7 +131,25 @@ def _bigquery(self) -> bigquery.Client:
             return bigquery.Client(project=self._source.project)
 
     def _run_sql(self, query: str) -> BigQueryRows:
-        return self._bigquery.query(query)
+        delays = (10, 20, 40, 80, None)
+        max_attempts = len(delays)
+        assert sum(delays[:-1]) < config.contribution_lambda_timeout
```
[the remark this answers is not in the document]
amarjandu: `enumerate((*delays, None))` is clever to get around this

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -131,7 +131,25 @@ def _bigquery(self) -> bigquery.Client:
...
+                if 'Exceeded rate limits' in e.message:
+                    msg = 'Exceeded BigQuery rate limit during attempt %i/%i.'
+                    if delay is not None:
+                        log.warning(msg + ' Retrying in %is.', attempt, max_attempts, delay, exc_info=e)
+                        time.sleep(delay)
+                    else:
+                        log.warning(msg, attempt, max_attempts, exc_info=e)
+                else:
+                    raise e
+        raise RuntimeError(f'Unable to complete query after {max_attempts} '
```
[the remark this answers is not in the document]
amarjandu: It seems like a good general case when calling on external libraries,  so that the original exception can be handled by the caller. An additional benefit is that that the caller does not have to additionally look for the modified exception, and have to figure out what went on.

src/azul/plugins/repository/tdr/__init__.py
amarjandu: Almost lost to temptation there. 👀

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -131,7 +131,20 @@ def _bigquery(self) -> bigquery.Client:
...
+            try:
+                return job.result()
+            except Forbidden as e:
+                if 'Exceeded rate limits' in e.message and delay is not None:
+                    log.warning('Exceeded BigQuery rate limit during attempt %i/%i. Retrying in %is.',
+                                attempt + 1, len(delays), delay, exc_info=e)
+                    time.sleep(delay)
+                else:
+                    raise e
+        assert False
```
amarjandu, line 147: It seems that this `assert False` here just to ensure we never exit without either: 
1. returning a query_job with a valid result
2. raise an exception/error back to the caller 

But even in the case where we set the delays to `()` and there is no job.result(),  ln 146 ` raise e ` should  still be hit, is this an intentional extra precaution?

hannes-ucsc: Yes. The `assert False` should never be hit. That's exactly why it is there. Assertions are used to communicate to readers of the code which invariants the developer intends to establish. With this assert I am communicating to you that the loop never exits naturally.

hannes-ucsc: Yes. The `assert False` should never be hit. That's exactly why it is there. Assertions are used to communicate to readers of the code which invariants the developer intends to establish. With this assert I am communicating to you that the loop never exits naturally.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -140,7 +140,7 @@ def _run_sql(self, query: str) -> BigQueryRows:
             except Forbidden as e:
                 if 'Exceeded rate limits' in e.message and delay is not None:
                     log.warning('Exceeded BigQuery rate limit during attempt %i/%i. Retrying in %is.',
-                                attempt + 1, len(delays), delay, exc_info=e)
+                                attempt + 1, len(delays) + 1, delay, exc_info=e)
```
amarjandu, line 143: @hannes-ucsc I changed the warnings to ensure the number of retries in the log looks correct.

do something; attempt 1/5; wait 10s 
do something; attempt 2/5; wait 20s 
do something; attempt 3/5; wait 40s 
do something; attempt 4/5; wait 80s 
do something; exception

rather than `attempt 4/4` and there is no mention of the 5th retry.

hannes-ucsc: Looks good.

hannes-ucsc (approved):

hannes-ucsc:
For demo, show a bunch of the retry messages in the logs for `dev`.

amarjandu:
https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/jobs/7067

Issue #2345, opened 2020-10-12
`make requirements_update` yields inconsistent diffs to `requirements*.trans` between machines


hannes-ucsc:
`make requirements_update` may not produce the same changes to the transitive files between machines if the image layer caches reflect different states on pypy. This led to the [reopening](https://github.com/databiosphere/azul/issues/2261#issuecomment-707278386) of #2261.

To repro run `make requirements_update` on machine A. Run `git diff > a1`. Wait a week. Run it on machine B. Run `git diff > b` on B. Then run it again on machine A. `git diff > a2`. The second run on A will reuse the cached layers from the first run on that machine. So a2 == a1 but `b != a2`. Ideally, `b == a2`.

hannes-ucsc:
The solution is to add the ability to run `make requirements_update` without image layer caching enabled.

natanlao:
Selectively disabling cache for a single layer [isn't supported directly in Docker](https://github.com/moby/moby/issues/1996) but we can get close by:

1. Adding a build argument (say `ARG INVALIDATE_CACHE`) directly before the last command in the Dockerfile.
2. Modifying the Makefile to pass a default value to `docker build` for that build argument, unless an environment variable is set.

Requirements could then be regenerated without the problematic cache layer like so:

```console
$ INVALIDATE_CACHE=$(date +%s) make requirements_update
```

`$(date +%s)` being seconds since epoch so that a new, uncached value for INVALIDATE_CACHE is "always" provided to `docker run`.

PR #2360, opened 2020-10-19
Optionally disable caching for `make requirements_update` (#2345)

natanlao:

- [x] Added `DCP/1` label <sub>or this PR does not target an `hca/*` branch</sub>
- [x] This PR is blocked by previous PR in chain <sub>or this PR is not chained to another PR</sub>
- [x] Added `chain` label to the blocking PR <sub>or this PR is not chained to another PR</sub>
- [x] Ran `make requirements_update` <sub>or this PR does not change requirements*.txt</sub>
- [x] Mentioned `make requirements` in UPGRADING.rst <sub>or this PR does not change requirements*.txt</sub>
- [x] Documented upgrade of personal deployments in UPGRADING.rst <sub>or this PR does not require upgrading</sub>

- [x] Sanity-checked history
- [x] Build passes in `sandbox` <sub>or added `no sandbox` label</sub>

- [ ] Moved freebies to Merged column <sub>or there are no freebies in this PR</sub> 
- [ ] Pushed merge commit to Gitlab <sub>or this changes can be pushed later, together with another PR</sub>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ include common.mk
 
 DOCKER_IMAGE ?= azul
 DOCKER_TAG ?= latest
+CACHE_SEED ?=
 
 .PHONY: virtualenv
 virtualenv: check_env
@@ -34,6 +35,7 @@ define docker
 docker$1: check_docker
 	docker build \
 		--build-arg make_target=requirements$2 \
+		--build-arg cache_seed=${CACHE_SEED} \
 		-t $$(DOCKER_IMAGE)$3:$$(DOCKER_TAG) \
 		.
 
@@ -65,6 +67,10 @@ requirements_update: check_venv check_docker
 	python scripts/manage_requirements.py \
 		--image=$(DOCKER_IMAGE)/deps:$(DOCKER_TAG) \
 		--build-image=$(DOCKER_IMAGE)/dev-deps:$(DOCKER_TAG)
+
+.PHONY: requirements_update_force
+requirements_update_force: check_venv check_docker
+	CACHE_SEED=$$(python -c 'import uuid; print(uuid.uuid4())') $(MAKE) requirements_update
 
 .PHONY: hello
 hello: check_python
--- a/README.md
+++ b/README.md
@@ -242,16 +242,19 @@ deployments with  `_select`.
    config file:
 
    ```
-   $ mkdir /Users/alice/.gcp
-   $ mv /Users/alice/Downloads/example-project-name_key.json /Users/alice/.gcp/
+   mkdir /Users/alice/.gcp
+   mv /Users/alice/Downloads/example-project-name_key.json /Users/alice/.gcp/
    ```
 
-6. Edit the `environment.local.py` file for your personal deployment and
-   modify the `GOOGLE_APPLICATION_CREDENTIALS` variable:
+6. Edit the `environment.local.py` file for your personal deployment:
 
    ```
-   $ vim /Users/alice/azul/deployments/alice.local/environment.local.py
-   
+   vim /Users/alice/azul/deployments/alice.local/environment.local.py
+   ```
+
+   and modify the `GOOGLE_APPLICATION_CREDENTIALS` variable:
+
+   ```
    'GOOGLE_APPLICATION_CREDENTIALS': '/Users/alice/.gcp/example-project-name_key.json'
    ```
 
@@ -259,10 +262,15 @@ deployments with  `_select`.
    create a symlink to your deployment's `environment.local.py` file:
 
    ```
-   $ cd /Users/alice/azul/deployments/dev
-   $ vim environment.local.py
-     (or)
-   $ ln -snf ../alice.local/environment.local.py environment.local.py
+   cd /Users/alice/azul/deployments/dev
+   vim environment.local.py
+   ```
+
+   or, alternatively:
+
+   ```
+   cd /Users/alice/azul/deployments/dev
+   ln -snf ../alice.local/environment.local.py environment.local.py
    ```
 
    As always, you can also create an `environment.local.py` in the project
--- a/README.md
+++ b/README.md
@@ -746,6 +746,17 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
 updated to reflect that change. To fix this, redeploy the lambdas (`make 
 package`) in the affected deployment.
 
+## `make requirements_update` does not update transitive requirements
+
+In some cases, `make requirements_update` might not produce any updates to
+transitive requirements, even if you expect them. For example, a sandbox build
+on Gitlab might identify updated transitive requirements even though doing `make
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update_force` instead.
+
 
 ##  Unable to re-register service account with SAM
 
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ generic with minimal need for project-specific behavior.
 - The `bash` shell
 
 - [Docker] for running the tests (the community edition is sufficient).
-  The minimal required version is uncertain, but 19.03, 18.09, and 17.09 are 
+  The minimal required version is uncertain, but 19.03, 18.09, and 17.09 are
   known to work.
 
 - Terraform (optional, to create new deployments). Refer the official
@@ -139,16 +139,16 @@ end.
    make requirements
    ```
 
-   Linux users whose distribution does not offer Python 3.8 should consider 
-   installing [pyenv] and then Python 3.8 using `pyenv install 3.8.3` and 
-   setting `PYENV_VERSION` to `3.8.3`. You may need to update pyenv itself 
+   Linux users whose distribution does not offer Python 3.8 should consider
+   installing [pyenv] and then Python 3.8 using `pyenv install 3.8.3` and
+   setting `PYENV_VERSION` to `3.8.3`. You may need to update pyenv itself
    before it recognizes the given Python version. Even if a distribution
    provides the  required minor version of Python natively, using pyenv is
    generally preferred because it offers every patch-level release of Python,
    supports an arbitrary number of different Python versions to be installed
    concurrently and allows for easily switching between them.
 
-   Ubuntu users using their system's default Python 3.8 installation must 
+   Ubuntu users using their system's default Python 3.8 installation must
    install `python3-dev` before the wheel requirements can be built.
 
    ```
@@ -175,17 +175,17 @@ end.
 
 5. Run `make`. It should say `Looking good!` If one of the check target fails,
    address the failure and repeat. Most check targets are defined in `common.mk`.
-   
-6. Make sure Docker works without explicit root access. Run the following 
+
+6. Make sure Docker works without explicit root access. Run the following
    command *without `sudo`*:
-   
+
    ```
    docker ps
    ```
-   
+
    If that fails, you're on your own.
 
-8. Finally, confirm that everything is configured properly on your machine by 
+8. Finally, confirm that everything is configured properly on your machine by
    running the unit tests:
 
    ```
@@ -195,7 +195,7 @@ end.
 
 ### 2.3.1 AWS credentials
 
-You should have been issued AWS credentials. Typically, those credentials 
+You should have been issued AWS credentials. Typically, those credentials
 require assuming a role in an account other than the one defining your IAM
 user.  Just set that up normally in `~/.aws/config` and `~/.aws/credentials`.
 If the  assumed role additionally requires an MFA token, you should run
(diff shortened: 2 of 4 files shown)


jessebrennan (commented):
The code and documentation looks good.

I'm imagining a potentially problematic scenario. Suppose the local cache is older than the version that's in the sandbox build, then a transitive dependency is updated before we update our local cache. This means that the version of that dep is newer than the one stored in the cache in the sandbox. Doesn't this mean the sandbox build will fail? How can we run `INVALIDATE_CACHE` on the sandbox to fix this?

Also I think this will have a slightly strange behaviour. If you run `make requirements_update` again, right after running it with `INVALIDATE_CACHE`, the second build won't use the cache, right? If you run it a third time though, I think it will. This isn't breaking anything, but I wouldn't expect this behaviour.

natanlao:
> Suppose the local cache is older than the version that's in the sandbox build, then a transitive dependency is updated before we update our local cache. This means that the version of that dep is newer than the one stored in the cache in the sandbox. Doesn't this mean the sandbox build will fail? How can we run INVALIDATE_CACHE on the sandbox to fix this?

This might be a question for Hannes. If I'm understanding this correctly, the non-transitive requirements remain the same through this scenario, and both the sandbox and local machine agree on what the set of dependencies are. There would/should be no reason to update the local cache in the first place.

> If you run make requirements_update again, right after running it with INVALIDATE_CACHE, the second build won't use the cache, right? If you run it a third time though, I think it will. This isn't breaking anything, but I wouldn't expect this behaviour.

If I understand correctly, you're referring to this scenario:

```console
$ INVALIDATE_CACHE=$(date +%s) make requirements_update
$ make requirements_update
$ make requirements_update
```

In this case:

* The first invocation will ignore the current build cache no matter what, but also store the current build context in the cache.
* The second invocation will use the build cache if...
  1. The previous invocation of `make requirements_update` did not result in any changes to any of the requirements files.
  2. `make requirements_update` was run some time prior to the previous invocation with the same set of requirements files.
* The third invocation will use the build cache under the same criteria as the second invocation.

I think this behavior (detached from how I implemented) makes sense. That said, I think naming the flag `INVALIDATE_CACHE` may not be the clearest choice since it implies that the cache is being emptied when that's not the case. I'll add a fixup that changes this to `IGNORE_CACHE`.

jessebrennan (approved):
Which ever name you choose is :+1: with me.

You alleviated my previous concerns on a call by reassuring me that the changes to `requirements.trans.txt` would cause a cash miss in the sandbox, and requirements would need to be re-generated.

Makefile
```diff
@@ -5,6 +5,7 @@ include common.mk
 
 DOCKER_IMAGE ?= azul
 DOCKER_TAG ?= latest
+IGNORE_CACHE ?= False
```
hannes-ucsc, line 8: `False` could be misconstrued by a developer think they need to set this to True. Can we use empty string as a default.

hannes-ucsc, line 8:
```suggestion
CACHE_SEED ?= ''
```

README.md
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
...
+might identify updated transitive requirements even though doing `make
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update` with caching disabled for the requirements generation
+step:
+
+```console
+$ IGNORE_CACHE=$(date +%s) make requirements_update
```
hannes-ucsc, line 762: System time is not a reliable source of unique values. Think daylight saving fall back and ntpd slowly adjusting clock skew and so on. We know python is installed so why not `import uuid; print(uuid.uuid4())`

hannes-ucsc (changes requested):

README.md
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
...
+transitive requirements, even if you expect them. For example, a sandbox build
+might identify updated transitive requirements even though doing `make
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update` with caching disabled for the requirements generation
+step:
+
+```console
```
hannes-ucsc, line 761: For the sake of consistency, I'd leave `console` off. It doesn't buy as anything, if we go with the _force target because there is nothing left to syntax-highlight. I'd welcome one commit that changes all of the preformatted blocks to use this where appropriate.

README.md
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
...
+$ CACHE_SEED=$(python -c 'import uuid; print(uuid.uuid4())') make requirements_update
```
hannes-ucsc, line 762:
```suggestion
$ make requirements_update_force
```

The corresponding recipe could just recursively invoke make. Look for `$(MAKE)` in the Makefile.

Makefile
```diff
@@ -5,6 +5,7 @@ include common.mk
 
 DOCKER_IMAGE ?= azul
 DOCKER_TAG ?= latest
+CACHE_SEED ?= ''
```
hannes-ucsc, line 8: Sure this is the right syntax for creating an empty string, and doesn't create string with two single quotes in it?

hannes-ucsc (changes requested):
No bump.

README.md
natanlao: Just to be clear, when you say

> I'd welcome one commit that changes all of the preformatted blocks to use this where appropriate.

are you instructing me to

* Adjust all preformatted blocks to ensure there is no explicit syntax-highlighting directive

* Adjust all preformatted blocks to remove the `console` syntax-highlighting directive where it exists

* Adjust all preformatted blocks to add syntax-highlighting directives where appropriate

I am assuming the first option, but I just wanted to double check to make sure.

hannes-ucsc: To be precise, I was only asking you to remove `console` here. If you WANT, you can add a separate commit that adds syntax-highlighting to all pre-formatted blocks in this file but there is no "instruction" to do so.

natanlao: Okay, thanks. I'm happy to make that adjustment.

hannes-ucsc: In fact, I changed my mind. I just looked at the file. There are so many preformatted blocks with short commands where syntax-highlighting would not add any clarity or ease of use. So don't do that separate commit, only remove `console` here to be consistent with the rest of the file.

natanlao: Ah, I'm sorry -- I didn't see your comment where you said

> But don't feel obliged.

My apologies.

hannes-ucsc: No problem.

Makefile
hannes-ucsc: I saw the fixup but would still like an answer.

README.md
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
...
+## `make requirements_update` does not update transitive requirements
+
+In some cases, `make requirements_update` might not produce any updates to
+transitive requirements, even if you expect them. For example, a sandbox build
+might identify updated transitive requirements even though doing `make
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update` with caching disabled for the requirements generation
```
hannes-ucsc, line 758: This is a bit awkward, since it seems to contradict the command below.

README.md
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
 updated to reflect that change. To fix this, redeploy the lambdas (`make 
 package`) in the affected deployment.
 
+## `make requirements_update` does not update transitive requirements
+
+In some cases, `make requirements_update` might not produce any updates to
+transitive requirements, even if you expect them. For example, a sandbox build
```
hannes-ucsc, line 752:
```suggestion
transitive requirements, even if you expect them. For example, a sandbox build on Gitlab
```

README.md
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
...
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update` with caching disabled for the requirements generation
+step:
+
+```
+$ make requirements_update_force
+```
```
hannes-ucsc, line 763:
```suggestion
This is a side effect of the Docker build cache on two different machines
diverging to reflect different states on PyPI. This can be fixed by running
`make requirements_update_force` instead.

File: README.md [hunk 9]
```diff
@@ -746,6 +746,22 @@ Check whether the DSS switched buckets. If so, the lambda policy may need to be
...
+might identify updated transitive requirements even though doing `make
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update` with caching disabled for the requirements generation
+step:
+
+```
+$ make requirements_update_force
```
hannes-ucsc, line 762: I don't think we use the $ convention in this file. Correct me if I am wrong.

hannes-ucsc (changes requested):

README.md
natanlao: It seems that the majority of the file does not include `$`, but with exceptions:
* https://github.com/DataBiosphere/azul/blame/develop/README.md#L245
* https://github.com/DataBiosphere/azul/blame/develop/README.md#L918
* https://github.com/DataBiosphere/azul/blame/develop/README.md#L1681

I'll add a commit that removes these for consistency. I'll also add, separately, a commit that strips trailing whitespace from the file (that can be dropped if it's not important).

README.md
```diff
@@ -737,28 +744,43 @@ configuration bucket.
...
+on Gitlab might identify updated transitive requirements even though doing `make
+requirements_update` locally doesn't.
+
+This is a side effect of the Docker build cache on two different machines
+diverging to reflect different states on PyPI. This can be fixed by running
+`make requirements_update_force` instead.
+
+```
+make requirements_update_force
+```
```
hannes-ucsc, line 769: This is redundant. My previous comment https://github.com/DataBiosphere/azul/pull/2360#discussion_r512863923 included the removal.

README.md
```diff
@@ -242,27 +242,34 @@ deployments with  `_select`.
...
    ```
-   $ cd /Users/alice/azul/deployments/dev
-   $ vim environment.local.py
-     (or)
-   $ ln -snf ../alice.local/environment.local.py environment.local.py
+   cd /Users/alice/azul/deployments/dev
+   vim environment.local.py
+   ```
+
+   or, alternatively:
```
hannes-ucsc, line 269: This makes it look like the `cd` is only part of the first alternative. Just duplicate it in the second.

README.md
```diff
@@ -242,27 +242,34 @@ deployments with  `_select`.
    config file:
 
    ```
-   $ mkdir /Users/alice/.gcp
```
hannes-ucsc, line 245: The title of this commit should include `…. in README`

README.md
```diff
@@ -97,7 +97,7 @@ generic with minimal need for project-specific behavior.
 - The `bash` shell
 
 - [Docker] for running the tests (the community edition is sufficient).
-  The minimal required version is uncertain, but 19.03, 18.09, and 17.09 are
```
hannes-ucsc, line 100: The title of this commit should include `…. in README`

hannes-ucsc (changes requested):

README.md
natanlao: My oversight.

README.md
```diff
@@ -242,27 +242,35 @@ deployments with  `_select`.
...
+   mkdir /Users/alice/.gcp
+   mv /Users/alice/Downloads/example-project-name_key.json /Users/alice/.gcp/
    ```
 
-6. Edit the `environment.local.py` file for your personal deployment and
-   modify the `GOOGLE_APPLICATION_CREDENTIALS` variable:
+6. Edit the `environment.local.py` file for your personal deployment:
 
    ```
    $ vim /Users/alice/azul/deployments/alice.local/environment.local.py
```
hannes-ucsc, line 252:
```suggestion
   vim /Users/alice/azul/deployments/alice.local/environment.local.py
```

hannes-ucsc (changes requested):

hannes-ucsc (approved):

hannes-ucsc:
For demo, run target once, run again and show that cache is used. Run a 3rd time with force and show that cache is not being used. Run a fourth time without force to show that cache is used again. Run a 5th time to show that cache is not used.

natanlao:
Why wouldn't the cache be used on the fifth time? By my understanding, assuming no initial cache:

1. `make requirements_update` sees a cache miss, generates new requirements.
2. `make requirements_update` sees a cache hit, returns previously-generated requirements.
3. `make requirements_update_force` ignores cache, generates new requirements without checking cache.
4. `make requirements_update` sees a cache hit assuming step 3 produces same requirements as step 2.
5. `make requirements_update` sees a cache hit assuming step 4 produces same requirements as step 2 or 3.

Issue #2416, opened 2020-11-04
CGM stratification tree building needs more unit test coverage


hannes-ucsc:
The current test handles exactly one stratification, there are many more that should be covered.

PR #2432, opened 2020-11-06
Add CGM stratification tree unit tests (#2416)

danielsotirhos:
Author (check before every review)

- [x] Added `DCP/1` label                                   <sub>or this PR does not target an `hca/*` branch</sub>
- [x] Created issue to track porting these changes          <sub>or this PR does not need porting</sub> 
- [x] Rebased branch on `develop`

- [x] Sanity-checked history
- [x] Reindexed `sandbox`                                   <sub>or this PR does not require reindexing</sub>

- [ ] Commented on demo expectations                        <sub>or labeled as `no demo`</sub>
- [ ] Reindexed `dev`                                       <sub>or this PR does not require reindexing</sub>

### DIFF
--- a/src/azul/plugins/metadata/hca/contributor_matrices.py
+++ b/src/azul/plugins/metadata/hca/contributor_matrices.py
@@ -1,0 +1,111 @@
+from collections import (
+    defaultdict,
+)
+from itertools import (
+    chain,
+    product,
+)
+
+from furl import (
+    furl,
+)
+
+from azul import (
+    config,
+)
+from azul.collections import (
+    NestedDict,
+)
+
+
+class ContributorMatrices:
+
+    @classmethod
+    def make_contributor_matrices(cls, catalog, files):
+        """
+        >>> from azul.doctests import assert_json
+        >>> def f(files):
+        ...     return assert_json(ContributorMatrices.make_contributor_matrices('dcp', files))
+
+        >>> f([{'uuid':'1', 'version': '2', 'name': 3, 'strata': "y=5,6;x=4\\nx=7;y=8"}])
+        {
+            "x": {
+                "4": {
+                    "y": {
+                        "5": {
+                            "url": [
+                                "https://service.dev.singlecell.gi.ucsc.edu/fetch/repository/files/1?version=2&catalog=dcp"
+                            ]
+                        },
+                        "6": {
+                            "url": [
+                                "https://service.dev.singlecell.gi.ucsc.edu/fetch/repository/files/1?version=2&catalog=dcp"
+                            ]
+                        }
+                    }
+                },
+                "7": {
+                    "y": {
+                        "8": {
+                            "url": [
+                                "https://service.dev.singlecell.gi.ucsc.edu/fetch/repository/files/1?version=2&catalog=dcp"
+                            ]
+                        }
+                    }
+                }
+            }
+        }
+        """
+        for key in 'uuid', 'name':
+            assert len(set(file[key] for file in files)) == len(files), files
+
+        files = [
+            {
+                # Each line in the stratification string represents a stratum,
+                # each stratum is a list of points, each point has a dimension
+                # and a list of values. Transform that string into a list of
+                # dictionaries. Each entry in those dictionaries maps the
+                # dimension to a value in that dimension. If dimension in a
+                # stratum has multiple values, the stratum is expanded into
+                # multiple strata, one per value. The strata are identical
+                # except in the dimension that had the multiple values.
+                'strata': list(chain.from_iterable(
+                    map(dict, product(*(
+                        [(dimension, value) for value in values.split(',')]
+                        for dimension, values in (point.split('=') for point in stratum.split(';'))
+                    )))
+                    for stratum in file['strata'].split('\n')
+                )),
+                'url': furl(config.service_endpoint(),
+                            path=('fetch', 'repository', 'files', file['uuid']),
+                            args=dict(version=file['version'],
+                                      catalog=catalog))
+            }
+            for file in files
+        ]
+
+        # To produce a tree with the most shared base branches possible we sort
+        # the dimensions by number of distinct values on that dimension.
+        distinct_values = defaultdict(set)
+        for file in files:
+            for stratum in file['strata']:
+                for dimension, value in stratum.items():
+                    distinct_values[dimension].add(value)
+        sorted_dimensions = sorted(distinct_values,
+                                   key=lambda k: len(distinct_values[k]))
+
+        # Build the tree, as a nested dictionary. The keys in the dictionary
+        # alternate between dimensions and values. The leaves of the tree are
+        # lists of matrix file URLs. If a matrix covers multiple strata, its
+        # URL will occur multiple times in the tree.
+        tree = NestedDict(2 * len(sorted_dimensions), list)
+        for file in files:
+            for stratum in file['strata']:
+                node = tree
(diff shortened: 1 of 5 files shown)


hannes-ucsc (changes requested):
You must not have seen the doctests I added. Please add to that.

src/azul/service/hca_response_v5.py
```diff
@@ -412,6 +412,11 @@ def _make_contributor_matrices(cls, catalog, files):
         sorted_dimensions = sorted(distinct_values,
                                    key=lambda k: len(distinct_values[k]))
 
+        # Verify that every stratum uses the same dimensions
```
hannes-ucsc, line 415: Where does this requirement come from?

src/azul/service/hca_response_v5.py
```diff
@@ -412,6 +412,11 @@ def _make_contributor_matrices(cls, catalog, files):
...
+        for file in files:
+            for stratum in file['strata']:
+                assert set(sorted_dimensions) == set(stratum.keys()), sorted_dimensions
```
hannes-ucsc, line 418: And dict.keys() returns a set.

src/azul/service/hca_response_v5.py
danielsotirhos: This is a requirement for the construction of the tree due to the set depth of the NestedDict. If a strata doesn't have the same dimensions as the other strata, the node where the file URL would go will be a NestedDict instead of a List.

hannes-ucsc: That's not a requirement, but an implementation concern. It could be that the use of NestedDict or aspects of its implementation are inadequate to meet the requirements. Is there a an actual requirement for this?

danielsotirhos: Created #2443 to allow stratification tree with varying dimensions.

src/azul/plugins/metadata/hca/contributor_matrices.py
```diff
@@ -0,0 +1,165 @@
...
+
+from azul import (
+    config,
+)
+from azul.collections import (
+    NestedDict,
+)
+
+
+class ContributorMatrices:
```
hannes-ucsc, line 21: No need for a class

src/azul/plugins/metadata/hca/contributor_matrices.py
```diff
@@ -0,0 +1,165 @@
...
+                # multiple strata, one per value. The strata are identical
+                # except in the dimension that had the multiple values.
+                'strata': list(chain.from_iterable(
+                    map(dict, product(*(
+                        [(dimension, value) for value in values.split(',')]
+                        for dimension, values in (point.split('=') for point in stratum.split(';'))
+                    )))
+                    for stratum in file['strata'].split('\n')
+                )),
+                'url': furl(config.service_endpoint(),
```
hannes-ucsc, line 126: This cannot live here. The plugin cannot know about the service. We need to parameterize this function with a factory for urls. Look for `file_url` in app.py.

src/azul/plugins/metadata/hca/contributor_matrices.py
```diff
@@ -0,0 +1,165 @@
...
+
+        >>> f([{'uuid': 'u1', 'version': 'v1', 'name': 'n1', 'strata': 'a=1;b=2'},
+        ...    {'uuid': 'u2', 'version': 'v2', 'name': 'n2', 'strata': 'a=1;b=2'}])
+        {
+            "a": {
+                "1": {
+                    "b": {
+                        "2": {
+                            "url": [
+                                "https://service.dev.singlecell.gi.ucsc.edu/fetch/repository/files/u1?version=v1&catalog=dcp",
```
hannes-ucsc, line 53: The factory will also allow you to make these assertions shorter and independent of the deployment. I'm actually a bit surprised this test works on other deployments. Please find out why.

src/azul/plugins/metadata/hca/contributor_matrices.py
```diff
@@ -0,0 +1,165 @@
...
+class ContributorMatrices:
+
+    @classmethod
+    def make_contributor_matrices(cls, catalog, files):
+        """
+        >>> from azul.doctests import assert_json
+        >>> def f(files):
+        ...     return assert_json(ContributorMatrices.make_contributor_matrices('dcp', files))
+
+        >>> f([{'uuid': 'u1', 'version': 'v1', 'name': 'n1', 'strata': 'a=1;b=2'}])
```
hannes-ucsc, line 30:
```suggestion
        >>> f([{'uuid': 'u', 'version': 'v', 'name': 'n', 'strata': 'a=1;b=2'}])
```

minimize

hannes-ucsc (changes requested):
Commits should be split as described in C guide.

src/azul/plugins/metadata/hca/contributor_matrices.py
danielsotirhos: This test does not work on other deployments. An oversight on my part, I had `develop` selected when testing as did Travis. Implementing suggested `file_url_func` parameter on the function and removing deployment specifics from doctest.

src/azul/service/hca_response_v5.py
```diff
@@ -317,107 +313,23 @@ def make_projects(self, entry):
...
-
-        return tree
+        return make_contributor_matrices_tree(self.catalog, files, self.make_file_url)
+
+    def make_file_url(self,
+                      catalog: CatalogName,
+                      file_id: str,
+                      file_version: str) -> str:
+        return furl(config.service_endpoint(),
+                    path=('fetch', 'repository', 'files', file_id),
```
hannes-ucsc, line 330: This is still brittle. Only the app module and maybe a controller should be aware of request routing. When we write code like this, we need to be aware of the dependencies it introduces, if they are brittle and what the failure mode would be if they broke. If someone came along and modified the route in app.py, they would in all likelihood miss to update this. We wouldn't learn about the breakage until a user actually tried to access this URL. An integration test might catch this but we have none for this right now and often they aren't run until a PR is tested in the sandbox.

And even if we had an IT, this violates the principle of separation of concerns and make future maintenance of this code harder.

hannes-ucsc (changes requested):
Change in plans. Look at how file_url is used in the service to patch the response. Use the same technique here. This means that the tree leafs would need to be dictionaries with `uuid` and `version` keys.

src/azul/service/index_query_service.py
```diff
@@ -74,10 +76,35 @@ def get_data(self,
                     file['url'] = file_url_func(catalog=catalog,
                                                 file_uuid=file['uuid'],
                                                 version=file['version'])
+        if entity_type == 'projects':
+            for hit in response['hits']:
+                # Compose a URL for each file in stratification trees
+                for project in hit['projects']:
+                    self.transform_stratification_file_nodes(tree=project['contributorMatrices'],
```
hannes-ucsc, line 83: This is ok but we should try to make this even more generic. I'm thinking of something that can scan the entire response for file references and inject a URL to the file. A file reference could be a dictionary containing both `file_uuid` and `file_version` key. A recursive traversal would solve this new use case and the preexisting one right above this section. The only issue is that the file references for the preexisting use case use the `uuid` and `version` keys instead of `file_uuid` and `file_version` but I'm sure that can be fixed.

hannes-ucsc (approved):

hannes-ucsc:
For demo, run test in PyCharm and show test coverage.

Issue #2538, opened 2020-11-25
Curl manifest for zarray matrices does not include related files


achave11:

achave11:
Findings:

Manifest generation for `curl` format only includes placeholder postfix `/.zattrs` and no auxiliary files (`related_files`). The `full` manifest format has the same issue as `curl`, described in ticket #2468.

theathorn:
Not a DCP 2.0 MVP show-stopper.

theathorn:
http://broadinstitute.zendesk.com/agent/tickets/116461

PR #2551, opened 2020-11-30
Curl manifest for zarray file format does not include auxiliary files (#2538)

achave11:
Author (check before every review)

- [x] Added `DCP/1` label                                   <sub>or this PR does not target an `hca/*` branch</sub>
- [x] Created issue to track porting these changes          <sub>or this PR does not need porting</sub> 
- [x] Rebased branch on `develop`

- [x] Sanity-checked history
- [x] Reindexed `sandbox`                                   <sub>or this PR does not require reindexing</sub>

- [x] Commented on demo expectations                        <sub>or labeled as `no demo`</sub>
- [ ] Reindexed `dev`                                       <sub>or this PR does not require reindexing</sub>

### DIFF
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -667,6 +667,13 @@ class CurlManifestGenerator(StreamingManifestGenerator):
     def entity_type(self) -> str:
         return 'files'
 
+    @cached_property
+    def source_filter(self) -> SourceFilters:
+        return [
+            *super().source_filter,
+            'contents.files.related_files'
+        ]
+
     @classmethod
     def manifest_properties(cls, url: str) -> JSON:
         return {
@@ -705,13 +712,8 @@ class CurlManifestGenerator(StreamingManifestGenerator):
         return '"' + s.replace('\\', '\\\\').replace('"', '\\"') + '"'
 
     def write_to(self, output: IO[str]) -> Optional[str]:
-        output.write('--create-dirs\n\n'
-                     '--compressed\n\n'
-                     '--location\n\n'
-                     '--continue-at -\n\n')
-        for hit in self._create_request().scan():
-            doc = self._hit_to_doc(hit)
-            file = one(doc['contents']['files'])
+
+        def _write(file: dict):
             uuid, name, version = file['uuid'], file['name'], file['version']
             url = furl(config.service_endpoint(),
                        path=f'/repository/files/{uuid}',
@@ -724,6 +726,17 @@ class CurlManifestGenerator(StreamingManifestGenerator):
             output_name = self._sanitize_path(bundle['uuid'] + '/' + name)
             output.write(f'url={self._option(url.url)}\n'
                          f'output={self._option(output_name)}\n\n')
+
+        output.write('--create-dirs\n\n'
+                     '--compressed\n\n'
+                     '--location\n\n'
+                     '--continue-at -\n\n')
+        for hit in self._create_request().scan():
+            doc = self._hit_to_doc(hit)
+            file = one(doc['contents']['files'])
+            _write(file)
+            for related_file in file['related_files']:
+                _write(related_file)
         return None
 
     # Disallow control characters and backslash as they likely indicate an
--- a/test/service/test_manifest.py
+++ b/test/service/test_manifest.py
@@ -329,6 +329,89 @@ class TestManifestEndpoints(ManifestTestCase, DSSUnitTestCase):
         Test that when downloading a manifest with a zarr, all of the files are added into the manifest even
         if they are not listed in the service response.
         """
+        expected = [
+            # Original file
+            {
+                'file_crc32c': '4e75003e',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/.zattrs',
+                'file_uuid': 'c1c4a2bc-b5fb-4083-af64-f5dec70d7f9d',
+                'specimen_from_organism_organ': 'brain'
+            },
+            # Related files from zarray store (auxiliary files)
+            {
+                'file_crc32c': '444a7707',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/.zgroup',
+                'file_uuid': '54541cc5-9010-425b-9037-22e43948c97c',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': '444a7707',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/.zgroup',
+                'file_uuid': '66b8f976-6f1e-45b3-bd97-069658c3c847',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': 'c6ab0701',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/cell_id/.zarray',
+                'file_uuid': 'ac05d7fb-d6b9-4ab1-8c04-6211450dbb62',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': 'cd2fd51f',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/cell_id/0.0',
+                'file_uuid': '0c518a52-f315-4ea2-beed-1c9d8f2d802b',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': 'b89e6723',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/expression/.zarray',
+                'file_uuid': '136108ab-277e-47a4-acc3-1feed8fb2f25',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': 'caaefa77',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/expression/0.0',
+                'file_uuid': '0bef5419-739c-4a2c-aedb-43754d55d51c',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': 'f629ec34',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/gene_id/.zarray',
+                'file_uuid': '3a5f7299-1aa1-4060-9631-212c29b4d807',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': '59d86b68',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/gene_id/0.0',
+                'file_uuid': 'a8f0dc39-6019-4fc7-899d-4e34a48d03e5',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': '25d193cf',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/qc_metric/.zarray',
+                'file_uuid': '68ba4711-1447-42ac-aa40-9c0e4cda1666',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': '17a84191',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/qc_metric/0.0',
+                'file_uuid': '27e66328-e337-4bcd-ba15-7893ecaf841f',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': '25d193cf',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/qc_values/.zarray',
+                'file_uuid': '2ab1a516-ef36-41b6-a78f-513361658feb',
+                'specimen_from_organism_organ': 'brain'
+            },
+            {
+                'file_crc32c': 'bdc30523',
+                'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/qc_values/0.0',
+                'file_uuid': '351970aa-bc4c-405e-a274-be9e08e42e98',
+                'specimen_from_organism_organ': 'brain'
+            }
+        ]
+
         bundle_fqid = self.bundle_fqid(uuid='587d74b4-1075-4bbf-b96a-4d1ede0481b2',
                                        version='2018-10-10T022343.182000Z')
         self._index_canned_bundle(bundle_fqid)
@@ -347,11 +430,11 @@ class TestManifestEndpoints(ManifestTestCase, DSSUnitTestCase):
                     lines = response.content.decode().splitlines()
                     tsv_file = csv.DictReader(lines, delimiter='\t')
                     rows = list(tsv_file)
-                    self.assertEqual(len(rows), 13)  # 12 related file, one original
-                    self.assertEqual(len(rows), len({row['file_uuid'] for row in rows}))
-                    # 2 pairs of files out of the 13 have matching checksums
-                    self.assertEqual(11, len({row['file_sha256'] for row in rows}))
-                    self.assertEqual(11, len({row['file_crc32c'] for row in rows}))
+                    rows = [dict(file_crc32c=row['file_crc32c'],
+                                 file_name=row['file_name'],
+                                 file_uuid=row['file_uuid'],
+                                 specimen_from_organism_organ=row['specimen_from_organism.organ']) for row in rows]
+                    self.assertEqual(expected, rows)
 
     @manifest_test
     def test_terra_bdbag_manifest(self):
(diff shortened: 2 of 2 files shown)


test/service/test_manifest.py
```diff
@@ -350,6 +350,13 @@ def test_manifest_zarr(self):
                     self.assertEqual(11, len({row['file_sha256'] for row in rows}), 'Incorrect unique sha256 count')
                     self.assertEqual(11, len({row['file_crc32c'] for row in rows}), 'Incorrect unique crc32c count')
 
+                    response = self._get_manifest(ManifestFormat.curl, filters)
+                    self.assertEqual(200, response.status_code, 'Unable to download manifest')
+
+                    lines = response.content.decode('utf-8').splitlines()
+                    output_paths = {line.partition('=')[-1] for line in lines if line.startswith('output')}
```
jessebrennan, line 357: I found the `[-1]` tuple index here confusing. I think usually it's better to explicitly unpack a tuple. Indexing here is convenient for the comprehension, but I'm not convinced that it is necessary since you're not actually using the output_paths. You could remove the index and just assert the line count?

jessebrennan (approved):
Looks good! I had one minor style thing, but it wasn't blocking.

If you haven't yet, I think it would be a good idea to run the integration tests with this change.

test/service/test_manifest.py
achave11: Done, I missed that! :)

test/service/test_manifest.py
```diff
@@ -351,6 +351,13 @@ def test_manifest_zarr(self):
                     self.assertEqual(11, len({row['file_sha256'] for row in rows}), 'Incorrect unique sha256 count')
                     self.assertEqual(11, len({row['file_crc32c'] for row in rows}), 'Incorrect unique crc32c count')
 
+                    response = self._get_manifest(ManifestFormat.curl, filters)
+                    self.assertEqual(200, response.status_code, 'Unable to download manifest')
```
hannes-ucsc, line 355:
```suggestion
                    self.assertEqual(200, response.status_code)
```

Here and above.

This is fluff. If someone needs that message to diagnose the assertion failure, they are working in the wrong profession.

hannes-ucsc: In fact, remove all of these messages in this test.

test/service/test_manifest.py
```diff
@@ -351,6 +351,13 @@ def test_manifest_zarr(self):
...
+
+                    lines = response.content.decode('utf-8').splitlines()
+                    output_paths = {line.partition('=') for line in lines if line.startswith('output')}
```
hannes-ucsc, line 358: The set elements are tuples. That can't be right.

hannes-ucsc (changes requested):
Many lines look longer than 80 stripped.

test/service/test_manifest.py
```diff
@@ -340,16 +340,24 @@ def test_manifest_zarr(self):
...
                     # 2 pairs of files out of the 13 have matching checksums
-                    self.assertEqual(11, len({row['file_sha256'] for row in rows}), 'Incorrect unique sha256 count')
-                    self.assertEqual(11, len({row['file_crc32c'] for row in rows}), 'Incorrect unique crc32c count')
+                    self.assertEqual(11, len({row['file_sha256'] for row in rows}))
+                    self.assertEqual(11, len({row['file_crc32c'] for row in rows}))
+
+                    response = self._get_manifest(ManifestFormat.curl, filters)
+                    self.assertEqual(200, response.status_code)
+
+                    lines = response.content.decode('utf-8').splitlines()
```
hannes-ucsc, line 357: utf-8 is the default encoding. Fix all places in this method in separate commit.

test/service/test_manifest.py
```diff
@@ -340,16 +340,24 @@ def test_manifest_zarr(self):
...
+                    output_paths = {line for line in lines if line.startswith('output')}
```
hannes-ucsc, line 358:
```suggestion
                    output_paths = {line for line in lines if line.startswith('output=')}
```

test/service/test_manifest.py
```diff
@@ -340,16 +340,24 @@ def test_manifest_zarr(self):
...
+                    # 12 related, one original, all unique files
+                    self.assertEqual(13, len(output_paths))
```
hannes-ucsc, line 360: How do I know that these actually are auxiliary files from a zarray store? That's a general issue with this test, I think, that I must have missed when reviewing the PR that added this.

hannes-ucsc (commented):
No bump.

test/service/test_manifest.py
achave11: Had to make some additional modifications to adapt the files check.

test/service/test_manifest.py
jessebrennan: This file is full of lines like 
```
```

Fixing them all might be beyond the scope of this PR, but consider adding another commit that removed the "fluff" from all such lines.

test/service/test_manifest.py
```diff
@@ -340,16 +423,16 @@ def test_manifest_zarr(self):
...
                     rows = list(tsv_file)
-                    self.assertEqual(len(rows), 13)  # 12 related file, one original
-                    self.assertEqual(len(rows), len({row['file_uuid'] for row in rows}), 'Rows are not unique')
+                    # self.assertEqual(len(rows), 13)  # 12 related file, one original
+                    self.assertEqual({f['uuid'] for f in expected}, {row['file_uuid'] for row in rows})
                     # 2 pairs of files out of the 13 have matching checksums
-                    self.assertEqual(11, len({row['file_sha256'] for row in rows}), 'Incorrect unique sha256 count')
-                    self.assertEqual(11, len({row['file_crc32c'] for row in rows}), 'Incorrect unique crc32c count')
+                    self.assertEqual({f['sha256'] for f in expected}, {row['file_sha256'] for row in rows})
+                    self.assertEqual({f['crc32c'] for f in expected}, {row['file_crc32c'] for row in rows})
```
jessebrennan, line 435: Delete all other asserts that use `expected` and replace with something like:

```
self.assertEqual(rows, expected)
```

test/service/test_manifest.py
```diff
@@ -340,16 +423,16 @@ def test_manifest_zarr(self):
...
-                    self.assertEqual(200, response.status_code, 'Unable to download manifest')
+                    self.assertEqual(200, response.status_code)
                     # Cannot use response.iter_lines() because of https://github.com/psf/requests/issues/3980
-                    lines = response.content.decode('utf-8').splitlines()
+                    lines = response.content.decode().splitlines()
                     tsv_file = csv.DictReader(lines, delimiter='\t')
                     rows = list(tsv_file)
-                    self.assertEqual(len(rows), 13)  # 12 related file, one original
-                    self.assertEqual(len(rows), len({row['file_uuid'] for row in rows}), 'Rows are not unique')
+                    # self.assertEqual(len(rows), 13)  # 12 related file, one original
```
jessebrennan, line 431: Also you left a commented-out line in here.

src/azul/service/manifest_service.py
```diff
@@ -701,13 +708,8 @@ def _option(cls, s: str):
...
     def write_to(self, output: IO[str]) -> Optional[str]:
-        output.write('--create-dirs\n\n'
-                     '--compressed\n\n'
-                     '--location\n\n'
-                     '--continue-at -\n\n')
-        for hit in self._create_request().scan():
-            doc = self._hit_to_doc(hit)
-            file = one(doc['contents']['files'])
+
+        def _write(file: dict):
```
jessebrennan, line 712: This suggestion is very optional, because the diff is very small, but you're combining refactoring (creating the `_write()` function) with functional changes. It would be easier to understand the diff if you moved the refactoring into a separate, prior commit.

test/service/test_manifest.py
```diff
@@ -433,6 +433,17 @@ def test_manifest_zarr(self):
                     # 2 pairs of files out of the 13 have matching checksums
                     self.assertEqual({f['sha256'] for f in expected}, {row['file_sha256'] for row in rows})
                     self.assertEqual({f['crc32c'] for f in expected}, {row['file_crc32c'] for row in rows})
+                    # Manifest for `curl` format  should also contain related files
```
jessebrennan, line 436: I personally don't mind it, but the little Hannes in my head is saying that this comment doesn't add anything that can't already be deduced from the code. Consider deleting it.

test/service/test_manifest.py
```diff
@@ -433,6 +433,17 @@ def test_manifest_zarr(self):
...
+                    response = self._get_manifest(ManifestFormat.curl, filters)
+                    self.assertEqual(200, response.status_code)
+                    lines = response.content.decode().splitlines()
+                    output_paths = set()
+                    for line in lines:
+                        if line.startswith('output'):
```
jessebrennan, line 442: I thought I saw in a previous review comment from Hannes that `'output'` be changed to `'output='`?

jessebrennan (changes requested):
Notice this comment here, which was accidentally excluded from the review. https://github.com/DataBiosphere/azul/pull/2551#discussion_r565617622

test/service/test_manifest.py
```diff
@@ -428,11 +428,11 @@ def test_manifest_zarr(self):
                     lines = response.content.decode().splitlines()
                     tsv_file = csv.DictReader(lines, delimiter='\t')
                     rows = list(tsv_file)
-                    # self.assertEqual(len(rows), 13)  # 12 related file, one original
-                    self.assertEqual({f['uuid'] for f in expected}, {row['file_uuid'] for row in rows})
-                    # 2 pairs of files out of the 13 have matching checksums
-                    self.assertEqual({f['sha256'] for f in expected}, {row['file_sha256'] for row in rows})
-                    self.assertEqual({f['crc32c'] for f in expected}, {row['file_crc32c'] for row in rows})
+                    files = [dict(crc32c=row['file_crc32c'],
```
jessebrennan, line 431: I think you should augment `expected` to contain all of the fields that are in rows. Then you don't have to do this comprehension, and also the assert is stronger.

test/service/test_manifest.py
```diff
@@ -428,11 +428,11 @@ def test_manifest_zarr(self):
...
+                                  sha256=row['file_sha256'],
+                                  name=row['file_name'],
+                                  uuid=row['file_uuid']) for row in rows]
+                    self.assertEqual(expected, files)
```
jessebrennan, line 435:
```suggestion
                    self.assertEqual(expected, rows)
```

jessebrennan (commented):
Just one thing I'm unsure about.

test/service/test_manifest.py
achave11: Good call. However, I think that augmenting `expected` to include all the manifest `csv` is beyond the scope of this test. I discussed your suggested changes with Hannes, to which he pointed out there are already tests that assert the `csv` response verbatim. Hannes did recommend adding an additional field (`specimen_from_organism.organ`) not obtained from the `files` entity, given that we are testing for inclusion of `related_files` in the manifests.

jessebrennan (approved):

src/azul/service/manifest_service.py
hannes-ucsc: I'm actually having a harder time understanding this if the history is split up like it is now. The refactoring doesn't make much sense in isolation. Why would I extract a function with just one call site? If we roll back the functional change, we'd also want to roll back the refactoring.

test/service/test_manifest.py
```diff
@@ -340,16 +436,28 @@ def test_manifest_zarr(self):
...
+                    lines = response.content.decode().splitlines()
                     tsv_file = csv.DictReader(lines, delimiter='\t')
                     rows = list(tsv_file)
-                    self.assertEqual(len(rows), 13)  # 12 related file, one original
-                    self.assertEqual(len(rows), len({row['file_uuid'] for row in rows}), 'Rows are not unique')
-                    # 2 pairs of files out of the 13 have matching checksums
-                    self.assertEqual(11, len({row['file_sha256'] for row in rows}), 'Incorrect unique sha256 count')
-                    self.assertEqual(11, len({row['file_crc32c'] for row in rows}), 'Incorrect unique crc32c count')
+                    rows = [dict(file_crc32c=row['file_crc32c'],
+                                 file_sha256=row['file_sha256'],
```
hannes-ucsc, line 445:
```suggestion
```

One checksum is sufficient. We're not trying to assert the completeness of the columns, only of the rows. The UUID, name and one checksum are enough to pin point a file.

test/service/test_manifest.py
```diff
@@ -340,16 +436,28 @@ def test_manifest_zarr(self):
...
+                    self.assertEqual(expected, rows)
+
+                    response = self._get_manifest(ManifestFormat.curl, filters)
+                    self.assertEqual(200, response.status_code)
+                    lines = response.content.decode().splitlines()
+                    file_output = 'output="587d74b4-1075-4bbf-b96a-4d1ede0481b2/'
+                    output_paths = set()
+                    for line in lines:
+                        if line.startswith(file_output):
+                            _, _, file_name = line.partition(file_output)
```
hannes-ucsc, line 458: Partition performs a search. If you know the prefix, you know where to slice the string, eliminating the need for a search.

test/service/test_manifest.py
```diff
@@ -340,16 +436,28 @@ def test_manifest_zarr(self):
...
+                            output_paths.add(file_name.strip('"'))
```
hannes-ucsc, line 459: This would pass with or without a quote at the end. It should be stricter.

test/service/test_manifest.py
```diff
@@ -340,16 +436,28 @@ def test_manifest_zarr(self):
...
+                                 file_name=row['file_name'],
+                                 file_uuid=row['file_uuid'],
+                                 specimen_from_organism_organ=row['specimen_from_organism.organ']) for row in rows]
+                    self.assertEqual(expected, rows)
+
+                    response = self._get_manifest(ManifestFormat.curl, filters)
+                    self.assertEqual(200, response.status_code)
+                    lines = response.content.decode().splitlines()
+                    file_output = 'output="587d74b4-1075-4bbf-b96a-4d1ede0481b2/'
+                    output_paths = set()
```
hannes-ucsc, line 455: With a set you'd let duplicated rows in the curl manifest pass.

hannes-ucsc (changes requested):
If you're splitting this into test&fix pair of commits (which is advisable) you would need to use the [m/n] convention in the commit title and push them separately.

src/azul/service/manifest_service.py
hannes-ucsc: For larger changes, @jessebrennan's suggestion would make sense, especially if the commits are pushed separately. It's just that this change is so small.

test/service/test_manifest.py
```diff
@@ -340,16 +423,28 @@ def test_manifest_zarr(self):
...
+                self.assertEqual(200, response.status_code)
+                lines = response.content.decode().splitlines()
+                file_prefix = 'output="587d74b4-1075-4bbf-b96a-4d1ede0481b2/'
+                curl_files = []
+                for line in lines:
+                    if line.startswith(file_prefix) and line.endswith('"'):
+                        # Slice out file_prefix; -1 slice out postfix double quotes char
+                        file_name = line[len(file_prefix):-1]
+                        curl_files.append(file_name)
+                self.assertEqual([f['file_name'] for f in expected], curl_files)
```
hannes-ucsc, line 447: This is sensitive to insignificant ordering differences.

hannes-ucsc (changes requested):
The [1/2] and [2/2] commit messages should have a distinguishing body.

test/service/test_manifest.py
```diff
@@ -444,7 +444,8 @@ def test_manifest_zarr(self):
                         # Slice out file_prefix; -1 slice out postfix double quotes char
                         file_name = line[len(file_prefix):-1]
                         curl_files.append(file_name)
-                self.assertEqual([f['file_name'] for f in expected], curl_files)
+                self.assertEqual(sorted(freeze([f['file_name'] for f in expected])),
```
hannes-ucsc, line 447: I don't think freezing is necessary. Why do you use it?

test/service/test_manifest.py
```diff
@@ -340,16 +423,29 @@ def test_manifest_zarr(self):
...
+                                 specimen_from_organism_organ=row['specimen_from_organism.organ']) for row in rows]
+                    self.assertEqual(expected, rows)
+
+                response = self._get_manifest(ManifestFormat.curl, filters)
+                self.assertEqual(200, response.status_code)
+                lines = response.content.decode().splitlines()
+                file_prefix = 'output="587d74b4-1075-4bbf-b96a-4d1ede0481b2/'
+                curl_files = []
+                for line in lines:
+                    if line.startswith(file_prefix) and line.endswith('"'):
```
hannes-ucsc, line 443: If a line starts with `file_prefix` it MUST end in `"`. Your current implementation would still fail the assertion below if it came across a line lacking the trailing quotation mark. That's acceptable, but the failure wouldn't be very informative since it conflates the absence of a line with a malformed line.

hannes-ucsc: Actually, the current implementation is wrong. If the response contained the same line twice, once with and once without trailing quotation marks, the test would pass.

hannes-ucsc (changes requested):

test/service/test_manifest.py
achave11: As mentioned during stand up, it was a misconception from my part that `freeze` was the standard for this type of situations.

hannes-ucsc (approved):

hannes-ucsc:
For demo, show curl download of zarr files and show that zarray store is complete and valid, using https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.tree, for example.

achave11:
During demo, we found out that auxiliary files are not provisioned correctly in the `curl` manifest.
They all contain the following error message:
```
{"Code":"NotFoundError","Message":"NotFoundError: Unable to find file 'bce942f6-5b2a-4471-aa85-fbe2b0690d32', version '2019-05-16T02:07:09.988140Z' in catalog 'dcp1'"}
```

achave11:
Check for same problem in `compact` format.

Issue #2629, opened 2020-12-15
Specify in field types whether a field is wrapped in a list


jessebrennan:

melainalegaspi:
@jessebrennan: "error with assert is related to #2071"

Issue #2071, opened 2020-07-24
Extraneous `null` values in `hits[].samples[].*`


danielsotirhos:
As a followup to #2043 fix the issue of extraneous `null` values in `hits[].samples[].*` due to the polymorphic nature of the `samples` inner entity.

![Screen Shot 2020-07-24 at 4 39 42 PM](https://user-images.githubusercontent.com/46407464/88443329-b57b4a80-cdcc-11ea-815c-bc5cbe901d60.png)

danielsotirhos:
Depending on response to https://github.com/humancellatlas/data-browser/issues/1534#issuecomment-823607101 we can remove `do not promote` label or coordinate promotion with DB team.

danielsotirhos:
> Depending on response to [HumanCellAtlas/data-browser#1534 (comment)](https://github.com/HumanCellAtlas/data-browser/issues/1534#issuecomment-823607101) we can remove `do not promote` label or coordinate promotion with DB team.

"Data Browser does not require any code changes to support the updated grouped format of hits[].samples" https://github.com/HumanCellAtlas/data-browser/issues/1534#issuecomment-823625691

PR #1382, opened 2019-10-15
[R] Fix: Indexer ignores mismatch between _foo and _foo_types (#1344)

danielsotirhos:

PR #2697, opened 2021-01-11
Specify in field types if a field is wrapped in a list (#2629)

jessebrennan:
Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history
- [x] Reindexed `sandbox`                                   <sub>or this PR does not require reindexing `sandbox`</sub>

- [x] Commented on demo expectations                        <sub>or labeled as `no demo`</sub>

Primary reviewer (reindex) 

Primary reviewer

- [ ] Unassign PR from reviewer and assign to PM

### DIFF
--- a/src/azul/indexer/document.py
+++ b/src/azul/indexer/document.py
@@ -423,10 +423,14 @@ class Document(Generic[C]):
                 # FIXME: Assert that a non-list field_type implies a non-list
                 #        doc (only possible for contributions).
                 #        https://github.com/DataBiosphere/azul/issues/2689
-                # Samples are an exception since they are a composite type.
-                # Unused fields are left as None, even if field_types would
-                # normally dictate otherwise.
-                assert isinstance(doc, list) or path[:2] == ('contents', 'samples')
+                # FIXME: Samples are an exception since they are polymorphic.
+                #        Unused fields are left as None, even if field_types
+                #        would normally dictate otherwise.
+                #        https://github.com/DataBiosphere/azul/issues/2071
+                assert (isinstance(doc, list)
+                        or (doc is not None if forward else doc != NullableString.null_string)
+                        or path[:2] == ('contents', 'samples'))
+
                 field_types = one(field_types)
             if isinstance(field_types, FieldType):
                 field_type = field_types
--- a/src/azul/indexer/document.py
+++ b/src/azul/indexer/document.py
@@ -19,12 +19,16 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Sequence,
     Tuple,
     TypeVar,
     Union,
 )
 
 import attr
+from more_itertools import (
+    one,
+)
 
 from azul import (
     CatalogName,
@@ -316,10 +320,10 @@ class NullableBool(NullableNumber[bool]):
 
 null_bool: NullableBool = NullableBool()
 
-FieldTypes4 = Union[Mapping[str, FieldType], FieldType]
-FieldTypes3 = Union[Mapping[str, FieldTypes4], FieldType]
-FieldTypes2 = Union[Mapping[str, FieldTypes3], FieldType]
-FieldTypes1 = Union[Mapping[str, FieldTypes2], FieldType]
+FieldTypes4 = Union[Mapping[str, FieldType], Sequence[FieldType], FieldType]
+FieldTypes3 = Union[Mapping[str, FieldTypes4], Sequence[FieldType], FieldType]
+FieldTypes2 = Union[Mapping[str, FieldTypes3], Sequence[FieldType], FieldType]
+FieldTypes1 = Union[Mapping[str, FieldTypes2], Sequence[FieldType], FieldType]
 FieldTypes = Mapping[str, FieldTypes1]
 CataloguedFieldTypes = Mapping[CatalogName, FieldTypes]
 
@@ -414,8 +418,19 @@ class Document(Generic[C]):
                 return [cls.translate_fields(val, field_types, forward=forward, path=path) for val in doc]
             else:
                 assert False, (path, type(doc))
-        elif isinstance(field_types, FieldType):
-            field_type = field_types
+        elif isinstance(field_types, (list, FieldType)):
+            if isinstance(field_types, list):
+                field_type = one(field_types)
+                assert isinstance(field_type, FieldType)
+
+                # In english: if forward and doc is not None, then if
+                # field_types is a list, doc should also be a list. We would
+                # also like to assert that a non-list field_type implies a
+                # non-list doc. Such an assertion could only be done for
+                # contributions, and only after https://github.com/DataBiosphere/azul/issues/2689
+                assert not forward or doc is None or isinstance(doc, list)
+            else:
+                field_type = field_types
             if forward:
                 if isinstance(doc, list):
                     if not doc and field_type.allow_sorting_by_empty_lists:
--- a/src/azul/indexer/document.py
+++ b/src/azul/indexer/document.py
@@ -418,19 +418,18 @@ class Document(Generic[C]):
                 return [cls.translate_fields(val, field_types, forward=forward, path=path) for val in doc]
             else:
                 assert False, (path, type(doc))
-        elif isinstance(field_types, (list, FieldType)):
+        else:
             if isinstance(field_types, list):
-                field_type = one(field_types)
-                assert isinstance(field_type, FieldType)
-
-                # In english: if forward and doc is not None, then if
-                # field_types is a list, doc should also be a list. We would
-                # also like to assert that a non-list field_type implies a
-                # non-list doc. Such an assertion could only be done for
-                # contributions, and only after https://github.com/DataBiosphere/azul/issues/2689
-                assert not forward or doc is None or isinstance(doc, list)
-            else:
+                # FIXME: Assert that a non-list field_type implies a non-list
+                #        doc (only possible for for contributions).
+                #        https://github.com/DataBiosphere/azul/issues/2689
+                if forward:
+                    assert doc is None or isinstance(doc, list)
+                field_types = one(field_types)
+            if isinstance(field_types, FieldType):
                 field_type = field_types
+            else:
+                assert False, (path, type(field_types))
             if forward:
                 if isinstance(doc, list):
                     if not doc and field_type.allow_sorting_by_empty_lists:
@@ -451,8 +450,6 @@ class Document(Generic[C]):
                     return [field_type.from_index(value) for value in doc]
                 else:
                     return field_type.from_index(doc)
-        else:
-            assert False, (path, type(field_types))
 
     def to_source(self) -> JSON:
         return dict(entity_id=self.coordinates.entity.entity_id,
--- a/src/azul/indexer/document.py
+++ b/src/azul/indexer/document.py
@@ -421,10 +421,12 @@ class Document(Generic[C]):
         else:
             if isinstance(field_types, list):
                 # FIXME: Assert that a non-list field_type implies a non-list
-                #        doc (only possible for for contributions).
+                #        doc (only possible for contributions).
                 #        https://github.com/DataBiosphere/azul/issues/2689
-                if forward:
-                    assert doc is None or isinstance(doc, list)
+                # Samples are an exception since they are a composite type.
+                # Unused fields are left as None, even if field_types would
+                # normally dictate otherwise.
+                assert isinstance(doc, list) or path[:2] == ('contents', 'samples')
                 field_types = one(field_types)
             if isinstance(field_types, FieldType):
                 field_type = field_types
--- a/src/azul/indexer/document_service.py
+++ b/src/azul/indexer/document_service.py
@@ -5,6 +5,10 @@ from typing import (
     Type,
 )
 
+from more_itertools import (
+    one,
+)
+
 from azul import (
     CatalogName,
     cache,
@@ -68,6 +72,8 @@ class DocumentService:
                 raise KeyError(f'Path {path} not represented in field_types')
             except TypeError:
                 raise TypeError(f'Path {path} not represented in field_types')
+            if isinstance(field_types, list):
+                field_types = one(field_types)
             if field_types is None:
                 return None
         return field_types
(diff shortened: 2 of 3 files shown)


src/azul/indexer/document.py
```diff
@@ -396,8 +400,19 @@ def translate_fields(cls,
                 return [cls.translate_fields(val, field_types, forward=forward, path=path) for val in doc]
             else:
                 assert False, (path, type(doc))
-        elif isinstance(field_types, FieldType):
-            field_type = field_types
+        elif isinstance(field_types, list) or isinstance(field_types, FieldType):
```
danielsotirhos, line 403:
```suggestion
        elif isinstance(field_types, (list, FieldType)):
```

danielsotirhos (approved):
Looks good!

src/azul/indexer/document.py
```diff
@@ -396,8 +400,19 @@ def translate_fields(cls,
                 return [cls.translate_fields(val, field_types, forward=forward, path=path) for val in doc]
             else:
                 assert False, (path, type(doc))
-        elif isinstance(field_types, FieldType):
-            field_type = field_types
+        elif isinstance(field_types, (list, FieldType)):
+            if isinstance(field_types, list):
```
hannes-ucsc, line 404: The repeated type checks are annoying. Try to sculpt the control flow such that you check every case exactly once.

src/azul/indexer/document.py
```diff
@@ -396,8 +400,19 @@ def translate_fields(cls,
...
+                field_type = one(field_types)
+                assert isinstance(field_type, FieldType)
+
+                # In english: if forward and doc is not None, then if
```
hannes-ucsc, line 408: Not only is "In english" obvious, so is what follows it. Document the why, not the how.

src/azul/indexer/document_service.py
```diff
@@ -68,6 +72,8 @@ def field_type(self, catalog: CatalogName, path: Tuple[str, ...]) -> FieldType:
                 raise KeyError(f'Path {path} not represented in field_types')
             except TypeError:
                 raise TypeError(f'Path {path} not represented in field_types')
+            if isinstance(field_types, list):
```
hannes-ucsc, line 75: This would handle inner nodes that are lists, which we currently don't support. I think this function is broken in other ways, like it doesnt handle partial paths, or the return type is inaccurate, but you won't need to fix that here.

src/azul/indexer/document_service.py
```diff
@@ -68,6 +72,8 @@ def field_type(self, catalog: CatalogName, path: Tuple[str, ...]) -> FieldType:
...
+                field_types = one(field_types)
             if field_types is None:
                 return None
```
hannes-ucsc, line 78: PyCharm emits a type warning here. Should the return type annotation be adjusted?

src/azul/plugins/metadata/hca/transform.py
```diff
@@ -934,6 +934,15 @@ def _contribution(self, contents: MutableJSON, entity_id: api.UUID4) -> Contribu
 
     @classmethod
     def field_types(cls) -> FieldTypes:
+        """
+        Field types outlines the general shape of our Elasticsearch documents.
+
+        Not all information is captured. Lists containing primitive types are
```
hannes-ucsc, line 940: This is a FIXME.

src/azul/indexer/document.py
```diff
@@ -396,8 +400,19 @@ def translate_fields(cls,
...
+                # field_types is a list, doc should also be a list. We would
+                # also like to assert that a non-list field_type implies a
+                # non-list doc. Such an assertion could only be done for
+                # contributions, and only after https://github.com/DataBiosphere/azul/issues/2689
+                assert not forward or doc is None or isinstance(doc, list)
```
hannes-ucsc, line 413: Why do we assert this only in forward mode?

hannes-ucsc (changes requested):
I'll push fixes for some of these.

hannes-ucsc:
Forgot to assign and unassign myself.

src/azul/indexer/document_service.py
jessebrennan: I did that intentionally, but I see why it would make sense to wait until that feature is actually implemented.

src/azul/indexer/document_service.py
jessebrennan: I ran tests and check the coverage report and it looks like this line is never touched https://codecov.io/gh/DataBiosphere/azul/pull/2697/diff?src=pr&el=tree#D3-74.

I traced it back to this PR https://github.com/DataBiosphere/azul/pull/1382/files#diff-73be64e9409fcaf8430fde4dd1a7e10f31d269606de6438161de8034281228d5R76.

I'll ask @danielsotirhos about this, but otherwise I think it should go.

src/azul/indexer/document.py
jessebrennan: The assert should apply in both directions. It required a small tweak to make this work.

src/azul/indexer/document_service.py
jessebrennan: From the Slack conversation with @danielsotirhos, I'm concluding that this is dead code and have removed it https://ucsc-gi.slack.com/archives/C705Y6G9Z/p1610739151024300.

src/azul/indexer/document_service.py
hannes-ucsc: You make it sound as if there is a trade-off here. If it was your intention to handle a case that shouldn't actually occur, that intention was misguided. Your implementation is unsound (its parts are inconsistent) and therefore confusing at best, and incorrect at worst.

jessebrennan: All I was trying to do in my last comment was admit my mistake / misunderstanding. Any other interpretations were unintended.

src/azul/indexer/document.py
```diff
@@ -403,10 +403,9 @@ def translate_fields(cls,
         else:
             if isinstance(field_types, list):
                 # FIXME: Assert that a non-list field_type implies a non-list
-                #        doc (only possible for for contributions).
+                #        doc (only possible for contributions).
                 #        https://github.com/DataBiosphere/azul/issues/2689
-                if forward:
-                    assert doc is None or isinstance(doc, list)
+                assert doc in (None, NullableString.null_string) or isinstance(doc, list)
```
hannes-ucsc, line 408: To tighten the assert, it should take the direction of the translation into account.

hannes-ucsc (changes requested):

src/azul/indexer/document.py
```diff
@@ -396,8 +400,19 @@ def translate_fields(cls,
...
             else:
                 assert False, (path, type(doc))
-        elif isinstance(field_types, FieldType):
-            field_type = field_types
+        else:
+            if isinstance(field_types, list):
+                # FIXME: Assert that a non-list field_type implies a non-list
+                #        doc (only possible for contributions).
+                #        https://github.com/DataBiosphere/azul/issues/2689
+                assert ((forward and doc is None)
```
hannes-ucsc, line 408: Should probably use `if` expression to avoid the symmetric guards (`forward` and `not forward`).

src/azul/plugins/metadata/hca/transform.py
```diff
@@ -948,6 +948,14 @@ def _contribution(self, contents: MutableJSON, entity_id: api.UUID4) -> Contribu
 
     @classmethod
     def field_types(cls) -> FieldTypes:
+        """
+        Field types outline the general shape of our documents.
+        """
+        # FIXME: Not all information is captured. Lists of primitive types are
+        #        represented, but lists of container types are not. Eventually,
+        #        we want field_types to more accurately describe the snape of
```
hannes-ucsc, line 956:
```suggestion
        #        we want field_types to more accurately describe the shape of
```

hannes-ucsc (changes requested):

src/azul/indexer/document.py
```diff
@@ -396,8 +400,18 @@ def translate_fields(cls,
...
             else:
                 assert False, (path, type(doc))
-        elif isinstance(field_types, FieldType):
-            field_type = field_types
+        else:
+            if isinstance(field_types, list):
+                # FIXME: Assert that a non-list field_type implies a non-list
+                #        doc (only possible for contributions).
+                #        https://github.com/DataBiosphere/azul/issues/2689
+                assert ((doc is None if forward else doc == NullableString.null_string)
```
hannes-ucsc, line 408: I find it odd that this mentions a specific null representation, those of nullable strings. I'm not even sure we support nullable lists. If you omit this term, does the assertion fire during unit or integration tests? If so where?

hannes-ucsc (commented):
No bump.

src/azul/indexer/document.py
jessebrennan: You're right, that is odd. It should have raised a flag for me.

It causes some unit tests to fail.

One such failure is `test_hca_indexer.py`'s `test_cell_line_sample()`:
```
  File "/home/jesse/.pyenv/versions/3.8.3/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/home/jesse/.pyenv/versions/3.8.3/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/home/jesse/.pyenv/versions/3.8.3/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/jesse/gi/azul/test/indexer/test_hca_indexer.py", line 1091, in test_cell_line_sample
    self._index_canned_bundle(bundle_fqid)
  File "/home/jesse/gi/azul/test/indexer/__init__.py", line 100, in _index_canned_bundle
    cls._index_bundle(bundle, delete=delete)
  File "/home/jesse/gi/azul/test/indexer/__init__.py", line 107, in _index_bundle
    cls.index_service.index(cls.catalog, bundle)
  File "/home/jesse/gi/azul/src/azul/indexer/index_service.py", line 147, in index
    self.aggregate(tallies)
  File "/home/jesse/gi/azul/src/azul/indexer/index_service.py", line 244, in aggregate
    contributions = self._read_contributions(total_tallies)
  File "/home/jesse/gi/azul/src/azul/indexer/index_service.py", line 362, in _read_contributions
    contributions = [
  File "/home/jesse/gi/azul/src/azul/indexer/index_service.py", line 363, in <listcomp>
    Contribution.from_index(self.catalogued_field_types(), hit)
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 472, in from_index
    source = cls.translate_fields(hit['_source'], field_types[coordinates.entity.catalog], forward=False)
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 393, in translate_fields
    new_doc[key] = cls.translate_fields(val, field_type, forward=forward, path=(*path, key))
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 393, in translate_fields
    new_doc[key] = cls.translate_fields(val, field_type, forward=forward, path=(*path, key))
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 400, in translate_fields
    return [cls.translate_fields(val, field_types, forward=forward, path=path) for val in doc]
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 400, in <listcomp>
    return [cls.translate_fields(val, field_types, forward=forward, path=path) for val in doc]
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 393, in translate_fields
    new_doc[key] = cls.translate_fields(val, field_type, forward=forward, path=(*path, key))
  File "/home/jesse/gi/azul/src/azul/indexer/document.py", line 408, in translate_fields
    assert doc is None or isinstance(doc, list)
AssertionError
```

jessebrennan: I will provide more info on what is doc and path. Possibly provide a solution.

jessebrennan: When this fails, doc is `~null`, field_types is `[NullableString]`, path is `contents.samples.disease`.

jessebrennan: The reason we get a None value even though, from field_types we expect at least a list of None is because we're looking at `samples`. Samples contain all of the fields for it's composite parts (so samples have a consistent document shape). If the type that samples represents doesn't contain a field (as is the case of disease) the field is stored as `None`.

In the particular case I was debugging, `Samples` was representing a `CellLine`, not a `Specimen` so the disease field was stored as `None`.

jessebrennan: One solution is to ignore the assertion of we're looking at a samples. I think the assert could then be changed to something like:
```python
assert is_samples() or isinstance(doc, list)
```
or even stronger:
```python
assert not is_samples() == isinstance(doc, list)
```
although I haven't tested that (and won't tonight :moon: )

jessebrennan: I confirmed this works, I'm pushing a version for review.

src/azul/indexer/document.py
```diff
@@ -405,8 +405,10 @@ def translate_fields(cls,
                 # FIXME: Assert that a non-list field_type implies a non-list
                 #        doc (only possible for contributions).
                 #        https://github.com/DataBiosphere/azul/issues/2689
-                assert ((doc is None if forward else doc == NullableString.null_string)
-                        or isinstance(doc, list))
+                # Samples are an exception since they are a composite type.
+                # Unused fields are left as None, even if field_types would
+                # normally dictate otherwise.
+                assert isinstance(doc, list) or path[:2] == ('contents', 'samples')
```
hannes-ucsc, line 411: The second term should be stricter. Convert the comment to a FIXME referencing #2071.

Composite is not the right term, BTW. Samples are polymorphic.

hannes-ucsc (changes requested):

src/azul/indexer/document.py
jessebrennan: I plan to ask during parking lot: stricter how?

jessebrennan: Combine with previous assertion: samples implies None or nullstr.

hannes-ucsc (approved):

hannes-ucsc:
@DataBiosphere/azul-operators, note chain label.

hannes-ucsc:
For demo, discuss diff, show changes made to _foo_types methods.

Issue #2766, opened 2021-02-05
Add ability to index a single source


hannes-ucsc:
We'll be asked to support "updated" snapshot. This requires removing one snapshot from the catalog and adding another one to replace it.

Add an argument to `reindex.py` such that it only sends remote reindex notification for a single source.

Parameterize the Gitlab reindex job so that it can optionally reindex a single source.

PR #2777, opened 2021-02-09
Add ability to index a single source (#2766)

noah-aviel-dove:
Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history

Operator (reindex) 

Operator

- [ ] Unassigned PR

### DIFF
--- a/scripts/reindex.py
+++ b/scripts/reindex.py
@@ -3,6 +3,10 @@ Command line utility to trigger indexing of bundles from DSS into Azul
 """
 
 import argparse
+from collections import (
+    defaultdict,
+)
+import fnmatch
 import logging
 import sys
 from typing import (
@@ -14,6 +18,7 @@ from args import (
 )
 from azul import (
     config,
+    require,
 )
 from azul.azulclient import (
     AzulClient,
@@ -64,6 +69,12 @@ parser.add_argument('--catalogs',
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--sources',
+                    default=config.reindex_sources,
+                    nargs='+',
+                    help='Limit remote reindexing to a subset of the configured sources. '
+                         'Supports shell-style wildcards to match multiple sources per argument. '
+                         'Must be * for local reindexing i.e., if --partition-prefix-length is not given.')
 parser.add_argument('--delete',
                     default=False,
                     action='store_true',
@@ -109,28 +120,59 @@ def main(argv: List[str]):
 
     azul = AzulClient(num_workers=args.num_workers)
 
+    source_globs = set(args.sources)
+    if args.partition_prefix_length:
+        sources_by_catalog = defaultdict(set)
+        globs_matched = set()
+        for catalog in args.catalogs:
+            sources = azul.catalog_sources(catalog)
+            for source_glob in source_globs:
+                matches = fnmatch.filter(sources, source_glob)
+                if matches:
+                    globs_matched.add(source_glob)
+                logger.debug('Source glob %r matched sources %r in catalog %r',
+                             source_glob, matches, catalog)
+                sources_by_catalog[catalog].update(matches)
+        unmatched = source_globs - globs_matched
+        if unmatched:
+            logger.warning('Source(s) not found in any catalog: %r', unmatched)
+        require(any(sources_by_catalog.values()),
+                'No valid sources specified for any catalog')
+    else:
+        if source_globs == {'*'}:
+            sources_by_catalog = {
+                catalog: azul.catalog_sources(catalog)
+                for catalog in args.catalogs
+            }
+        else:
+            parser.error('Cannot specify sources when performing a local reindex')
+            assert False
+
     azul.reset_indexer(args.catalogs,
                        purge_queues=args.purge,
                        delete_indices=args.delete,
                        create_indices=args.create or args.index and args.delete)
 
     if args.index:
-        slot_manager = None
         logger.info('Queuing notifications for reindexing ...')
+        slot_manager = None
         num_notifications = 0
-        for catalog in args.catalogs:
-            if (
-                args.manage_slots
-                and slot_manager is None
-                and isinstance(azul.repository_plugin(catalog), tdr.Plugin)
-            ):
-                slot_manager = SlotManager()
-                slot_manager.ensure_slots_active()
-            if args.partition_prefix_length:
-                azul.remote_reindex(catalog, args.prefix, args.partition_prefix_length)
-                num_notifications = None
+        for catalog, sources in sources_by_catalog.items():
+            if sources:
+                if (
+                    args.manage_slots
+                    and slot_manager is None
+                    and isinstance(azul.repository_plugin(catalog), tdr.Plugin)
+                ):
+                    slot_manager = SlotManager()
+                    slot_manager.ensure_slots_active()
+                if args.partition_prefix_length:
+                    azul.remote_reindex(catalog, args.prefix, args.partition_prefix_length, sources)
+                    num_notifications = None
+                else:
+                    num_notifications += azul.reindex(catalog, args.prefix)
             else:
-                num_notifications += azul.reindex(catalog, args.prefix)
+                logger.info('Skipping catalog %r (no matching sources)', catalog)
         if args.wait:
             if num_notifications == 0:
                 logger.warning('No notifications for prefix %r and catalogs %r were sent',
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -2,6 +2,7 @@ import functools
 import logging
 import os
 import re
+import shlex
 from typing import (
     AbstractSet,
     ClassVar,
@@ -855,6 +856,12 @@ class Config:
     @property
     def dynamo_object_version_table_name(self) -> str:
         return self.qualified_resource_name('object_versions')
+
+    @property
+    def reindex_sources(self) -> List[str]:
+        sources = shlex.split(os.environ.get('azul_reindex_sources', '*'))
+        require(sources, 'Sources cannot be empty', sources)
+        return sources
 
     terms_aggregation_size = 99999
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -26,8 +26,8 @@ PEP
 Line length
 -----------
 
-* For prose (documentation, messages, comments) wrap lines at the word boundary
-  closest to or at, but not beyond, column 79. The first column is column 0.
+* For prose (documentation, comments) wrap lines at the word boundary closest to
+  or at, but not beyond, column 79. The first column is column 0.
 
 * For code, we keep the trimmed line length under 81. A trimmed line is a line
   in the source with leading and trailing whitespace removed. This means a line
--- a/src/azul/azulclient.py
+++ b/src/azul/azulclient.py
@@ -230,14 +230,14 @@ class AzulClient(object):
     def remote_reindex(self,
                        catalog: CatalogName,
                        prefix: str,
-                       partition_prefix_length: int):
+                       partition_prefix_length: int,
+                       sources: AbstractSet[str]):
         validate_uuid_prefix(prefix)
         partition_prefixes = [
             prefix + ''.join(partition_prefix)
             for partition_prefix in product('0123456789abcdef',
                                             repeat=partition_prefix_length)
         ]
-        sources = self.repository_plugin(catalog).sources
 
         def message(source: str, partition_prefix: str) -> JSON:
             logger.info('Remotely reindexing prefix %r of source %r into catalog %r',


noah-aviel-dove:
Tests pass, build is red due to codecov not finding cache files

scripts/reindex.py
```diff
@@ -105,9 +109,20 @@ def main(argv: List[str]):
 
     if args.index:
         logger.info('Queuing notifications for reindexing ...')
+        if args.source:
+            if not args.partition_prefix_length:
+                parser.error('Cannot specify a source when performing a local reindex')
+            for catalog in args.catalogs:
+                require(args.source in azul.catalog_sources(catalog),
+                        f'Source {args.source!r} not found in catalog '
+                        f'{catalog!r}')
```
danielsotirhos, line 118:
```suggestion
                        f'Source {args.source!r} not found in catalog {catalog!r}')
```

danielsotirhos (approved):
Just an unnecessary new line, otherwise looks great.

scripts/reindex.py
noah-aviel-dove: I think this count as a "message" and thus should be wrapped at 80 chars (absolute) according to the contributing guide : 

> For prose (documentation, messages, comments) wrap lines at the word boundary closest to or at, but not beyond, column 79. The first column is column 0.

Let me know if you think I'm misinterpreting this section.

Makefile
```diff
@@ -102,6 +102,8 @@ subscribe: check_python check_branch
 unsubscribe: check_python check_branch
 	python scripts/subscribe.py --unsubscribe
 
+gitlab_reindex_source = $(shell [ -z $${reindex_source+x} ] || echo "--source=$${reindex_source}")
```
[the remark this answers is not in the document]
hannes-ucsc: I don't like to have elaborate shell logic in the Makefile. See if you can implement this such that `--source` (w/o args) behaves like --source wasn't passed. We could also make the source option a positional argument.

scripts/reindex.py
```diff
@@ -58,6 +59,9 @@
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--source',
+                    nargs='?',
```
hannes-ucsc, line 63: Seems like 
```suggestion
                    nargs='+',
```

is low hanging fruit. If it can be done easily please do so.

hannes-ucsc (changes requested):
I don't see any implementation of the 

> Parameterize the Gitlab reindex job so that it can optionally reindex a single source.

part of the ticket specification.

Take a look at https://docs.gitlab.com/ee/ci/variables/#create-a-custom-variable-in-gitlab-ciyml for details.

scripts/reindex.py
noah-aviel-dove: `'+'` seems to be incompatible with allowing an empty `--source` arg but `'*'` works.

Makefile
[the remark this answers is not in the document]
noah-aviel-dove: Not an azul-defined variable, my intent was for this to be the parameter entered on gitlab. But apparently I misunderstood that part of the instructions.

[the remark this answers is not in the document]
hannes-ucsc: Right. Let me explain: Any environment variable that we introduce, that has no meaning outside of Azul, counts as an Azul-defined variable. By that definition `reindex_source` is such a variable.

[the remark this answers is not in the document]
hannes-ucsc: The reason for the prefix is to avoid conflicts with other systems.

hannes-ucsc (commented):
The UI for passing variables to jobs is extremely lacking in Gitlab. There is no selection from a list of supported variables. I just assumed there would be since every other CI system had some form of defining job parameters. But worse, there is https://gitlab.com/gitlab-org/gitlab/-/issues/37268 which makes recovering from mistakes with UI impossible.

I don't think this is the right direction since it will be error prone to specify a source in the Gitlab UI an no way to fix mistakes.

I'll leave this in the review column for now until I have a better idea.

scripts/reindex.py
```diff
@@ -58,6 +62,13 @@
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--sources',
+                    nargs='*',
+                    default=(),
```
hannes-ucsc, line 67:
```suggestion
                    default=('*'),
```

And then we'll use globbing to match. This will make it easier to specify multiple source that share a name prefix. Special care must be taken to avoid these globs from being consumed by the shell.

.gitlab-ci.yml
```diff
@@ -1,6 +1,7 @@
 variables:
   DOCKER_IMAGE: $CI_REGISTRY_IMAGE
   DOCKER_TAG: $CI_PIPELINE_ID
+  AZUL_INDEX_SOURCES: ''
```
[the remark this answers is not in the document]
hannes-ucsc: And the name should have `reindex` in it. So ultimately it should be 

`azul_reindex_sources`

scripts/reindex.py
hannes-ucsc: It'll help to make this script aware of `azul_reindex_sources`, so we don't need to modify the makefile. So something along the lines of

```suggestion
                    default=os.environ.get('azul_reindex_sources', '*').split(','),
```

hannes-ucsc (changes requested):
After thinking about this a bit more, I think we can go ahead with this implementation and see how it works in practice.

hannes-ucsc:
No bump.

scripts/reindex.py
```diff
@@ -58,6 +61,13 @@
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--sources',
+                    default=os.environ.get('azul_reindex_sources', '*'),
```
hannes-ucsc, line 65: We should probably go through azul.Config

```suggestion
                    default=config.reindex_sources,
```

scripts/reindex.py
```diff
@@ -96,8 +106,24 @@ def main(argv: List[str]):
 
     configure_script_logging(logger)
 
+    sources = set(args.sources.split(','))
```
hannes-ucsc, line 109: I know I suggested comma as the separator but I would like to try shell-style word splitting on consecutive whitespace instead. It's more consistent with using a shell-like mechanism for globbing and it would allow us to use `nargs='+'` again.

hannes-ucsc (changes requested):
No bump.

hannes-ucsc (changes requested):
A few nits. I haven't tested these.

```diff
Index: scripts/reindex.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/scripts/reindex.py b/scripts/reindex.py
--- a/scripts/reindex.py	(revision 7239b2b4f437c9197e350b0763393f1e9107ec4b)
+++ b/scripts/reindex.py	(date 1614301628212)
@@ -3,6 +3,7 @@
 """
 
 import argparse
+from collections import defaultdict
 import fnmatch
 import logging
 import sys
@@ -105,19 +106,18 @@
 
     configure_script_logging(logger)
 
-    sources = set(args.sources)
-    if sources != {'*'} and not args.partition_prefix_length:
+    source_globs = set(args.sources)
+    if source_globs != {'*'} and not args.partition_prefix_length:
         parser.error('Cannot specify sources when performing a local reindex')
 
     azul = AzulClient(num_workers=args.num_workers)
 
-    sources_by_catalog = {}
+    sources_by_catalog = defaultdict(set)
     for catalog in args.catalogs:
-        sources_by_catalog[catalog] = set()
-        configured_sources = azul.catalog_sources(catalog)
-        for source in sources:
-            matches = fnmatch.filter(configured_sources, source)
-            require(matches, 'No matching source in catalog', source, catalog)
+        sources = azul.catalog_sources(catalog)
+        for source_glob in source_globs:
+            matches = fnmatch.filter(sources, source_glob)
+            require(matches, 'No matching source in catalog', source_glob, catalog)
             sources_by_catalog[catalog].update(matches)
     assert all(sources_by_catalog.values())
 
```

noah-aviel-dove:
Tested changed and fixed an issue with the reindex failing after resetting the indexer when `azul_reindex_sources` is set to an empty string.

scripts/reindex.py
```diff
@@ -58,6 +63,13 @@
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--sources',
+                    default=config.reindex_sources,
+                    nargs='+',
+                    help='Limit remote reindexing to a subset of the configured sources. '
+                         'Supports shell-style * globbing to match multiple sources per argument. '
+                         'Each argument must match at least one configured source for every selected catalog. '
+                         'Default: %(default)s')
```
hannes-ucsc, line 72: Is this necessary?

scripts/reindex.py
```diff
@@ -96,8 +108,21 @@ def main(argv: List[str]):
...
+        parser.error('Cannot specify sources when performing a local reindex')
+
     azul = AzulClient(num_workers=args.num_workers)
 
+    sources_by_catalog = defaultdict(set)
+    for catalog in args.catalogs:
+        sources = azul.catalog_sources(catalog)
+        for source_glob in source_globs:
+            matches = fnmatch.filter(sources, source_glob)
+            require(matches, 'No matching source in catalog', source_glob, catalog)
```
hannes-ucsc, line 122: I think we should allow there to be no matches. If I understand this correctly, I'd be forced to specify the catalog of a source if I specify a source because the default value for the catalogs option would likely reference a catalog that doesn't match the source. That's impractical, I think.

hannes-ucsc (changes requested):

scripts/reindex.py
noah-aviel-dove: Not necessarily I suppose. Checking that it occurs in at least one catalog should be sufficient.

hannes-ucsc: I meant the inclusion of the default in the help string. The formatter should do that for us, no?

noah-aviel-dove: Ah, you are correct/

scripts/reindex.py
```diff
@@ -96,8 +106,29 @@ def main(argv: List[str]):
...
+    globs_matched = set()
+    for catalog in args.catalogs:
+        sources = azul.catalog_sources(catalog)
+        for source_glob in source_globs:
+            matches = fnmatch.filter(sources, source_glob)
+            if matches:
+                globs_matched.add(source_glob)
+            logger.debug('Matched sources %r in catalog %r', matches, catalog)
+            sources_by_catalog[catalog].update(matches)
+        reject(not sources_by_catalog[catalog],
```
hannes-ucsc, line 125: Doesn't quite capture my intention behind https://github.com/DataBiosphere/azul/pull/2777#discussion_r583911217

If the user specifies a very specific glob that only matches one source in one catalog, the user shouldn't have to then also specify the catalog. What's the point in doing that? The glob alone is already unambiguous. 

If a glob matches two or more sources from different catalogs (for example if two catalog use the same source) and the user doesn't want to reindex both catalogs, the user can then opt to disambiguate via `--catalogs`.

hannes-ucsc (changes requested):

scripts/reindex.py
```diff
@@ -96,8 +106,31 @@ def main(argv: List[str]):
...
+        sources = azul.catalog_sources(catalog)
+        for source_glob in source_globs:
+            matches = fnmatch.filter(sources, source_glob)
+            if matches:
+                globs_matched.add(source_glob)
+            logger.debug('Matched sources %r in catalog %r', matches, catalog)
+            sources_by_catalog[catalog].update(matches)
+    unmatched = source_globs - globs_matched
+    if unmatched:
+        logger.warning('Source(s) not found in any catalog: %r', unmatched)
```
hannes-ucsc, line 127:
```suggestion
```

It's not a problem to for a catalog not to match. Note even a potential one.

scripts/reindex.py
```diff
@@ -96,8 +106,31 @@ def main(argv: List[str]):
...
+    for catalog, sources in sources_by_catalog.items():
+        if not sources:
+            logger.info('Skipping catalog %r (no applicable sources)', catalog)
+            args.catalogs.remove(catalog)
```
hannes-ucsc, line 131: I'd rather not modify args.sources

hannes-ucsc (changes requested):

scripts/reindex.py
noah-aviel-dove: @hannes-ucsc: "I misunderstood, ignore the diff suggestion."

scripts/reindex.py
noah-aviel-dove: (Assuming you meant `args.catalogs`)

scripts/reindex.py
```diff
@@ -120,16 +120,15 @@ def main(argv: List[str]):
             matches = fnmatch.filter(sources, source_glob)
             if matches:
                 globs_matched.add(source_glob)
-            logger.debug('Matched sources %r in catalog %r', matches, catalog)
+            logger.debug('Source glob %r matched sources %r in catalog %r',
+                         source_glob, matches, catalog)
```
noah-aviel-dove, line 124: While testing, I observed this log message to be useless without explaining which source was being matched.

hannes-ucsc (changes requested):
I think we should do something like this? It makes the bifurcation more clear.

```diff
Index: scripts/reindex.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/scripts/reindex.py b/scripts/reindex.py
--- a/scripts/reindex.py	(revision c1f6ccc03cd7a438a8044bf1bd46e5cab674a618)
+++ b/scripts/reindex.py	(date 1616189367366)
@@ -106,29 +106,35 @@
 
     configure_script_logging(logger)
 
-    source_globs = set(args.sources)
-    if source_globs != {'*'} and not args.partition_prefix_length:
-        parser.error('Cannot specify sources when performing a local reindex')
-
     azul = AzulClient(num_workers=args.num_workers)
 
-    sources_by_catalog = defaultdict(set)
-    globs_matched = set()
-    for catalog in args.catalogs:
-        sources = azul.catalog_sources(catalog)
-        for source_glob in source_globs:
-            matches = fnmatch.filter(sources, source_glob)
-            if matches:
-                globs_matched.add(source_glob)
-            logger.debug('Source glob %r matched sources %r in catalog %r',
-                         source_glob, matches, catalog)
-            sources_by_catalog[catalog].update(matches)
-    unmatched = source_globs - globs_matched
-    if unmatched:
-        logger.warning('Source(s) not found in any catalog: %r', unmatched)
-
-    require(any(sources_by_catalog.values()),
-            'No valid sources specified for any catalog')
+    source_globs = set(args.sources)
+    if args.partition_prefix_length:
+        sources_by_catalog = defaultdict(set)
+        globs_matched = set()
+        for catalog in args.catalogs:
+            sources = azul.catalog_sources(catalog)
+            for source_glob in source_globs:
+                matches = fnmatch.filter(sources, source_glob)
+                if matches:
+                    globs_matched.add(source_glob)
+                logger.debug('Source glob %r matched sources %r in catalog %r',
+                             source_glob, matches, catalog)
+                sources_by_catalog[catalog].update(matches)
+        unmatched = source_globs - globs_matched
+        if unmatched:
+            logger.warning('Source(s) not found in any catalog: %r', unmatched)
+        require(any(sources_by_catalog.values()),
+                'No valid sources specified for any catalog')
+    else:
+        if source_globs == {'*'}:
+            sources_by_catalog = {
+                azul.catalog_sources(catalog)
+                for catalog in args.catalogs
+            }
+        else:
+            parser.error('Cannot specify sources when performing a local reindex')
+            assert False
 
     azul.reset_indexer(args.catalogs,
                        purge_queues=args.purge,
```

noah-aviel-dove:
Patch applied and tested.

scripts/reindex.py
```diff
@@ -98,18 +108,52 @@ def main(argv: List[str]):
...
+                             source_glob, matches, catalog)
+                sources_by_catalog[catalog].update(matches)
+        unmatched = source_globs - globs_matched
+        if unmatched:
+            logger.warning('Source(s) not found in any catalog: %r', unmatched)
+        require(any(sources_by_catalog.values()),
+                'No valid sources specified for any catalog')
+    else:
+        if source_globs == {'*'}:
+            sources_by_catalog = {
```
hannes-ucsc, line 131: This looks like it creates a set of dictionaries. That can't be right. Was this branch of the control flow tested?

hannes-ucsc (changes requested):

scripts/reindex.py
noah-aviel-dove: Seems I overlooked the case where `*` is used without any additional sources.

noah-aviel-dove: I misinterpreted the context here, I did test `*` by itself, it was the local reindex I overlooked.

scripts/reindex.py
```diff
@@ -58,6 +63,11 @@
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--sources',
+                    default=config.reindex_sources,
+                    nargs='+',
+                    help='Limit remote reindexing to a subset of the configured sources. '
+                         'Supports shell-style * globbing to match multiple sources per argument.')
```
hannes-ucsc, line 70:
```suggestion
                    help='Limit remote reindexing to a subset of the configured sources. '
                         'Supports shell-style wildcards to match multiple sources per argument.
                         Must be * for local reindexing i.e., if --partition-prefix-length is not given.')
```

src/azul/__init__.py
```diff
@@ -850,6 +851,12 @@ def lambda_layer_key(self) -> str:
     def dynamo_object_version_table_name(self) -> str:
         return self.qualified_resource_name('object_versions')
 
+    @property
+    def reindex_sources(self) -> List[str]:
+        sources = shlex.split(os.environ.get('azul_reindex_sources', '*'))
+        require(sources, 'Sources cannot be empty')
```
hannes-ucsc, line 857:
```suggestion
        require(sources, 'Sources cannot be empty', sources)
```

hannes-ucsc (changes requested):

scripts/reindex.py
```diff
@@ -58,6 +63,12 @@
                     ],
                     choices=config.catalogs,
                     help='The names of the catalogs to reindex.')
+parser.add_argument('--sources',
+                    default=config.reindex_sources,
+                    nargs='+',
+                    help='Limit remote reindexing to a subset of the configured sources. '
+                         'Supports shell-style * globbing to match multiple sources per argument.'
```
hannes-ucsc, line 70: https://github.com/DataBiosphere/azul/pull/2777#discussion_r603466521

hannes-ucsc (changes requested):

hannes-ucsc (changes requested):
Conflicts. Rest looks good.

hannes-ucsc (approved):

hannes-ucsc:
For demo, run `reindex.py` 

- with a glob other than `*` that matches a source in more than one catalog
- with a glob that matches no source in any catalog
- show the mutual exclusivity of specifying a glob with local reindex

amarjandu:
Screen Shot 2021-04-06 at 3 44 54 PM

@hannes-ucsc looks like you might have misses these check-boxes, I'll click them for a small fee.

Issue #2776, opened 2021-02-09
test_fetch_bundle fails in `prod` deployment


hannes-ucsc:
… due to a brittle coupling with the value of the `AZUL_TDR_SERVICE_URL` variable.

https://gitlab.azul.data.humancellatlas.org/ucsc/azul/-/jobs/388

PR #2780, opened 2021-02-10
Fix: test_fetch_bundle fails in `prod` deployment (#2776)

noah-aviel-dove:
Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history

Operator (reindex) 

Operator

- [ ] Unassigned PR


Makefile
```diff
@@ -190,3 +190,11 @@ check_autosquash: check_env
 readme: check_docker
 	docker pull evkalinin/gh-md-toc:0.7.0
 	docker run -it -v $$PWD:/build evkalinin/gh-md-toc:0.7.0 --no-backup /build/README.md
+
+.PHONY: tdr_canned_bundle
+tdr_canned_bundle: check_python
```
jessebrennan, line 195: Is the point of this target just to make sure that the bundles are always generated in the same order? When is the target supposed to run?

If someone calls the script manually, instead of through the Makfile, you'll have a duplicate UUID, right? That would probably cause problems. If this is the case, consider putting a note in the script that warns it should only be called via the Makefile.

Finally, consider
```suggestion
tdr_canned_bundles: check_python
```

noah-aviel-dove: > Is the point of this target just to make sure that the bundles are always generated in the same order?

Only bundle is generated. The point is to avoid the tedium/potential source of error of having to manually match the snapshot and bundle UUID to their values in `test_tdr.py`, as I frequently had to do while testing my changes. My other changes in that commit ensure that the _entities_ are consistently sorted within the canned bundle, which helps keep the diff small when changes to the recanning script are made, as in the second commit.

> When is the target supposed to run?

When changes are made to the recanning script that affect its output.

> If someone calls the script manually, instead of through the Makfile, you'll have a duplicate UUID, right?

I'm not entirely sure what you mean by this, but I think you're referring to the eventuality of generating a second canned bundle from a different underlying DSS bundle, thus resulting in two canned TDR bundles where different entities have the same UUID? I hadn't considered that, but I can see how it could be a potential problem. I don't see what it has to do with the make target though. It strikes me as a consequence of making the generated UUIDs deterministic; the duplicates would occur whenever running the script with multiple bundles, regardless of whether the script is invoked through the make target or not.

Have I correctly interpreted your concern?

> Finally, consider ...

The script only outputs a single bundle per invocation, and the target only invokes the script once. I believe the target is correctly named. I think using a plural here would lead to confusion regarding the format of the script's output.

noah-aviel-dove:
No changes, just rebase and reply to comments.

Makefile
jessebrennan: Assuming you meant
> Only **one** bundle is generated
I overlooked that, sorry. The name definitely makes sense as is.

Thanks for the thorough explanation in regards to everything else.

> Have I correctly interpreted your concern?

Yes.

> the duplicates would occur whenever running the script with multiple bundles, regardless of whether the script is invoked through the make target or not.

You're correct, I was mistaken. I assumed the script could (and was) operating on multiple bundles.

We agree though, if the script is run multiple times on different bundles, the outputs will share "random" UUIDs. If you think that could lead to some unexpected bugs (perhaps if both bundles are used in the same test) then consider either fixing the behavior, or document this in the `recan_bundle_tdr.py` script.

jessebrennan (approved):
I had just one optional thing for you to think about, so I'll approve as is.

Nice work :+1:

Makefile
noah-aviel-dove: Yeah, having "UUID"s that aren't unique feels like an accident waiting to happen. I think I can easily fix this by using the bundle UUID as the namespace instead of hardcoding an arbitrary value.

Oh, and I was mistaken about my changes causing the entities to be sorted, I was thinking of a different PR. The ordering is deterministic, not sorted.

scripts/recan_bundle_tdr.py
```diff
@@ -108,27 +109,26 @@ def content_length(content: JSON) -> int:
...
-
-
 def incrementing_version() -> str:
     return tdr.Plugin.format_version(next(incrementing_versions))
 
 
+namespace_counts: Dict[uuid.UUID, int] = {}
+
+
+def random_uuid(namespace: str) -> str:
```
jessebrennan, line 129: Random doesn't quite feel accurate here anymore, but I don't have any better ideas. `random_deterministic_uuid` is more accurate, but verbose.

scripts/recan_bundle_tdr.py
```diff
@@ -108,27 +109,26 @@ def content_length(content: JSON) -> int:
...
 
-def random_uuid() -> str:
-    return next(random_deterministic_uuids)
-
-
 def incrementing_version() -> str:
     return tdr.Plugin.format_version(next(incrementing_versions))
 
 
+namespace_counts: Dict[uuid.UUID, int] = {}
```
jessebrennan, line 126: What you wrote is perfectly functional, but the ideal data structure already exists in the standard lib https://docs.python.org/3/library/collections.html#collections.Counter

jessebrennan (approved):

hannes-ucsc (changes requested):

scripts/recan_bundle_tdr.py
```diff
@@ -281,20 +263,26 @@ def dump_tables(bundle: TDRBundle) -> MutableJSON:
...
+                   *,
+                   num_files: int,
+                   version: str
+                   ) -> None:
     links_json = bundle.metadata_files['links.json']['links']
     links_manifest = one(e for e in bundle.manifest if e['name'] == 'links.json')
     project_id = bundle.metadata_files['project_0.json']['provenance']['document_id']
 
-    metadata_ids = sorted(random_deterministic_uuid(bundle.uuid) for _ in range(num_files))
+    # UUIDs and document names must BOTH be sorted for consistency with plugin.
```
hannes-ucsc, line 276: They can't both be sorted. Sorting by one yields a different ordering than sorting by the other. Whatever it is you're trying to say is not clear to me.

scripts/recan_bundle_tdr.py
```diff
@@ -110,25 +107,9 @@ def content_length(content: JSON) -> int:
...
-namespace_counts = Counter()
-
-
-def random_deterministic_uuid(namespace: str) -> str:
-    namespace = uuid.UUID(namespace)
-    count = namespace_counts[namespace]
-    namespace_counts.update([namespace])
-    return str(uuid.uuid5(namespace, str(count)))
+def deterministic_uuid(name: str) -> str:
+    namespace = uuid.UUID('8d400234-9f95-4682-8a02-7bfc63bb9d10')
```
hannes-ucsc, line 111: This will allocate the same UUID for different entities in different bundles.

hannes-ucsc (changes requested):
Re title of main commit:

> TDR unit test does not depend on configured service url 

URL is an acronym. Because of the negation it sounds like something is wrong. Generally, I'd avoid describing a change using the outcome, especially when there is a negation.

For example:

`Foo does this` is inferior to `Stop having Foo do that`

and especially

`Foo does not do that` is inferior to `Change Foo to not do that` or `Foo now does this instead of that`.

scripts/recan_bundle_tdr.py
noah-aviel-dove: What I'm trying to say is: when the plugin retrieves entities from bigquery, it sorts them by UUID, and then enumerates them to generate the numerical component of the document names (e.g. the "2" in "cell_supension_2.json"). So both the entity UUIDs and the document names _are_ sorted, albeit not by the same key (document names are ordered numerically, not lexicographically). So, the script needs to follow the same logic, obtain a sequence of sorted UUIDs and then enumerate then, to generate the expected document names. Comment has been updated with an attempt to express this more succinctly.

scripts/recan_bundle_tdr.py
```diff
@@ -329,18 +340,32 @@ def main(argv):
...
     parser.add_argument('--input-dir', '-I',
                         default=os.path.join(config.project_root, 'test', 'indexer', 'data'),
-                        help='The path to the input directory (default: %(default)s).')
+                        help='The path to the input directory '
+                             '(default: %(default)s).')
     parser.add_argument('--add-supp-files',
                         type=int,
                         default=0,
-                        help='How many fake supplementary files to add to the output (default: %(default)s).')
+                        help='How many fake supplementary files to add to the '
```
hannes-ucsc, line 367: Slightly awkward

```suggestion
                        help='The number of fake supplementary files to add to the '
```

test/indexer/test_tdr.py
```diff
@@ -139,6 +145,7 @@ def test_fetch_bundle(self):
         with self.assertRaises(RequirementError):
             self._test_fetch_bundle(self.source, load_tables=False)
 
+    @patch('azul.Config.tdr_service_url', new=PropertyMock(return_value=mock_service_url))
```
hannes-ucsc, line 148: line too long

scripts/recan_bundle_tdr.py
```diff
@@ -109,8 +112,9 @@ def content_length(content: JSON) -> int:
     return len(json.dumps(content).encode('UTF-8'))
 
 
-def random_uuid() -> str:
-    return str(uuid.uuid4())
+def deterministic_uuid(bundle_uuid: str, entity_name: str) -> str:
+    namespace = uuid.UUID(bundle_uuid)
+    return str(uuid.uuid5(namespace, entity_name))
```
hannes-ucsc, line 117: For future reference: At this point it would help if the bundle uuid was contextual but that can only be done by converting this to an class so this would read:

```suggestion
def deterministic_uuid(self, entity_name: str) -> str:
    namespace = uuid.UUID(self.bundle_uuid)
    return str(uuid.uuid5(namespace, entity_name))
```

But leave it as is.

scripts/recan_bundle_tdr.py
```diff
@@ -263,20 +268,26 @@ def dump_tables(bundle: TDRBundle) -> MutableJSON:
...
+                   num_files: int,
+                   version: str
+                   ) -> None:
     links_json = bundle.metadata_files['links.json']['links']
     links_manifest = one(e for e in bundle.manifest if e['name'] == 'links.json')
     project_id = bundle.metadata_files['project_0.json']['provenance']['document_id']
 
-    metadata_ids = sorted(random_uuid() for _ in range(num_files))
+    # Entities must be sorted by UUID for the incrementing document names to
+    # match the plugin's output.
```
hannes-ucsc, line 281: Better but not ideal. The best explanation so far is actually in your reply to my previous comment on this.

scripts/recan_bundle_tdr.py
```diff
@@ -263,20 +268,26 @@ def dump_tables(bundle: TDRBundle) -> MutableJSON:
...
+    metadata_ids = sorted(deterministic_uuid(bundle.uuid, str(i) + 'supp file')
```
hannes-ucsc, line 282: See if you can unify this key and the file name.

scripts/recan_bundle_tdr.py
```diff
@@ -263,20 +268,26 @@ def dump_tables(bundle: TDRBundle) -> MutableJSON:
...
+                          for i in range(num_files))
 
     for i, metadata_id in enumerate(metadata_ids):
```
hannes-ucsc, line 285: This loop would then read 

```suggestion
    for document_name, metadata_id in enumerate(metadata_ids):
```

scripts/recan_bundle_tdr.py
```diff
@@ -109,8 +112,9 @@ def content_length(content: JSON) -> int:
     return len(json.dumps(content).encode('UTF-8'))
 
 
-def random_uuid() -> str:
-    return str(uuid.uuid4())
+def deterministic_uuid(bundle_uuid: str, entity_name: str) -> str:
```
hannes-ucsc, line 115:
```suggestion
def deterministic_uuid(bundle_uuid: str, *name: str) -> str:
```

So you can use a separator, and don't have to do all the string concatenation in the call sites. Ensure that the separator does not occur in the keys. That's crucial for the correctness of such hashing schemes.

hannes-ucsc (changes requested):

scripts/recan_bundle_tdr.py
```diff
@@ -109,8 +113,11 @@ def content_length(content: JSON) -> int:
     return len(json.dumps(content).encode('UTF-8'))
 
 
-def random_uuid() -> str:
-    return str(uuid.uuid4())
+def deterministic_uuid(bundle_uuid: str, *names: str, sep='\0') -> str:
```
hannes-ucsc, line 116: There is no need to make sep a parameter.

scripts/recan_bundle_tdr.py
```diff
@@ -109,8 +113,11 @@ def content_length(content: JSON) -> int:
...
+    namespace = uuid.UUID(bundle_uuid)
+    for name in names:
+        reject(sep in name, 'Separator must not occur in name elements', sep, name)
```
hannes-ucsc, line 119: It's not a user error if this fires. It should be an assert and doesn't need a special message.

```suggestion
        assert sep not in name, (sep, name)
```

scripts/recan_bundle_tdr.py
```diff
@@ -329,18 +353,35 @@ def main(argv):
     Load a canned bundle from DCP/1 and write *.manifest.tdr and *.metadata.tdr
     files showing the desired output for DCP/2.
     """
+    default_version = tdr.Plugin.format_version(datetime(
```
hannes-ucsc, line 356: shouldn't this be aligned indent? Try extracting a variable to eliminate the wrapping altogether

hannes-ucsc (changes requested):

scripts/recan_bundle_tdr.py
```diff
@@ -329,18 +353,33 @@ def main(argv):
     Load a canned bundle from DCP/1 and write *.manifest.tdr and *.metadata.tdr
     files showing the desired output for DCP/2.
     """
+    default_version = datetime(year=2021,
```
hannes-ucsc, line 356: This looks like it fits one line. If it does, it would look better. That's why I requested you extract the variable.

scripts/recan_bundle_tdr.py
```diff
@@ -329,18 +353,33 @@ def main(argv):
...
+                               month=1,
+                               day=17,
+                               hour=0)
+
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument('--bundle-uuid', '-b',
-                        help='The UUID of the existing DCP/1 canned bundle.')
+                        default=TestTDRPlugin.bundle_uuid,
+                        help='The UUID of the existing DCP/1 canned bundle '
+                             '(default: %(default)s).')
```
hannes-ucsc, line 365: There's a help formattter that prints the default automatically. Some of our scripts already use that and so should this one, if possible.

scripts/recan_bundle_tdr.py
```diff
@@ -329,18 +353,33 @@ def main(argv):
...
                         default=os.path.join(config.project_root, 'test', 'indexer', 'data'),
-                        help='The path to the input directory (default: %(default)s).')
+                        help='The path to the input directory '
+                             '(default: %(default)s).')
     parser.add_argument('--add-supp-files',
                         type=int,
                         default=0,
-                        help='How many fake supplementary files to add to the output (default: %(default)s).')
+                        help='The number of fake supplementary files to add to '
+                             'the output (default: %(default)s).')
```
hannes-ucsc, line 382:
```suggestion
                        help='The number of mock supplementary files to add to '
                             'the output (default: %(default)s).')
```

scripts/recan_bundle_tdr.py
```diff
@@ -329,18 +353,33 @@ def main(argv):
...
+    parser.add_argument('--version', '-v',
+                        default=tdr.Plugin.format_version(default_version),
+                        help='The version for any fake entities synthesized by '
+                             'the script (default: %(default)s).')
     parser.add_argument('--input-dir', '-I',
                         default=os.path.join(config.project_root, 'test', 'indexer', 'data'),
-                        help='The path to the input directory (default: %(default)s).')
+                        help='The path to the input directory '
+                             '(default: %(default)s).')
     parser.add_argument('--add-supp-files',
```
hannes-ucsc, line 378:
```suggestion
    parser.add_argument('--mock-supplementary-files', '-S'
```

hannes-ucsc (changes requested):

hannes-ucsc:
For demo, select `prod` and run `indexer.test_tdr.TestTDRPlugin.test_fetch_bundle`

hannes-ucsc (approved):

Issue #2846, opened 2021-03-02
Compact manifest provision invalid zarr related files


achave11:

achave11:
https://github.com/DataBiosphere/azul/pull/2841#discussion_r583945093

hannes-ucsc:
The assignment to @jessebrennan was consistent with my [instructions](https://github.com/DataBiosphere/azul/pull/2841#discussion_r583945093) on PR #2841. The test added by Jesse did not detect this issue but it could and should have.

However, the test was since then improved by @achave11 (just not enough to reproduce this issue) and he is more deeply involved in this now, so I'll be assigning it to him.

hannes-ucsc:
#2538 was reopened because of a similar issue with related files in curl manifests. It needs to be fully addressed before we can attempt to address this one.

achave11:
Unit test failed on [prod deployment in Gitlab](https://gitlab.azul.data.humancellatlas.org/ucsc/azul/-/jobs/1065#L3836).

PR #2841, opened 2021-02-25
Fix: curl manifest entries for related files point at main file (#2538)

achave11:
Author

- [x] PR title references issue
- [x] Title of main commit references issue
- [x] PR is linked to Zenhub issue

Author (reindex)

- [x] Added `r` tag to commit title                         <sub>or this PR does not require reindexing</sub>
- [x] Added `reindex` label to PR                           <sub>or this PR does not require reindexing</sub>

Author (freebies & chains)

- [x] Freebies are blocked on this PR                       <sub>or there are no freebies in this PR</sub>
- [x] Freebies are referenced in commit titles              <sub>or there are no freebies in this PR</sub>
- [x] This PR is blocked by previous PR in the chain        <sub>or this PR is not chained to another PR</sub>
- [x] Added `chain` label to the blocking PR                <sub>or this PR is not chained to another PR</sub>

Author (upgrading)

- [x] Documented upgrading of deployments in UPGRADING.rst  <sub>or this PR does not require upgrading</sub>
- [x] Added `u` tag to commit title                         <sub>or this PR does not require upgrading</sub>
- [x] Added `upgrade` label to PR                           <sub>or this PR does not require upgrading</sub>

Author (requirements, before every review)

- [x] Ran `make requirements_update`                        <sub>or this PR leaves requirements*.txt, common.mk and Makefile untouched</sub>
- [x] Added `R` tag to commit title                         <sub>or this PR leaves requirements*.txt untouched</sub>
- [x] Added `reqs` label to PR                              <sub>or this PR leaves requirements*.txt untouched</sub>

Author (before every review)

- [x] `make integration_test` passes in personal deployment <sub>or this PR does not touch functionality that could break the IT</sub>
- [x] Rebased branch on `develop`, squashed old fixups

Primary reviewer (after approval)

- [x] Commented on demo expectations                        <sub>or labelled PR as `no demo`</sub>
- [x] Decided if PR can be labeled `no sandbox`
- [x] PR title is appropriate as title of merge commit
- [x] Moved ticket to Approved column
- [x] Assigned PR to an operator

Operator (before pushing merge the commit)

- [x] Checked `reindex` label and `r` commit title tag
- [x] Checked that Demo expectations are clear              <sub>or PR is labeled as `no demo`</sub>
- [x] Rebased and squashed branch
- [x] Sanity-checked history
- [x] Pushed PR branch to Github
- [x] Branch pushed to Gitlab and build passes in `sandbox` <sub>or added `no sandbox` label</sub>
[...]

PR #2847, opened 2021-03-02
Fix: Compact manifest rows for related files have main file's DRS URI (#2846)

achave11:
Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history

Operator (reindex) 

Operator

- [ ] Unassigned PR

### DIFF
--- a/src/azul/service/manifest_service.py
+++ b/src/azul/service/manifest_service.py
@@ -41,7 +41,6 @@ import time
 from typing import (
     Any,
     IO,
-    Iterable,
     List,
     Mapping,
     MutableMapping,
@@ -110,7 +109,6 @@ from azul.service.storage_service import (
 from azul.types import (
     JSON,
     JSONs,
-    MutableJSON,
 )
 from azul.vendored.frozendict import (
     frozendict,
@@ -911,25 +909,26 @@ class CompactManifestGenerator(StreamingManifestGenerator):
             for bundle in list(doc['bundles'])[0:100]:  # iterate over copy …
                 doc['bundles'] = [bundle]  # … to facilitate this in-place modification
                 row = {}
+                related_rows = []
                 for doc_path, column_mapping in self.manifest_config.items():
                     entities = [
                         dict(e, file_url=file_url)
                         for e in self._get_entities(doc_path, doc)
                     ]
                     self._extract_fields(entities, column_mapping, row)
+                    if doc_path == 'contents.files':
+                        entity = one(entities)
+                        if 'related_files' in entity:
+                            for file in entity['related_files']:
+                                related_row = {}
+                                entity.update(file)
+                                self._extract_fields([entity], column_mapping, related_row)
+                                related_rows.append(related_row)
                 writer.writerow(row)
-                writer.writerows(self._get_related_rows(doc, row))
+                for related in related_rows:
+                    row.update(related)
+                    writer.writerow(row)
         return None
-
-    def _get_related_rows(self, doc: JSON, row: MutableJSON) -> Iterable[MutableJSON]:
-        file_ = one(doc['contents']['files'])
-        for related in file_['related_files']:
-            # FIXME: Properly provision related_files in row
-            #        https://github.com/DataBiosphere/azul/issues/2846
-            del related['drs_path']
-            new_row = row.copy()
-            new_row.update({'file_' + k: v for k, v in related.items()})
-            yield new_row
 
 
 class FullManifestGenerator(StreamingManifestGenerator):
--- a/test/service/test_manifest.py
+++ b/test/service/test_manifest.py
@@ -342,6 +342,7 @@ class TestManifestEndpoints(ManifestTestCase, DSSUnitTestCase):
                 'file_crc32c': '4e75003e',
                 'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/.zattrs',
                 'file_uuid': 'c1c4a2bc-b5fb-4083-af64-f5dec70d7f9d',
+                'file_drs_uri': '/c1c4a2bc-b5fb-4083-af64-f5dec70d7f9d?version=2018-10-10T031037.983672Z',
                 'specimen_from_organism_organ': 'brain'
             },
             # Related files from zarray store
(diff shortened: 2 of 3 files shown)


achave11:
Assigning @jessebrennan as requested. https://github.com/DataBiosphere/azul/pull/2841#discussion_r583945093

Note:
Feel free to modify and improve the mechanism by which the related file row in the manifest is generated.

src/azul/service/manifest_service.py
```diff
@@ -885,23 +884,32 @@ def write_to(self, output: IO[str]) -> Optional[str]:
...
+                        try:
+                            related_files = entity['related_files']
+                        except KeyError:
+                            # Disregard file entities without related_files
+                            continue
+                        else:
+                            for file in related_files:
+                                related_row = {}
+                                for k, v in file.items():
+                                    entity[k] = v if v else entity[k]
```
amarjandu, line 903: just a question: Do all the related files share the same `content-type`, `file_format`, `file_type` as the entity? 
It looks like `related_files` are only applicable to files in a zarry.

amarjandu: if a file is missing a value, and we use the value from the entity, can that create malformed data?

Say if for some reason the `uuid` was `''`, then we would use the `uuid` from the entity, it seems like this can cause malformed data in the manifest. 

You might be able to rework this to compose a dictionary that contains all the stuff from the `files` and cherry pick some stuff out of the entity if it is not present in the file keys.

amarjandu (changes requested):
Just a small thing, and a question.

hannes-ucsc (commented):
You don't appear to have much luck in choosing good commit and PR titles. Just stick with `Fix: ` + issue title for bugs or just issue title for enhancements.

src/azul/service/manifest_service.py
achave11: > Do all the related files share the same `content-type`, `file_format`, `file_type` as the entity?

I believe so, by looking at the [object](https://github.com/DataBiosphere/azul/blob/develop/src/azul/plugins/metadata/hca/transform.py#L757-L765) returned by the `_related_file` function, which drives the format of the `related_files` content …
```
{
    'name': file.manifest_entry.name,
    'crc32c': file.manifest_entry.crc32c,
    'sha256': file.manifest_entry.sha256,
    'size': file.manifest_entry.size,
    'uuid': file.manifest_entry.uuid,
    'drs_path': self.bundle.drs_path(file.manifest_entry.json),
    'version': file.manifest_entry.version,
}
```
… and compare it with the object returned by the `_file` [object](https://github.com/DataBiosphere/azul/blob/develop/src/azul/plugins/metadata/hca/transform.py#L718-L744) …
```
{
    'content-type': file.manifest_entry.content_type,
    'indexed': file.manifest_entry.indexed,
    'name': file.manifest_entry.name,
    'crc32c': file.manifest_entry.crc32c,
    'sha256': file.manifest_entry.sha256,
    'size': file.manifest_entry.size,
    'uuid': file.manifest_entry.uuid,
    'drs_path': self.bundle.drs_path(file.manifest_entry.json),
    'version': file.manifest_entry.version,
    'document_id': str(file.document_id),
    'file_type': file.schema_name,
    'file_format': file.file_format,
    'content_description': sorted(file.content_description),
    'source': Submitter.title_for_id(file.submitter_id),
    '_type': 'file',
    'related_files': list(map(self._related_file, related_files)),
    **(
        {
             'read_index': file.read_index,
             'lane_index': file.lane_index
        } if isinstance(file, api.SequenceFile) else {
        }
    ),
}
```
… we can determine that fields not specified in `_related_files` function are copied over to the related file row. It doesn't only use the fields you mention, it also copies all the other entities from the original file, projects, samples etc., as you provably saw in line 913.

>It looks like related_files are only applicable to files in a zarry.

Agreed, however, if I recall correctly, Hannes had previously mentioned that at the moment the `related_files` hierarchy is only exhibited in zarry (as you point out) but in the future other file formats might also have related files.

>if a file is missing a value, and we use the value from the entity, can that create malformed data?

🤔 seems like it, but related files are indexed with a very small subset of fields compared to a full file document in the index. I think that the fields indexed are just enough to locate the file, as such, if the related file had a malformed (missing `uuid` or `sha`) field, something very file metadata specific, Azul would fail to index it.

>…  rework this to compose a dictionary that contains all the stuff from the files and cherry pick some stuff out of the entity if it is not present in the file keys.

If for some reason the the fields that identify the related files were allowed to be absent and the indexer doesn't complain (which can cause the malformation you point out), it might be inevitable to recycle all the other entity fields from the original document. Original and related files together form a single source of data, which is why, I think, they share all the other entities. The other option, which I had accidentally implemented first, was to only include in the row, fields which were populated in the object returned by the `_related_file` function. But that felt like the related files were alienated, since they had less than seven fields populated per row. I am not necessarily sure how to proceed, since this malformations might still occur either way.

achave11:
> You don't appear to have much luck in choosing good commit and PR titles. 

Haha, that same luck is the reason why I don't gamble. 😆 

>Just stick with `Fix: ` + issue title for bugs or just issue title for enhancements.

Will do, thanks for the suggestion.

src/azul/service/manifest_service.py
amarjandu: Some of the fields for the related files are not shared with the entity, one example would be the content-type.
In the example below the entity content type is `'application/json; dcp-type=data'` while one of the related files has `'application/octet-stream; dcp-type=data'`. 

The manifest rows will have mixed fields from the entity and the related files, because https://github.com/DataBiosphere/azul/blob/develop/src/azul/plugins/metadata/hca/transform.py#L757-L765 does not specify all the fields that the file has. 

It might be worthwhile to discuss this with @hannes-ucsc, but I would think that blending the fields like this is not desirable.

I wonder why we even have `_related_files`, and why we cant just use another call to `_files` for this mapping.
https://github.com/DataBiosphere/azul/blob/96105bce814cf18063916020fb48d1f12c733938/src/azul/plugins/metadata/hca/transform.py#L736
 

Screen Shot 2021-03-11 at 3 15 07 PM

Sorry about the screenshot size, you might have to use the zoom feature, and get a magnifying glass out too.

amarjandu (commented):
Discussion continues....  https://github.com/DataBiosphere/azul/pull/2847#discussion_r592810589

src/azul/service/manifest_service.py
achave11: Include entries that begin with `file.manifest_entry` (included in `_file`) which are currently missing in related files.

achave11:
@amarjandu As you indicated, I pointed the "[Some of the fields for the related files are not shared with the entity](https://github.com/DataBiosphere/azul/pull/2847#discussion_r592810589)" issue to Hannes. His recommendation was to include more of the file fields under `file.manifest_entry`, present in the `_file` but not in `_related_file`. This helps the related files to not be overwritten by the original file fields, which might be different (as you indicated). Thank you for pointing that out 😄 .

amarjandu (approved):
I think this has addressed all my concerns, there are two kinda nit things that I think should be inspected. Consider the nit about the amount of data that is being copied back to the entity, as that one might have a performance hit.

Make sure to squash before review :)

src/azul/service/manifest_service.py
```diff
@@ -890,23 +889,35 @@ def write_to(self, output: IO[str]) -> Optional[str]:
...
+                            continue
+                        else:
+                            for file in related_files:
+                                related_row = {}
+                                invalid_manifest_entry = file.keys() - entity.keys()
+                                for field, value in file.items():
+                                    if field in invalid_manifest_entry:
+                                        continue
+                                    else:
+                                        entity[field] = value
```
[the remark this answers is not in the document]
achave11: > tbh im not even sure this would make a performance impact, it seems like the less manipulation of objects the better.

Agreed. I was able to replace the whole section by calling on …
```
entity.update(file)
```
… so entity entries that are present in related `file` will be overwritten in one call. However, I removed the `indexed` field of the `related_file` object. For one, it isn't represented in the compact manifest, so it would not be a disadvantage or a potential problem. Also, I do not think it is significant for anything other than indicating wether the content of the document was indexed, which we know for related files, they do not get indexed, so we don't loose or gain information.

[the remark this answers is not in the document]
amarjandu: entity.update() certainly makes the code more compact and legible. The downside to the approach is that we don't get the flexibility of being able to handle fields in the related file that we don't need in the file entity, also the data overwriting is still mostly occurring. 

> For one, it isn't represented in the compact manifest, so it would not be a disadvantage or a potential problem.
 
The PR is focused on compact manifests, but some of the changes are going to land in the indexer transformations. IIRC, by removing the `_indexed` field from the `_related_files` we also loose that field within the related_files object during a file hits response from ES.

It probably something we can live with, but if the field was added in the future would this cause problems with the compact manifest generation? Would there be an extra `_indexed` column in the manifest row? 

I'm not sure how high these concerns are, it might be something you want @hannes-ucsc to chime in on, lots of trade offs happening.

amarjandu (approved):
Comments here https://github.com/DataBiosphere/azul/pull/2847#discussion_r595334787
Because this PR touches the transformer, I think we need a reindex label on it. `[r]`

src/azul/service/manifest_service.py
```diff
@@ -896,23 +895,29 @@ def write_to(self, output: IO[str]) -> Optional[str]:
...
                 for doc_path, column_mapping in self.manifest_config.items():
                     entities = self._get_entities(doc_path, doc)
                     self._extract_fields(entities, column_mapping, row)
+                    if doc_path == 'contents.files':
+                        entity = one(entities)
+                        try:
+                            related_files = entity['related_files']
+                        except KeyError:
+                            # Disregard file entities without related_files
+                            continue
```
hannes-ucsc, line 908: https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#118control-flow

hannes-ucsc (changes requested):
Wait with next review until base PR in chain has been merged.

test/service/test_manifest.py
```diff
@@ -430,6 +443,7 @@ def test_manifest_zarr(self):
                     rows = [dict(file_crc32c=row['file_crc32c'],
                                  file_name=row['file_name'],
                                  file_uuid=row['file_uuid'],
+                                 file_drs_uri_path=furl(row['file_drs_uri']).path,
```
hannes-ucsc, line 446: We can assert more of the URL. Utilize

```python
>>> furl('drs://unpredictable/bar').set(host='known')
furl('drs://known/bar')
```

hannes-ucsc (changes requested):

test/service/test_manifest.py
achave11: > We can assert more of the URL.

Done.

I don't understand the purpose of modifying the host name of the `file_drs_uri` field in the manifest.

test/service/test_manifest.py
```diff
@@ -332,83 +332,97 @@ def test_manifest_zarr(self):
...
             {
                 'file_crc32c': 'bdc30523',
                 'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/qc_values/0.0',
                 'file_uuid': '351970aa-bc4c-405e-a274-be9e08e42e98',
+                'file_drs_uri': '/351970aa-bc4c-405e-a274-be9e08e42e98?version=2018-10-10T031041.135992Z',
                 'specimen_from_organism_organ': 'brain'
             }
         ]
-
+        for i in range(len(expected)):
```
hannes-ucsc, line 424: Since the items in the list are dicts there is no need to access them by index in order to be able to modify them.

test/service/test_manifest.py
```diff
@@ -332,83 +332,97 @@ def test_manifest_zarr(self):
...
+            expected[i]['file_drs_uri'] = furl(f'drs://{self.drs_domain}{expected[i]["file_drs_uri"]}').url
```
hannes-ucsc, line 425: There is no point in using `furl(h+p).url`. It is only subtly different to `h + p`. Use the latter or exploit the composition facilities that furl offers:

```
furl('/foo').set(scheme='drs', host='bar')
furl('drs://bar/foo')
```

hannes-ucsc (changes requested):

test/service/test_manifest.py
```diff
@@ -340,83 +340,98 @@ def test_manifest_zarr(self):
...
                 'file_name': '377f2f5a-4a45-4c62-8fb0-db9ef33f5cf0.zarr/expression_matrix/qc_values/0.0',
                 'file_uuid': '351970aa-bc4c-405e-a274-be9e08e42e98',
+                'file_drs_uri': '/351970aa-bc4c-405e-a274-be9e08e42e98?version=2018-10-10T031041.135992Z',
                 'specimen_from_organism_organ': 'brain'
             }
         ]
-
+        for entry in expected:
+            entry['file_drs_uri'] = furl(entry["file_drs_uri"]).set(scheme='drs',
+                                                                    host=config.drs_domain).url
```
hannes-ucsc, line 434:
```suggestion
for entry in expected:
    url = furl(entry['file_drs_uri'])
    entry['file_drs_uri'] = url.set(scheme='drs',
                                    host=config.drs_domain).url
```

hannes-ucsc: Note the stray double quotes in your version.

hannes-ucsc (changes requested):

hannes-ucsc (approved):

hannes-ucsc:
The ticket doesn't have a good reproduction so we need to first reproduce it in the demo,

For demo, show the problem in a deployment that doesn't have the fix yet (prod or personal), then show the fix in a deployment that does (dev or prod).

Issue #1848, opened 2020-05-29
Explicitly couple names of handler and lambda function


hannes-ucsc:
PR #1773 adds FIXME in  `lambdas/indexer/.chalice/config.json.template.py`

hannes-ucsc:
At least one brittle coupling was missed:

https://github.com/DataBiosphere/azul/blob/eb61eee2652b57a117e768d112604a1948fc3663/src/azul/queues.py#L440

PR #2731, opened 2021-01-26
Automate purchase and deletion of BQ slots (#2651)

noah-aviel-dove:
Author

- [ ] PR title references issue
- [ ] Title of main commit references issue
- [ ] PR is linked to Zenhub issue

Author (reindex)

- [ ] Added `r` tag to commit title                         <sub>or this PR does not require reindexing</sub>
- [ ] Added `reindex` label to PR                           <sub>or this PR does not require reindexing</sub>

Author (freebies & chains)

- [ ] Freebies are blocked on this PR                       <sub>or there are no freebies in this PR</sub>
- [ ] Freebies are referenced in commit titles              <sub>or there are no freebies in this PR</sub>
- [ ] This PR is blocked by previous PR in chain            <sub>or this PR is not chained to another PR</sub>
- [ ] Added `chain` label to the blocking PR                <sub>or this PR is not chained to another PR</sub>

Author (upgrading)

- [ ] Documented upgrading of deployments in UPGRADING.rst  <sub>or this PR does not require upgrading</sub>
- [ ] Added `u` tag to commit title                         <sub>or this PR does not require upgrading</sub>
- [ ] Added `upgrade` label to PR                           <sub>or this PR does not require upgrading</sub>

Author (requirements, before every review)

- [ ] Ran `make requirements_update`                        <sub>or this PR leaves requirements*.txt, common.mk and Makefile untouched</sub>
- [ ] Added `R` tag to commit title                         <sub>or this PR leaves requirements*.txt untouched</sub>
- [ ] Added `reqs` label to PR                              <sub>or this PR leaves requirements*.txt untouched</sub>

Author (before every review)

- [ ] `make integration_test` passes in personal deployment <sub>or this PR does not touch functionality that could break the IT</sub>
- [ ] Rebased branch on `develop`, squashed old fixups

Primary reviewer (before pushing merge commit)

- [ ] Checked `reindex` label and `r` commit title tag
- [ ] Rebased and squashed branch
- [ ] Sanity-checked history
- [ ] Build passes in `sandbox`                             <sub>or added `no sandbox` label</sub>
- [ ] Reindexed `sandbox`                                   <sub>or this PR does not require reindexing `sandbox`</sub>
- [ ] Added PR reference to merge commit
- [ ] Moved linked issue to Merged
- [ ] Pushed merge commit to Github

Primary reviewer (after pushing merge commit)

- [ ] Moved freebies to Merged column                       <sub>or there are no freebies in this PR</sub> 
[...]

PR #2929, opened 2021-04-01
Fix brittle coupling with lambda names in manage_queues.py (#1848) 

jessebrennan:
~To reviewers: Notice that there are two commits. The first is a minimal approach, and the second is something a it more elaborate. I'm unsure which approach is preferred.~

Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history

Operator (reindex) 

Operator

- [ ] Unassigned PR

### DIFF
--- a/scripts/sell_unused_slots.py
+++ b/scripts/sell_unused_slots.py
@@ -88,8 +88,8 @@ class ReindexDetector:
     def _contribution_lambda_names(self) -> FrozenSet[str]:
         indexer = load_app_module('indexer')
         return frozenset((
-            indexer.contribute.lambda_name,
-            indexer.contribute_retry.lambda_name
+            indexer.contribute.name,
+            indexer.contribute_retry.name
         ))
 
     def _is_contribution_lambda(self, function_name: str) -> bool:
--- a/scripts/sell_unused_slots.py
+++ b/scripts/sell_unused_slots.py
@@ -95,8 +95,10 @@ class ReindexDetector:
     def _is_contribution_lambda(self, function_name: str) -> bool:
         for lambda_name in self._contribution_lambda_names:
             try:
+                # FIXME: Eliminate hardcoded separator
+                #        https://github.com/databiosphere/azul/issues/2964
                 resource_name, _ = config.unqualified_resource_name(function_name,
-                                                                    suffix=lambda_name)
+                                                                    suffix='-' + lambda_name)
             except RequirementError:
                 pass
             else:
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -346,7 +346,7 @@ class Config:
 
     def unqualified_resource_name(self,
                                   qualified_resource_name: str,
-                                  suffix: str = None
+                                  suffix: str = ''
                                   ) -> Tuple[str, str]:
         """
         >>> config.unqualified_resource_name('azul-foo-dev')
@@ -364,17 +364,13 @@ class Config:
         Traceback (most recent call last):
         ...
         azul.RequirementError: ['azul', 'object', 'versions', 'dev']
-
-        >>> config.unqualified_resource_name('azul-indexer-dev-contribute', suffix='contribute')
-        ('indexer', 'dev')
         """
-        sep = '-'
-        if suffix is not None:
-            require(len(suffix) > 0, suffix)
-            suffix = sep + suffix
-            require(qualified_resource_name.endswith(suffix))
+        require(qualified_resource_name.endswith(suffix))
+        if suffix:
             qualified_resource_name = qualified_resource_name[:-len(suffix)]
-        components = qualified_resource_name.split(sep)
+        # FIXME: Eliminate hardcoded separator
+        #        https://github.com/databiosphere/azul/issues/2964
+        components = qualified_resource_name.split('-')
         require(len(components) == 3, components)
         prefix, resource_name, deployment_stage = components
         require(prefix == self.resource_prefix)
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -412,11 +412,25 @@ class Config:
 
     @property
     def indexer_name(self) -> str:
-        return self.qualified_resource_name('indexer')
+        return self.indexer_function_name()
 
     @property
     def service_name(self) -> str:
-        return self.qualified_resource_name('service')
+        return self.service_function_name()
+
+    def indexer_function_name(self, handler_name: Optional[str] = None):
+        return self._function_name('indexer', handler_name)
+
+    def service_function_name(self, handler_name: Optional[str] = None):
+        return self._function_name('service', handler_name)
+
+    def _function_name(self, lambda_name: str, handler_name: Optional[str]):
+        if handler_name is None:
+            return self.qualified_resource_name(lambda_name)
+        else:
+            # FIXME: Eliminate hardcoded separator
+            #        https://github.com/databiosphere/azul/issues/2964
+            return self.qualified_resource_name(lambda_name, suffix='-' + handler_name)
 
     deployment_name_re = re.compile(r'[a-z][a-z0-9]{1,16}')
 
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -380,22 +380,6 @@ class Config:
         require(prefix == self.resource_prefix)
         return resource_name, deployment_stage
 
-    def unqualified_resource_name_or_none(self,
-                                          qualified_resource_name: str,
-                                          suffix: Optional[str] = None
-                                          ) -> Tuple[Optional[str], Optional[str]]:
-        """
-        >>> config.unqualified_resource_name_or_none('azul-foo-dev')
-        ('foo', 'dev')
-
-        >>> config.unqualified_resource_name_or_none('invalid-foo-dev')
-        (None, None)
-        """
-        try:
-            return self.unqualified_resource_name(qualified_resource_name, suffix=suffix)
-        except RequirementError:
-            return None, None
-
     def subdomain(self, lambda_name):
         return os.environ['AZUL_SUBDOMAIN_TEMPLATE'].replace('*', lambda_name)
 
--- a/src/azul/chalice.py
+++ b/src/azul/chalice.py
@@ -183,10 +183,10 @@ class AzulChaliceApp(Chalice):
         # Our handlers reference the name of the corresponding Lambda function
         # which allows the handler to be the single source of truth when
         # configuring Terraform, etc.
-        if hasattr(wrapped_handler, 'lambda_name'):
-            assert wrapped_handler.lambda_name == name
+        if hasattr(wrapped_handler, 'name'):
+            assert wrapped_handler.name == name
         else:
-            wrapped_handler.lambda_name = name
+            wrapped_handler.name = name
 
     # Some type annotations to help with auto-complete
     lambda_context: LambdaContext
--- a/src/azul/chalice.py
+++ b/src/azul/chalice.py
@@ -76,9 +76,6 @@ class AzulChaliceApp(Chalice):
 
             def decorator(view_func):
                 self._register_spec(path, path_spec, method_spec, methods)
-                # Stash the URL path a view function is bound to as an attribute of
-                # the function itself.
-                view_func.path = path
                 return chalice_decorator(view_func)
 
             return decorator
@@ -171,6 +168,8 @@ class AzulChaliceApp(Chalice):
         else:
             log.info('Returning %i response. To log headers and body, set AZUL_DEBUG to 1.', response.status_code)
 
+    absent = object()
+
     def _register_handler(self,
                           handler_type,
                           name,
@@ -182,11 +181,25 @@ class AzulChaliceApp(Chalice):
                                   wrapped_handler, kwargs, options)
         # Our handlers reference the name of the corresponding Lambda function
         # which allows the handler to be the single source of truth when
-        # configuring Terraform, etc.
-        if hasattr(wrapped_handler, 'name'):
-            assert wrapped_handler.name == name
-        else:
-            wrapped_handler.name = name
+        # configuring Terraform, etc. We store other parameters used to
+        # configure the handler for the same reason.
+        for attribute, new_value, is_additive in [
+            ('name', name, False),
+            ('queue', kwargs.get('queue', self.absent), False),
+            ('path', kwargs.get('path', self.absent), True)
+        ]:
+            if new_value is not self.absent:
+                try:
+                    old_value = getattr(wrapped_handler, attribute)
+                except AttributeError:
+                    if is_additive:
+                        new_value = [new_value]
+                    setattr(wrapped_handler, attribute, new_value)
+                else:
+                    if is_additive:
+                        old_value.append(new_value)
+                    else:
+                        assert old_value == new_value
 
     # Some type annotations to help with auto-complete
     lambda_context: LambdaContext
--- a/src/azul/queues.py
+++ b/src/azul/queues.py
@@ -43,6 +43,9 @@ from azul.files import (
 from azul.lambdas import (
     Lambdas,
 )
+from azul.modules import (
+    load_app_module,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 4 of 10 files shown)


src/azul/queues.py
```diff
@@ -430,19 +433,24 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
...
-                elif queue_name == config.tallies_queue_name(retry=True):
-                    # FIXME: Brittle coupling between the string literal below and
-                    #        the handler function name in app.py
-                    #        https://github.com/DataBiosphere/azul/issues/1848
-                    submit(self._manage_sqs_push, config.indexer_name + '-aggregate_retry', queue, enable)
+                if queue_name in queues_to_lambda:
+                    lambda_name = f'{config.indexer_name}-{queues_to_lambda[queue_name].lambda_name}'
+                    submit(self._manage_lambda, lambda_name, enable)
+                else:
+                    assert 'fail' in queue_name
```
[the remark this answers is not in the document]
noah-aviel-dove: Ideally we would have a method in `Config` to construct this name without having to specify the separator (`'-'`) here, but no existing method seems to accomplish this. I suspect `Config.qualified_resource_name` will need to be modified analogous to how `unqualified_resource_name` had to be changed to accommodate these same lambda names in https://github.com/DataBiosphere/azul/issues/2731, but I'm not sure how it should be done.

noah-aviel-dove (changes requested):

src/azul/queues.py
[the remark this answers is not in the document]
jessebrennan: I spent some time thinking about the best way to do this, but the fifo queues suffix prevents a neat solution. Since this is the only usage, I'm tempted to leave this as it is. We can see if Hannes thinks it's worth adding.

noah-aviel-dove (approved):
Looks great!

src/azul/chalice.py
```diff
@@ -187,6 +187,13 @@ def _register_handler(self,
             assert wrapped_handler.lambda_name == name
         else:
             wrapped_handler.lambda_name = name
+        # We store any other parameters used to configure the handler for the
+        # same reason.
+        for k, v in kwargs.items():
+            if hasattr(wrapped_handler, k):
```
hannes-ucsc, line 193: Some duplication here that could be eliminated by appending a tuple to what .items() returns or by extracting a nested function.

Also, I would like to understand when we expect this condition to ever be True.

src/azul/queues.py
```diff
@@ -430,19 +433,26 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
             def submit(f, *args, **kwargs):
                 futures.append(tpe.submit(f, *args, **kwargs))
 
+            indexer = load_app_module('indexer')
+            lambdas = [
```
hannes-ucsc, line 437: Any way we can deduce this by checking which lambdas have a `.queue` attribute rather than hard-coding this list?

src/azul/queues.py
```diff
@@ -430,19 +433,26 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
...
+                indexer.contribute,
+                indexer.contribute_retry,
+                indexer.aggregate,
+                indexer.aggregate_retry
+            ]
+            queues_to_lambda = {lambda_.queue: lambda_ for lambda_ in lambdas}
             for queue_name, queue in queues.items():
                 if queue_name == config.notifications_queue_name():
+                    # Since the indexer lambda writes to the notifications
```
hannes-ucsc, line 446: Couldn't we implement this exception more elegantly by patching `queues_to_lambda`?

src/azul/queues.py
```diff
@@ -430,19 +433,26 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
...
-                    # FIXME: Brittle coupling between the string literal below and
-                    #        the handler function name in app.py
-                    #        https://github.com/DataBiosphere/azul/issues/1848
-                    submit(self._manage_sqs_push, config.indexer_name + '-aggregate_retry', queue, enable)
+                try:
+                    lambda_ = queues_to_lambda[queue_name]
+                except KeyError:
+                    assert 'fail' in queue_name
+                else:
+                    lambda_name = f'{config.indexer_name}-{lambda_.lambda_name}'
```
hannes-ucsc, line 454: Surely there is sth in config for this purpose?

hannes-ucsc (changes requested):

src/azul/queues.py
jessebrennan: This is somewhat complicated, would like to discuss in parking lot.

jessebrennan: My idea about bypassing suffix was wrong and doesn't work. There's still not an elegant way to do this. I'd like to code through this with you @hannes-ucsc so that we can come up with a satisfying answer.

src/azul/chalice.py
jessebrennan: This condition is true when a handler has multiple decorators, for example https://github.com/DataBiosphere/azul/blob/65970a0947f38fae439a3bf8fd960d351787b7a3/lambdas/service/app.py#L1184

jessebrennan: One weird side-effect of this is that the `path` attribute of handlers is being overwritten. The path attribute was added a long time ago and is used only by `file_url()` as far as I can tell. I had to add some logic to work around this.

scripts/sell_unused_slots.py
```diff
@@ -87,8 +87,9 @@ def _contribution_lambda_names(self) -> FrozenSet[str]:
     def _is_contribution_lambda(self, function_name: str) -> bool:
         for lambda_name in self._contribution_lambda_names:
             try:
+                # FIXME: https://github.com/databiosphere/azul/issues/2964
```
hannes-ucsc, line 90: Title missing. I shouldn't have to click the link to understand the FIXME.

src/azul/__init__.py
```diff
@@ -364,25 +364,19 @@ def unqualified_resource_name(self,
...
-        >>> config.unqualified_resource_name('azul-indexer-dev-contribute', suffix='contribute')
-        ('indexer', 'dev')
         """
-        sep = '-'
-        if suffix is not None:
-            require(len(suffix) > 0, suffix)
-            suffix = sep + suffix
-            require(qualified_resource_name.endswith(suffix))
+        require(qualified_resource_name.endswith(suffix))
+        if len(suffix) > 0:
```
hannes-ucsc, line 369:
```suggestion
        if suffix:
```

src/azul/__init__.py
```diff
@@ -430,13 +424,18 @@ def drs_endpoint(self):
     def lambda_names(self) -> List[str]:
         return ['indexer', 'service']
 
-    @property
-    def indexer_name(self) -> str:
-        return self.qualified_resource_name('indexer')
+    def indexer_name(self, lambda_name: Optional[str] = None) -> str:
```
hannes-ucsc, line 427:
```suggestion
    def indexer_name(self, handler_name: Optional[str] = None) -> str:
```

`handler_name` is the name of the handler function
`lambda_name` is `'indexer'` or `'service'`
`function_name` is the fully qualified name of the AWS Lambda function

and let's leave this as a property but add `indexer_function_name` and `service_function_name` methods that call …

src/azul/__init__.py
```diff
@@ -430,13 +424,18 @@ def drs_endpoint(self):
...
+        return self._app_name('indexer', lambda_name)
 
-    @property
-    def service_name(self) -> str:
-        return self.qualified_resource_name('service')
+    def service_name(self, lambda_name: Optional[str] = None) -> str:
+        return self._app_name('service', lambda_name)
+
+    def _app_name(self, app: str, lambda_name: Optional[str]):
```
hannes-ucsc, line 433: this method

```suggestion
    def _function_name(self, lambda_name: str, handler_name: str):
```

If you have a better idea, let me know.

src/azul/__init__.py
```diff
@@ -430,13 +424,18 @@ def drs_endpoint(self):
...
+        if lambda_name is None:
+            return self.qualified_resource_name(app)
+        else:
+            # FIXME: https://github.com/databiosphere/azul/issues/2964
```
hannes-ucsc, line 437: see above

src/azul/chalice.py
hannes-ucsc: Not seeing any of the stuff we talked about.

hannes-ucsc (changes requested):

src/azul/__init__.py
jessebrennan: I'm making this change, but for the record, I don't understand why this is better. Also I just wanted to point out that I reverted Noah's diff exactly, so checking length was how this was done originally.

src/azul/__init__.py
jessebrennan: I like this.

src/azul/__init__.py
hannes-ucsc: It's better because it is shorter.

src/azul/chalice.py
jessebrennan: I have since addressed this.

src/azul/__init__.py
jessebrennan: I like this renaming scheme, and calling the parameter `handler_name` makes sense. I'd gone with `lambda_name`, in part because that's the name of the property of the function, for example, `contribute.lambda_name`. Should I change the name of the property as well?

My plan for now is to not change the property name, and push to review as is.

jessebrennan:
@hannes-ucsc, notice the two consecutive `fixup!`s at the end of the PR history. I tried implementing your suggestion in two different ways. Both are functionally equivalent.

Also see this comment https://github.com/DataBiosphere/azul/pull/2929#discussion_r612724795

src/azul/chalice.py
```diff
@@ -182,11 +179,22 @@ def _register_handler(self,
...
+                # Attributes may already exist when a handler function has
+                # multiple Chalice decorators (usually route decorators).
+                assert getattr(wrapped_handler, attribute) == value
+            else:
+                setattr(wrapped_handler, attribute, value)
+        if 'path' in kwargs:
+            if hasattr(wrapped_handler, 'path'):
+                wrapped_handler.paths.append(kwargs['path'])
+            else:
+                wrapped_handler.paths = [kwargs['path']]
```
hannes-ucsc, line 197: This is an odd mix of general and specific code and it doesn't use EAFP. Can you try this:

```python
Index: src/azul/chalice.py
IDEA additional info:
<+>UTF-8
diff --git a/src/azul/chalice.py b/src/azul/chalice.py
--- a/src/azul/chalice.py	(revision d327e6a6aeeb0dbe6930695cd24c9d0c8410dd54)
+++ b/src/azul/chalice.py	(date 1618509062574)
@@ -168,6 +168,8 @@
         else:
             log.info('Returning %i response. To log headers and body, set AZUL_DEBUG to 1.', response.status_code)
 
+    absent = object()
+
     def _register_handler(self,
                           handler_type,
                           name,
@@ -181,20 +183,23 @@
         # which allows the handler to be the single source of truth when
         # configuring Terraform, etc. We store other parameters used to
         # configure the handler for the same reason.
-        attributes = ('queue',)
-        for attribute, value in (('lambda_name', name),
-                                 *((key, kwargs[key]) for key in attributes if key in kwargs)):
-            if hasattr(wrapped_handler, attribute):
-                # Attributes may already exist when a handler function has
-                # multiple Chalice decorators (usually route decorators).
-                assert getattr(wrapped_handler, attribute) == value
-            else:
-                setattr(wrapped_handler, attribute, value)
-        if 'path' in kwargs:
-            if hasattr(wrapped_handler, 'path'):
-                wrapped_handler.paths.append(kwargs['path'])
-            else:
-                wrapped_handler.paths = [kwargs['path']]
+        for attribute, new_value, is_additive in [
+            ('queue', kwargs.get('queue', self.absent), False),
+            ('lambda_name', name, False),
+            ('path', kwargs.get('path', self.absent), True)
+        ]:
+            if new_value is not self.absent:
+                try:
+                    old_value = getattr(wrapped_handler, attribute)
+                except AttributeError:
+                    if is_additive:
+                        new_value = [new_value]
+                    setattr(wrapped_handler, attribute, new_value)
+                else:
+                    if is_additive:
+                        old_value.append(new_value)
+                    else:
+                        assert old_value == new_value
 
     # Some type annotations to help with auto-complete
     lambda_context: LambdaContext
```

If you think we can conflate None and absence, you can drop the `self.absent` stuff.

hannes-ucsc (changes requested):

src/azul/chalice.py
jessebrennan: I like this way better.

> If you think we can conflate None and absence, you can drop the self.absent stuff.

Probably conflating is OK, but better safe than sorry. I like this pattern anyway.

src/azul/queues.py
```diff
@@ -377,19 +380,24 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
...
-                    submit(self._manage_sqs_push, config.indexer_name + '-aggregate', queue, enable)
-                elif queue_name == config.tallies_queue_name(retry=True):
-                    # FIXME: Brittle coupling between the string literal below and
-                    #        the handler function name in app.py
-                    #        https://github.com/DataBiosphere/azul/issues/1848
-                    submit(self._manage_sqs_push, config.indexer_name + '-aggregate_retry', queue, enable)
+                try:
+                    lambda_names = queues_to_lambda_names[queue_name]
+                except KeyError:
+                    assert 'fail' in queue_name
```
hannes-ucsc, line 397: The `"-fail`" suffix is encapsulated in `azul.config` and allowing the substring to occur anywhere in the string, not anchored at boundaries, might lead to false positives depending on the substring. Can we make this both stricter and less brittle?

src/azul/queues.py
```diff
@@ -377,19 +380,24 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
             def submit(f, *args, **kwargs):
                 futures.append(tpe.submit(f, *args, **kwargs))
 
+            indexer = load_app_module('indexer')
+            queues_to_lambda_names = {
+                lambda_.queue: [config.indexer_function_name(lambda_.lambda_name)]
```
hannes-ucsc, line 385:
```suggestion
                lambda_.queue: [config.indexer_function_name(lambda_.handler_name)]
```

In a separate commit at the beginning of the history.

Or even better

```suggestion
                lambda_.queue: [config.indexer_function_name(lambda_.name)]
```

hannes-ucsc (changes requested):

src/azul/queues.py
jessebrennan: For the record, I had originally called this attribute `.name` in the first PR, but remember you asking me to change it to `lambda_name` (unfortunately I couldn't find a written record of this).

I'm happy to make this change.

src/azul/queues.py
```diff
@@ -377,19 +380,24 @@ def manage_lambdas(self, queues: Mapping[str, Queue], enable: bool):
             def submit(f, *args, **kwargs):
                 futures.append(tpe.submit(f, *args, **kwargs))
 
+            indexer = load_app_module('indexer')
+            queues_to_lambda_names = {
```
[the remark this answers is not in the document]
jessebrennan: Assuming you meant `functions_by_queue`

hannes-ucsc (changes requested):
I'm trying something new. I've added a [commit](https://github.com/DataBiosphere/azul/pull/2929/commits/de1ab7945188ac0f03fcb7400fb2b0f1fc77f89d) titled `REVIEW` to the branch. You should soft-reset the branch to the parent of that commit and then re-commit as fixups.

hannes-ucsc (changes requested):
The commits that add a fixme should refer to the fixme issue in their title. Some do some don't in your history.

A commit that reverts a change should refer to the issue whose PR contains the commit being reverted.

We tend to put the commit id of the reverted commit in the body of the reverting commit, even if it's just a reversion "in spirit", not a literal one.

hannes-ucsc (approved):

Issue #2750, opened 2021-02-02
Make length of reindex partition prefix specific to source


hannes-ucsc:
Right now all sources use the same hard-coded length of 2. Eventually there will be a source (snapshot) per project and projects can vary in size considerably so we should configure the prefix length for each source. Something like:

```diff
<+>UTF-8
===================================================================
diff --git a/deployments/dev/environment.py b/deployments/dev/environment.py
--- a/deployments/dev/environment.py	(revision 08265a7f42655e5179040d9715982a97e2a4f658)
+++ b/deployments/dev/environment.py	(date 1611869186220)
@@ -40,8 +40,8 @@
             'hca:it2ebi:repository/tdr:metadata/hca'
         ]),
 
-        'AZUL_TDR_SOURCE': 'tdr:broad-jade-dev-data:snapshot/hca_dev_20201217_test4___20210126',
-        'AZUL_TDR_DCP2EBI_SOURCE': 'tdr:broad-jade-dev-data:snapshot/hca_dev_20201023_ebiv4___20201121',
+        'AZUL_TDR_SOURCE': 'tdr:broad-jade-dev-data:2:snapshot/hca_dev_20201217_test4___20210126',
+        'AZUL_TDR_DCP2EBI_SOURCE': 'tdr:broad-jade-dev-data:1:snapshot/hca_dev_20201023_ebiv4___20201121',
         'AZUL_TDR_IT2EBI_SOURCE': 'tdr:broad-jade-dev-data:snapshot/hca_dev_20201023_ebiv4___20201121',
         'AZUL_TDR_SERVICE_URL': 'https://jade.datarepo-dev.broadinstitute.org',
         'AZUL_SAM_SERVICE_URL': 'https://sam.dsde-dev.broadinstitute.org',
```

melainalegaspi:
@amarjandu to spike and quickly evaluate how difficult this will be. Goal of the spike is to come up with an estimate.

hannes-ucsc:
Other team members will observe this change when they follow the upgrade instructions. If it breaks something, we will undoubtedly find out. No demo.

PR #2827, opened 2021-02-23
Allow prefix to be specified per source (#2650)

amarjandu:
Author

- [x] PR title references issue
- [x] Title of main commit references issue
- [x] PR is linked to Zenhub issue

Author (reindex)

- [x] Added `r` tag to commit title                         <sub>or this PR does not require reindexing</sub>
- [x] Added `reindex` label to PR                           <sub>or this PR does not require reindexing</sub>

Author (freebies & chains)

- [x] Freebies are blocked on this PR                       <sub>or there are no freebies in this PR</sub>
- [x] Freebies are referenced in commit titles              <sub>or there are no freebies in this PR</sub>
- [x] This PR is blocked by previous PR in the chain        <sub>or this PR is not chained to another PR</sub>
- [x] Added `chain` label to the blocking PR                <sub>or this PR is not chained to another PR</sub>

Author (upgrading)

- [x] Documented upgrading of deployments in UPGRADING.rst  <sub>or this PR does not require upgrading</sub>
- [x] Added `u` tag to commit title                         <sub>or this PR does not require upgrading</sub>
- [x] Added `upgrade` label to PR                           <sub>or this PR does not require upgrading</sub>

Author (requirements, before every review)

- [x] Ran `make requirements_update`                        <sub>or this PR leaves requirements*.txt, common.mk and Makefile untouched</sub>
- [x] Added `R` tag to commit title                         <sub>or this PR leaves requirements*.txt untouched</sub>
- [x] Added `reqs` label to PR                              <sub>or this PR leaves requirements*.txt untouched</sub>

Author (before every review)

- [x] `make integration_test` passes in personal deployment <sub>or this PR does not touch functionality that could break the IT</sub>
- [x] Rebased branch on `develop`, squashed old fixups

Primary reviewer (after approval)

- [x] Commented on demo expectations                        <sub>or labelled PR as `no demo`</sub>
- [x] Decided if PR can be labeled `no sandbox`
- [x] PR title is appropriate as title of merge commit
- [x] Moved ticket to Approved column
- [x] Assigned PR to an operator

Operator (before pushing merge the commit)

- [x] Checked `reindex` label and `r` commit title tag
- [x] Checked that Demo expectations are clear              <sub>or PR is labeled as `no demo`</sub>
- [x] Rebased and squashed branch
- [x] Sanity-checked history
- [x] Pushed PR branch to Github
- [x] Branch pushed to Gitlab and build passes in `sandbox` <sub>or added `no sandbox` label</sub>
[...]

PR #2931, opened 2021-04-01
Add prefix_length to source (#2750)

amarjandu:
https://github.com/databiosphere/azul/issues/2750

Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history

Operator (reindex) 

Operator

- [ ] Unassigned PR

### DIFF
--- a/UPGRADING.rst
+++ b/UPGRADING.rst
@@ -11,6 +11,19 @@ reverted. This is all fairly informal and loosely defined. Hopefully we won't
 have too many entries in this file.
 
 
+#2750 Add partition_prefix_length to sources
+============================================
+
+The syntax of the ``AZUL_TDR_SOURCES`` and ``AZUL_TDR_…_SOURCES`` environment
+variables was modified to include a partition prefix length. To specify a
+partition prefix length within a source, append a slash delimiter ``/`` followed
+by a partition length (e.g., ``/2``) to the source entry in the
+deployment's ``environment.py`` . If the partition prefix length is not
+specified in one of the above variables, the default value from
+``AZUL_PARTITION_PREFIX_LENGTH`` will be used.
+As always, use the sandbox deployment's ``environment.py`` as a template.
+
+
 #2865 Allow catalog.internal to be configurable
 ===============================================
 
--- a/deployments/prod/environment.py
+++ b/deployments/prod/environment.py
@@ -57,6 +57,8 @@ def env() -> Mapping[str, Optional[str]]:
             }
         }),
 
+        'AZUL_PARTITION_PREFIX_LENGTH': '2',
+
         'AZUL_TDR_SOURCES': ','.join([
             'tdr:broad-datarepo-terra-prod-hca2:snapshot/hca_prod_20201118_dcp1___20201209:',
         ]),
--- a/deployments/sandbox/environment.py
+++ b/deployments/sandbox/environment.py
@@ -59,19 +59,21 @@ def env() -> Mapping[str, Optional[str]]:
             ]
         }),
 
+        'AZUL_PARTITION_PREFIX_LENGTH': '1',
+
         'AZUL_TDR_SOURCES': ','.join([
             'tdr:broad-jade-dev-data:snapshot/hca_dev_20201203___20210524_lattice:42',
             'tdr:broad-jade-dev-data:snapshot/hca_dev_20210621_managedaccess_4298b4de92f34cbbbbfe5bc11b8c2422__20210622:'
         ]),
         **{
             f'AZUL_TDR_{catalog.upper()}_SOURCES': ','.join([
-                f'tdr:broad-jade-dev-data:snapshot/hca_dev_20201023_ebiv4___20210302:4'
+                f'tdr:broad-jade-dev-data:snapshot/hca_dev_20201023_ebiv4___20210302:4/0'
             ])
             for catalog in ('dcp2ebi', 'it2ebi')
         },
         **{
             f'AZUL_TDR_{catalog.upper()}_SOURCES': ','.join([
-                'tdr:broad-jade-dev-data:snapshot/lungmap_dev_20210412__20210414:',
+                'tdr:broad-jade-dev-data:snapshot/lungmap_dev_20210412__20210414:/0',
             ])
             for catalog in ('lungmap', 'it3lungmap')
         },
--- a/environment.py
+++ b/environment.py
@@ -338,6 +338,16 @@ def env() -> Mapping[str, Optional[str]]:
         # indexed.
         'AZUL_DSS_QUERY_PREFIX': '',
 
+        # The length of the subgraph UUID prefix by which to partition the set
+        # of subgraphs matching the query during remote reindexing. Partition
+        # prefixes that are too long result in many small or even empty
+        # partitions and waste some amount of resources. Partition prefixes that
+        # are too short result in few large partitions that could exceed the
+        # memory and running time limitations of the AWS Lamba function that
+        # processes them. If in doubt err on the side of too many small
+        # partitions.
+        'AZUL_PARTITION_PREFIX_LENGTH': None,
+
         # A short string (no punctuation allowed) that identifies a Terraform
         # component i.e., a distinct set of Terraform resources to be deployed
         # together but separately from resources in other components. They are
@@ -372,11 +382,12 @@ def env() -> Mapping[str, Optional[str]]:
         #
         # sources = source (',', source )* ;
         #
-        # source = GitHub URL ;
+        # source = GitHub URL,
+        #          ':', [ UUID prefix [ '/', Partition prefix length ] ] ;
         #
         # Example:
         #
-        # https://github.com/HumanCellAtlas/schema-test-data/tree/de355ca/tests
+        # https://github.com/HumanCellAtlas/schema-test-data/tree/de355ca/tests:
         #
         # The GitHub URL must have the syntax
         #
@@ -395,11 +406,12 @@ def env() -> Mapping[str, Optional[str]]:
         #
         # source = 'tdr:', Google Cloud project name,
         #          ':', ( 'dataset' | 'snapshot' ),
-        #          '/', 'TDR dataset or snapshot name' ;
+        #          '/', TDR dataset or snapshot name,
+        #          ':', [ UUID prefix [ '/', Partition prefix length ] ] ;
         #
         # Example:
         #
-        # tdr:broad-jade-dev-data:dataset/hca_mvp,tdr:broad-jade-dev-data:snapshot/hca_mvp
+        # tdr:broad-jade-dev-data:dataset/hca_mvp:2/1,tdr:broad-jade-dev-data:snapshot/hca_mvp:2
         #
         # To configure a source specific to a particular catalog, include the
         # catalog name in all upper case in the variable name. For example, to
--- a/scripts/recan_bundle_tdr.py
+++ b/scripts/recan_bundle_tdr.py
@@ -35,6 +35,7 @@ from azul.files import (
 )
 from azul.indexer import (
     Bundle,
+    Prefix,
     SimpleSourceSpec,
     SourcedBundleFQID,
 )
@@ -387,7 +388,7 @@ def main(argv):
         metadata = json.load(f)
 
     dss_source = DSSSourceRef(id='',
-                              spec=SimpleSourceSpec(prefix='',
+                              spec=SimpleSourceSpec(prefix=Prefix.of_everything,
                                                     name=config.dss_endpoint))
     dss_bundle = DSSBundle(fqid=SourcedBundleFQID(source=dss_source,
                                                   uuid=args.bundle_uuid,
@@ -396,7 +397,8 @@ def main(argv):
                            metadata_files=metadata)
 
     tdr_source = TDRSourceRef(id=args.source_id,
-                              spec=TDRSourceSpec(project='test_project',
+                              spec=TDRSourceSpec(prefix=Prefix.of_everything,
+                                                 project='test_project',
                                                  name='test_name',
                                                  is_snapshot=True))
     tdr_bundle = dss_bundle_to_tdr(dss_bundle, tdr_source)
--- a/scripts/reindex.py
+++ b/scripts/reindex.py
@@ -50,16 +50,18 @@ parser.add_argument('--workers',
                     default=defaults.num_workers,
                     type=int,
                     help='The number of workers that will be sending bundles to the indexer concurrently')
-parser.add_argument('--partition-prefix-length',
-                    metavar='NUM',
-                    default=0,
-                    type=int,
-                    help='The length of the bundle UUID prefix by which to partition the set of bundles matching the '
-                         'query. Each query partition is processed independently and remotely by the indexer lambda. '
-                         'The lambda queries the repository and queues a notification for each matching bundle. If 0 '
-                         '(the default) no partitioning occurs, the repository is queried locally and the indexer '
-                         'notification endpoint is invoked for each bundle individually and concurrently using worker'
-                         'threads. This is magnitudes slower than remote i.e., partitioned indexing.')
+parser.add_argument('--local',
+                    default=False,
+                    action='store_true',
+                    help='Do not offload the listing of subgraphs to the indexer Lambda function. When this option is '
+                         'used, this script queries the repository without partitioning, and the indexer notification '
+                         'endpoint is invoked for each subgraph individually and concurrently using worker threads. '
+                         'This is magnitudes slower than remote (i.e. partitioned) indexing. If this option is not '
+                         'used (the default), the set of subgraphs matching the query is partitioned using the '
+                         'partition prefix length configured for each of the catalog sources being reindexed. Each '
+                         'query partition is processed independently and remotely by the indexer lambda. The index '
+                         'Lambda function queries the repository for each partition and queues a notification for each '
+                         'matching subgraph in the partition.')
 parser.add_argument('--catalogs',
                     nargs='+',
                     metavar='NAME',
@@ -72,9 +74,9 @@ parser.add_argument('--catalogs',
 parser.add_argument('--sources',
                     default=config.reindex_sources,
                     nargs='+',
-                    help='Limit remote reindexing to a subset of the configured sources. '
+                    help='Limit reindexing to a subset of the configured sources. '
                          'Supports shell-style wildcards to match multiple sources per argument. '
-                         'Must be * for local reindexing i.e., if --partition-prefix-length is not given.')
+                         'Must be * for local reindexing i.e., if --local is given.')
 parser.add_argument('--delete',
                     default=False,
                     action='store_true',
@@ -128,7 +130,7 @@ def main(argv: List[str]):
     azul = AzulClient(num_workers=args.num_workers)
 
     source_globs = set(args.sources)
-    if args.partition_prefix_length or args.deindex:
+    if not args.local or args.deindex:
         sources_by_catalog = defaultdict(set)
         globs_matched = set()
         for catalog in args.catalogs:
@@ -183,8 +185,8 @@ def main(argv: List[str]):
                 ):
                     reservation = BigQueryReservation()
                     reservation.activate()
-                if args.partition_prefix_length:
-                    azul.remote_reindex(catalog, args.partition_prefix_length, sources)
+                if not args.local:
+                    azul.remote_reindex(catalog, sources)
                     num_notifications = None
                 else:
                     num_notifications += azul.reindex(catalog, args.prefix)
--- a/src/azul/__init__.py
+++ b/src/azul/__init__.py
@@ -205,6 +205,10 @@ class Config:
     def dss_query_prefix(self) -> str:
         return os.environ.get('AZUL_DSS_QUERY_PREFIX', '')
 
+    @property
+    def partition_prefix_length(self) -> int:
+        return int(os.environ['AZUL_PARTITION_PREFIX_LENGTH'])
+
     # Remove once https://github.com/HumanCellAtlas/data-store/issues/1837 is resolved
 
     def dss_deployment_stage(self, dss_endpoint: str) -> str:
--- a/src/azul/azulclient.py
+++ b/src/azul/azulclient.py
@@ -10,8 +10,6 @@ from functools import (
 )
 from itertools import (
     groupby,
-    product,
-    starmap,
 )
 import json
 import logging
@@ -228,40 +226,39 @@ class AzulClient(object):
 
     def remote_reindex(self,
                        catalog: CatalogName,
-                       partition_prefix_length: int,
                        sources: AbstractSet[str]):
-        partition_prefixes = [
-            ''.join(partition_prefix)
-            for partition_prefix in product('0123456789abcdef',
-                                            repeat=partition_prefix_length)
-        ]
 
-        def message(source: str, partition_prefix: str) -> JSON:
-            logger.info('Remotely reindexing prefix %r of source %r into catalog %r',
-                        partition_prefix, source, catalog)
-            return dict(action='reindex',
-                        catalog=catalog,
-                        source=str(source),
-                        prefix=partition_prefix)
+        plugin = self.repository_plugin(catalog)
+        for source in sources:
+            source = plugin.resolve_source(source)
+            plugin.verify_source(source)
 
-        messages = starmap(message, product(sources, partition_prefixes))
-        for batch in chunked(messages, 10):
-            entries = [
-                dict(Id=str(i), MessageBody=json.dumps(message))
-                for i, message in enumerate(batch)
-            ]
-            self.notifications_queue.send_messages(Entries=entries)
+            def message(partition_prefix: str) -> JSON:
+                logger.info('Remotely reindexing prefix %r of source %r into catalog %r',
+                            partition_prefix, str(source.spec), catalog)
+                return dict(action='reindex',
+                            catalog=catalog,
+                            source=source.to_json(),
+                            prefix=partition_prefix)
+
+            messages = map(message, source.spec.prefix.partition_prefixes())
+            for batch in chunked(messages, 10):
+                entries = [
+                    dict(Id=str(i), MessageBody=json.dumps(message))
+                    for i, message in enumerate(batch)
+                ]
+                self.notifications_queue.send_messages(Entries=entries)
 
     def remote_reindex_partition(self, message: JSON) -> None:
         catalog = message['catalog']
         prefix = message['prefix']
         validate_uuid_prefix(prefix)
-        source = message['source']
-        bundle_fqids = self.list_bundles(catalog, source, prefix)
+        source = self.repository_plugin(catalog).source_from_json(message['source'])
+        bundle_fqids = self.list_bundles(catalog, str(source.spec), prefix)
         bundle_fqids = self.filter_obsolete_bundle_versions(bundle_fqids)
         logger.info('After filtering obsolete versions, '
                     '%i bundles remain in prefix %r of source %r in catalog %r',
-                    len(bundle_fqids), prefix, source, catalog)
+                    len(bundle_fqids), prefix, str(source.spec), catalog)
         messages = (
             {
                 'action': 'add',
@@ -292,8 +289,9 @@ class AzulClient(object):
         each bundle UUID.
         >>> AzulClient.filter_obsolete_bundle_versions([])
         []
-        >>> from azul.indexer import SimpleSourceSpec, SourceRef
-        >>> s = SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n'))
+        >>> from azul.indexer import SimpleSourceSpec, SourceRef, Prefix
+        >>> p = Prefix.parse('/2')
+        >>> s = SourceRef(id='i', spec=SimpleSourceSpec(prefix=p, name='n'))
         >>> def b(u, v):
         ...     return SourcedBundleFQID(source=s, uuid=u, version=v)
         >>> AzulClient.filter_obsolete_bundle_versions([
@@ -303,32 +301,53 @@ class AzulClient(object):
         ... ]) # doctest: +NORMALIZE_WHITESPACE
         [SourcedBundleFQID(uuid='c',
                            version='0',
-                           source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n'))),
+                           source=SourceRef(id='i',
+                                            spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                                partition=2),
+                                                                  name='n'))),
         SourcedBundleFQID(uuid='b',
                           version='3',
-                          source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n'))),
+                          source=SourceRef(id='i',
+                                           spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                               partition=2),
+                                                                 name='n'))),
         SourcedBundleFQID(uuid='a',
                           version='1',
-                          source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n')))]
+                          source=SourceRef(id='i',
+                                           spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                               partition=2),
+                                                                 name='n')))]
         >>> AzulClient.filter_obsolete_bundle_versions([
         ...     b('C', '0'), b('a', '1'), b('a', '0'),
         ...     b('a', '2'), b('b', '1'), b('c', '2')
         ... ]) # doctest: +NORMALIZE_WHITESPACE
         [SourcedBundleFQID(uuid='c',
                            version='2',
-                           source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n'))),
+                           source=SourceRef(id='i',
+                                            spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                                partition=2),
+                                                                  name='n'))),
         SourcedBundleFQID(uuid='b',
-                           version='1',
-                           source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n'))),
+                          version='1',
+                          source=SourceRef(id='i',
+                                           spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                               partition=2),
+                                                                 name='n'))),
         SourcedBundleFQID(uuid='a',
-                           version='2',
-                           source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n')))]
+                          version='2',
+                          source=SourceRef(id='i',
+                                           spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                               partition=2),
+                                                                 name='n')))]
         >>> AzulClient.filter_obsolete_bundle_versions([
         ...     b('a', '0'), b('A', '1')
         ... ]) # doctest: +NORMALIZE_WHITESPACE
         [SourcedBundleFQID(uuid='A',
                            version='1',
-                           source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n')))]
+                           source=SourceRef(id='i',
+                                            spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                                                partition=2),
+                                                                  name='n')))]
         """
 
         # Sort lexicographically by source and FQID. I've observed the DSS
--- a/src/azul/indexer/__init__.py
+++ b/src/azul/indexer/__init__.py
@@ -2,6 +2,10 @@ from abc import (
     ABC,
     abstractmethod,
 )
+from itertools import (
+    product,
+)
+import logging
 from threading import (
     RLock,
 )
@@ -10,6 +14,7 @@ from typing import (
     ClassVar,
     Dict,
     Generic,
+    Iterator,
     Optional,
     Protocol,
     TYPE_CHECKING,
@@ -25,8 +30,9 @@ from more_itertools import (
 )
 
 from azul import (
+    cached_property,
+    config,
     reject,
-    require,
 )
 from azul.types import (
     JSON,
@@ -48,6 +54,8 @@ else:
 
         def __lt__(self, __other: Any) -> bool: ...
 
+log = logging.getLogger(__name__)
+
 BundleUUID = str
 BundleVersion = str
 
@@ -57,6 +65,113 @@ class BundleFQID(SupportsLessThan):
     uuid: BundleUUID
     version: BundleVersion
 
+
+@attr.s(frozen=True, auto_attribs=True, kw_only=True)
+class Prefix:
+    common: str = ''
+    partition: Optional[int]
+    of_everything: ClassVar['Prefix']
+
+    def __attrs_post_init__(self):
+        validate_uuid_prefix(self.common)
+        assert ':' not in self.common, self.common
+        if self.partition:
+            assert isinstance(self.partition, int), self.partition
+            # Version 4 UUIDs specify fixed bits in the third dash-seperated
+            # group. To ensure that any concatenation of common and
+            # partition_prefix is a valid UUID prefix, we restrict the number of
+            # characters from the concatenation to be within the first
+            # dash-seperated group.
+            reject(len(self.common) + self.partition > 8,
+                   'Invalid common prefix and partition length', self)
+
+    @classmethod
+    def parse(cls, prefix: str) -> 'Prefix':
+        """
+        >>> Prefix.parse('aa/1')
+        Prefix(common='aa', partition=1)
+
+        >>> p = Prefix.parse('a')
+        >>> print(p.partition)
+        None
+        >>> p.effective.partition == config.partition_prefix_length
+        True
+
+        >>> Prefix.parse('aa/')
+        Traceback (most recent call last):
+        ...
+        azul.RequirementError: ('Prefix source cannot end in a delimiter.', 'aa/', '/')
+
+        >>> Prefix.parse('8f538f53/1').partition_prefixes() # doctest: +NORMALIZE_WHITESPACE
+        Traceback (most recent call last):
+        ...
+        azul.RequirementError: ('Invalid common prefix and partition length',
+                                Prefix(common='8f538f53', partition=1))
+
+        >>> list(Prefix.parse('8f538f53/0').partition_prefixes())
+        ['']
+        """
+        source_delimiter = '/'
+        reject(prefix.endswith(source_delimiter),
+               'Prefix source cannot end in a delimiter.', prefix, source_delimiter)
+        if prefix == '':
+            entry = ''
+            partition = None
+        else:
+            try:
+                entry, partition = prefix.split(source_delimiter)
+            except ValueError:
+                entry = prefix
+                partition = None
+            if partition:
+                try:
+                    partition = int(partition)
+                except ValueError:
+                    raise ValueError('Partition prefix length must be an integer.', partition)
+        validate_uuid_prefix(entry)
+        return cls(common=entry, partition=partition)
+
+    @cached_property
+    def effective(self) -> 'Prefix':
+        if self.partition is None:
+            return attr.evolve(self, partition=config.partition_prefix_length)
+        else:
+            return self
+
+    def partition_prefixes(self) -> Iterator[str]:
+        """
+        >>> list(Prefix.parse('/0').partition_prefixes())
+        ['']
+
+        >>> list(Prefix.parse('/1').partition_prefixes())
+        ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
+
+        >>> len(list(Prefix.parse('/2').partition_prefixes()))
+        256
+        """
+        partition_prefixes = map(''.join, product('0123456789abcdef',
+                                                  repeat=self.partition))
+        for partition_prefix in partition_prefixes:
+            validate_uuid_prefix(self.common + partition_prefix)
+            yield partition_prefix
+
+    def __str__(self):
+        """
+        >>> s = 'aa'
+        >>> s == str(Prefix.parse(s))
+        True
+
+        >>> s = 'aa/1'
+        >>> s == str(Prefix.parse(s))
+        True
+        """
+        if self.partition is None:
+            return self.common
+        else:
+            return f'{self.common}/{self.partition}'
+
+
+Prefix.of_everything = Prefix.parse('').effective
 
 SOURCE_SPEC = TypeVar('SOURCE_SPEC', bound='SourceSpec')
 
@@ -70,11 +185,7 @@ class SourceSpec(ABC, Generic[SOURCE_SPEC]):
     have simple unstructured names may want to use :class:`SimpleSourceSpec`.
     """
 
-    prefix: str = ''
-
-    def __attrs_post_init__(self):
-        validate_uuid_prefix(self.prefix)
-        assert ':' not in self.prefix, self.prefix
+    prefix: Prefix
 
     @classmethod
     @abstractmethod
@@ -84,6 +195,10 @@ class SourceSpec(ABC, Generic[SOURCE_SPEC]):
     @abstractmethod
     def __str__(self) -> str:
         raise NotImplementedError
+
+    @cached_property
+    def effective(self) -> SOURCE_SPEC:
+        return attr.evolve(self, prefix=self.prefix.effective)
 
 
 @attr.s(frozen=True, auto_attribs=True, kw_only=True)
@@ -96,8 +211,10 @@ class SimpleSourceSpec(SourceSpec['SimpleSourceSpec']):
     @classmethod
     def parse(cls, spec: str) -> 'SimpleSourceSpec':
         """
-        >>> SimpleSourceSpec.parse('https://foo.edu:12')
-        SimpleSourceSpec(prefix='12', name='https://foo.edu')
+        >>> SimpleSourceSpec.parse('https://foo.edu:12') # doctest: +NORMALIZE_WHITESPACE
+        SimpleSourceSpec(prefix=Prefix(common='12',
+                                       partition=None),
+                         name='https://foo.edu')
 
         >>> SimpleSourceSpec.parse('foo')
         Traceback (most recent call last):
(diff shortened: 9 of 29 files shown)


src/azul/azulclient.py
```diff
@@ -228,29 +227,32 @@ def notifications_queue(self):
...
+                            partition_prefix, source, catalog)
+                return dict(action='reindex',
+                            catalog=catalog,
+                            source=str(source),
+                            prefix=partition_prefix)
+
+            partition_prefix = [''.join(partition_prefix)
+                                for partition_prefix in product('0123456789abcdef',
+                                                                repeat=source_name.partition_prefix_length)
+                                ]
```
noah-aviel-dove, line 247: Nice job reorganizing this section, but the wrapping here isn't right:
```suggestion
            partition_prefix = [
                ''.join(partition_prefix)
                for partition_prefix in product('0123456789abcdef',
                                                               repeat=source_name.partition_prefix_length)
            ]
```

That said, this could easily be a generator:

```suggestion
            partition_prefix = map(''.join, product('0123456789abcdef',
                                                                       repeat=source_name.partition_prefix_length))
```

And I also can't help but feel that the variable name should be plural:

```suggestion
            partition_prefixes = ...
```

But I might be being a little overzealous here. Your call.

noah-aviel-dove (changes requested):

src/azul/plugins/repository/dss/__init__.py
```diff
@@ -70,6 +70,7 @@
 class DSSSourceName(SourceName):
     endpoint: str
     prefix: str = config.dss_query_prefix
+    prefix_partition_length: int = config.partition_prefix_length
```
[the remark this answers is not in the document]
amarjandu: dude another straggler. i thought i changed all of these. 🐛 🔫

src/azul/azulclient.py
amarjandu: actually the mapping on the `partition_prefixes` is a good call, if the length is 4, then the length of the list is 65536 values.... so def an improvement to use the generator.

It should have been plural in the first place.

UPGRADING.rst
```diff
@@ -10,6 +10,14 @@ branch that does not have the listed changes, the steps would need to be
...
 have too many entries in this file.
 
+#2750 Add partition_prefix_length to sources
+============================================
+
+The syntax of ``AZUL_TDR_SOURCES`` and ``AZUL_TDR_…_SOURCES`` environment
+variables was modified to include the partition prefix length.  To upgrade a
+deployment, append every source entry in the deployment's `environment.py`
+with a colon delimiter ``:`` followed by the partition length e.g., ``:2``.
+As always, use the sandbox deployment's ``environment.py`` as a template.
```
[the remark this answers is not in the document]
amarjandu: I borrowed the majority of the upgrading instructions from this review comment https://github.com/DataBiosphere/azul/pull/2827#discussion_r590713556 in the chained PR. 

Your suggestion is easier to read, I'm going to change it here, and change it in the prefix PR as well.

scripts/reindex.py
```diff
@@ -108,6 +98,11 @@
                     default=True,
                     action='store_false',
                     help='Suppress management of BigQuery slot commitments.')
+parser.add_argument('--remote',
```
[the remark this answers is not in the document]
amarjandu: i moved it and added some more details to the `help` argument.

src/azul/terra.py
```diff
@@ -134,14 +134,19 @@ def parse(cls, source: str) -> 'TDRSourceName':
...
-        if partition_prefix_length == '':
+        try:
+            partition_prefix_length = int(partition_prefix_length)
+        except ValueError:
+            require(partition_prefix_length == '',
+                    'Must be number or empty',
+                    partition_prefix_length)
             partition_prefix_length = config.partition_prefix_length
-            source = f'{source}{partition_prefix_length}'
+            source += str(partition_prefix_length)
```
noah-aviel-dove: I'm not convinced I made the right call here. Up to you whether to keep this version or revert to what you had originally.

scripts/reindex.py
```diff
@@ -50,16 +50,17 @@
...
+                    dest='remote',
+                    default=False,
+                    action='store_true',
+                    help='Perform the reindexing remotely. The partition prefix length for the catalog source is used '
+                         'to form queries that are processed independently and remotely by the indexer lambda. The '
+                         'lambda queries the repository and queues a notification for each matching subgraph. If the '
+                         'flag is set false (the default) no partitioning occurs, the repository is queried locally '
+                         'and the indexer notification endpoint is invoked for each bundle individually and '
+                         'concurrently using worker threads. This is magnitudes slower than remote i.e., partitioned'
+                         ' indexing.')
```
noah-aviel-dove, line 63:
```suggestion
                    help='Perform the reindexing remotely. The set of subgraphs matching the query is partitioned using the ' 
                             'partition prefix length(s) configured for the catalog sources(s). Each query partition is processed '
                             'independently and remotely by the indexer lambda. The lambda queries the repository and '
                             'queues a notification for each matching subgraph. If this flag is not set, no partitioning occurs, '
                             'the repository is queried locally, and the indexer notification endpoint is invoked for each bundle '
                             'individually and concurrently using worker threads. This is magnitudes slower than remote (i.e. '
                             partitioned) indexing.')
```
```

Comment (noah-aviel-dove) [reply scripts/reindex.py:63]: I've re-written this to try and preserve more of the original language which I think was clearer. I also made the following specific changes:
 - Replaced a lingering reference to "bundle" with subgraph".
 - Pluralized references to catalog sources since there can be more than one per catalog.
 - Added oxford comma and changed another comma I felt was awkward to parentheses. These changes are admittedly a bit subjective.
 - Moved leading whitespace to the previous line.

Comment (noah-aviel-dove) [reply scripts/reindex.py:63]: Good thing you added this paragraph, it was an oversight on my part not to point out that you initially omitted it.

**Review by noah-aviel-dove — CHANGES_REQUESTED** [review]
Code looks good, just some prose suggestions regrading the argument help paragraph.

File: src/azul/terra.py (hunk 5 above)
Comment (amarjandu) [reply src/azul/terra.py]: I preferred the old way personally. reverted.

File: scripts/reindex.py (hunk 6 above)
Comment (amarjandu) [reply scripts/reindex.py:63]: Noticed it because of the small diff by moving `--remote` to where `--partition-prefix-length` was defined. :)

**Review by noah-aviel-dove — APPROVED** [review]
Changes look good! Just make sure everything (including the previous branch in the chain) is properly rebased, since you have `drop!` commit here that's now on develop.

**Comment by amarjandu** [comment]
@noah-aviel-dove i realized that I did not update the env files for any of the deployments, I've added a fixup commit that addresses this.

**Comment by noah-aviel-dove** [comment]
> @noah-aviel-dove i realized that I did not update the env files for any of the deployments, I've added a fixup commit that addresses this.

Good catch

**Review by noah-aviel-dove — APPROVED** [review]

File: src/azul/indexer/__init__.py [hunk 7]
```diff
@@ -63,6 +63,9 @@ class SourceName(ABC, Generic[SOURCE_NAME]):
     have simple unstructured names may want to use :class:`StringSourceName`.
     """
 
+    prefix: str
```
amarjandu, line 66: Here

src/azul/__init__.py
```diff
@@ -202,6 +202,10 @@ def sam_service_url(self):
     def dss_query_prefix(self) -> str:
         return os.environ.get('AZUL_DSS_QUERY_PREFIX', '')
 
+    @property
+    def partition_prefix_length(self) -> int:
+        return int(os.environ.get('AZUL_PARTITION_PREFIX_LENGTH', 2))
```
hannes-ucsc, line 207: Defaults go only into the top-level `environment.py` and shouldn't be duplicated here.

deployments/prod/environment.py
```diff
@@ -45,11 +45,11 @@ def env() -> Mapping[str, Optional[str]]:
         ]),
 
         'AZUL_TDR_SOURCES': ','.join([
-            'tdr:broad-datarepo-terra-prod-hca2:snapshot/hca_prod_20201118_dcp1___20201209:',
+            'tdr:broad-datarepo-terra-prod-hca2:snapshot/hca_prod_20201118_dcp1___20201209::2',
```
hannes-ucsc, line 48:
```suggestion
            'tdr:broad-datarepo-terra-prod-hca2:snapshot/hca_prod_20201118_dcp1___20201209:-2',
```

So that both prefix-related settings are in the same field. `-` is not allowed to occur in a UUID prefix so we can use it as a delimiter. `…:aa-1` means use `aa` for prefix and a partition prefix of length one. `…:aa` means to use prefix `aa` and the default prefix length.  `…:aa-` is not allowed. `…:` means to use no prefix and the default partition prefix length.

The ticket specifies 

> I don't think that there is a reasonable default that works for all projects so we shouldn't hard-code one anywhere. 

but the addition of `AZUl_PARTITION_PREFIX_LENGTH` goes against that. That had me rethink this.

hannes-ucsc: And since 2 is actually the default, we shouldn't have to duplicate it here:

```suggestion
            'tdr:broad-datarepo-terra-prod-hca2:snapshot/hca_prod_20201118_dcp1___20201209:',
```

deployments/sandbox/environment.py
```diff
@@ -55,17 +55,17 @@ def env() -> Mapping[str, Optional[str]]:
         # FIXME: Add tooling to aid in prefix choice
         #        https://github.com/DataBiosphere/azul/issues/3027
         'AZUL_TDR_SOURCES': ','.join([
-            'tdr:broad-jade-dev-data:snapshot/hca_dev_20201203___20210426:42',
+            'tdr:broad-jade-dev-data:snapshot/hca_dev_20201203___20210426:42:2',
         ]),
         **{
             f'AZUL_TDR_{catalog.upper()}_SOURCES': ','.join([
-                f'tdr:broad-jade-dev-data:snapshot/hca_dev_20201023_ebiv4___20210302:4'
+                f'tdr:broad-jade-dev-data:snapshot/hca_dev_20201023_ebiv4___20210302:4:2'
```
hannes-ucsc, line 62: IIRC, the prefix `4` gives 200 or so bundles so a partition prefix of two digits doesn't really make sense.

deployments/sandbox/environment.py
```diff
@@ -55,17 +55,17 @@ def env() -> Mapping[str, Optional[str]]:
...
             ])
             for catalog in ('dcp2ebi', 'it2ebi')
         },
         **{
             f'AZUL_TDR_{catalog.upper()}_SOURCES': ','.join([
-                'tdr:broad-jade-dev-data:snapshot/lungmap_dev_20210412__20210414:',
+                'tdr:broad-jade-dev-data:snapshot/lungmap_dev_20210412__20210414::2',
```
hannes-ucsc, line 68: Same here. Is a partition prefix length of 0 supported? If it is easy to do I would want to include support for it and consider lungmap as a test case of that support.

environment.py
```diff
@@ -330,6 +330,10 @@ def env() -> Mapping[str, Optional[str]]:
         # indexed.
         'AZUL_DSS_QUERY_PREFIX': '',
 
+        # The length of the subgraph UUID prefix by which to partition the set of
+        # subgraphs matching the query.
```
hannes-ucsc, line 334:
```suggestion
        # subgraphs matching the query during remote reindexing. Partition prefixes that are too long result in many small or even empty partitions and waste some amount of resources. Partition prefixes that are too short result in few large partitions that could exceed the memory and running time limitations of the AWS Lamba function that processes them. If in doubt err on the side of too many small partitions.
```

scripts/reindex.py
```diff
@@ -50,16 +50,17 @@
...
+parser.add_argument('--remote',
+                    dest='remote',
+                    default=False,
+                    action='store_true',
+                    help='Perform the reindexing remotely. The set of subgraphs matching the query is partitioned '
+                         'using the partition prefix length(s) configured for the catalog sources(s). Each query '
+                         'partition is processed independently and remotely by the indexer lambda. The lambda queries '
+                         'the repository and queues a notification for each matching subgraph. If this flag is not '
+                         'set, no partitioning occurs, the repository is queried locally, and the indexer '
+                         'notification endpoint is invoked for each bundle individually and concurrently using worker '
```
hannes-ucsc, line 62:
```suggestion
                         'notification endpoint is invoked for each subgraph individually and concurrently using worker '
```

scripts/reindex.py
```diff
@@ -50,16 +50,17 @@
...
-                    help='The length of the bundle UUID prefix by which to partition the set of bundles matching the '
-                         'query. Each query partition is processed independently and remotely by the indexer lambda. '
-                         'The lambda queries the repository and queues a notification for each matching bundle. If 0 '
-                         '(the default) no partitioning occurs, the repository is queried locally and the indexer '
-                         'notification endpoint is invoked for each bundle individually and concurrently using worker'
-                         'threads. This is magnitudes slower than remote i.e., partitioned indexing.')
+parser.add_argument('--remote',
+                    dest='remote',
+                    default=False,
+                    action='store_true',
```
hannes-ucsc, line 56: Since `--remote` is the normal case, I think we should

 ```suggestion
parser.add_argument('--local',
                    dest='local',
                    default=False,
                    action='store_true',
```

**Review by hannes-ucsc — CHANGES_REQUESTED** [review]

File: deployments/prod/environment.py (hunk 9 above)
Comment (amarjandu) [reply deployments/prod/environment.py:48]: >So that both prefix-related settings are in the same field. - is not allowed to occur in a UUID prefix so we can use it as a delimiter.

someone COULD choose to use `6cab9e80-0` as a prefix, maybe we should use an alternative delimiter, how about `~`?

File: deployments/sandbox/environment.py (hunk 11 above)
Comment (amarjandu) [reply deployments/sandbox/environment.py:68]: partition_prefix_length of `0` would send one notification to the indexer lambda to remote reindex the entire prefix by itself.
```
> list(map(''.join, itertools.product('0123456789abcdef', repeat=0)))
['']
```

A notification would be passed to `remote_reindex_partition` where the bundles under the prefix + the empty partition would be listed out, then queued up for indexing.

This might defeat the purpose of a remote reindex as we want the bundle enumeration to be parallelized by the lambdas. Rather than preventing this, we can add a warning if the partition_prefix_length is set to `0`. 

> consider lungmap as a test case of that support.

For testing would you like to see another IT, or just confirm that the expected notifications will be passed to the lambdas in a unit test? I think a unit test will be sufficient as the rest of the code has coverage.

What about when the lungmap source changes?

File: deployments/prod/environment.py (hunk 9 above)
Comment (hannes-ucsc) [reply deployments/prod/environment.py:48]: Valid point. I thought I wrote code to only allow hex characters but I can't find it. So let's use `/`

Independently, we should disallow prefixes that end in a dash. I don't like the ambiguity: `6cab9e80-` and `6cab9e80` match the same set of UUIDs.

File: deployments/sandbox/environment.py (hunk 11 above)
Comment (hannes-ucsc) [reply deployments/sandbox/environment.py:68]: > partition_prefix_length of `0` would send one notification to the indexer lambda to remote reindex the entire prefix by itself.

That's the desired behavior.

> A notification would be passed to `remote_reindex_partition` where the bundles under the prefix + the empty partition would be listed out, then queued up for indexing.
> 
> This might defeat the purpose of a remote reindex as we want the bundle enumeration to be parallelized by the lambdas.

I don't think it does. If the source is so small that it can be handled by one lambda invocation, there is no need to further break it up. Partitions smaller than a certain size aren't going to deliver a speed-up. I don't know what that size is, but a partition size that packs one lambda invocation is a good approximation. And even if there is just one partition, remote reindexing makes sense, because it still gets to benefit from the implicit retry mechanism and the fact that the diagnostic output is kept in a central place.

> What about when the lungmap source changes?

That's no different to the question about what happens when any source changes. The person configuring the source needs to know how many bundles it contains in order to make a decision on which partition prefix length to configure. The source is immutable, so a valid choice will always remain valid.

File: src/azul/indexer/__init__.py [hunk 15]
```diff
@@ -55,6 +60,78 @@ class BundleFQID(SupportsLessThan):
     version: BundleVersion
 
 
+@attr.s(frozen=True, auto_attribs=True, kw_only=True)
+class Prefix:
+    prefix: str = ''
```
hannes-ucsc, line 65:
```suggestion
    global: str = ''
```

src/azul/indexer/__init__.py
```diff
@@ -55,6 +60,78 @@ class BundleFQID(SupportsLessThan):
...
+    partition_prefix_length: int
```
hannes-ucsc, line 66:
```suggestion
    partition: int
```

It's imprecise but we need the `repr` result to be shorter for all those doc tests

src/azul/uuids.py
```diff
@@ -57,28 +61,36 @@ def validate_uuid_prefix(uuid_prefix: str) -> None:
...
-    >>> uuid = '8f53d355-b2fa-4bab-a2f2-6852d852d2ec'
-    >>> # noinspection PyUnresolvedReferences
-    >>> prefixes = [uuid[:n] for n in range(len(uuid))]
-    >>> for prefix in prefixes: validate_uuid_prefix(prefix)
+    # A complete UUID is a valid prefix
+    >>> validate_uuid_prefix('8f53d355-b2fa-4bab-a2f2-6852d852d2ec')
 
-    >>> validate_uuid_prefix('aaaaaaaa-aaaa-')
-    >>> validate_uuid_prefix('aaaaaaaa-aaaa-a')
+    >>> validate_uuid_prefix('8F53')
```
hannes-ucsc, line 86: Can you move this back so as to minimize the diff?

test/indexer/__init__.py
```diff
@@ -100,7 +100,7 @@ def _load_canned_bundle(cls, bundle: SourcedBundleFQID) -> Bundle:
 
 class IndexerTestCase(ElasticsearchTestCase, CannedBundleTestCase):
     index_service: IndexService
-    source = DSSSourceRef.for_dss_endpoint('test')
+    source = DSSSourceRef.for_dss_endpoint(endpoint='test')
```
hannes-ucsc, line 103: why this change? Doesn't add any clarity.

src/azul/terra.py
```diff
@@ -119,17 +124,22 @@ def parse(cls, spec: str) -> 'TDRSourceSpec':
         # BigQuery (and by extension the TDR) does not allow : or / in dataset names
         # FIXME: Move parsing of prefix to SourceSpec
         #        https://github.com/DataBiosphere/azul/issues/3073
-        service, project, name, prefix = spec.split(':')
-        type, name = name.split('/')
+        service, project, _name, _prefix = spec.split(':')
```
hannes-ucsc, line 127:
```suggestion
        service, project, name, prefix = spec.split(':')
```

Reassigning is ok.

src/azul/terra.py
```diff
@@ -119,17 +124,22 @@ def parse(cls, spec: str) -> 'TDRSourceSpec':
...
             is_snapshot = True
         elif type == cls._type_dataset:
             is_snapshot = False
         else:
             assert False, type
-        validate_uuid_prefix(prefix)
-        self = cls(prefix=prefix, project=project, name=name, is_snapshot=is_snapshot)
+        prefix = Prefix.parse(_prefix)
+        if _prefix != str(prefix):
+            spec = ':'.join((service, project, _name, str(prefix)))
```
hannes-ucsc, line 138: This special case to just make the assertion pass is a bad smell. Parsing and formatting should be the 100% inverse of each other.

src/azul/azulclient.py
```diff
@@ -104,6 +102,7 @@ def synthesize_notification(self,
             'source': {
                 'id': bundle_fqid.source.id,
                 'spec': str(bundle_fqid.source.spec),
+                'prefix': str(bundle_fqid.source.spec.prefix)
```
hannes-ucsc, line 105: Why do we need this? Isn't the prefix included in `str(bundle_fqid.source.spec)`?

test/indexer/test_indexer_controller.py
```diff
@@ -89,12 +90,41 @@ def test_invalid_notification(self):
         self.assertRaises(AssertionError, self.controller.contribute, event)
 
     def test_remote_reindex(self):
-        event = [self._mock_sqs_record(action='reindex', prefix='ff')]
-        with mock.patch.object(target=AzulClient,
-                               attribute=AzulClient.remote_reindex_partition.__name__) as mock_method:
-            mock_method.return_value = True
-            self.controller.contribute(event)
-            mock_method.assert_called_once_with(dict(action='reindex', prefix='ff'))
+        mock_ppl = mock.patch.dict(os.environ, AZUL_PARTITION_PREFIX_LENGTH='0')
```
hannes-ucsc, line 93: patch.dict accepts multiple kwargs

hannes-ucsc: Even if it didn't you don't need to assign individual variables. A single tuple-typed variable would work. or you could just inline all three and wrap the `with` statement.

test/indexer/test_indexer_controller.py
```diff
@@ -89,12 +90,41 @@ def test_invalid_notification(self):
...
+        mock_dqp = mock.patch.dict(os.environ, AZUL_DSS_QUERY_PREFIX='ff')
+        mock_dss = mock.patch.dict(os.environ, AZUL_DSS_ENDPOINT='foo_source')
+        with mock_ppl, mock_dqp, mock_dss:
```
hannes-ucsc, line 96: Why did this get so much longer? Is this a remnant from another PR?

hannes-ucsc (changes requested):

hannes-ucsc (changes requested):
Also, shouldn't there be changes to the deployment configuration of at least the sandbox?

test/indexer/test_indexer_controller.py
amarjandu: It's been a while since we had this discussion but if i recall correctly; this test modification came from a previous discussion about having a `partition_prefix_length` of 0. We wanted to ensure that the synthesized notification work as expected. Rather than creating a new test that had this functionality we opted to extend the `test_remote_reindex`.

src/azul/azulclient.py
amarjandu: Removed.

src/azul/terra.py
amarjandu: From this comment it seemed like the delimiter was optional https://github.com/DataBiosphere/azul/pull/2931#discussion_r626111147

> `-` is not allowed to occur in a UUID prefix so we can use it as a delimiter. …:aa-1 means use aa for prefix and a partition prefix of length one. …:aa means to use prefix aa and the default prefix length. …:aa- is not allowed. …: means to use no prefix and the default partition prefix length.

```
  File "/Users/amar/dev/hca/azul/src/azul/terra.py", line 143, in parse
    assert spec == str(self), (spec, str(self), self)
AssertionError: ('tdr:foo:snapshot/bar:', 'tdr:foo:snapshot/bar:/2', TDRSourceSpec(prefix=Prefix(prefix='', partition_prefix_length=2), project='foo', name='bar', is_snapshot=True))
```
Would you prefer that the assertion would be more loose?

src/azul/uuids.py
amarjandu: self-note: inspect diff when other fixups land.

src/azul/indexer/__init__.py
amarjandu: Cant use `global` its a keyword.

amarjandu: change to `initial`

amarjandu:
```
|                             | dcp2   | dcp2ebi | lungmap |
|-----------------------------|--------|---------|---------|
| Total Bundles in Snapshot   | 137935 | 2119    | 55      |
| Prefixed Bundles in Sandbox | 8666   | 15      | 55      |
```

We want to try different partitioning on a smaller dataset, maybe lungmap with partition of `1`?

amarjandu:
self note: pre-rebase hash 5a2507b

src/azul/plugins/__init__.py
```diff
@@ -233,7 +233,7 @@ def resolve_source(self, spec: str) -> SOURCE_REF:
         ref_cls = self._source_ref_cls
         spec = ref_cls.spec_cls().parse(spec)
         id = self.lookup_source_id(spec)
-        return ref_cls(id=id, spec=spec)
+        return ref_cls(id=id, spec=spec).effective
```
amarjandu, line 236: @hannes-ucsc highlighting this change.

amarjandu:
There is an issue with the `.effective` method in that it modifies the spec instance within a SourceRef, this breaks the `intern` handling that the SourceRef provides because the Prefix.partition attribute can/will change from `None` to the default value.

Attempted to expand the keys that are taken into consideration for the interning of SourceRef instances, the issue with expanding additional entries into the lookup key is that it breaks the principal that an ID refers to a unique SourceSpec.
If we expand the keys to include (cls, id, spec) the lookup would allow the same ID to have unique/different Specs. 

It is possible that we have a nested lookup that first looks at the (cls, id) then performs another lookup at a `spec.name` which finally points at a set of SourceRef instances. It's fairly complex to manage all the instances that are present.

amarjandu:
self note: IT passes with .effective changes + changes to intern keys

amarjandu:
Note to self: `make reindex` completes with no failures in the queues on local deployment.

hannes-ucsc (changes requested):

environment.py
```diff
@@ -330,6 +330,16 @@ def env() -> Mapping[str, Optional[str]]:
...
 
+        # The length of the subgraph UUID prefix by which to partition the set
+        # of subgraphs matching the query during remote reindexing. Partition
+        # prefixes that are too long result in many small or even empty
+        # partitions and waste some amount of resources. Partition prefixes that
+        # are too short result in few large partitions that could exceed the
+        # memory and running time limitations of the AWS Lamba function that
+        # processes them. If in doubt err on the side of too many small
+        # partitions.
+        'AZUL_PARTITION_PREFIX_LENGTH': '2',
```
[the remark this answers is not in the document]
amarjandu: Created #3260, added FIXME

scripts/recan_bundle_tdr.py
```diff
@@ -396,7 +397,8 @@ def main(argv):
                            metadata_files=metadata)
 
     tdr_source = TDRSourceRef(id=args.source_id,
-                              spec=TDRSourceSpec(project='test_project',
+                              spec=TDRSourceSpec(prefix=Prefix.parse(''),
```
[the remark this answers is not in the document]
amarjandu: I missed it because i was looking for usages of Spec.parse, not Prefix.parse. This is my fault.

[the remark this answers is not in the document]
amarjandu: Also calling .effective here may also throw in an operators config.parition_prefix_length into the canned bundle, This might not be a problem the canned bundles are expected to be modified before they are committed. 

I recall in a past conversation, "I expect an emtpy string `''` to be insignificant", when talking talking about fields in a canned bundle.

src/azul/indexer/__init__.py
```diff
@@ -59,6 +67,101 @@ class BundleFQID(SupportsLessThan):
...
+    def partition_prefixes(self) -> Iterator[str]:
+        """
+        >>> list(Prefix.parse('/0').partition_prefixes())
+        ['']
+        >>> sorted(Prefix.parse('/1').partition_prefixes())
+        ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
+        >>> len(set(Prefix.parse('/2').partition_prefixes()))
+        256
+        >>> sorted(Prefix.parse('8f538f53/1').partition_prefixes())
+        Traceback (most recent call last):
```
hannes-ucsc, line 147: This exception is raised by `parse`, not `partition_prefixes` so this doctest case should be in `parse` not here.

src/azul/indexer/__init__.py
```diff
@@ -121,12 +225,12 @@ def parse(cls, spec: str) -> 'SimpleSourceSpec':
...
         reject(sep == '',
                'Source specifications must end in a colon followed by an optional UUID prefix')
-        return cls(prefix=prefix, name=name)
+        return cls(prefix=Prefix.parse(prefix), name=name)
 
     def __str__(self) -> str:
         """
-        >>> str(SimpleSourceSpec(prefix='12', name='foo:bar/baz'))
-        'foo:bar/baz:12'
+        >>> str(SimpleSourceSpec(prefix=Prefix.parse('12/2'), name='foo:bar/baz'))
```
hannes-ucsc, line 232: As previously stated, every class that defines the `parse`, `__str__` pair should have a doctest in `__str__` proving that one is the inverse of the other. 

To be more specific, that test should look like 

```python
>>> s = 'string_representation'
>>> s == str(Foo.parse(s))
True
```

That test alone should cover all the code in `parse` and `__str__` except for code that handles invalid inputs. If this test case leaves code uncovered in those two methods, more cases for different string repesentations may need to be added until all non-error handling code in those two methods is covered.

scripts/recan_bundle_tdr.py
```diff
@@ -387,7 +388,7 @@ def main(argv):
         metadata = json.load(f)
 
     dss_source = DSSSourceRef(id='',
-                              spec=SimpleSourceSpec(prefix='',
+                              spec=SimpleSourceSpec(prefix=Prefix.of_everything().effective,
```
hannes-ucsc, line 391: This is not exactly what I [asked](https://github.com/DataBiosphere/azul/pull/2931#discussion_r673377097) for. If there's a reason to deviate from that, I would like to know.

hannes-ucsc (changes requested):

src/azul/indexer/__init__.py
amarjandu: The exception is not being raised by `parse`, its being raised within `partition_prefixes`, take a look at the 
error that is raised: `azul.uuids.InvalidUUIDPrefixError: '8f538f530' is not a valid UUID prefix.` The 0 was not present within the common prefix, it was added by the map during the method, which then fails to validate before its yielded out.

scripts/recan_bundle_tdr.py
```diff
@@ -388,7 +388,7 @@ def main(argv):
         metadata = json.load(f)
 
     dss_source = DSSSourceRef(id='',
-                              spec=SimpleSourceSpec(prefix=Prefix.of_everything().effective,
+                              spec=SimpleSourceSpec(prefix=Prefix.of_everything,
```
amarjandu, line 391: this throws pycharm warnings for attribute not found.

scripts/recan_bundle_tdr.py
amarjandu: I'm a little confused with what you're asking for. You want to add a constant attribute to Prefix that returns an instance of the prefix class, which has the prefix set to `''`. 

If the fixme does not address this we should talk during parking lot.

hannes-ucsc: I asked for a constant attribute but a method was added in response.

hannes-ucsc (changes requested):
https://github.com/DataBiosphere/azul/pull/2931#discussion_r674989559
https://github.com/DataBiosphere/azul/pull/2931#discussion_r674988031

amarjandu:
Self Note: testing top level comment.

amarjandu:
Changed primary reviewer to @danielsotirhos

src/azul/indexer/__init__.py
```diff
@@ -59,6 +68,101 @@ class BundleFQID(SupportsLessThan):
...
+        256
+        """
+        for partition_prefix in self._partition_generator(self.partition):
+            validate_uuid_prefix(self.common + partition_prefix)
+            yield partition_prefix
+
+    def _partition_generator(self, partition) -> Iterator[str]:
+        return map(''.join, product('0123456789abcdef', repeat=partition))
+
+    def __str__(self):
```
danielsotirhos, line 157: Missing doctest for https://github.com/DataBiosphere/azul/pull/2931#discussion_r674388377

src/azul/terra.py
```diff
@@ -161,7 +168,12 @@ def bq_name(self):
 
     def __str__(self) -> str:
```
danielsotirhos, line 169: Missing doctest for https://github.com/DataBiosphere/azul/pull/2931#discussion_r674388377

danielsotirhos (changes requested):

environment.py
```diff
@@ -330,6 +330,19 @@ def env() -> Mapping[str, Optional[str]]:
...
+        # of subgraphs matching the query during remote reindexing. Partition
+        # prefixes that are too long result in many small or even empty
+        # partitions and waste some amount of resources. Partition prefixes that
+        # are too short result in few large partitions that could exceed the
+        # memory and running time limitations of the AWS Lamba function that
+        # processes them. If in doubt err on the side of too many small
+        # partitions.
+        # FIXME: Make targets should fail if this is not set within a
+        #        deployments environment file.
+        #        https://github.com/DataBiosphere/azul/issues/3260
```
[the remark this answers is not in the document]
amarjandu: I think @hannes-ucsc wanted an explicit failure/assertion to fire. I'll double check.

src/azul/terra.py
```diff
@@ -150,18 +155,34 @@ def parse(cls, spec: str) -> 'TDRSourceSpec':
...
+        """
+        >>> s = 'tdr:foo:snapshot/bar:'
+        >>> s == str(TDRSourceSpec.parse(s))
+        True
+
+        >>> s = 'tdr:foo:snapshot/bar:22'
+        >>> s == str(TDRSourceSpec.parse(s))
+        True
+        """
```
danielsotirhos, line 178:
```suggestion
        >>> s = 'tdr:foo:snapshot/bar:22/2'
        >>> s == str(TDRSourceSpec.parse(s))
        True
        """
```

danielsotirhos (changes requested):

src/azul/indexer/__init__.py
```diff
@@ -58,6 +67,114 @@ class BundleFQID(SupportsLessThan):
...
+                entry, partition = prefix.split(source_delimiter)
+            except ValueError:
+                entry = prefix
+                partition = None
+            if partition:
+                try:
+                    partition = int(partition)
+                except ValueError:
+                    raise ValueError('Partition prefix length must be an integer.',
+                                     partition)
```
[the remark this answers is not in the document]
amarjandu: @danielsotirhos i think there was a typo on this line, i assume that you wanted this to all be on one line.....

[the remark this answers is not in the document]
danielsotirhos: No typo, Git is forcing the wrapping to fit the width of the suggestion box (note no '+' on the second line).
Your assumption is correct however, please put on 1 line.

src/azul/terra.py
```diff
@@ -105,23 +103,30 @@ class TDRSourceSpec(SourceSpec):
...
+        '/{partition_prefix_length}'.
 
         >>> s = TDRSourceSpec.parse('tdr:foo:snapshot/bar:')
-        >>> s
-        TDRSourceSpec(prefix='', project='foo', name='bar', is_snapshot=True)
+        >>> s # doctest: +NORMALIZE_WHITESPACE
+        TDRSourceSpec(prefix=Prefix(common='', partition=None),
+              project='foo',
+              name='bar',
+              is_snapshot=True)
```
danielsotirhos, line 114:
```suggestion
        TDRSourceSpec(prefix=Prefix(common='', partition=None),
                      project='foo',
                      name='bar',
                      is_snapshot=True)
```

danielsotirhos (changes requested):
Sorry, just one last alignment nitpick.

environment.py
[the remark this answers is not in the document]
amarjandu: fixme dropped

amarjandu:
self note: hash before large commit changes is 1a4b8c2

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -171,7 +171,7 @@ class Plugin(RepositoryPlugin[TDRSourceSpec, TDRSourceRef]):
     @classmethod
     def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
         return cls(sources=frozenset(
-            TDRSourceSpec.parse(spec).effective
+            TDRSourceSpec.parse(spec)
```
amarjandu: @hannes-ucsc 
removing .effective here causes failures during the `.list_bundle` operation, we cannot assert that the effective source from the notification matches the sources from the config class because its possible that the partition is set to None.

amarjandu: https://github.com/DataBiosphere/azul/blob/develop/src/azul/plugins/repository/tdr/__init__.py#L226

Should this assert be changed to inspect the SourceRef.id rather than the spec?

amarjandu:
self note: IT and `make reindex` pass

src/azul/terra.py
amarjandu: https://github.com/DataBiosphere/azul/pull/2931/commits/1d588613533a1b3d2b7a8d70babadd68eae6752f

danielsotirhos (changes requested):
The fixme in `environment.py` still needs to be removed.

danielsotirhos (approved):

src/azul/azulclient.py
```diff
@@ -300,32 +298,46 @@ def filter_obsolete_bundle_versions(cls,
         ... ]) # doctest: +NORMALIZE_WHITESPACE
         [SourcedBundleFQID(uuid='c',
                            version='0',
-                           source=SourceRef(id='i', spec=SimpleSourceSpec(prefix='42', name='n'))),
+                           source=SourceRef(id='i',
+                                            spec=SimpleSourceSpec(prefix=Prefix(common='', partition=2),
```
hannes-ucsc, line 302: Wrap the args to `Prefix`  for consistency

hannes-ucsc (changes requested):
I'll push an additional commit, too.

hannes-ucsc:
I pushed out of turn to fix the build.

amarjandu:
https://github.com/DataBiosphere/azul/pull/2931#discussion_r687262561

>  I'm assuming the warning is now gone.

Screen Shot 2021-08-12 at 10 37 04 AM


Yes, no warnings on my end.

src/azul/plugins/repository/dss/__init__.py
```diff
@@ -103,7 +105,9 @@ def create(cls, catalog: CatalogName) -> RepositoryPlugin:
     @property
     def sources(self) -> AbstractSet[SimpleSourceSpec]:
         assert config.dss_endpoint is not None
-        return {SimpleSourceSpec(name=config.dss_endpoint)}
+        return {
+            SimpleSourceSpec.parse(f'{config.dss_endpoint}:{config.dss_query_prefix}').effective
```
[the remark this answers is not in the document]
amarjandu: @hannes-ucsc 

Is the only reason for using the constructor the line length? If we use the line length, there are not many actual usages of `SimpleSourceSpec.parse`, the only case where its explicitly used is within the canned repository plugin...

Just a heads up, this will also change once we move from `dss_endpoint:dss_query_prefix` to `dss_source`

src/azul/indexer/__init__.py
```diff
@@ -58,6 +66,105 @@ class BundleFQID(SupportsLessThan):
...
+        Prefix(common='aa', partition=1)
+
+        >>> p = Prefix.parse('aa')
+        >>> p
+        Prefix(common='aa', partition=None)
+        >>> from unittest.mock import patch
+        >>> import os
+        >>> with patch.dict(os.environ, AZUL_PARTITION_PREFIX_LENGTH='2'):
+        ...     p.effective
+        Prefix(common='aa', partition=2)
```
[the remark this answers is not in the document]
amarjandu: changed it a bit...

hannes-ucsc (changes requested):
Conflicts. No bump.

amarjandu:
Resolved conflicts with `scripts/reindex.py` tested `make reindex` as well.

src/azul/plugins/repository/dss/__init__.py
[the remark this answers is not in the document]
hannes-ucsc: Makes sense.

src/azul/uuids.py
```diff
@@ -67,18 +71,28 @@ def validate_uuid_prefix(uuid_prefix: str) -> None:
...
-    >>> validate_uuid_prefix('aaaaaaaa-aaaa-a')
+    >>> validate_uuid_prefix('8f538f53a')
     Traceback (most recent call last):
     ...
-    azul.uuids.InvalidUUIDPrefixError: 'aaaaaaaa-aaaa-a' is not a valid UUID prefix.
+    azul.uuids.InvalidUUIDPrefixError: '8f538f53a' is not a valid UUID prefix.
     """
     valid_uuid_str = '26a8fccd-bbd2-4342-9c19-6ed7c9bb9278'
+    reject(uuid_prefix.endswith('-'),
+           f'Remove the trailing `-` from the uuid prefix: {uuid_prefix}')
```
hannes-ucsc, line 95: It's not a good idea to give specific remedies. Just describe what is wrong.

UUID is an acronym. LIke ID and URL.

Also, it looks like `8f538f53-a` would be considered valid. I thought we agreed to limit prefixes to the first dash-separated group in the hexadecimal representation of UUIDs because the third group has some fixed bits that can't be randomized.

test/indexer/test_indexer_controller.py
```diff
@@ -89,12 +91,43 @@ def test_invalid_notification(self):
...
+                SourcedBundleFQID(source=source,
+                                  uuid='ffa338fe-7554-4b5d-96a2-7df127a7640b',
+                                  version='2018-03-28T151023.074974Z')
+            ]
+            self.controller.repository_plugin(self.catalog)._assert_source(source)
+            mock_list_bundles = mock.patch('azul.plugins.repository.dss.Plugin.list_bundles',
+                                           return_value=bundle_fqids)
+            with mock_list_bundles:
+                self.controller.contribute(events)
+            notification = one(
```
hannes-ucsc, line 122:
```suggestion
            
            notification = one(
```

test/indexer/test_indexer_controller.py
```diff
@@ -89,12 +91,43 @@ def test_invalid_notification(self):
...
+                self.queue_manager.read_messages(self.controller._notifications_queue)
+            )
+            expected_source = {
+                "id": source.id,
```
hannes-ucsc, line 126: Q

test/indexer/test_indexer_controller.py
```diff
@@ -89,12 +91,43 @@ def test_invalid_notification(self):
...
+            self.assertEqual(expected_notification, notification)
+            events = [self._mock_sqs_record(**notification)]
+
+            bundle_fqids = [
+                SourcedBundleFQID(source=source,
+                                  uuid='ffa338fe-7554-4b5d-96a2-7df127a7640b',
+                                  version='2018-03-28T151023.074974Z')
+            ]
+            self.controller.repository_plugin(self.catalog)._assert_source(source)
+            mock_list_bundles = mock.patch('azul.plugins.repository.dss.Plugin.list_bundles',
```
hannes-ucsc, line 118: Only one use of variable. Inline or assert calls on it.

test/indexer/test_tdr.py
```diff
@@ -78,11 +78,8 @@ class TestTDRPlugin(CannedBundleTestCase):
     bundle_uuid = '1b6d8348-d6e9-406a-aa6a-7ee886e52bf9'
 
     mock_service_url = 'https://azul_tdr_service_url_testing.org'
-
     source = TDRSourceRef(id='test_id',
-                          spec=TDRSourceSpec(project='test_project',
-                                             name='snapshot',
-                                             is_snapshot=True))
+                          spec=TDRSourceSpec.parse('tdr:test_project:snapshot/snapshot:').effective)
```
hannes-ucsc, line 82: Why don't we hard-code the prefix length here? It seems that using .effective here introduces unnecessary variability.

hannes-ucsc (changes requested):

src/azul/uuids.py
amarjandu: When validating UUID prefixes that start to specify values into the

src/azul/indexer/__init__.py
```diff
@@ -170,13 +285,29 @@ def __new__(cls: Type[SOURCE_REF], *, id: str, spec: SOURCE_SPEC) -> SOURCE_REF:
...
-        ...
-        azul.RequirementError: ('Ambiguous source specs for same ID.',
-                                SimpleSourceSpec(prefix='', name='a'),
-                                SimpleSourceSpec(prefix='', name='b'),
-                                '1')
+        S(id='1', spec=SimpleSourceSpec(prefix=Prefix(common='',
+                                                      partition=None),
+                                        name='b'))
+
+        Two specs with same name and id but different prefix string
```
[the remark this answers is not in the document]
amarjandu: Removing these, these tests don't add any coverage for edge cases now that SourceRefs with the same IDs but different Specs are permitted. These were added in when I was ignoring the Prefix.partition_prefix_length when comparing SourceSpecs....

src/azul/uuids.py
hannes-ucsc: `8f538f53/0` should be considered valid. `8f538f53/1` should not. `8f538f5/1` should be considered valid. `8f538f5/2` should not. And so on. Any combination of common prefix and partition prefix length should be validated against that rule and invalid combinations should be rejected as early as possible, so ideally not so late as when the individual partitions are actually enumerated during a reindex.

test/indexer/test_tdr.py
amarjandu: Note I added another commit to this PR `Change subclass for CannedBundleTestCase`.

scripts/reindex.py
```diff
@@ -50,16 +50,17 @@
...
+                    help='Do not offload the listing of subgraphs to the lambdas. When this option is used, this '
+                         'script queries the repository without partitioning, and the indexer notification endpoint is '
+                         'invoked for each subgraph individually and concurrently using worker threads. This is '
+                         'magnitudes slower than remote (i.e. partitioned) indexing. If the flag is not set, the set '
+                         'of subgraphs matching the query is partitioned using the partition prefix length(s) '
+                         'configured for the catalog sources(s). Each query partition is processed independently and '
+                         'remotely by the indexer lambda. The lambda queries the repository and queues a notification '
+                         'for each matching subgraph.')
```
hannes-ucsc, line 63:
```suggestion
                    help='Do not offload the listing of subgraphs to the indexer Lambda function. When this option is used, this '
                         'script queries the repository without partitioning, and the indexer notification endpoint is '
                         'invoked for each subgraph individually and concurrently using worker threads. This is '
                         'magnitudes slower than remote (i.e. partitioned) indexing. If this option is not used (the default), the set '
                         'of subgraphs matching the query is partitioned using the partition prefix length '
                         'configured for each of the catalog sources being reindexed. Each query partition is processed independently and '
                         'remotely by the indexer lambda. The index Lambda function queries the repository for each partition and queues a notification '
                         'for each matching subgraph in the partition.')
```

environment.py
```diff
@@ -372,11 +382,13 @@ def env() -> Mapping[str, Optional[str]]:
         #
         # sources = source (',', source )* ;
         #
-        # source = GitHub URL ;
+        # source = GitHub URL: ,
```
hannes-ucsc, line 385: The colon should be an EBNF literal aka terminal symbol.

environment.py
```diff
@@ -395,11 +407,12 @@ def env() -> Mapping[str, Optional[str]]:
         #
         # source = 'tdr:', Google Cloud project name,
         #          ':', ( 'dataset' | 'snapshot' ),
-        #          '/', 'TDR dataset or snapshot name' ;
+        #          '/', 'TDR dataset or snapshot name',
```
hannes-ucsc, line 410:
```suggestion
        #          '/', TDR dataset or snapshot name,
```

src/azul/indexer/__init__.py
```diff
@@ -94,6 +97,12 @@ def parse(cls, prefix: str) -> 'Prefix':
         Traceback (most recent call last):
         ...
         azul.RequirementError: ('Prefix source cannot end in a delimiter.', 'aa/', '/')
+
+        >>> Prefix.parse('8f538f53/1').partition_prefixes() # doctest: +NORMALIZE_WHITESPACE
```
hannes-ucsc, line 101: Add case `8f538f53/0` for contrast.

src/azul/indexer/__init__.py
```diff
@@ -77,6 +78,8 @@ def __attrs_post_init__(self):
         assert ':' not in self.common, self.common
         if self.partition:
             assert isinstance(self.partition, int), self.partition
+            require(len(self.common) + self.partition <= 8,
```
hannes-ucsc, line 81: Explain why `8` in a comment

hannes-ucsc:
```suggestion
            reject(len(self.common) + self.partition > 8,
```

hannes-ucsc (changes requested):

hannes-ucsc (approved):

noah-aviel-dove:
PR checklist is out of date, following the steps in the current version

noah-aviel-dove:
Need to update requirements since the `Makefile` was touched. The build passed on GitHub but failed a few minutes later on GitLab so I don't think PR author could have caught this.

noah-aviel-dove:
Assigning Hannes for review the updated commit history

hannes-ucsc:
> Needed to update requirements since the Makefile was touched. The build passed on GitHub but failed a few minutes later on GitLab so I don't think PR author could have caught this.

The checklist item reads 

> Ran make requirements_update or this PR leaves requirements*.txt, common.mk and Makefile untouched

so, yes, the author could and should have caught this. @amarjandu 

Aside from that, the GH build does not check for the need to update requirements, only the GL build does.

hannes-ucsc (approved):

amarjandu:
@hannes-ucsc  @noah-aviel-dove The pipeline failure due to the not running `make requirements_update` is my fault. Sorry for the inconvenience @noah-aviel-dove, I know it can be annoying to deal with when being the operator.

amarjandu:
Self Note: need to talk to @hannes-ucsc about workflow for new fixups, currently using `git restore --staged` to be able to commit non-conflicting changes disables PyCharms conflict resolution tooling, it works for 1 file, might be painful if there are lots of changes.

Commit hash is 51b628adc01c1b6224e87d0db1c95063d53809f1 try to rebase onto github/develop.

hannes-ucsc (approved):

jessebrennan:
This [passed in `sandbox`](https://gitlab.dev.singlecell.gi.ucsc.edu/ucsc/azul/-/pipelines/7524), but I updated a commit title that was missing the `[u]` tag after the build passed.

Issue #2953, opened 2021-04-07
Invoke TDR's list snapshots endpoint and cache the response as necessary


theathorn:
For every incoming Azul request call the snapshots endpoint and make that response available globally for subsequent requests to use.

theathorn:
Spike to determine latency and throughput of list snapshots endpoint to determine caching location (memory or persistent).

noah-aviel-dove:
> * if endpoint can be invoked anonymously

The endpoint cannot be accessed without authenticating.

noah-aviel-dove:
> * latency and throughput of list snapshots endpoint

By default, the `/api/repository/v1/snapshots` endpoint limits its response to 10 items. A higher limit can be specified, but there is no documented way to avoid the limit entirely (using -1 gives a 400 response). When the limit is set high enough to retrieve all accessible snapshots* in a single round trip, the average latency is 254 milliseconds (min 130, max 459, n=50). The API specifies that the limit is a 32 bit integer and the maximum accepted limit appears to be 2**31-1 (any higher gives a 400 response).

*When I hit the endpoint using the swagger API and authenticated using my TDR burner account, the endpoint lists 66 snapshots, but when using `azul.terra.TDRClient` it only lists 22. Probably due to different access permissions? Latency data is for the 22 snapshot response.

noah-aviel-dove:
Should use page size of 100, fail if there are more than 100 snapshots

noah-aviel-dove:
https://github.com/DataBiosphere/azul/issues/2978

noah-aviel-dove:
1. Is there a significant difference in latency for a single request to DynamoDB vs. S3?
2. Is there a significant difference in cost between those requests?
3. Is there a significant difference in cost for storing small objects between the two?

noah-aviel-dove:
1. DynamoDB is substantially faster:
```
2021-04-14 00:50:32,120 INFO    MainThread: It took 0.504s for cache miss in 'S3'
2021-04-14 00:50:32,351 INFO    MainThread: It took 0.231s for cache set in 'S3'
2021-04-14 00:50:32,636 INFO    MainThread: It took 0.285s for cache hit in 'S3'
2021-04-14 00:50:33,065 INFO    MainThread: It took 0.428s for cache miss in 'DynamoDB'
2021-04-14 00:50:33,170 INFO    MainThread: It took 0.105s for cache set in 'DynamoDB'
[...]

Issue #2978, opened 2021-04-13
Determine set of publicly accessible snapshots via TDR


noah-aviel-dove:
When a user is signed in, we will hit TDR on their behalf to determine the snapshots they have access to. This will include publicly accessible snapshots and the access-controlled ones the user has been granted access to. When the user is not signed in (anonymous) they should still see al publicly accessible snapshots. Unfortunately, it is not possible to list snapshots without authenticating to TDR so we need to find another way. One possible solution is to create another service account that represents anonymous Azul users and ask the TDR team to assign anonymously accessible snapshots to a group of which said service account is a member of.

Another is to hard-code the IDs of those snapshots.

Edit: Based on the discussion below we'll go with a second SA

```diff
Index: environment.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/environment.py b/environment.py
--- a/environment.py	(revision 80ac68ed80c3b9f2be72aca4e5eabff4f4990552)
+++ b/environment.py	(date 1619632813991)
@@ -278,10 +278,15 @@
         'AZUL_DEPLOYMENT_INCARNATION': '0',
 
         # The name of the Google Cloud service account to be created and used
-        # in conjunction with DSS subscriptions. If unset, a canonical resource
-        # name will be used. That default allows one such account per Azul
-        # deployment and Google Cloud project.
-        'AZUL_GOOGLE_SERVICE_ACCOUNT': 'azul-ucsc-{AZUL_DEPLOYMENT_INCARNATION}-{AZUL_DEPLOYMENT_STAGE}',
+        # for accessing all (meta)data in Google-based repositories.
+        #
+        'AZUL_SERVICE_ACCOUNT': 'azul-main{AZUL_DEPLOYMENT_INCARNATION}-{AZUL_DEPLOYMENT_STAGE}',
+
+        # The name of the Google Cloud service account to be created and used
+        # for accessing public (not access-controlled) (meta)data in Google-
+        # based repositories anonymously i.e., without authentication.
+        #
+        'AZUL_PROXY_SERVICE_ACCOUNT': 'azul-proxy{AZUL_DEPLOYMENT_INCARNATION}-{AZUL_DEPLOYMENT_STAGE}',
 
         # The number of concurrently running indexer lambda executions. Chalice
         # creates one Lambda function for handling HTTP requests from API Gateway

```

theathorn:
[...]

PR #3023, opened 2021-04-29
List and cache repository sources (#2953)

noah-aviel-dove:
Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Sanity-checked history
- [x] Branch pushed to Gitlab                               <sub>or PR is labeled `no sandbox`</sub>
- [x] Build passes in sandbox and added `sandbox` label     <sub>or PR is labeled `no sandbox`</sub>

Operator (reindex) 

- [ ] Started reindex in `dev`                              <sub>or this PR does not require reindexing or does not target `dev`</sub>
- [ ] Checked for failures in `dev`                         <sub>or this PR does not require reindexing or does not target `dev`</sub>
- [ ] Started reindex in `prod`                             <sub>or this PR does not require reindexing or does not target `prod`</sub>
- [ ] Checked for failures in `prod`                        <sub>or this PR does not require reindexing or does not target `prod`</sub>

Operator

- [ ] Unassigned PR

### DIFF
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -16,7 +16,6 @@ from typing import (
 )
 import urllib.parse
 
-import attr
 from botocore.exceptions import (
     ClientError,
 )
@@ -46,8 +45,10 @@ from azul import (
     config,
     drs,
 )
+from azul.auth import (
+    OAuth2,
+)
 from azul.chalice import (
-    Authentication,
     AzulChaliceApp,
 )
 from azul.drs import (
@@ -388,13 +389,6 @@ class ServiceApp(AzulChaliceApp):
                     args=dict(catalog=catalog,
                               **params)).url
 
-    @attr.s(auto_attribs=True, frozen=True)
-    class OAuth2(Authentication):
-        access_token: str
-
-        def identity(self) -> str:
-            return self.access_token
-
     def _authenticate(self) -> Optional[OAuth2]:
         try:
             header = self.current_request.headers['Authorization']
@@ -407,7 +401,7 @@ class ServiceApp(AzulChaliceApp):
                 raise UnauthorizedError(header)
             else:
                 if auth_type.lower() == 'bearer':
-                    return self.OAuth2(auth_token)
+                    return OAuth2(auth_token)
                 else:
                     raise UnauthorizedError(header)
 
--- a/lambdas/service/app.py
+++ b/lambdas/service/app.py
@@ -1751,6 +1751,42 @@ def _repository_files(file_uuid: str, fetch: bool) -> MutableJSON:
                                                    headers=headers)
 
 
+@app.route('/repository/sources', methods=['GET'], cors=True, method_spec={
+    'summary': 'List available data sources',
+    'tags': ['Repository'],
+    'parameters': [catalog_param_spec],
+    'responses': {
+        '200': {
+            'description': format_description('''
+                List the sources the currently authenticated user is authorized
+                to access in the underlying data repository.
+            '''),
+            **responses.json_content(
+                schema.object(sources=schema.array(
+                    schema.object(
+                        sourceId=str,
+                        sourceSpec=str,
+                    )
+                ))
+            )
+        },
+        # FIXME: Determine public snapshots
+        #        https://github.com/DataBiosphere/azul/issues/2978
+        '401': {
+            'description': format_description('''
+                This endpoint requires authentication.
+            ''')
+        }
+    }
+})
+def list_sources() -> Response:
+    validate_params(app.current_request.query_params or {},
+                    catalog=validate_catalog)
+    sources = app.repository_controller.list_sources(app.catalog,
+                                                     app.current_request)
+    return Response(body={'sources': sources}, status_code=200)
+
+
 @app.route('/url', methods=['POST'], cors=True)
 def shorten_query_url():
     """
--- a/src/azul/chalice.py
+++ b/src/azul/chalice.py
@@ -1,4 +1,3 @@
-import abc
 import json
 from json import (
     JSONEncoder,
@@ -26,6 +25,9 @@ from furl import (
 from azul import (
     config,
 )
+from azul.auth import (
+    Authentication,
+)
 from azul.json import (
     copy_json,
     json_head,
@@ -37,17 +39,6 @@ from azul.types import (
 )
 
 log = logging.getLogger(__name__)
-
-
-class Authentication(abc.ABC):
-
-    @abc.abstractmethod
-    def identity(self) -> str:
-        """
-        A string uniquely identifying the authenticated entity, for at least
-        some period of time.
-        """
-        raise NotImplementedError
 
 
 class AzulRequest(Request):
--- a/src/azul/hmac.py
+++ b/src/azul/hmac.py
@@ -3,7 +3,6 @@ from typing import (
     Optional,
 )
 
-import attr
 import chalice
 import requests
 from requests_http_signature import (
@@ -13,8 +12,8 @@ from requests_http_signature import (
 from azul import (
     require,
 )
-from azul.chalice import (
-    Authentication,
+from azul.auth import (
+    HMACAuthentication,
 )
 from azul.deployment import (
     aws,
@@ -23,26 +22,19 @@ from azul.deployment import (
 logger = logging.getLogger(__name__)
 
 
-@attr.s(auto_attribs=True, frozen=True)
-class HMACAuthentication(Authentication):
-    key_id: str
-
-    def identity(self) -> str:
-        return self.key_id
-
-    @classmethod
-    def from_request(cls, request: chalice.app.Request) -> Optional[Authentication]:
-        try:
-            header = request.headers['Authorization']
-        except KeyError:
-            return None
+def auth_from_request(request: chalice.app.Request
+                      ) -> Optional[HMACAuthentication]:
+    try:
+        header = request.headers['Authorization']
+    except KeyError:
+        return None
+    else:
+        prefix = 'Signature '
+        if header.startswith(prefix):
+            key_id = verify(request)
+            return HMACAuthentication(key_id)
         else:
-            prefix = 'Signature '
-            if header.startswith(prefix):
-                key_id = verify(request)
-                return cls(key_id)
-            else:
-                raise chalice.UnauthorizedError(header)
+            raise chalice.UnauthorizedError(header)
 
 
 def verify(current_request: chalice.app.Request) -> str:
--- a/src/azul/indexer/__init__.py
+++ b/src/azul/indexer/__init__.py
@@ -16,9 +16,13 @@ from typing import (
     Tuple,
     Type,
     TypeVar,
+    get_args,
 )
 
 import attr
+from more_itertools import (
+    one,
+)
 
 from azul import (
     reject,
@@ -201,6 +205,16 @@ class SourceRef(Generic[SOURCE_SPEC, SOURCE_REF]):
 
     def to_json(self):
         return dict(id=self.id, spec=str(self.spec))
+
+    @classmethod
+    def from_json(cls, ref: JSON) -> 'SourceRef':
+        return cls(spec=cls.spec_cls().parse(ref['spec']), id=ref['id'])
+
+    @classmethod
+    def spec_cls(cls) -> Type[SourceSpec]:
+        base_cls = one(getattr(cls, '__orig_bases__'))
+        spec_cls, ref_cls = get_args(base_cls)
+        return spec_cls
 
 
 @attr.s(auto_attribs=True, frozen=True, kw_only=True, order=True)
--- a/src/azul/plugins/__init__.py
+++ b/src/azul/plugins/__init__.py
@@ -29,8 +29,12 @@ from more_itertools import (
 
 from azul import (
     CatalogName,
+    cached_property,
     config,
     require,
+)
+from azul.chalice import (
+    Authentication,
 )
 from azul.drs import (
     DRSClient,
@@ -191,26 +195,53 @@ class RepositoryPlugin(Generic[SOURCE_SPEC, SOURCE_REF], Plugin):
         """
         raise NotImplementedError
 
-    def resolve_source(self, *, spec: str, id: Optional[str] = None) -> SOURCE_REF:
+    @abstractmethod
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> Iterable[SOURCE_REF]:
+        """
+        The sources the plugin is configured to read metadata from.
+        Retrieving this information may require a round-trp to the underlying
+        repository. Implementations should raise PermissionError if the provided
+        authentication is insufficient to access the repository.
+        """
+        raise NotImplementedError
+
+    @cached_property
+    def _source_ref_cls(self) -> Type[SOURCE_REF]:
+        cls = type(self)
+        base_cls = one(getattr(cls, '__orig_bases__'))
+        spec_cls, ref_cls = get_args(base_cls)
+        require(issubclass(spec_cls, SourceSpec))
+        require(issubclass(ref_cls, SourceRef))
+        assert ref_cls.spec_cls() is spec_cls
+        return ref_cls
+
+    def source_from_json(self, ref: JSON) -> SOURCE_REF:
+        """
+        Instantiate a :class:`SourceRef` from its JSON representation. The
+        expected input format matches the output format of `SourceRef.to_json`.
+        """
+        return self._source_ref_cls.from_json(ref)
+
+    def resolve_source(self, spec: str) -> SOURCE_REF:
         """
         Return an instance of :class:`SourceRef` for the repository source
         matching the given specification or raise an exception if no such source
-        exists. If an ID is given, ensure that the source matching the
-        specification has the given ID.
+        exists.
         """
-        cls = type(self)
-        base_cls = one(getattr(cls, '__orig_bases__'))
-        source_spec_cls, source_ref_cls = get_args(base_cls)
-        require(issubclass(source_spec_cls, SourceSpec))
-        require(issubclass(source_ref_cls, SourceRef))
-        spec = source_spec_cls.parse(spec)
-        actual_id = self.lookup_source_id(spec)
-        if id is None:
-            id = actual_id
-        else:
-            require(id == actual_id,
-                    'Source ID changed unexpectedly', spec, id, actual_id)
-        return source_ref_cls(id=id, spec=spec)
+        ref_cls = self._source_ref_cls
+        spec = ref_cls.spec_cls().parse(spec)
+        id = self.lookup_source_id(spec)
+        return ref_cls(id=id, spec=spec)
+
+    def verify_source(self, ref: SOURCE_REF) -> None:
+        """
+        Verify that the source's ID matches that defined in the
+        repository for the source's spec.
+        """
+        actual_id = self.lookup_source_id(ref.spec)
+        require(ref.id == actual_id, 'Source ID changed unexpectedly', ref, actual_id)
 
     @abstractmethod
     def lookup_source_id(self, spec: SOURCE_SPEC) -> str:
--- a/src/azul/plugins/repository/dss/__init__.py
+++ b/src/azul/plugins/repository/dss/__init__.py
@@ -36,6 +36,9 @@ from azul import (
     cached_property,
     config,
 )
+from azul.auth import (
+    Authentication,
+)
 from azul.collections import (
     adict,
 )
@@ -103,6 +106,13 @@ class Plugin(RepositoryPlugin[SimpleSourceSpec, DSSSourceRef]):
 
     def lookup_source_id(self, spec: SimpleSourceSpec) -> str:
         return DSSSourceRef.id_from_spec(spec)
+
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> List[DSSSourceRef]:
+        # FIXME: De-conflate str and SimpleSourceName in dss.Plugin.sources
+        #        https://github.com/DataBiosphere/azul/issues/3155
+        return []
 
     @cached_property
     def dss_client(self):
--- a/src/azul/plugins/repository/tdr/__init__.py
+++ b/src/azul/plugins/repository/tdr/__init__.py
@@ -48,6 +48,10 @@ from azul import (
     reject,
     require,
 )
+from azul.auth import (
+    Authentication,
+    OAuth2,
+)
 from azul.bigquery import (
     BigQueryRow,
     BigQueryRows,
@@ -174,6 +178,30 @@ class Plugin(RepositoryPlugin[TDRSourceSpec, TDRSourceRef]):
     @property
     def sources(self) -> AbstractSet[TDRSourceSpec]:
         return self._sources
+
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> List[TDRSourceRef]:
+        if isinstance(authentication, OAuth2):
+            tdr = TDRClient.with_user_credentials(authentication)
+            configured_specs_by_name = {spec.name: spec for spec in self.sources}
+            snapshot_ids_by_name = {
+                name: id
+                for id, name in tdr.snapshot_names_by_id().items()
+                if name in configured_specs_by_name
+            }
+            return [
+                TDRSourceRef(id=id,
+                             spec=configured_specs_by_name[name])
+                for name, id in snapshot_ids_by_name.items()
+            ]
+        elif authentication is None:
+            # FIXME: Determine public snapshots
+            #        https://github.com/DataBiosphere/azul/issues/2978
+            raise PermissionError('Authentication required')
+        else:
+            raise PermissionError('Unsupported authentication format',
+                                  type(authentication))
 
     @property
     def tdr(self):
--- a/src/azul/plugins/repository/tdr/__init__.py
+++ b/src/azul/plugins/repository/tdr/__init__.py
@@ -76,7 +76,6 @@ from azul.plugins import (
 from azul.terra import (
     TDRClient,
     TDRSourceSpec,
-    TerraDRSClient,
 )
 from azul.types import (
     JSON,
@@ -195,7 +194,7 @@ class Plugin(RepositoryPlugin[TDRSourceSpec, TDRSourceRef]):
     @classmethod
     @cache_per_thread
     def _tdr(cls):
-        return TDRClient()
+        return TDRClient.with_service_account_credentials()
 
     def _assert_source(self, source: TDRSourceRef):
         assert source.spec in self.sources, (source, self.sources)
@@ -464,7 +463,7 @@ class Plugin(RepositoryPlugin[TDRSourceSpec, TDRSourceRef]):
             return root
 
     def drs_client(self) -> DRSClient:
-        return TerraDRSClient()
+        return self.tdr.drs_client()
 
     def file_download_class(self) -> Type[RepositoryFileDownload]:
         return TDRFileDownload
--- a/src/azul/service/repository_controller.py
+++ b/src/azul/service/repository_controller.py
@@ -2,6 +2,7 @@ import json
 import logging
 import time
 from typing import (
+    Iterable,
     Mapping,
     Optional,
     Sequence,
@@ -11,6 +12,7 @@ from typing import (
 from chalice import (
     BadRequestError,
     NotFoundError,
+    UnauthorizedError,
 )
 
 from azul import (
@@ -19,6 +21,12 @@ from azul import (
     cached_property,
     config,
     reject,
+)
+from azul.chalice import (
+    AzulRequest,
+)
+from azul.indexer import (
+    SourceRef,
 )
 from azul.plugins import (
     RepositoryPlugin,
@@ -29,6 +37,13 @@ from azul.service import (
 from azul.service.index_query_service import (
     IndexQueryService,
 )
+from azul.service.source_cache_service import (
+    CacheMiss,
+    SourceCacheService,
+)
+from azul.types import (
+    JSONs,
+)
 
 log = logging.getLogger(__name__)
 
@@ -38,6 +53,10 @@ class RepositoryController(Controller):
     @cached_property
     def service(self):
         return IndexQueryService()
+
+    @cached_property
+    def _source_cache_service(self) -> SourceCacheService:
+        return SourceCacheService()
 
     @classmethod
     @cache
@@ -224,3 +243,41 @@ class RepositoryController(Controller):
             }
         else:
             assert False
+
+    def list_sources(self,
+                     catalog: CatalogName,
+                     request: AzulRequest,
+                     ) -> JSONs:
+        authentication = request.authentication
+        plugin = self.repository_plugin(catalog)
+
+        def list_sources() -> Iterable[SourceRef]:
+            try:
+                return plugin.list_sources(authentication)
+            except PermissionError:
+                raise UnauthorizedError
+
+        if authentication is None:
+            # FIXME: Determine public snapshots
+            #        https://github.com/DataBiosphere/azul/issues/2978
+            sources = list_sources()
+        else:
+            cache_key = f'{catalog}:{authentication.identity()}'
+            try:
+                source_jsons = self._source_cache_service.get(cache_key)
+            except CacheMiss:
+                sources = list_sources()
+                source_jsons = [source.to_json() for source in sources]
+                self._source_cache_service.put(cache_key, source_jsons)
+            else:
+                sources = [
+                    plugin.source_from_json(source_json)
+                    for source_json in source_jsons
+                ]
+        return [
+            {
+                'sourceId': source.id,
+                'sourceSpec': str(source.spec)
+            }
+            for source in sources
+        ]
--- a/src/azul/terra.py
+++ b/src/azul/terra.py
@@ -8,10 +8,15 @@ from time import (
     sleep,
 )
 from typing import (
+    Dict,
     Sequence,
+    Union,
 )
 
 import attr
+from chalice import (
+    UnauthorizedError,
+)
 from furl import (
     furl,
 )
@@ -33,8 +38,11 @@ from google.cloud.bigquery import (
 from google.cloud.bigquery.table import (
     TableListItem,
 )
+from google.oauth2.credentials import (
+    Credentials as TokenCredentials,
+)
 from google.oauth2.service_account import (
-    Credentials,
+    Credentials as ServiceAccountCredentials,
 )
 from more_itertools import (
     one,
@@ -47,6 +55,9 @@ from azul import (
     cached_property,
     config,
     require,
+)
+from azul.auth import (
+    OAuth2,
 )
 from azul.bigquery import (
     BigQueryRows,
@@ -75,6 +86,8 @@ from azul.uuids import (
 )
 
 log = logging.getLogger(__name__)
+
+Credentials = Union[ServiceAccountCredentials, TokenCredentials]
 
 
 @attr.s(frozen=True, auto_attribs=True, kw_only=True)
@@ -193,15 +206,16 @@ class ServiceAccountCredentialsProvider(CredentialsProvider):
     @cache
     def credentials(self) -> Credentials:
         with aws.service_account_credentials() as file_name:
-            credentials = Credentials.from_service_account_file(file_name)
+            credentials = ServiceAccountCredentials.from_service_account_file(file_name)
         credentials = credentials.with_scopes(self.oauth2_scopes())
         credentials.refresh(Request())  # Obtain access token
         return credentials
 
     def identity(self) -> str:
         # When authenticated using the service account credentials, each
-        # instance of ServiceAccountCredentialsProvider obtains a fresh token, making the token
-        # inconsistent across lambda invocations and thus unsuitable as a key.
+        # instance of ServiceAccountCredentialsProvider obtains a fresh token,
+        # making the token inconsistent across lambda invocations and thus
+        # unsuitable as a key.
         return self.credentials().service_account_email
 
     def insufficient_access(self, resource: str):
@@ -210,6 +224,33 @@ class ServiceAccountCredentialsProvider(CredentialsProvider):
             f'authorized to access {resource} or that resource does not exist. Make sure '
             f'that it exists, that the SA is registered with SAM and has been granted read '
             f'access to the resource.'
+        )
+
+
+class UserCredentialsProvider(CredentialsProvider):
+
+    def __init__(self, token: OAuth2):
+        self.token = token
+
+    def oauth2_scopes(self) -> Sequence[str]:
+        return ['email']
+
+    @cache
+    def credentials(self) -> TokenCredentials:
+        # FIXME: this assumes the user has selected all required scopes.
+        return TokenCredentials(self.token.identity(), scopes=self.oauth2_scopes())
+
+    def identity(self) -> str:
+        return self.token.identity()
+
+    def insufficient_access(self, resource: str):
+        scopes = ', '.join(self.oauth2_scopes())
+        return UnauthorizedError(
+            f'The current user is not authorized to access {resource} or that '
+            f'resource does not exist. Make sure that it exists, that the user '
+            f'is registered with Terra, that the provided access token is not '
+            f'expired, and that the following access scopes were granted when '
+            f'authenticating: {scopes}.'
         )
 
 
@@ -229,7 +270,14 @@ class TerraClient:
         """
         A urllib3 HTTP client with OAuth 2.0 credentials.
         """
-        return AuthorizedHttp(self.credentials, http_client())
+        # By default, AuthorizedHTTP attempts to refresh the credentials on a 401
+        # response, which is never helpful. When using service account
+        # credentials, a fresh token is obtained for every lambda invocation,
+        # which will never persist long enough for the token to expire. User
+        # tokens can expire, but attempting to refresh them raises
+        # `google.auth.exceptions.RefreshError` due to the credentials not being
+        # configured with (among other fields) the client secret.
+        return AuthorizedHttp(self.credentials, http_client(), refresh_status_codes=())
 
     def _request(self,
                  method,
@@ -417,9 +465,29 @@ class TDRClient(SAMClient):
         else:
             raise RequirementError('Unexpected response from TDR API', response.status)
 
+    def snapshot_names_by_id(self) -> Dict[str, str]:
+        """
+        List the TDR snapshots accessible to the current credentials.
+        """
+        limit = 100
+        endpoint = self._repository_endpoint('snapshots')
+        response = self._request('GET', endpoint, fields=dict(limit=str(limit)))
+        response = self._check_response(endpoint, response)
+        # FIXME: Implement paging
+        #        https://github.com/DataBiosphere/azul/issues/3093
+        require(response['total'] <= limit, 'Too many snapshots', response['total'])
+        return {
+            snapshot['id']: snapshot['name']
+            for snapshot in response['items']
+        }
+
     @classmethod
     def with_service_account_credentials(cls) -> 'TDRClient':
         return cls(credentials_provider=ServiceAccountCredentialsProvider())
+
+    @classmethod
+    def with_user_credentials(cls, token: OAuth2) -> 'TDRClient':
+        return cls(credentials_provider=UserCredentialsProvider(token))
 
     def drs_client(self):
         return DRSClient(http_client=self._http_client)
--- a/src/azul/terra.py
+++ b/src/azul/terra.py
@@ -61,6 +61,7 @@ from azul.strings import (
 )
 from azul.types import (
     JSON,
+    MutableJSON,
 )
 from azul.uuids import (
     validate_uuid_prefix,
@@ -269,30 +270,25 @@ class TDRClient(SAMClient):
         endpoint = self._repository_endpoint(tdr_path)
         params = dict(filter=source.bq_name, limit='2')
         response = self._request('GET', endpoint, fields=params)
-        if response.status == 200:
-            response = json.loads(response.data)
-            try:
-                # FIXME: Once filteredTotal is deployed in Terra prod, we can
-                #        remove this try/except block.
-                #        https://github.com/DataBiosphere/azul/issues/3195
-                total = response['filteredTotal']
-            except KeyError:
-                total = response['total']
-            if total == 0:
-                raise self._insufficient_access(resource)
-            elif total == 1:
-                snapshot_id = one(response['items'])['id']
-                endpoint = self._repository_endpoint(tdr_path, snapshot_id)
-                response = self._request('GET', endpoint)
-                require(response.status == 200,
-                        f'Failed to access {resource} after resolving its ID to {snapshot_id!r}')
-                return json.loads(response.data)
-            else:
-                raise RequirementError('Ambiguous response from TDR API', endpoint, response)
-        elif response.status == 401:
-            raise self._insufficient_access(endpoint)
+        response = self._check_response(endpoint, response)
+        try:
+            # FIXME: Once filteredTotal is deployed in Terra prod, we can
+            #        remove this try/except block.
+            #        https://github.com/DataBiosphere/azul/issues/3195
+            total = response['filteredTotal']
+        except KeyError:
+            total = response['total']
+        if total == 0:
+            raise self._insufficient_access(resource)
+        elif total == 1:
+            snapshot_id = one(response['items'])['id']
+            endpoint = self._repository_endpoint(tdr_path, snapshot_id)
+            response = self._request('GET', endpoint)
+            require(response.status == 200,
+                    f'Failed to access {resource} after resolving its ID to {snapshot_id!r}')
+            return json.loads(response.data)
         else:
-            raise RequirementError('Unexpected response from TDR API', response.status)
+            raise RequirementError('Ambiguous response from TDR API', endpoint, response)
 
     def check_bigquery_access(self, source: TDRSourceSpec):
         """
@@ -360,6 +356,17 @@ class TDRClient(SAMClient):
     def _repository_endpoint(self, *path: str) -> str:
         return furl(config.tdr_service_url,
                     path=('api', 'repository', 'v1', *path)).url
+
+    def _check_response(self,
+                        endpoint: str,
+                        response: urllib3.HTTPResponse
+                        ) -> MutableJSON:
+        if response.status == 200:
+            return json.loads(response.data)
+        elif response.status == 401:
+            raise self._insufficient_access(endpoint)
+        else:
+            raise RequirementError('Unexpected response from TDR API', response.status)
 
 
 class TerraDRSClient(DRSClient, TerraClient):
--- a/src/azul/terra.py
+++ b/src/azul/terra.py
@@ -157,11 +157,15 @@ class TerraClient:
     @cached_property
     def credentials(self) -> Credentials:
         with aws.service_account_credentials() as file_name:
-            return Credentials.from_service_account_file(file_name)
+            credentials = Credentials.from_service_account_file(file_name)
+        credentials = credentials.with_scopes(self.oauth2_scopes)
+        credentials.refresh(Request())  # Obtain access token
+        return credentials
 
     oauth2_scopes = [
         'email',
         'openid',
+        'https://www.googleapis.com/auth/bigquery.readonly',
         'https://www.googleapis.com/auth/devstorage.read_only'
     ]
 
@@ -170,8 +174,7 @@ class TerraClient:
         """
         A urllib3 HTTP client with OAuth 2.0 credentials.
         """
-        return AuthorizedHttp(self.credentials.with_scopes(self.oauth2_scopes),
-                              http_client())
+        return AuthorizedHttp(self.credentials, http_client())
 
     def _request(self,
                  method,
--- a/src/azul/terra.py
+++ b/src/azul/terra.py
@@ -1,7 +1,14 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
 import json
 import logging
 from time import (
     sleep,
+)
+from typing import (
+    Sequence,
 )
 
 import attr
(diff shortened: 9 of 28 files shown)


src/azul/terra.py
```diff
@@ -348,6 +358,93 @@ def _repository_endpoint(self, *path: str) -> str:
...
+                        headers: JSON = None,
+                        params: JSON = None
+                        ) -> MutableJSON:
+        response = self._oauthed_request('GET', endpoint, headers=headers, fields=params)
+        if response.status == 200:
+            return json.loads(response.data)
+        elif response.status == 401:
+            raise self._insufficient_access(endpoint)
+        else:
+            raise RequirementError('Unexpected response from TDR API', response.status)
```
jessebrennan, line 372: It looks to me like the creation of this method is purely a refactoring, in which case maybe it should go in the previous commit.

test/indexer/test_tdr.py
```diff
@@ -206,5 +226,67 @@ def _full_table_name(self, source: TDRSourceName, table_name: str) -> str:
...
+    def _mock_list_snapshots(self, method, url, **kwargs):
+        self.assertEqual(method, 'GET')
+        self.assertEqual(furl(url).remove(query=True).url,
+                         TDRClient()._repository_endpoint('snapshots'))
+        headers = kwargs.get('headers')
+        self.assertIsNotNone(headers)
+        try:
+            token = (headers.get('Authorization') or headers['authorization']).split('Bearer ').pop()
+        except (KeyError, ValueError):
+            assert False, headers
```
jessebrennan, line 246: I don't understand why you except and then assert False. Why not remove the try block all together and let the test fail if an exception is raised?

jessebrennan (changes requested):
The name of the relevant commit does not match the PR title.

Also I think giving a more specific name to "Refactor `terra.py`" would improve the history.

Some options to consider: 
- "Optimize scope privileges for Terra clients"
- "Fix credential management for Terra clients"

but since there's several kinds of changes squished in there, maybe your simple title is a preferred common denominator.

Also It might be easier to go over some of the questions I asked over a call, since I lack some of the context for these changes. Please reach out after you've looked through them.

src/azul/chalice.py
```diff
@@ -181,6 +184,7 @@ def _register_spec(self,
 
     def _get_view_function_response(self, view_function, function_args):
         self._log_request()
+        self._cache_user_snapshots()
```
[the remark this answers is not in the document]
noah-aviel-dove: The ticket specified to cache the snapshots "For every incoming Azul request". I share your concerns regarding latency but I can't imagine another interpretation.

test/indexer/test_tdr.py
noah-aviel-dove: This is detritus left over from debugging test failures that I neglected to clean up before pushing.

src/azul/terra.py
noah-aviel-dove: True, but the refactoring was motivated by the addition of the `TDRClient.list_snapshots` method which would have otherwise had a lot of redundant code with `TDRClient._lookup_source`. The refactoring of the scopes and SAM auth make sense as a standalone code improvement but the motivation for this change wouldn't be clear without the implementation of `list_snapshots` so I think it should stay where it is.

src/azul/terra.py
```diff
@@ -348,6 +358,93 @@ def _repository_endpoint(self, *path: str) -> str:
...
+    ddb_value_name = 'snapshot_jsons'
+    ddb_ttl_name = 'token_ttl_seconds'
+
+    def list_snapshots(self) -> JSONs:
+        """
+        When authenticated using the service account credentials, each client
+        obtains a fresh token, making token-based caching ineffective between
+        lambda invocations. This can be addressed by using an externally
+        provided token instead, as is done here.
+        """
```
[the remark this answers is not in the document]
noah-aviel-dove: I've updated the comment but we should still discuss this further.

src/azul/terra.py
```diff
@@ -230,6 +225,14 @@ class TDRClient(SAMClient):
     A client for the Broad Institute's Terra Data Repository aka "Jade".
     """
 
+    @property
+    def credential_scopes(self):
+        return [
+            *super().credential_scopes,
+            'https://www.googleapis.com/auth/devstorage.read_only',
+            'https://www.googleapis.com/auth/bigquery.readonly'
```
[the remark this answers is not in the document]
noah-aviel-dove: We discussed and agreed this diff is fine but `bigquery.readonly` needs to be added to the list of all scopes in the previous commit.

noah-aviel-dove:
Note that the main commit (not the fixup) has been changed due to rebasing.

src/azul/terra.py
```diff
@@ -195,11 +199,6 @@ def _oauthed_request(self,
                       trunc_ellipses(response.data, 256))
         return response
 
-    def get_access_token(self) -> str:
-        credentials = self.credentials.with_scopes(self.oauth_scopes)
```
jessebrennan, line 192: Maybe I'm missing something here, but it looks like the old way didn't use the bigquery.readonly scope. If I recall correctly from the call, you said it used a more general scope, but it looks like that's not the case?

src/azul/terra.py
```diff
@@ -159,11 +159,15 @@ class TerraClient:
...
-            return Credentials.from_service_account_file(file_name)
+            credentials = Credentials.from_service_account_file(file_name)
+        credentials = credentials.with_scopes(self.credential_scopes)
+        credentials.refresh(Request())  # Obtain access token
+        return credentials
 
     oauth_scopes = [
         'email',
         'openid',
+        'https://www.googleapis.com/auth/bigquery.readonly',
```
jessebrennan: I tried deleting this line and running the repository proxy tests and it still passed. Maybe I'm running the wrong tests, but I'm thinking again this line should be deleted.

jessebrennan (changes requested):
The commit title "TerraClient subclasses use minimal scopes" is not in the imperative tense, but could be rephrased as "Use minimal scopes for TerraClient subclasses".

src/azul/terra.py
noah-aviel-dove: The unit tests load canned tables into TinyQuery and do not make any actual requests to BigQuery. The quickest way to verify that the client has sufficient authorizations is to run the `make sam` target:

```
$ git diff
diff --git a/src/azul/terra.py b/src/azul/terra.py
index 33315a3a..01df75d5 100644
--- a/src/azul/terra.py
+++ b/src/azul/terra.py
@@ -255,7 +255,6 @@ class TDRClient(SAMClient):
         return [
             *super().credential_scopes,
             'https://www.googleapis.com/auth/devstorage.read_only',
-            'https://www.googleapis.com/auth/bigquery.readonly'
         ]
 
     @cache
$ make sam
python /home/nadove/PycharmProjects/azul-sc/scripts/register_sam.py
2021-05-06 20:20:10,064 DEBUG   MainThread: _oauthed_request('POST', 'https://sam.dsde-dev.broadinstitute.org/register/user/v1', fields=None, headers=None, body='')
2021-05-06 20:20:10,085 INFO    MainThread: Found credentials in environment variables.
2021-05-06 20:20:11,132 DEBUG   MainThread: _oauthed_request(…) -> b'{"causes":[],"message":"user CreateWorkbenchUser(2620357612338d02b515c,116556158276818438280,azul-ucsc-4-nadove@platform-hca-dev.iam.gserviceaccount.com,None) already exists","source":"sam","stackTrace":[],"statusCode":409}'
2021-05-06 20:20:11,132 INFO    MainThread: Google service account previously registered with SAM.
2021-05-06 20:20:11,134 DEBUG   MainThread: _oauthed_request('GET', 'https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots', fields={'filter': 'lungmap_dev_20210412__20210414', 'limit': 2}, headers=None, body=None)
2021-05-06 20:20:11,590 DEBUG   MainThread: _oauthed_request(…) -> b'{"total":1,"items":[{"id":"215793b0-b991-4b73-9584-1103d4072f39","name":"lungmap_dev_20210412__20210414","description":"Create snapshot lungmap_dev_20210412__20210414","createdDate":"2021-04-14T19:45:16.972386Z","profileId":"390e7a85-d47f-4531-b612-165f...'
2021-05-06 20:20:11,591 DEBUG   MainThread: _oauthed_request('GET', 'https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots/215793b0-b991-4b73-9584-1103d4072f39', fields=None, headers=None, body=None)
2021-05-06 20:20:12,002 DEBUG   MainThread: _oauthed_request(…) -> b'{"id":"215793b0-b991-4b73-9584-1103d4072f39","name":"lungmap_dev_20210412__20210414","description":"Create snapshot lungmap_dev_20210412__20210414","createdDate":"2021-04-14T19:45:16.972386Z","source":[{"dataset":{"id":"10330a3a-d0e8-41c7-bc1a-2a5cb02c8...'
2021-05-06 20:20:12,003 DEBUG   MainThread: _oauthed_request('GET', 'https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots', fields={'filter': 'lungmap_dev_20210412__20210414', 'limit': 2}, headers=None, body=None)
2021-05-06 20:20:12,163 DEBUG   MainThread: _oauthed_request(…) -> b'{"total":1,"items":[{"id":"215793b0-b991-4b73-9584-1103d4072f39","name":"lungmap_dev_20210412__20210414","description":"Create snapshot lungmap_dev_20210412__20210414","createdDate":"2021-04-14T19:45:16.972386Z","profileId":"390e7a85-d47f-4531-b612-165f...'
2021-05-06 20:20:12,165 DEBUG   MainThread: _oauthed_request('GET', 'https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots/215793b0-b991-4b73-9584-1103d4072f39', fields=None, headers=None, body=None)
2021-05-06 20:20:12,617 DEBUG   MainThread: _oauthed_request(…) -> b'{"id":"215793b0-b991-4b73-9584-1103d4072f39","name":"lungmap_dev_20210412__20210414","description":"Create snapshot lungmap_dev_20210412__20210414","createdDate":"2021-04-14T19:45:16.972386Z","source":[{"dataset":{"id":"10330a3a-d0e8-41c7-bc1a-2a5cb02c8...'
2021-05-06 20:20:12,618 INFO    MainThread: TDR client is authorized for API access to tdr:broad-jade-dev-data:snapshot/lungmap_dev_20210412__20210414:.
  File "/home/nadove/PycharmProjects/azul-sc/src/azul/terra.py", line 319, in check_bigquery_access
    tables = list(bigquery.list_tables(source.bq_name, max_results=1))
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/api_core/page_iterator.py", line 212, in _items_iter
    for page in self._page_iter(increment=False):
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/api_core/page_iterator.py", line 243, in _page_iter
    page = self._next_page()
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/api_core/page_iterator.py", line 369, in _next_page
    response = self._get_next_page_response()
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/api_core/page_iterator.py", line 418, in _get_next_page_response
    return self.api_request(
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/cloud/bigquery/client.py", line 556, in _call_api
    return call()
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/api_core/retry.py", line 281, in retry_wrapped_func
    return retry_target(
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/api_core/retry.py", line 184, in retry_target
    return target()
  File "/home/nadove/PycharmProjects/azul-sc/.venv/lib/python3.8/site-packages/google/cloud/_http.py", line 483, in api_request
    raise exceptions.from_http_response(response)
google.api_core.exceptions.Forbidden: 403 GET https://bigquery.googleapis.com/bigquery/v2/projects/broad-jade-dev-data/datasets/lungmap_dev_20210412__20210414/tables?maxResults=1&prettyPrint=false: Request had insufficient authentication scopes.

During handling of the above exception, another exception occurred:

  File "/home/nadove/PycharmProjects/azul-sc/scripts/register_sam.py", line 43, in <module>
    main()
  File "/home/nadove/PycharmProjects/azul-sc/scripts/register_sam.py", line 39, in main
    tdr.check_bigquery_access(source)
  File "/home/nadove/PycharmProjects/azul-sc/src/azul/terra.py", line 330, in check_bigquery_access
    raise self._insufficient_access(resource)
azul.RequirementError: The service account (SA) 'azul-ucsc-4-nadove@platform-hca-dev.iam.gserviceaccount.com' is not authorized to access BigQuery dataset 'lungmap_dev_20210412__20210414' in Google Cloud project 'broad-jade-dev-data' or that resource does not exist. Make sure that it exists, that the SA is registered with SAM and has been granted read access to the resource.
Makefile:91: recipe for target 'sam' failed
make: *** [sam] Error 1
```

src/azul/terra.py
noah-aviel-dove: When the `Credentials` object is instantiated, it has no scopes, and is implicitly authorized for everything the Service Account has access to. The call to `with_scopes` reduces that authorization to the provided scopes. Note that `with_scopes` returns a new `Credentials` and does not mutate the original.

Prior to these changes, `with_scopes` was only called on the line you've selected, and only the local variable in `SAMClient.get_access_token` ever had its scopes limited. Thus, when `TDRClient` instantiated its `BigQuery` client, [here](https://github.com/DataBiosphere/azul/blob/d66e82be0c35510a4a865ae4599cfd85ec02e7d1/src/azul/terra.py#L314), it is providing it with the original, unscoped credentials, which implicitly contains the `bigquery.readonly` scope (because it has _no_ explicit scopes).

My changes move the call to `with_scopes` into the `credentials` property so that all `TerraClient` instances only ever have the credential scopes that they declare at all call sites. As a consequence, we must now declare certain scopes that previously were granted implicitly.

jessebrennan: I see. Thanks for explaining this so thoroughly!

noah-aviel-dove:
Re-requesting peer review post-approval since this has been rebased on some substantial changes to the upstream PR, which I'm confident is now close to landing.

src/azul/terra.py
jessebrennan: If I recall correctly you did end up splitting this out. Even still I just noticed a relevant section of the contributing guide so I thought to share it https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#34bloated-diffs

jessebrennan: I just noticed a relevant section of the contributing guide so I thought to share it https://github.com/DataBiosphere/azul/blob/develop/CONTRIBUTING.rst#34bloated-diffs.

I leave it to you to decide if it applies here.

If you don't want to split it up the changes, try moving the definition of `_check_response()` directly after where it is used. That may also be sufficient to minimize the diff.

src/azul/terra.py
```diff
@@ -239,6 +250,14 @@ class TDRClient(SAMClient):
     A client for the Broad Institute's Terra Data Repository aka "Jade".
     """
 
+    @property
+    def credential_scopes(self):
+        return [
+            *super().credential_scopes,
+            'https://www.googleapis.com/auth/devstorage.read_only',
+            'https://www.googleapis.com/auth/bigquery.readonly'
```
jessebrennan, line 258: You still need to change the title of the commit that makes these changes. I've left several comments about this already, but I suspect you've overlooked them.

jessebrennan (changes requested):
You still need to change the commit title.

src/azul/terra.py
noah-aviel-dove: Ack, sorry, thanks for being patient.

noah-aviel-dove:
Retracting review because this could be refactored to facilitate a much smaller diff fro #2978

noah-aviel-dove:
@jessebrennan I've decided to take a different approach since you last looked at this and I expect you'll have questions.

noah-aviel-dove:
Reassigning review since Jesse's OOO

src/azul/terra.py
```diff
@@ -353,8 +461,76 @@ def _repository_endpoint(self, *path: str) -> str:
...
+            log.info('Snapshot list cache is expired, %r', auth_key)
+            result = None
+        if result is None:
+            result = self._list_snapshots()
+            self._dynamodb.put_item(TableName=table_name,
+                                    Item={
+                                        self.ddb_key_name: {'S': auth_key},
+                                        self.ddb_value_name: {'S': json.dumps(result)},
+                                        self.ddb_ttl_name: {
+                                            'N': str(int(time()) + self.auth_provider.auth_key_lifespan())
```
amarjandu, line 499: This TTL might be too long, if we cache the snapshots and a user gets access right after we perform the caching, they have to wait for the the `auth_key_lifespan` in order to see the new snapshots they now have access to.

Might want to discuss this w/ @hannes-ucsc

amarjandu (changes requested):
Looks good, couple of questions. Also not sure if you know but @jessebrennan is back today. Unless the PR has drastic changes from what Jesse has seen in the past, there might be value in having him take a look at this as well.

src/azul/terra.py
```diff
@@ -353,8 +461,76 @@ def _repository_endpoint(self, *path: str) -> str:
...
+
+    def _list_snapshots(self) -> JSONs:
+        """
+        List the TDR snapshots accessible to the current credentials.
+        """
+        limit = 100
+        endpoint = self._repository_endpoint('snapshots')
+        response = self._check_response(endpoint,
+                                        params=dict(limit=limit))
+        require(response['total'] <= limit, 'Too many snapshots', response['total'])
```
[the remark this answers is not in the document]
noah-aviel-dove: This was the result of an old conversation with @hannes-ucsc. I don't remember the exact reasoning but it was related to endpoint throughput and latency. 

> Won't this become problematic if we use the result to assert if the user has access to a snapshot?

I don't understand why/when it would be appropriate to assert that.

terraform/dynamo.tf.json.template.py
```diff
@@ -23,6 +26,21 @@
                                 "type": "S"
                             }
                         ]
+                    },
+                    "tdr_user_snapshot_cache": {
```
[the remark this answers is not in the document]
noah-aviel-dove: Not related. I've deleted the table from my old personal deployment.

src/azul/terra.py
noah-aviel-dove: I'll bring it up at parking lot unless @hannes-ucsc replies here.

noah-aviel-dove: I decided not to bring this up at paring because I realized an obvious solution: since the snapshots are cached by access token and not by user email, users can reset/bypasss the cache simply by re-authenticating to obtain a fresh token. I think "logging out and logging back in again" is a sufficiently obvious course of action in the event of an authorization update that we don't need to change the implementation here.

noah-aviel-dove:
Reviewers: note https://github.com/databiosphere/azul/issues/2953#issuecomment-849886848

noah-aviel-dove:
Note that I've assigned two peer reviewers as I would like to take advantage of my time off on Friday to get as much feedback as possible. Also because I dislike seeing that red +/- icon every time I check this page.

src/azul/terra.py
```diff
@@ -513,7 +513,13 @@ def _list_snapshots(self) -> JSONs:
...
                                         params=dict(limit=limit))
         require(response['total'] <= limit, 'Too many snapshots', response['total'])
-        return response['items']
+        return [
+            {
+                'source_id': snapshot['id'],
+                'source_name': snapshot['name']
+            }
+            for snapshot in response['items']
+        ]
```
jessebrennan, line 522: Why this change?

jessebrennan (changes requested):

jessebrennan:
I reviewed and cleared @amarjandu's assignment so he doesn't waste effort reviewing already reviewed code.

src/azul/terra.py
amarjandu: It an interesting UX case.. I would like to see how we can tackle this, might have to do some combination of reduced TTL and a fallback to  `logging out and logging back in again`.

How can we make it apparent that a client should `logging out and logging back in again`? I would think by the time they email support, and support pings the team, the cache could be invalidated. 

On a side note, maybe the new `/repository/sources` endpoint can take a `refresh` url parameter. Presence of this parameter would return the accessible sources from the repository rather than from our cache, and update the cache in the process.

src/azul/terra.py
noah-aviel-dove: For comparison, this is what an element of `response['items']` looks like:

```
{
      "id": "9449a9f4-85aa-45cb-9fbf-cead88bf8f7d",
      "name": "hca_dev_20201203___20210219",
      "description": "Create snapshot hca_dev_20201203___20210219",
      "createdDate": "2021-03-03T20:17:01.095509Z",
      "profileId": "390e7a85-d47f-4531-b612-165fc977d3bd",
      "storage": [
        {
          "region": "us",
          "cloudResource": "bigquery",
          "cloudPlatform": "gcp"
        },
        {
          "region": "us-central1",
          "cloudResource": "bucket",
          "cloudPlatform": "gcp"
        },
        {
          "region": "us-central1",
          "cloudResource": "firestore",
          "cloudPlatform": "gcp"
        }
      ]
    },
```

Previously the cached snapshots were only used internally. Now that they're exposed to users via `/repository/sources` I decided to put more thought into how the JSON structure should be organized/presented. The main motivation for this change was to present the fields in a way that felt appropriate for the endpoint, but it also has the benefit of reducing our DynamoDB storage footprint.

test/integration_test.py
```diff
@@ -670,6 +686,28 @@ def _assert_indices_exist(self, catalog: CatalogName):
...
+                 path='/repository/sources',
+                 query={'catalog': catalog}).url
+        )
+        response = set(freeze(json.loads(response.data)['sources']))
+        expected_sources = {
+            frozendict(
+                source_name=source_name.name,
+                source_id=azul_tdr_client.lookup_source_id(source_name)
+            )
+            for source_name in map(TDRSourceName.parse, config.tdr_sources(catalog))
```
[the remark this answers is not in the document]
noah-aviel-dove: I really like this idea but unfortunately referencing `TDRSourceName` in `src/azul/__init__.py` would create a circular import.

src/azul/terra.py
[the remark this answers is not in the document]
noah-aviel-dove: @hannes-ucsc: "Add FIXME to implement paging of response"

src/azul/terra.py
noah-aviel-dove: @hannes-ucsc: "Add FIXME to streamline cache line invalidation. Two events should invalidate the cache line: addition or removal of a snapshot, and a change in the set of snapshots that the user has access to."

src/azul/drs.py
```diff
@@ -74,9 +74,10 @@ class Access:
     headers: Optional[Mapping[str, str]] = None
 
 
-@attr.s(auto_attribs=True, kw_only=True, frozen=True)
 class DRSClient:
-    http_client: RequestMethods
+
+    def __init__(self, *, http_client: RequestMethods):
+        self.http_client = http_client
```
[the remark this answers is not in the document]
noah-aviel-dove: It's a consequence of the change to `TerraDRSClient` and issues with mixing `attrs` and non-`attrs` classes with multiple inheritance. That said, I've found a less intrusive way to change this, although it still has a bad smell to me.

jessebrennan (commented):
I was trying to come up with a way to avoid the inheritance ugliness. What do you think of this

```
Index: src/azul/drs.py
IDEA additional info:
<+>UTF-8
diff --git a/src/azul/drs.py b/src/azul/drs.py
--- a/src/azul/drs.py	(revision 509197e6cd21fbcee99ae9d071ee02b2fb072985)
+++ b/src/azul/drs.py	(date 1622661029487)
@@ -74,8 +74,7 @@
     headers: Optional[Mapping[str, str]] = None
 
 
-# Have to set frozen=False to facilitate multiple inheritance in TerraDRSClient
-@attr.s(auto_attribs=True, kw_only=True, frozen=False)
+@attr.s(auto_attribs=True, kw_only=True, frozen=True)
 class DRSClient:
     http_client: RequestMethods
 
Index: src/azul/terra.py
IDEA additional info:
<+>UTF-8
diff --git a/src/azul/terra.py b/src/azul/terra.py
--- a/src/azul/terra.py	(revision 509197e6cd21fbcee99ae9d071ee02b2fb072985)
+++ b/src/azul/terra.py	(date 1622660905464)
@@ -275,13 +275,12 @@
         )
 
 
+@attr.s(auto_attribs=True, kw_only=True, frozen=True)
 class TerraClient:
     """
     A client to a service in the Broad Institute's Terra ecosystem.
     """
-
-    def __init__(self, auth_provider: AuthProvider):
-        self.auth_provider = auth_provider
+    auth_provider: AuthProvider
 
     @property
     def credentials(self) -> Credentials:
@@ -534,11 +533,10 @@
     @classmethod
     @cache
     def for_token(cls, token: OAuth2) -> 'TDRClient':
-        return cls(UserTokenAuthProvider(token))
+        return cls(auth_provider=UserTokenAuthProvider(token))
 
 
-class TerraDRSClient(DRSClient, TerraClient):
-
-    def __init__(self):
-        TerraClient.__init__(self, AzulAuthProvider())
-        super().__init__(http_client=self._http_client)
+@attr.s(auto_attribs=True, kw_only=True, frozen=True)
+class TerraDRSClient(DRSClient):
+    terra_client: TerraClient = TerraClient(auth_provider=AzulAuthProvider())
+    http_client: RequestMethods = terra_client._http_client
```

noah-aviel-dove:
> I was trying to come up with a way to avoid the inheritance ugliness. What do you think of this

Clever! I think it is an improvement. However, this raises a warning about access to a protected member (since `_http_client` is no longer a direct attribute of `TerraDRSClient`) and also has the effect of forcing every instance of `TerraDRSClient` to share an instance of `TerraClient`. I think I've found a solution that avoids boilerplate while maintaining the current inheritance hierarchy. It does require one small hack, albeit one [endorsed by the attrs devs](https://www.attrs.org/en/19.3.0/init.html#post-init-hook)

jessebrennan:
>  I think I've found a solution that avoids boilerplate while maintaining the current inheritance hierarchy.

I've used `__post_init__` a few times in the past and Hannes isn't very crazy about it in review, especially if there is a way to accomplish the goal using a `@classmethod` constructor. But if you prefer it, you're very welcome to take it up with him.

For those reasons I was trying to find something with composition.

noah-aviel-dove:
@jessebrennan I changed the JSON keys again, this time to camel case, to match the structure of `hits[].sources[]` in the existing index endpoints.

jessebrennan (approved):

hannes-ucsc (changes requested):
This draws the wrong abstraction lines in the wrong places. Where we deal with sources generically and where we can assume TDR snapshots and OAuth2 is all wrong.

I know getting abstractions right is difficult, and I think the best way for me to correct this is by refactoring it myself. It would take longer for me to explain, than to actually do it. If you understand intuitively where I'm coming from, give it your best shot. Otherwise, I'm happy to refactor it.

The second commit does more than what the title alludes. Clean up or squash.

I don't need to see fixups next time. But I'd like a clean history, not overly fragmented, and I'd like to see proof that each commit (except cosmetic ones) passes the build on it's own, at least once.

src/azul/terra.py
```diff
@@ -353,8 +452,88 @@ def _repository_endpoint(self, *path: str) -> str:
...
+        require(response['total'] <= limit, 'Too many snapshots', response['total'])
+        return [
+            {
+                'sourceId': snapshot['id'],
+                'sourceName': snapshot['name']
+            }
+            for snapshot in response['items']
+        ]
+
+    ddb_key_name = 'auth_key'
```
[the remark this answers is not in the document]
noah-aviel-dove: > Where else do we refer to DynamoDB as "ddb"?
https://github.com/DataBiosphere/azul/blob/17c53547ec814c09117a2acced40bc96e49268b5/test/version_table_test_case.py#L28

src/azul/service/repository_controller.py
```diff
@@ -224,3 +237,13 @@ def download_file(self,
...
+
+    def accessible_sources(self,
+                           catalog: CatalogName,
+                           authentication: Authentication,
+                           ) -> JSONs:
+        plugin = self.repository_plugin(catalog)
+        lister = plugin.source_lister(authentication)
+        sources = self.source_cache_service.list_sources(lister)
+        catalog_sources = {source.name for source in plugin.sources}
+        return [source for source in sources if source['sourceName'] in catalog_sources]
```
noah-aviel-dove, line 249: @hannes-ucsc : "Business logic of filtering by catalog belongs in RepositoryService"

lambdas/service/app.py
```diff
@@ -1752,6 +1746,49 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
+                This endpoint requires authentication.
+            ''')
+        }
+    }
+})
+def repository_sources() -> Response:
+    validate_params(app.current_request.query_params, catalog=validate_catalog)
+    catalog = app.catalog
+    authentication = app.current_request.authentication
+    if authentication is None:
```
noah-aviel-dove, line 1783: Move this conditional into `repository_controller.accessible_sources`

src/azul/terra.py
```diff
@@ -151,31 +178,116 @@ def qualify_table(self, table_name: str) -> str:
...
+        authorization. Should be consistent across lambda invocations.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def insufficient_access(self, resource: str) -> Exception:
+        raise NotImplementedError
+
+
+class AzulAuthenticator(Authenticator):
```
noah-aviel-dove, line 204: Rename to ServiceAccountAuthenticator

src/azul/terra.py
```diff
@@ -151,31 +178,116 @@ def qualify_table(self, table_name: str) -> str:
...
+    def insufficient_access(self, resource: str):
+        return RequirementError(
+            f'The service account (SA) {self.credentials.service_account_email!r} is not '
+            f'authorized to access {resource} or that resource does not exist. Make sure '
+            f'that it exists, that the SA is registered with SAM and has been granted read '
+            f'access to the resource.'
+        )
+
+
+class UserTokenAuthenticator(Authenticator):
```
noah-aviel-dove, line 237: Rename `UserCredentialsProvider`

lambdas/service/app.py
```diff
@@ -1752,6 +1746,36 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
+        # FIXME Determine public snapshots
+        #       https://github.com/DataBiosphere/azul/issues/2978
+        '401': {
+            'description': format_description('''
+                This endpoint requires authentication.
+            ''')
+        }
+    }
+})
+def repository_sources() -> Response:
```
hannes-ucsc, line 1772:
```suggestion
def list_sources() -> Response:
```

lambdas/service/app.py
```diff
@@ -1752,6 +1746,36 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
+    validate_params(app.current_request.query_params, catalog=validate_catalog)
+    sources = app.repository_controller.accessible_sources(app.catalog,
```
hannes-ucsc, line 1774:
```suggestion
    sources = app.repository_controller.list_sources(app.catalog,
```

lambdas/service/app.py
```diff
@@ -1752,6 +1746,36 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
+                                                           app.current_request.authentication)
```
hannes-ucsc, line 1775: Let the controller unpack the request. At some point we should move `app.catalog` to `app.current_request.catalog`

```suggestion
                                                           app.current_request)
```

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -48,10 +48,16 @@
     reject,
     require,
 )
+from azul.auth import (
+    OAuth2,
+)
```
hannes-ucsc, line 53:
```suggestion
from azul.auth import (
    Authentication,
    OAuth2,
)
```

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -48,10 +48,16 @@
...
 from azul.bigquery import (
     BigQueryRow,
     BigQueryRows,
 )
+from azul.chalice import (
+    Authentication,
+)
```
hannes-ucsc, line 60:
```suggestion
```

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,6 +182,21 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
     def sources(self) -> AbstractSet[TDRSourceName]:
         return self._sources
 
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> JSONs:
+        if isinstance(authentication, OAuth2):
+            sources = TDRClient.with_user_credentials(authentication).list_snapshots()
```
hannes-ucsc, line 189:
```suggestion
            tdr = TDRClient.with_user_credentials(authentication)
            sources = tdr.list_snapshots()
```

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,6 +182,21 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
...
+            catalog_sources = {source.name for source in self.sources}
+            return [source for source in sources if source['sourceName'] in catalog_sources]
```
hannes-ucsc, line 191: Use set.intersection?

src/azul/service/repository_controller.py
```diff
@@ -224,3 +240,23 @@ def download_file(self,
             }
         else:
             assert False
+
+    def accessible_sources(self,
+                           catalog: CatalogName,
+                           authentication: Optional[Authentication],
+                           ) -> JSONs:
+        def retrieve_sources():
```
hannes-ucsc, line 248:
```suggestion
        def list_sources():
```

test/indexer/test_tdr.py
```diff
@@ -219,5 +234,42 @@ def _full_table_name(self, source: TDRSourceName, table_name: str) -> str:
...
 
+class TestTDRSourceList(AzulTestCase):
+
+    def _mock_snapshots(self, access_token: str, *, translate: bool) -> JSONs:
+        return [{
+            'sourceId' if translate else 'id': 'foo',
+            'sourceName' if translate else 'name': f'{access_token}_snapshot'
+        }]
+
+    def _mock_list_snapshots(self,
```
hannes-ucsc, line 245:
```suggestion
    def _mock_urlopen(self,
```

test/indexer/test_tdr.py
```diff
@@ -219,5 +234,42 @@ def _full_table_name(self, source: TDRSourceName, table_name: str) -> str:
...
+                             tdr_client: TDRClient
+                             ) -> Callable[..., HTTPResponse]:
+        def _mock_response(http_client, method, url, **kwargs):
```
hannes-ucsc, line 248:
```suggestion
        def _mock_urlopen(http_client, method, url, **kwargs):
```

test/indexer/test_tdr.py
```diff
@@ -219,5 +234,42 @@ def _full_table_name(self, source: TDRSourceName, table_name: str) -> str:
...
+            self.assertEqual(method, 'GET')
+            self.assertEqual(furl(url).remove(query=True).url,
+                             tdr_client._repository_endpoint('snapshots'))
+            headers = {k.capitalize(): v for k, v in kwargs.get('headers', {}).items()}
```
hannes-ucsc, line 252: Couldn't `headers` be an explicit kwarg?

hannes-ucsc (changes requested):
The service and controller changes aren't covered.

https://coveralls.io/builds/40467806/source?filename=src%2Fazul%2Fservice%2Fsource_cache_service.py#L28

https://coveralls.io/builds/40467806/source?filename=src%2Fazul%2Fservice%2Frepository_controller.py#L244

src/azul/plugins/repository/tdr/__init__.py
noah-aviel-dove: I don't think `set.intersection` can be efficiently applied here because `sources` is a `Sequence[dict]` and `catalog_sources` is a `Sequence[str]`. I've renamed `catalog_sources` to `catalog_source_names` to make this more clear.

lambdas/service/app.py
```diff
@@ -1757,6 +1751,36 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
+    'tags': ['Repository'],
+    'parameters': [catalog_param_spec],
+    'responses': {
+        '200': {
+            'description': format_description('''
+                List the sources in the underlying data repository that the
+                current authenticated user is authorized to access.
+            '''),
+            **responses.json_content(
+                schema.object(sources=array_of_object_spec)
```
hannes-ucsc, line 1765: Too generic. This endpoint should return the sources in a format that's consistent with how they are returned by the /index endpoints, and the spec should reflect that specific format.

lambdas/service/app.py
```diff
@@ -1757,6 +1751,36 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
                                                    headers=headers)
 
 
+@app.route('/repository/sources', methods=['GET'], cors=True, method_spec={
+    'summary': 'List available data sources for the selected catalog',
```
hannes-ucsc, line 1755: I don't think we use the term "selected" elsewhere. Also, what if no catalog is passed?

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,10 +184,37 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
...
+        else:
+            raise RequirementError('Unsupported authentication format',
+                                   type(authentication))
+
     @property
     def tdr(self):
         return self._tdr()
 
+    @property
+    def _source_cache(self):
```
hannes-ucsc, line 215: Type hint

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -195,7 +230,12 @@ def tdr(self):
     @classmethod
     @cache_per_thread
     def _tdr(cls):
-        return TDRClient()
+        return TDRClient.with_service_account_credentials()
+
+    @classmethod
+    @cache_per_thread
+    def __source_cache(cls):
```
hannes-ucsc, line 237: Typically, services are used by controllers. The plugins don't use services. Think about why that is desirable.

Return type hint missing.

hannes-ucsc: Motivate the caching per thread. Service typically are NOT cached per thread.

hannes-ucsc (changes requested):
Looking at the coverage data, the case of an unauthenticated user hitting a TDR catalog doesn't seem to be covered.

src/azul/plugins/repository/tdr/__init__.py
noah-aviel-dove: I don't remember what my motivation was.

noah-aviel-dove:
Retracting review request, issues with the approach came up while rebasing the chained PR's

noah-aviel-dove:
Retracting review _again_, more problems came up on the _other_ chained PR

noah-aviel-dove:
Pushing a quick second fixup to remove an unused method left over from a more complicated approach

hannes-ucsc (changes requested):
See REVIEW commit.

src/azul/plugins/repository/dss/__init__.py
```diff
@@ -104,6 +107,11 @@ def sources(self) -> AbstractSet[str]:
     def lookup_source_id(self, spec: SimpleSourceSpec) -> str:
         return DSSSourceRef.id_from_spec(spec)
 
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> List[DSSSourceRef]:
+        return []
```
hannes-ucsc, line 113: Add FIXME

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,6 +182,38 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
     def sources(self) -> AbstractSet[TDRSourceSpec]:
         return self._sources
 
+    @cached_property
```
hannes-ucsc, line 185: State shouldn't be mutable unless it has to be. This is even more so the case for a cache, which is meant to be shared and therefore would be very exotic if it were mutable.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,6 +182,38 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
...
+    def _configured_source_specs_by_name(self) -> Dict[str, TDRSourceSpec]:
```
hannes-ucsc, line 186: I can see how we would want to constrain the name to be unique
per catalog, but across catalogs, no. I think list_sources
should be parameterized with the catalog. I also don't think
we need to cache this.

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,6 +182,38 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
...
+            source.name: source
+            for source in map(TDRSourceSpec.parse, catalog_sources)
+        }
+
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> List[TDRSourceRef]:
+        if isinstance(authentication, OAuth2):
+            tdr = TDRClient.with_user_credentials(authentication)
+            snapshot_ids_by_name = tdr.snapshot_ids_by_name()
```
hannes-ucsc, line 201: Same concern here. While snapshot names are de-facto unique at the moment and the DCP/2 spec requires them to be, the called method does not only return DCP/2 catalogs so I wouldn't want to rely on names in that list being unique. In this case it's easy to invert the mapping.

src/azul/plugins/__init__.py
```diff
@@ -191,16 +195,31 @@ def sources(self) -> AbstractSet[SOURCE_SPEC]:
...
+                     authentication: Optional[Authentication]
+                     ) -> Iterable[SOURCE_REF]:
+        """
+        List all sources in the underlying repository. Implementations should
+        raise RequirementError if the provided authentication is insufficient
+        to access the repository.
+        """
+        raise NotImplementedError
+
+    def source_classes(self) -> Tuple[Type[SOURCE_SPEC], Type[SOURCE_REF]]:
```
hannes-ucsc, line 209: This is a rather complicated concern that ideally I wouldn't want to bother clients with. It's ok to have this magic as an implementation detail of a method, but exposing it externally complicates the client interface and is brittle.

lambdas/service/app.py
```diff
@@ -1757,6 +1751,42 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
 
 
+@app.route('/repository/sources', methods=['GET'], cors=True, method_spec={
+    'summary': 'List available data sources',
+    'tags': ['Repository'],
+    'parameters': [catalog_param_spec],
+    'responses': {
+        '200': {
+            'description': format_description('''
+                List the sources in the underlying data repository that the
```
hannes-ucsc, line 1761:
```suggestion
                List the sources the currently authenticated user is authorized 
                to access in the underlying data repository.
```

src/azul/service/repository_controller.py
```diff
@@ -224,3 +243,42 @@ def download_file(self,
...
+            plugin_name = config.plugin_name(catalog, 'repository')
+            authentication_id = f'{plugin_name}:{authentication.identity()}'
+            try:
+                source_jsons = self._source_cache_service.get(authentication_id)
+            except CacheMiss:
+                sources = list_sources()
+                source_jsons = [source.to_json() for source in sources]
+                self._source_cache_service.put(authentication_id, source_jsons)
+            else:
+                _, source_ref_cls = plugin.source_classes()
```
hannes-ucsc, line 274: That's what I mean by bothering. See if you can hide this detail in `from_json`. In OO, there is no requirement that `type(cls.from_json()) is cls` only that `isinstance(cls.from_json(), cls)`.

src/azul/terra.py
```diff
@@ -352,8 +439,48 @@ def _repository_endpoint(self, *path: str) -> str:
...
+
+    @classmethod
+    def with_user_credentials(cls, token: OAuth2) -> 'TDRClient':
+        return cls(credentials_provider=UserCredentialsProvider(token))
+
 
-class TerraDRSClient(DRSClient, TerraClient):
+@attr.s(auto_attribs=True, kw_only=True, frozen=True)
+class TerraDRSClient(DRSClient, TDRClient):
+    credentials_provider: CredentialsProvider = ServiceAccountCredentialsProvider()
```
hannes-ucsc, line 480: Given that ServiceAccountCredentialsProvider isn't stateless and that its state is user-specific, I don't think it's a good idea to initialize this at the class level. It might even be a gaping security hole.

src/azul/terra.py
```diff
@@ -352,8 +439,48 @@ def _repository_endpoint(self, *path: str) -> str:
...
+    http_client: urllib3.PoolManager = attr.ib(init=False)
 
-    def __init__(self) -> None:
-        super().__init__(http_client=self._http_client)
+    def __attrs_post_init__(self):
+        # Bypass FrozenInstanceError on setting attribute post-init
+        # https://www.attrs.org/en/19.3.0/init.html#post-init-hook
+        object.__setattr__(self, 'http_client', self._http_client)
```
hannes-ucsc, line 486: See if you can find a less hacky way to model this. For example, http_client could be an abstract property in the super class. You should also make it a habit of documenting as to why a hack is employed. It took me a while to catch that there is an ordering issue since _http_client depends on `credentials_provider`. If attrs is getting in the way, you don't have to use it. I think you should dissolve the provider hierarchy and merge it into the TerraClient hierarchy. Abstract properties may be the way to go here. SAMClient can remain abstract since it isn't instantiated anywhere.

hannes-ucsc (changes requested):

src/azul/service/repository_controller.py
noah-aviel-dove: I don't think that's possible because `SourceRef` cannot know from the JSON alone which subclass to instantiate (and it seems bad/circular for the superclass to even be aware of all of its subclasses). At present, I'm convinced this conversion has to involve the plugin, but I think I can minimize the extent to which the plugin leaks these details.

src/azul/terra.py
noah-aviel-dove: I think I found a simpler and cleaner solution that we should discuss during parking lot.

src/azul/service/repository_controller.py
hannes-ucsc: > I don't think that's possible because SourceRef cannot know from the JSON alone which subclass to instantiate

Generally speaking it can, it would have to be encoded in the JSON.

> and it seems bad/circular for the superclass to even be aware of all of its subclasses

That's a myth. It's only bad if that awareness introduces a brittle coupling. Python has a mechanism for dealing with subclasses generically (`__init_subclass__`) and the parametric polymorphism employed by the SourceRef hierarchy is another way of dealing with this.

I didn't object to the involvement of the plugin, I objected to client code having to deal with an complicated implementation detail that should have be hidden inside SourceRef or the plugin.

src/azul/plugins/__init__.py
```diff
@@ -191,25 +194,44 @@ def sources(self) -> AbstractSet[SOURCE_SPEC]:
...
+        repository. Implementations should raise PermissionError if the provided
+        authentication is insufficient to access the repository.
+        """
+        raise NotImplementedError
+
+    def resolve_source(self,
+                       *,
+                       spec: str,
+                       id: Optional[str] = None,
+                       verify_id: bool = True) -> SOURCE_REF:
```
hannes-ucsc, line 213: With the new argument set to False, this alters the behavior of the method beyond recognition. It's name suggests that it resolves something, and the ID lookup is that resolution. See REVIEW commit.

src/azul/service/repository_controller.py
```diff
@@ -224,3 +243,42 @@ def download_file(self,
...
+            try:
+                source_jsons = self._source_cache_service.get(cache_key)
+            except CacheMiss:
+                sources = list_sources()
+                source_jsons = [source.to_json() for source in sources]
+                self._source_cache_service.put(cache_key, source_jsons)
+            else:
+                sources = [
+                    # Skip ID verification to avoid O(n) round-trips to TDR
+                    plugin.resolve_source(**source_json, verify_id=False)
```
hannes-ucsc, line 275: This is assumes that the `resolve_source` argument names match the keys that source.to_json emits. See REVIEW commit.

hannes-ucsc (changes requested):
See REVIEW commit.

src/azul/indexer/__init__.py
```diff
@@ -202,6 +205,16 @@ def __new__(cls: Type[SOURCE_REF], *, id: str, spec: SOURCE_SPEC) -> SOURCE_REF:
...
         return dict(id=self.id, spec=str(self.spec))
 
+    @classmethod
+    def from_json(cls, ref: JSON) -> 'SourceRef':
+        return cls(spec=cls.spec_cls().parse(ref['spec']), id=ref['id'])
+
+    @classmethod
+    def spec_cls(cls) -> Type[SourceSpec]:
+        base_cls = one(getattr(cls, '__orig_bases__'))
+        spec_cls, forward_ref = get_args(base_cls)
```
hannes-ucsc, line 215:
```suggestion
        spec_cls, ref_cls = get_args(base_cls)
```

src/azul/plugins/__init__.py
```diff
@@ -191,26 +195,45 @@ def sources(self) -> AbstractSet[SOURCE_SPEC]:
...
-                    'Source ID changed unexpectedly', spec, id, actual_id)
-        return source_ref_cls(id=id, spec=spec)
+        spec_cls, ref_cls = get_args(base_cls)
+        require(issubclass(spec_cls, SourceSpec))
+        require(issubclass(ref_cls, SourceRef))
+        assert ref_cls.spec_cls() is spec_cls
+        return ref_cls
+
+    def source_from_json(self, ref: JSON) -> SOURCE_REF:
+        return self._source_ref_cls.from_json(ref)
```
hannes-ucsc, line 221: Add docstring

src/azul/plugins/__init__.py
```diff
@@ -191,26 +195,45 @@ def sources(self) -> AbstractSet[SOURCE_SPEC]:
...
+        Return an instance of :class:`SourceRef` for the repository source
+        matching the given specification or raise an exception if no such source
+        exists.
+        """
+        ref_cls = self._source_ref_cls
+        spec = ref_cls.spec_cls().parse(spec)
+        id = self.lookup_source_id(spec)
+        return ref_cls(id=id, spec=spec)
+
+    def verify_source(self, ref: SOURCE_REF) -> None:
```
hannes-ucsc, line 234: Add docstring

src/azul/plugins/repository/tdr/__init__.py
```diff
@@ -176,6 +179,30 @@ def create(cls, catalog: CatalogName) -> 'RepositoryPlugin':
...
+                for id, name in tdr.snapshot_names_by_id().items()
+                if name in configured_specs_by_name
+            }
+            return [
+                TDRSourceRef(id=id,
+                             spec=configured_specs_by_name[name])
+                for name, id in snapshot_ids_by_name.items()
+            ]
+        elif authentication is None:
+            # FIXME Determine public snapshots
```
hannes-ucsc, line 199: formatting

src/azul/plugins/repository/dss/__init__.py
```diff
@@ -104,6 +107,13 @@ def sources(self) -> AbstractSet[str]:
     def lookup_source_id(self, spec: SimpleSourceSpec) -> str:
         return DSSSourceRef.id_from_spec(spec)
 
+    def list_sources(self,
+                     authentication: Optional[Authentication]
+                     ) -> List[DSSSourceRef]:
+        # FIXME De-conflate str and SimpleSourceName in dss.Plugin.sources
```
hannes-ucsc, line 113: formatting

test/service/test_repository_proxy.py
```diff
@@ -181,6 +186,59 @@ def test_repository_files_proxy(self):
...
+                    'sources': [
+                        {
+                            'sourceId': source['id'],
+                            'sourceSpec': source['spec']
+                        }
+                        for source in mock_source_jsons
+                    ]
+                })
+
+            # FIXME Determine public snapshots
```
hannes-ucsc, line 229: formatting

src/azul/plugins/__init__.py
```diff
@@ -206,33 +207,33 @@ def list_sources(self,
...
-            else:
-                require(id == actual_id,
-                        'Source ID changed unexpectedly', spec, id, actual_id)
-        else:
-            require(id is not None)
-        return source_ref_cls(id=id, spec=spec)
+        ref_cls = self._source_ref_cls
+        spec = ref_cls.spec_cls().parse(spec)
+        id = self.lookup_source_id(spec)
+        return ref_cls(id=id, spec=spec)
```
hannes-ucsc, line 232: Do you also get a PyCharm warning here?

hannes-ucsc (changes requested):

src/azul/plugins/__init__.py
noah-aviel-dove: I do not.

src/azul/plugins/__init__.py
```diff
@@ -218,6 +218,10 @@ def _source_ref_cls(self) -> Type[SOURCE_REF]:
         return ref_cls
 
     def source_from_json(self, ref: JSON) -> SOURCE_REF:
+        """
+        Instance a :class:`SourceRef` from its JSON representation. The expected
```
hannes-ucsc, line 222: Looking at MerriamWebster this isn't correct use of the verb `to instance`. What this returns is neither a citation nor an example.

hannes-ucsc (changes requested):
A grammar nit. 

The next time around, since I won't be doing PR reviews tomorrow, could you have a thorough peer comb through this one more time?

amarjandu (commented):
Some of the methods in the subclasses around `Authentication` and `CredentialsProvider` seem like that can be changed to properties.

It might be worth changing the name of the keys for the objects that are returned from `/repository/sources`.

lambdas/service/app.py
```diff
@@ -1751,6 +1751,42 @@ def validate_wait(wait: Optional[str]) -> Optional[int]:
...
+    'responses': {
+        '200': {
+            'description': format_description('''
+                List the sources the currently authenticated user is authorized
+                to access in the underlying data repository.
+            '''),
+            **responses.json_content(
+                schema.object(sources=schema.array(
+                    schema.object(
+                        sourceId=str,
```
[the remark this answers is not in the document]
hannes-ucsc: We're already commit to `sourceId` and `sourceSpec` in /index/responses. The naming should be consistent.

src/azul/plugins/__init__.py
hannes-ucsc: Because verification is expensive. It requires a round-trip to TDR.

src/azul/plugins/__init__.py
hannes-ucsc: I get that, too. PyCharm doesn't realize that `@cached_property` should be treated like `@property`.

src/azul/service/repository_controller.py
```diff
@@ -224,3 +243,41 @@ def download_file(self,
...
+                source_jsons = [source.to_json() for source in sources]
+                self._source_cache_service.put(cache_key, source_jsons)
+            else:
+                sources = [
+                    plugin.source_from_json(source_json)
+                    for source_json in source_jsons
+                ]
+        return [
+            {
+                'sourceId': source.id,
```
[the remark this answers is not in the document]
hannes-ucsc: See above comment on consistency with existing naming conventions.

src/azul/terra.py
```diff
@@ -150,25 +157,72 @@ def qualify_table(self, table_name: str) -> str:
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def insufficient_access(self, resource: str) -> Exception:
+        raise NotImplementedError
+
+
+class ServiceAccountCredentialsProvider(CredentialsProvider):
+
+    def oauth2_scopes(self) -> Sequence[str]:
```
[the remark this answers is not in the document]
hannes-ucsc: I'm ambivalent. It wouldn't save that much typing but require four mentions of `@property`.

src/azul/service/repository_controller.py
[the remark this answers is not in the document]
noah-aviel-dove: > The return type says that these contents should be SourceRefs

There is a local function on line 254 that returns `Iterable[SourceRef]`, but the return statement on line 277 is for the function with return type `JSONs`

amarjandu (changes requested):
Small stuff.

test/dynamodb_test_case.py
```diff
@@ -0,0 +1,41 @@
+from typing import (
```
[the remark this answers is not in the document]
noah-aviel-dove: Couldn't hurt.

src/azul/hmac.py
```diff
@@ -23,26 +22,18 @@
...
-    def identity(self) -> str:
-        return self.key_id
-
-    @classmethod
-    def from_request(cls, request: chalice.app.Request) -> Optional[Authentication]:
-        try:
-            header = request.headers['Authorization']
-        except KeyError:
-            return None
+def auth_from_request(request: chalice.app.Request) -> Optional[HMACAuthentication]:
```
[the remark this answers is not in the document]
noah-aviel-dove: Tried to fix this but the way our formatting looks with only one argument is pretty hideous IMO, would appreciate any suggestions

amarjandu (approved):

noah-aviel-dove:
Pulling from review to address a bug found while working on #3176

src/azul/terra.py
```diff
@@ -149,31 +170,123 @@ def qualify_table(self, table_name: str) -> str:
...
-
-    def _request(self, method, url, *, fields=None, headers=None, body=None) -> urllib3.HTTPResponse:
+        # By default, AuthorizedHTTP attempts to refresh the credentials on a 401
+        # response, which is never helpful. When using service account
+        # credentials, a fresh token is obtained for every lambda invocation,
+        # which will never persist long enough for the token to expire. User
+        # tokens can expire, but attempting to refresh them raises
+        # `google.auth.exceptions.RefreshError` due to the credentials not being
+        # configured with (among other fields) the client secret.
+        return AuthorizedHttp(self.credentials, http_client(), refresh_status_codes=())
```
amarjandu, line 280: 79 char lucky :p

amarjandu (approved):
Thanks for adding the explanation around the 401s.

hannes-ucsc (approved):

Issue #3290, opened 2021-08-03
Upgrade Chalice to at least 1.23.0 or later


jessebrennan:
- [x] Remove vendored wheels for grpcio and fastavro. To do this we need to update Chalice so that it allows wheels for `manylinux2010` and `manylinux2014`. We vendor the wheels currently to work around this.

@jessebrennan "This will break the way we do our request logging. Request logging should probably be changed to use [Chalice's middleware API](https://aws.github.io/chalice/topics/middleware.html)"

melainalegaspi:
@jessebrennan spike to estimate by exploring what other changes the chalice upgrade would entail.

hannes-ucsc:
For demo, show that `make clean` removes the three `vendor` directories and that `git status` shows a clean working copy before and after.

PR #3323, opened 2021-08-12
Upgrade Chalice to 1.24.1 (#3290)

jessebrennan:
https://github.com/DataBiosphere/azul/issues/3290

Author

Author (reindex)

Author (freebies & chains)

Author (upgrading)

Author (before every review)

- [x] Commented in issue about demo expectations            <sub>or labelled issue as `no demo`</sub>

- [x] Checked that demo expectations are clear              <sub>or issue is labeled as `no demo`</sub>
- [x] Sanity-checked history
- [x] Branch pushed to Gitlab and added `sandbox` label     <sub>or PR is labeled `no sandbox`</sub>
- [x] Build passed in sandbox                               <sub>or PR is labeled `no sandbox`</sub>

Operator (reindex) 

- [ ] Started reindex in `dev`                              <sub>or this PR does not require reindexing or does not target `dev`</sub>
- [ ] Checked for failures in `dev`                         <sub>or this PR does not require reindexing or does not target `dev`</sub>
- [ ] Started reindex in `prod`                             <sub>or this PR does not require reindexing or does not target `prod`</sub>
- [ ] Checked for failures in `prod`                        <sub>or this PR does not require reindexing or does not target `prod`</sub>

Operator

- [ ] Unassigned PR


src/azul/chalice.py
```diff
@@ -65,6 +66,30 @@ def __init__(self, app_name, unit_test=False, spec=None):
         else:
             self._specs: Optional[MutableJSON] = None
         super().__init__(app_name, debug=config.debug > 0, configure_logs=False)
+        # The last handler is called first. We want logging to surround all
```
jessebrennan, line 69: This comment is incorrect. I will fix it in the next cycle.

lambdas/layer/Makefile
```diff
@@ -8,10 +8,6 @@ include ../lambdas.mk
 wheels: check_env
 	rm -rf vendor/jsonobject*
```
jessebrennan, line 9: This needs to be changed to remove the whole contents of vendor or else add upgrading instructions to do that once so that we don't keep packaging the old wheels redundantly.

src/azul/chalice.py
```diff
@@ -65,6 +66,30 @@ def __init__(self, app_name, unit_test=False, spec=None):
...
+        # other handler calls.
+        self.register_middleware(self.logging_middleware, 'http')
```
amarjandu, line 71: IIRC the SQS -> Lambda interactions that the bundle notifications have use `sqs` as the event type rather than `http`, should we extend this logging middleware to have  `event_type=all` in order to record any type of event trigger that the lambdas? Not sure if we can do something like ('http', 'sqs') here.

amarjandu (changes requested):
Looks good, just a question...

src/azul/chalice.py
jessebrennan: Glancing at the indexer lambdas, it seems like we already log when a notification comes in. Also middleware here would have to be changed since the SQS lambdas don't return a `response`.

If you think the SQS lambdas have insufficient logging, I think that would make sense as a separate issue.

amarjandu (approved):

src/azul/chalice.py
jessebrennan: @amarjandu, I'm confused about the eyes emoji, lol.

hannes-ucsc: Me too.

hannes-ucsc (changes requested):

lambdas/layer/Makefile
jessebrennan: We should remove the whole contents of vendor because this caused a problem for @danielsotirhos where pyresist somehow ended up in there and caused duplicates.

src/azul/chalice.py
```diff
@@ -65,6 +66,30 @@ def __init__(self, app_name, unit_test=False, spec=None):
...
+    def logging_middleware(self, event, get_response):
+        self._log_request()
+        response = get_response(event)
+        self._log_response(response)
+        return response
+
+    def authentication_middleware(self, event, get_response):
+        # We use the handler's private method to generate the response in case
+        # authentication fails
+        handler = RestAPIEventHandler(self.routes,
```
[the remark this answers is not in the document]
jessebrennan: > The reason for switching to middleware is to reduce the hackyness of the current approach

This isn't the main reason for switching, but rather a hopeful side-effect. The reason we switched was because the method we were previously overriding in `Chalice()` no longer existed. Middleware seemed like the simplest way to get the same functionality.

> This is no less hacky than what we already do, it's less efficient and more surprising (two handler objects).

I agree, instantiating the handler again is hacky. The alternatives are:
- to duplicate the body of `_get_view_function_response()` (maybe not a bad idea) or
- subclass `RestAPIEventHander()` and change the behavior. 
I'll explain why below.

> I think the global error handler (_global_error_handler) is the outermost handler so the translation of any exception raised in _authenticate would still happen, even if we invoked it explicitly.

This is incorrect. `_global_error_hander()` simply packages the exception in a 500 response via `_unhandled_exception_to_response()`.

Anyway, assuming all of this stands, would you consider resetting the review counter?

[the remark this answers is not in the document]
jessebrennan: I'll bring this up in PL tomorrow.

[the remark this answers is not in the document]
hannes-ucsc: > > The reason for switching to middleware is to reduce the hackyness of the current approach
> 
> This isn't the main reason for switching, but rather a hopeful side-effect. The reason we switched was because the method we were previously overriding in `Chalice()` no longer existed. Middleware seemed like the simplest way to get the same functionality.

Let me rephrase that. Authentication is a text book use case of middleware. We should be able to implement authentication using middleware without resorting to hacks.

> > I think the global error handler (_global_error_handler) is the outermost handler so the translation of any exception raised in _authenticate would still happen, even if we invoked it explicitly.
> 
> This is incorrect. `_global_error_hander()` simply packages the exception in a 500 response via `_unhandled_exception_to_response()`.

But the "packaging" of the exception in a 500 response is exactly the translation I was talking about. I don't see where my statement is incorrect. What are we not getting from the global handler that we're getting from your current approach?

> Anyway, assuming all of this stands, would you consider resetting the review counter?

I'll decide when it is clear that A) there is no better implementation or B) the implementation we end up using couldn't possibly have been implemented by you initially, because information wasn't available to you or I changed the specification mid-PR.

deployments/sandbox/environment.py
```diff
@@ -59,8 +59,6 @@ def env() -> Mapping[str, Optional[str]]:
             ]
         }),
 
-        # FIXME: Add tooling to aid in prefix choice
-        #        https://github.com/DataBiosphere/azul/issues/3027
```
hannes-ucsc, line 63: COmmit needs ticket ref.

hannes-ucsc (changes requested):

src/azul/chalice.py
```diff
@@ -65,6 +66,24 @@ def __init__(self, app_name, unit_test=False, spec=None):
...
+        self._log_request()
+        response = get_response(event)
+        self._log_response(response)
+        return response
+
+    def authentication_middleware(self, event, get_response):
+        try:
+            self.__authenticate()
+        except ChaliceViewError as e:
+            response = Response(body={'Code': e.__class__.__name__, 'Message': str(e)},
```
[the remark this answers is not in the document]
jessebrennan: I'll make this change. For my own edification, I'm curious as to why. I had copied the above code more or less verbatim from Chalice.

test/service/test_app_logging.py
```diff
@@ -60,7 +60,9 @@ def test_request_logs(self):
                             level,
                             'Returning 200 response. To log headers and body, set AZUL_DEBUG to 1.'
                             if level == INFO else
-                            'Returning 200 response with headers {}. '
+                            'Returning 200 response with headers {"Access-Control-Allow-Origin": '
+                            '"*", "Access-Control-Allow-Headers": '
+                            '"Authorization,Content-Type,X-Amz-Date,X-Amz-Security-Token,X-Api-Key"}. '
```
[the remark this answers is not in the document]
jessebrennan: @hannes-ucsc: "Jesse should investigate why headers weren't present before; timebox 30min."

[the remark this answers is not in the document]
jessebrennan: Previously (before middleware), our insertion point for logging occurred by overriding the method called here: https://github.com/aws/chalice/blob/729b28543142636891ba01b2ac2e927cc38c3771/chalice/app.py#L1094

You can see, that the CORS headers were added soon after this method is called, so our logging statement was without them.

With middleware, the [_main_rest_api_handler()](https://github.com/aws/chalice/blob/ea779474e6dc193e21ea0a68c279e42634ad64b0/chalice/app.py#L1627) fully process the request before the middleware runs, thus the middleware is populated with all of the headers we expect.

hannes-ucsc (approved):

hannes-ucsc:
Operator, note drop! commit.

hannes-ucsc:
Re: https://github.com/DataBiosphere/azul/pull/3323#discussion_r699761632

> > Anyway, assuming all of this stands, would you consider resetting the review counter?

> I'll decide when it is clear that A) there is no better implementation or B) the implementation we end up using couldn't possibly have been implemented by you initially, because information wasn't available to you or I changed the specification mid-PR.Re: 

I determine that neither A nor B apply, so review count stands.

noah-aviel-dove:
@jessebrennan The commit title tagging is wrong. `[r]` is for reindex and `[R]` is for requirements updates. I believe the problem is perfunctory enough that I can fix it myself.

noah-aviel-dove:
Conflicts resolved by operator

Issue #3303, opened 2021-08-05
Add content_description for matrices to /project endpoint


theathorn:
Add content_description to the /project API response for projects[].matrices and projects[].contributorMatrices.

theathorn:
@NoopDog states "Alternatively we could query the files endpoint for files related to the project and use that to determine the CGMs and DCP matrices".

hannes-ucsc:
We should unify `_matrices` and `_file` in `transform.py` so that `contents.{matrices, contributor_matrices}` resembles `contents.files` in project contributions.

Issue #2462, opened 2020-11-17
Allow null values in PFBs 


jessebrennan:
Remove the work around for #2370 when https://broadworkbench.atlassian.net/browse/AS-521 is fixed.

hannes-ucsc:
I want to solve this, and thereby verify the upstream resolution, before we get deeper into the verbatim hand-over epic (#2693). In fact, I will add this to that epic.

nadove-ucsc:
For demo, create and download a reasonably small PFB manifest using the Swagger UI. Confirm the presence of null values using `pfb show`. Then perform a handover to Terra via the Data Browser using the same set of filters. Confirm presence of null values in the workspace.

PR #3380, opened 2021-08-27
Add content_description for matrices to /project endpoint (#3303)

dsotirho-ucsc:
https://github.com/DataBiosphere/azul/issues/3303

Author

- [x] PR title matches issue title (preceded by `Fix: ` for bugs)   <sub>or there is a good reason why they're different</sub>
- [x] PR is connected to Zenhub issue and description links to issue

Author (reindex)

Author (freebies & chains)

Author (upgrading)

- [x] Added announcement to PR description                  <sub>or this PR does not require announcement</sub>

Author (before every review)

- [x] Commented in issue about demo expectations            <sub>or labelled issue as `no demo`</sub>

- [x] Checked that demo expectations are clear              <sub>or issue is labeled as `no demo`</sub>
- [x] Sanity-checked history
- [x] Branch pushed to Gitlab and added `sandbox` label     <sub>or PR is labeled `no sandbox`</sub>
- [x] Build passed in sandbox                               <sub>or PR is labeled `no sandbox`</sub>

- [x] Made announcement requested by author                 <sub>or PR description does not contain an announcement</sub>
- [x] Moved freebies to Merged column                       <sub>or there are no freebies in this PR</sub> 
- [x] Shortened the PR chain                                <sub>or this PR is not the base of another PR</sub>
- [x] Verified that `N reviews` labelling is accurate
- [x] Pushed merge commit to Gitlab                         <sub>or this changes can be pushed later, together with another PR</sub>
- [x] Deleted PR branch from Github and Gitlab

Operator (reindex) 

- [x] Started reindex in `dev`                              <sub>or this PR does not require reindexing or does not target `dev`</sub>
- [x] Checked for failures in `dev`                         <sub>or this PR does not require reindexing or does not target `dev`</sub>
- [x] Started reindex in `prod`                             <sub>or this PR does not require reindexing or does not target `prod`</sub>
- [x] Checked for failures in `prod`                        <sub>or this PR does not require reindexing or does not target `prod`</sub>

Operator

- [x] Unassigned PR


src/azul/service/avro_pfb.py
```diff
@@ -153,24 +150,39 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
...
             if field['name'] not in object_:
-                if name == 'files':
-                    default_value = None
+                if name == 'files' or name == 'related_files':
+                    if 'string' in field['type']:
+                        default_value = None
+                    elif field['type']['type'] == 'array':
+                        default_value = [None]
+                    else:
+                        assert False, field
```
jessebrennan, line 168: I think this might be just a little bit stricter than necessary. If `size` was left out for some reason, for example, we would have to change this code to accommodate it. I think `default_value` should be `None` for everything unless it's an array, or other composite type.

jessebrennan (changes requested):
Nice work this all looks really good. I had just a few questions and points of feedback.

For the commit "[r] Keep datetime values as strings (#3303)" it might make sense to add some of the "why" for this change, possibly in the body of the commit message.

How come you had to add the namespace stuff? I vaguely remember it being necessary if there's an ambiguous path, or something like that..

src/azul/service/avro_pfb.py
danielsotirhos: All the `_nullable_to_pfb_types` include "string" which is why I thought this would be acceptable. e.g. `null_int: ['string', 'long']`.
With that said I've taken your comment and tried to make this section a bit more general as you suggested. The nature of the `field['type']` variable made this a little tricky though since I observed it either containing a list of types (`field['type'] == ['string']`) or a further nested dictionary (`field['type']['type'] == 'array'`)

danielsotirhos:
> For the commit "[r] Keep datetime values as strings (#3303)" it might make sense to add some of the "why" for this change, possibly in the body of the commit message.

+1

> How come you had to add the namespace stuff? I vaguely remember it being necessary if there's an ambiguous path, or something like that..

The `matrices` and `contributor_matrices` inner entities previously had FieldTypes that used a pass thru:

```
return {
    'document_id': null_str,
    # Pass through dict with file properties, will never be None
    'file': pass_thru_json,
}
``` 

After changing this so that the fields under `file` were individually defined:

```
return {
    'document_id': null_str,
    'file': {
        **cls._file_types(),
        'strata': null_str
    }
}
```

I ran into an issue with fastavro where it complained that I was using a "redefined named type".
There wasn't much to be found about this error, however I tracked it down in the fastavro [source code](https://github.com/fastavro/fastavro/blob/e4ce5237dd8e378fb8874e2f8a9ebb587c2b550f/fastavro/_schema.pyx#L206) and then looked into namespaces and found "A schema or protocol may not contain multiple definitions of a fullname." in the [fastavro documentation](https://avro.apache.org/docs/current/spec.html#names). 

This led me to try adding the namespaces which solved the issue.

src/azul/service/avro_pfb.py
jessebrennan: > All the _nullable_to_pfb_types include "string" which is why I thought this would be acceptable

You're totally right, I didn't think about this. What you wrote before is totally acceptable then. 

Possibly you already know this, but every type contains 'string' also because Terra doesn't allow `null`. Instead we represent it with an empty string. 

So the `'string'` you wrote here is the same as `null`. Consider adding a FIXME like `FIXME: Change 'string' to 'null' in https://github.com/DataBiosphere/azul/issues/2462`

jessebrennan (approved):
See my response to your comment, sorry for the misunderstanding there. I think what you had originally is better though both are acceptable.

Either way this looks good to go.

src/azul/indexer/aggregate.py
```diff
@@ -163,6 +167,30 @@ def get(self):
...
+    accumulation of dictionaries can be sorted by. (eg. `itemgetter('uuid')`)
+    """
+
+    def __init__(self, key: Callable):
+        self.key = key
+        self.value = {}
+
+    def accumulate(self, value):
+        require(isinstance(value, dict), value)
+        key = self.key(value)
```
hannes-ucsc, line 182: This use of self.key contradicts the documentation. Apparently the key is not just used for sorting.

src/azul/indexer/aggregate.py
```diff
@@ -163,6 +167,30 @@ def get(self):
...
+    An accumulator of dictionaries. The key should return a unique value the
+    accumulation of dictionaries can be sorted by. (eg. `itemgetter('uuid')`)
+    """
+
+    def __init__(self, key: Callable):
+        self.key = key
+        self.value = {}
+
+    def accumulate(self, value):
+        require(isinstance(value, dict), value)
```
hannes-ucsc, line 181: Other than this, nothing else requires `value` to be a dict. It is never used as a dict. It could be anything. I still like the name DictAccumulator, not because the accumulated values are dicts but because their are accumulated in a dict. In a way this is a generalized SetAccumulator. DictAccumulator can replace a SetAccumulator by using the identity function for the key (`lambda _: _`). Not saying we should do that, but it would help if the documentation mentioned that.

src/azul/plugins/metadata/hca/transform.py
```diff
@@ -1307,36 +1321,37 @@ def matrix_stratification_values(self, file: api.File) -> JSON:
             file_name = file.manifest_entry.name
             strata = parse_strata(file_description)
             for stratum in strata:
-                donors = {}
+                donor = {}
```
hannes-ucsc, line 1309: Rename makes sense but is on wrong commit.

hannes-ucsc (changes requested):
I don't think I would reference #3303 in the title of the first commit.

hannes-ucsc (changes requested):
I think I'm getting overwhelmed by the size of this. We should extract the first commit into an issue/PR combo on it's own. No bump.

danielsotirhos:
> We should extract the first commit into an issue/PR combo on it's own.

Created https://github.com/databiosphere/azul/issues/3416 and rebased this PR off of PR https://github.com/DataBiosphere/azul/pull/3419

hannes-ucsc (commented):
I want to review the base PR first.

danielsotirhos:
> Created #3416 and rebased this PR off of PR #3419

This ticket is not dependent on #3416. I've rebased this ticket on `develop` and removed blocking relationship.

src/azul/indexer/aggregate.py
```diff
@@ -163,6 +167,34 @@ def get(self):
         return thaw(super().get())
 
 
+class DictAccumulator(Accumulator):
+    """
+    Accumulate values into a dictionary, allowing one unique value per key.
+    In a way this is a generalized SetAccumulator. DictAccumulator can replace
+    a SetAccumulator by using the identity function for the key (lambda _: _).
```
hannes-ucsc, line 174:
```suggestion
    a SetAccumulator by using the identity function (``lambda _: _``) for the key.
```

src/azul/indexer/aggregate.py
```diff
@@ -163,6 +167,34 @@ def get(self):
...
+    """
+
+    def __init__(self, key: Callable):
+        """
+        :param key: The key to be used both for storing the accumulated value
```
hannes-ucsc, line 179:
```suggestion
        :param key: A function returning the key to be used both for storing the accumulated value
```

src/azul/service/avro_pfb.py
```diff
@@ -153,24 +154,40 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
...
+        elif schema['name'] == 'related_files':
+            entity_schema = schema['type']['items']
+        else:
+            assert False, schema
         for field in entity_schema['fields']:
             if field['name'] not in object_:
-                if name == 'files':
-                    default_value = None
+                if name == 'files' or name == 'related_files':
+                    # FIXME: Change 'string' to 'null' in https://github.com/DataBiosphere/azul/issues/2462
```
hannes-ucsc, line 167: F

src/azul/service/avro_pfb.py
```diff
@@ -153,24 +154,40 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
...
-                    assert field['type']['type'] == 'array'
+                    assert field['type']['type'] == 'array', field
                     # Default for a list of records is an empty list
                     if isinstance(field['type']['items'], dict):
-                        assert field['type']['items']['type'] == 'record'
+                        assert field['type']['items']['type'] == 'record', field
                         default_value = []
                     # All other types are lists of primitives where None is an accepted value
                     else:
                         # FIXME: Change 'string' to 'null' in https://github.com/DataBiosphere/azul/issues/2462
```
hannes-ucsc, line 182: F

src/azul/service/avro_pfb.py
```diff
@@ -153,24 +154,40 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
...
+            entity_schema = schema['type']['items']
+        else:
+            assert False, schema
         for field in entity_schema['fields']:
             if field['name'] not in object_:
-                if name == 'files':
-                    default_value = None
+                if name == 'files' or name == 'related_files':
+                    # FIXME: Change 'string' to 'null' in https://github.com/DataBiosphere/azul/issues/2462
+                    if 'string' in field['type']:
```
hannes-ucsc, line 168: Why is this using `in` and not `==`?

src/azul/service/avro_pfb.py
```diff
@@ -153,24 +154,40 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
...
+            entity_schema = one(e for e in object_schema['type'] if e['name'] == name)
+        elif schema['name'] == 'related_files':
+            entity_schema = schema['type']['items']
+        else:
+            assert False, schema
         for field in entity_schema['fields']:
             if field['name'] not in object_:
-                if name == 'files':
-                    default_value = None
+                if name == 'files' or name == 'related_files':
```
hannes-ucsc, line 166: Is there any way we can avoid hard-coding the field names here?

src/azul/service/avro_pfb.py
```diff
@@ -153,24 +154,40 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
         None is the default value, but because of https://github.com/DataBiosphere/azul/issues/2370
         this isn't currently reflected in the schema.
         """
-        object_schema = one(f for f in schema['fields'] if f['name'] == 'object')
-        entity_schema = one(e for e in object_schema['type'] if e['name'] == name)
+        if schema['type'] == 'record':
+            object_schema = one(f for f in schema['fields'] if f['name'] == 'object')
+            entity_schema = one(e for e in object_schema['type'] if e['name'] == name)
+        elif schema['name'] == 'related_files':
```
hannes-ucsc, line 160: and here

src/azul/service/avro_pfb.py
```diff
@@ -423,22 +442,26 @@ def string_to_datetime(data, *args):
...
         if isinstance(field_type, dict):
             yield {
                 "name": entity_type,
+                "namespace": '.'.join(path),
                 "type": {
                     # This is always an array, even if singleton is passed in
                     "type": "array",
                     "items": {
                         "name": entity_type,
+                        "namespace": '.'.join(path),
```
hannes-ucsc, line 460: What does this do? It seems unrelated.

hannes-ucsc: Also, shouldn't `'.'.join(path)` be extracted into a variable to avoid duplicating the expression?

hannes-ucsc (changes requested):
Still overwhelmed with the amount of changes. The monolithic history doesn't make understanding this easier. For example, you should probably extract the `_file_base` refactoring into a separate commit that is semantically neutral.

src/azul/service/avro_pfb.py
danielsotirhos: All `_nullable_to_pfb_types` are lists that contain `string`, but not all only contain `string`.

src/azul/service/avro_pfb.py
danielsotirhos: > What does this do? It seems unrelated.

https://github.com/DataBiosphere/azul/pull/3380#issuecomment-910465259

src/azul/service/avro_pfb.py
```diff
@@ -157,31 +157,28 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
         if schema['type'] == 'record':
             object_schema = one(f for f in schema['fields'] if f['name'] == 'object')
             entity_schema = one(e for e in object_schema['type'] if e['name'] == name)
-        elif schema['name'] == 'related_files':
+        elif isinstance(schema['type'], dict):
```
hannes-ucsc, line 160: I'd like @jessebrennan to review this PR again.

hannes-ucsc (changes requested):
No bump.

src/azul/service/avro_pfb.py
hannes-ucsc: Makes sense.

src/azul/service/avro_pfb.py
```diff
@@ -244,9 +245,11 @@ def pfb_schema_from_field_types(field_types: FieldTypes) -> JSON:
     entity_schemas = (
         {
             "name": entity_type,
+            "namespace": '',
             "type": "record",
             "fields": list(_entity_schema_recursive(field_type,
-                                                    files_entity=entity_type == 'files'))
+                                                    files_entity=entity_type == 'files',
```
hannes-ucsc, line 264: You should be able to derive that from the path inside the method.

src/azul/service/avro_pfb.py
```diff
@@ -423,22 +426,27 @@ def string_to_datetime(data, *args):
 
 
 def _entity_schema_recursive(field_types: FieldTypes,
-                             files_entity: bool = False) -> Iterable[JSON]:
+                             files_entity: bool = False,
+                             path: Tuple[str, ...] = ()) -> Iterable[JSON]:
```
hannes-ucsc, line 443: See if you can use `*path`

src/azul/service/avro_pfb.py
```diff
@@ -472,25 +482,30 @@ def _entity_schema_recursive(field_types: FieldTypes,
...
                         # FIXME: Change 'string' to 'null' in https://github.com/DataBiosphere/azul/issues/2462
                         "string",
                         {
                             "name": entity_type,
+                            "namespace": namespace,
                             "type": "record",
                             "fields": [
                                 {
                                     "name": name,
+                                    "namespace": '.'.join((namespace, entity_type)),
```
hannes-ucsc, line 522: Duplicates joining string.

More verbose and probably slower than `+`.

hannes-ucsc (changes requested):

hannes-ucsc (changes requested):
I don't think this should be a split commit.

The breakdown into commits is fine, it's just there is no reason to use the split M/N convention because some of the commits are worth keeping, even if we revert the one that actually adds content_description. A split commit indicates an all-or-none relationship.

src/azul/service/avro_pfb.py
```diff
@@ -167,24 +167,37 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
...
                     else:
-                        # FIXME: Change 'string' to 'null' in https://github.com/DataBiosphere/azul/issues/2462
-                        assert 'string' in field['type']['items']
+                        # FIXME: Change 'string' to 'null'
+                        #        https://github.com/DataBiosphere/azul/issues/2462
+                        assert 'string' in field['type']['items'], field
                         default_value = [None]
+                else:
+                    assert False, field
                 object_[field['name']] = default_value
```
hannes-ucsc, line 195: It seem that there is another semantically neutral refactoring in this hunk that could be a separate commit.

hannes-ucsc (changes requested):
No fixups next time but push commits individually.

src/azul/service/avro_pfb.py
```diff
@@ -167,8 +167,13 @@ def _add_missing_fields(cls, name: str, object_: MutableJSON, schema):
         None is the default value, but because of https://github.com/DataBiosphere/azul/issues/2370
         this isn't currently reflected in the schema.
         """
-        object_schema = one(f for f in schema['fields'] if f['name'] == 'object')
-        entity_schema = one(e for e in object_schema['type'] if e['name'] == name)
+        if schema['type'] == 'record':
```
hannes-ucsc, line 170: Don't understand why this is in the commit `Include all file_base fields in related_files`. PL item maybe?

danielsotirhos: Avro PFB change factored out into `Fix: Avro PFB doesn't handle missing fields in dict nested in inner entities`  commit

